PR #2, opened 2015-05-15
Adding pandas

msund:
Handles exception with timestamps


chriddyp:
oh yeah, nice, thx man!

PR #9, opened 2016-04-19
[WIP] [DO NOT MERGE] React

jdfreder:
This PR refactors Dash to use React under the hood.  It will also change the API to be more symmetric to the DOM structure it represents. i.e.

```python
div([
    h2('hello dash')
])
```

would be represented by something roughly resembling

```html
<div>
  <h2>hello dash</h2>
</div>
```

@chriddyp I pulled your commits from https://github.com/chriddyp/messin.git because I think there's a lot of good work there!  Also, I need to give credit where credit is due.  I'm opening this PR so you can keep track of what I'm doing as I go along and interject along the way if you need to.

cc @jackparmer


jdfreder:
@chriddyp if you'd like, I can squash your commits...

jdfreder:
Talking to @chriddyp has made me realize this should be put on hold for a while in favor of some more discussion / exploration.

PR #18, opened 2017-06-21
Updated requirements.txt for development

jackwluo:
Should work for both 2.x and 3.x


chriddyp:
I'm not sure if we even need this `requirements.txt` file. I have `dev-requirements.txt` for testing and if users want to install, they should use `pip` or just `python setup.py install`

chriddyp:
I think we should actually remove the `requirements.txt` file from this repo and just keep `dev-requirements.txt` for circle tests

PR #20, opened 2017-06-21
Update circleci

esaezgil:
Update python versions for the default circleci image (https://circleci.com/docs/1.0/build-image-trusty/#python). Other possibility would be to update the circleci config to install the missing python versions.


chriddyp:
Closed with #127 . Thanks for this!

PR #23, opened 2017-06-22
Fix unicode error on Py3.6.1 during install

xnoder:
Fix a unicode handling error during installation
on Python 3.6.1 by forcing Py27 and Py3 to use
the compatible io.open method on both versions.


chriddyp:
thank you @xnoder !

PR #32, opened 2017-06-23
Falsey values ("", 0, [], {},  False) are as bad as None

cclauss:
In these instances, all other falsey values ("", 0, [], {},  False, etc.) are just as dangerous as `None` so we should treat them as if they were `None`.

This approach replaces an empty `name` with 'dash' even if the caller provides a `server`.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -35,13 +35,8 @@ class Dash(object):
         csrf_protect=True
     ):
         # allow users to supply their own flask server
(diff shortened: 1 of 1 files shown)


yunake:
fwiw this looks good, and fixes the `name` bug too.

HorlogeSkynet:
Hey, why this has not been merged yet ? Bye 👋

PR #48, opened 2017-06-27
add missing import json (?)

latorrefabian:


chriddyp:
🙈  of course! Thank you @latorrefabian . I'll add a test for this condition soon too.

PR #70, opened 2017-07-05
Dash App with Multiple Flask Routes

rc1405:
Modified flask endpoints to allow for multiple routes to be added to the Flask server, such as serving multiple Dash apps, or additional content.  Simple example below.

```
import dash
from flask import Flask

server = Flask("my app")

app2 = dash.Dash(name='place1', sharing=True, server=server, url_base_pathname='/test1')

app2.layout = html.Div([
        html.H1('This is a test1')
    ])

app3 = dash.Dash(name='place2', sharing=True, server=server, url_base_pathname='/test2')

app3.layout = html.Div([
        html.H1('This is test2')
    ])

server.run(port=80, host='0.0.0.0')
```

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -94,34 +94,47 @@ class Dash(object):
 
         self.server.add_url_rule(
             '{}_dash-layout'.format(self.url_base_pathname),
-            view_func=self.serve_layout)
+            view_func=self.serve_layout,
+            endpoint='{}_dash-layout'.format(self.url_base_pathname)
+        )
 
         self.server.add_url_rule(
             '{}_dash-dependencies'.format(self.url_base_pathname),
-            view_func=self.dependencies)
+            view_func=self.dependencies,
+            endpoint='{}_dash-dependencies'.format(self.url_base_pathname)
+        )
 
         self.server.add_url_rule(
             '{}_dash-update-component'.format(self.url_base_pathname),
             view_func=self.dispatch,
+            endpoint='{}_dash-update-component'.format(self.url_base_pathname),
             methods=['POST'])
 
         self.server.add_url_rule((
             '{}_dash-component-suites'
             '/<string:package_name>'
             '/<path:path_in_package_dist>').format(self.url_base_pathname),
-            view_func=self.serve_component_suites)
+            view_func=self.serve_component_suites,
+            endpoint='{}_dash-component-suites'.format(self.url_base_pathname)
+        )
 
         self.server.add_url_rule(
             '{}_dash-routes'.format(self.url_base_pathname),
-            view_func=self.serve_routes
+            view_func=self.serve_routes,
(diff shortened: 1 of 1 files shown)


rc1405:
Haha, yeah I saw your PR after I committed the changes, and was a bit confused about how I missed it until I looked at the time stamp.  I thought the implementation was different enough to throw some more options out there.

chriddyp:
This is pretty cool! I like this. I don't see any drawbacks to this. Will using `endpoint` bite us down the line in any way?

chriddyp:
For a related topic on multi-page URLs, check out https://github.com/plotly/dash/issues/77

cnblevins:
I am looking at the structure that is coming together here and wondering how the selective authorization use-case can be implemented...
I must be dense, because I am still trying to piece together the dash equivalent for the following (one dyno yielding a publicly accessible app1 "page" and a restricted-access app2 "page")...

```
[...]
from flask_login import LoginManager, UserMixin, login_user, login_required, logout_user, current_user
[...]

(at)app.route('/app1')
[...]

(at)app.route('/app2')
(at)login_required
[...]
```

rc1405:
I haven't seen any drawbacks to using endpoint, from what I've seen and the related docs is that it changes how the function is stored.  By default, it is stored by function name which causes a conflict when you try to store the same name twice unless it is the same object as with the authentication.

rc1405:
#77 fits my use case better, although I can picture environments where they might be need to be separate

Volodymyrk:
#77 does not help my use-case.
I am planning to build multiple, independent apps and I don't to simplify the deployment/management. I would not have any links between them, but I want those apps to have different URLs, so they can be embedded into iframes in the intranet portal.
I would like this to be merged into master, if no-one found a problem with endpoints.

chriddyp:
> I want those apps to have different URLs, so they can be embedded into iframes in the intranet portal.

FWIW, #77 allows you to do this as well:

```
app.layout = html.Div([
    dcc.Location(id='location'),
    html.Div(id='content')
])

@app.callback(Output('content', 'children'), [Input('location', 'pathname')])
def display_page(pathname):
    if pathname == '/':
         return app1.layout
    elif pathname == '/app-2':
         return app2.layout
    # [...]
```

chriddyp:
This looks good to me though. Thanks @rc1405 and @Volodymyrk !

Volodymyrk:
@chriddyp what happens with callbacks in app1 and app2 ?
In my quick test - they disappear, since they were not registered in app..
#70 is addressing this. However, I can't get it to work with a proper webserver (gunicorn)

brylie:
Wow! Is this in the docs? I have been searching around and only came to this PR from a lengthy discussion thread.

sidd-hart:
Has anyone tried to implement flask-login with this solution. It doesn't seem to be working for me. Would be very helpful if anyone could suggest a solution to implement flask-login.

Issue #78, opened 2017-07-10
Any planned support for data* or aria-* options?


rc1405:
As the title indicates, any planned support for the above options add they are supported in react?  Main use case would be for using bootstrap within apps.

chriddyp:
Good idea. It seems like we should support this.

It'll be a little bit tricky because `data-*` and `aria-*` are regex properties. Our dynamic component generation code depends on a strict set of properties rather than a pattern of property names (see https://github.com/plotly/dash/blob/master/dash/development/base_component.py#L188-L259).

rc1405:
Full disclosure, I don't have access to my development area but wanted to throw some initial thoughts together.  The catch is you would have to replace the dash (-) with an underscore(_) when initially initiating.  There are some documentation flaws in doing it this way that would need to be corrected, but some thoughts to debate.

`>>> import dash_html_components as html
>>> html.Div("testing", data_test="#testing")
Div(children='testing', data-test='#testing')
>>> 
`

https://github.com/rc1405/dash/blob/bootstrap/dash/development/base_component.py

rc1405:
or even add the available properties for each class of: data_flags=[{"test": "#testing"] and aria_flags[{"key": "value"}], which can be propagated to data-test="#testing" and aria-key="value" in base_component.py

rc1405:
First method of https://github.com/rc1405/dash/blob/bootstrap/dash/development/base_component.py

```import dash
import dash_html_components as html

app = dash.Dash()

app.layout = html.Div([
    html.H1('Hello Dash'),

    html.Div('''
        Dash: A web application framework for Python.
    '''),

], data_stuff='blah')

if __name__ == '__main__':
    app.run_server(host='0.0.0.0', port=8080)```

Provides capability such as:

![image](https://user-images.githubusercontent.com/13261087/28243694-367d114e-699a-11e7-9c9e-3b7948c1579d.png)

javValverde:
I am also really interested in using bootstrap with dash! Are there any plans for implementing it?

chriddyp:
I wrote a guide on how to add support for this in Dash in https://community.plot.ly/t/can-data-attributes-be-created-in-dash/7222/2?u=chriddyp

chriddyp:
Done in https://github.com/plotly/dash/pull/237

PR #104, opened 2017-07-29
Draft aria-* and data-* additions for Bootstrap

rc1405:
Pull request to discuss implementation of aria-* and data-* options for bootstrap and other uses.  Proposed solution to #78

### DIFF
--- a/dash/development/base_component.py
+++ b/dash/development/base_component.py
@@ -16,7 +16,7 @@ class Component(collections.MutableMapping):
     def __init__(self, **kwargs):
         for k, v in list(kwargs.items()):
             if k[:5] in ["data_", "aria_"]:
-                k = k.replace('_','-')
+                k = k.replace('_','-',1)
                 self._prop_names.append(k)
             elif k not in self._prop_names:
                 # TODO - What's the right exception here?
--- a/dash/development/base_component.py
+++ b/dash/development/base_component.py
@@ -228,7 +228,8 @@ def generate_class(typename, props, description, namespace):
             if(any(getattr(self, c, None) is not None for c in self._prop_names
                    if (c is not self._prop_names[0] and (not c[:5] in ['data_', 'aria_'])))):
 
-                return '{typename}('+', '.join([c+'='+repr(getattr(self, c, None))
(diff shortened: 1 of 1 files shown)


NMRobert:
This is working for me - it might be good to alter it so that it affects dcc.* components as well - currently they don't throw an exception, but the attributes don't show up.

dash/development/base_component.py
```diff
@@ -223,7 +226,7 @@ def __init__(self, {default_argtext}):
 
         def __repr__(self):
             if(any(getattr(self, c, None) is not None for c in self._prop_names
-                   if c is not self._prop_names[0])):
+                   if (c is not self._prop_names[0] and (not c[:5] in ['data_', 'aria_'])))):
```
chriddyp: `__repr__` should print out a string that can be copied and pasted back into a python program to get the same result. In this case, it looks like we have:
```
>>> html.Div('test', data_foo='bar')
Div(children='test', data-foo='bar')
```

chriddyp:
>  it might be good to alter it so that it affects dcc.* components as well 

That's a good point. The `dash_html_components` will end up rendering these attributes but any non-HTML component will not. In those cases, those components should throw an exception if a property wasn't supplied and if the component doesn't accept it.

I'm not sure what the solution would be for that case. Perhaps the component authors can include a special prop-type like `'data-*': PropTypes.string` which would have a special interpretation by the dash abstract component.

dash/development/base_component.py
rc1405: __repr__ adjusted to provide output as described

rc1405: Output is correct at
`Div(children='test', data_foo='bar')`

chriddyp:
> I'm not sure what the solution would be for that case. Perhaps the component authors can include a special prop-type like 'data-*': PropTypes.string which would have a special interpretation by the dash abstract component.

As mentioned above, I think that we'll need a solution that allows general component authors to specify that these properties are valid, rather than monkey patching the base component.
I wrote a guide in the community forum on the steps required to make this happen: https://community.plot.ly/t/can-data-attributes-be-created-in-dash/7222/2?u=chriddyp

I'm going to close this for now in favor of a more comprehensive approach. Thanks for opening the discussion!

PR #112, opened 2017-08-14
added id to KeyError for get and set on Components

ned2:
When building up a Component tree layout using an id setter (eg `layout['some-id'] = some_component`), this propagates the missing id value into the traceback, making debugging much easier


chriddyp:
Ah yeah, that's great. Thanks @ned2 !

PR #128, opened 2017-09-07
Remove excessive calls to `layout()` 

### LINKED ISSUE #116: Functions running too often on page load and callback
On page load, Dash serves the initial layout (when provided in a function) more than once. I've seen it run one of my "initial" functions at least 20 times. Likewise, if you have 5 inputs, a callback's function is running 5 times even if not all of those inputs have changed or fired. Caching has helped, but it still seems really inefficient and others have noticed similar behaviour.

Have a look [here for more details](https://community.plot.ly/t/many-updates-on-one-page-load-dash-app/4743).

chriddyp:
fixes #116 

If `app.layout` is a function, then it is now only called once on startup and once for the requests


chriddyp:
Usage

```
import dash
import inspect

app = dash.Dash()

print('App start')


def serve_layout():
    print('Calling layout')
    return html.Div('hlo world')


app.layout = serve_layout

if __name__ == '__main__':
    app.run_server(debug=False)
```


```
$ python app_layout_function.py
App start
Calling layout
 * Running on http://127.0.0.1:8050/ (Press CTRL+C to quit)
127.0.0.1 - - [07/Sep/2017 11:26:31] "GET / HTTP/1.1" 200 -
Calling layout
127.0.0.1 - - [07/Sep/2017 11:26:32] "GET /_dash-layout HTTP/1.1" 200 -
127.0.0.1 - - [07/Sep/2017 11:26:32] "GET /_dash-dependencies HTTP/1.1" 200 -
127.0.0.1 - - [07/Sep/2017 11:26:33] "GET /favicon.ico HTTP/1.1" 200 -
```

PR #138, opened 2017-09-25
To make dash support Flask-Login

imkfyao:
Add one configure parameter to indicate whether to support Flask-Login.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -2,6 +2,7 @@ import flask
 import json
 import plotly
 from flask import Flask, Response
+from flask_login import login_required
 from flask_compress import Compress
 from flask_seasurf import SeaSurf
 import os
@@ -26,7 +27,8 @@ class Dash(object):
         server=None,
         static_folder=None,
         url_base_pathname='/',
-        csrf_protect=True
+        csrf_protect=True,
+        loginSupported=False
     ):
         # allow users to supply their own flask server
         if server is not None:
@@ -79,21 +81,20 @@ class Dash(object):
                 endpoint=name,
                 methods=methods
             )
-
         add_url(
             '{}_dash-layout'.format(self.config['routes_pathname_prefix']),
-            self.serve_layout)
+            self.serve_layout if not loginSupported else login_required(self.serve_layout))
 
         add_url(
             '{}_dash-dependencies'.format(
                 self.config['routes_pathname_prefix']),
-            self.dependencies)
+            self.dependencies if not loginSupported else login_required(self.dependencies))
 
         add_url(
             '{}_dash-update-component'.format(
                 self.config['routes_pathname_prefix']
             ),
-            self.dispatch,
+            self.dispatch if not loginSupported else login_required(self.dispatch),
             ['POST'])
 
         add_url((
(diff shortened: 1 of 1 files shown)


AfroMetal:
It is all good, and really required but please, please, please, fix that `loginSupported` to proper pythonic `login_supported`.

imkfyao:
Thanks for the review. My original idea for a full Flask-Login supported Dash App is like the one in this post https://community.plot.ly/t/flask-login-suport-for-dash/5934

AfroMetal:
But that will require writing html. That ruins the whole beauty and purpose of using Dash.

imkfyao:
That should be doable to render the login page with Dash components. I will think about it.

chriddyp:
I would prefer that this would be a separate package so that this `dash` would remain un-opinionated about which auth method to use.
There are some other approaches to this in https://github.com/plotly/dash-auth for now.

jkgenser:
@imkfyao: are you thinking of turning it into a separate package?

numpynewb:
@chriddyp this is still missing from dash-auth, I think. Might it be reasonable to merge this bit while dash-auth matures? Having access to the flask tool chain, specifically login/auth, is what keeps me from migrating from R/Shiny to Dash. I would love to contribute if there is a clear path toward accomplishing this!

rmarren1:
@numpynewb Dash.server is a plain flask instance, so you can add any auth to from the flask toolchain

numpynewb:
@rmarren1 Thank you for pointing this out. Can you possibly show me how it might work with a simple vignette? For example, how would one use Flask-Login to protect a simple dash app?

numpynewb:
@rmarren1 I fumbled around the internet, based on your very concise response, and was able to cobble something together that seems to work:

```{python}
from dash import Dash
from werkzeug.wsgi import DispatcherMiddleware
import flask
from flask import Flask, Response, redirect, url_for, request, session, abort
from werkzeug.serving import run_simple

from flask_login import LoginManager, UserMixin, login_required, login_user, logout_user 


server = flask.Flask(__name__)

server.config.update(
    DEBUG = True,
    SECRET_KEY = 'secret_xxx'
)

# flask-login
login_manager = LoginManager()
login_manager.init_app(server)
login_manager.login_view = "login"


# silly user model
class User(UserMixin):

    def __init__(self, id):
        self.id = id
        self.name = str(id)
        self.password = "secret"
        
    def __repr__(self):
        return "%d/%s/%s" % (self.id, self.name, self.password)


# create some users with ids 1 to 20       
users = [User("jason")]
 
# somewhere to login
@server.route("/login", methods=["GET", "POST"])
def login():
    if request.method == 'POST':
        username = request.form['username']
        password = request.form['password']        
        if password == "secret":
            id = username
            user = User(id)
            login_user(user)
            return flask.redirect(request.args.get("next"))
        else:
            return abort(401)
    else:
        return Response('''
        <form action="" method="post">
            <p><input type=text name=username>
            <p><input type=password name=password>
            <p><input type=submit value=Login>
        </form>
        ''')


# somewhere to logout
@server.route("/logout")
@login_required
def logout():
    logout_user()
    return Response('<p>Logged out</p>')


# handle login failed
@server.errorhandler(401)
def page_not_found(e):
    return Response('<p>Login failed</p>')
    
    
# callback to reload the user object        
@login_manager.user_loader
def load_user(userid):
    return User(userid)

dash_app1 = Dash(__name__, server = server, url_base_pathname='/dashboard' )
dash_app2 = Dash(__name__, server = server, url_base_pathname='/reports')
dash_app1.layout = html.Div([html.H1('Hi there, I am app1 for dashboards')])
dash_app2.layout = html.Div([html.H1('Hi there, I am app2 for reports')])

def protect_views(app):
    for view_func in app.server.view_functions:
        if view_func.startswith(app.url_base_pathname):
            app.server.view_functions[view_func] = login_required(app.server.view_functions[view_func])
    
    return app

dash_app1 = protect_views(dash_app1)
dash_app2 = protect_views(dash_app2)

@server.route('/')
@server.route('/hello')
@login_required
def hello():
    return 'hello world!'

@server.route('/dashboard')
def render_dashboard():
    return flask.redirect('/dash1')


@server.route('/reports')
def render_reports():
    return flask.redirect('/dash2')


app = DispatcherMiddleware(server, {
    '/dash1': dash_app1.server,
    '/dash2': dash_app2.server
})

run_simple('0.0.0.0', 8080, app, use_reloader=True, use_debugger=True)
```

rmarren1:
@numpynewb Interesting, I've never seen a multi page app deployed that way. I generally use [this](https://dash.plot.ly/urls) method to make a multi page app. I've put together an app with login required pages, and it ended up looking very similar to this except using the linked method. If you aren't sharing state between the two dash apps, and don't mind a small hit to performance when switching apps (initial page load usually takes the longest), this looks kosher to me!

PR #163, opened 2017-11-28
Standardise HTML link tags to use the HTML5 format as shown here:

jbampton:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link  as this
project uses a HTML5 DOCTYPE.


chriddyp:
ah yeah, that's much better. Thanks @jbampton , and welcome to Dash!

jbampton:
Hi @chriddyp  thanks ! 👍

PR #164, opened 2017-11-28
Removing outdated reference to NYC masterclass

ravintpillai:
Now that the masterclass has been and gone, I'm suggesting removing the reference to it from the README (I got confused and tried to sign up because I didn't read the dates on the link in the README carefully enough).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # Dash
 
-***
-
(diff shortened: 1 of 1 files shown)


chriddyp:
thanks @ravintpillai !

PR #165, opened 2017-11-28
Fix position of HTML footer tag to be inside the HTML body tag.  

jbampton:
You can see the error at https://validator.w3.org/ when validating a Dash app.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -298,12 +298,11 @@ class Dash(object):
                         Loading...
                     </div>
                 </div>
+                <footer>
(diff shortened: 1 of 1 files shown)


jbampton:
Another way to see the malformed HTML markup is in Firefox 57 desktop mode view source of a Dash app  -> view-source:https://plot.ly/dash/  and the HTML tags that are highlighted in red are the ones in the wrong place.

chriddyp:
Very nice. Thanks @jbampton !

PR #166, opened 2017-11-28
Fix HTML meta tag to be HTML5 compliant by removing the closing slash…

jbampton:
… as seen on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

Also HTML5 meta tags can be seen  here: https://www.w3schools.com/tags/tag_meta.asp

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -288,7 +288,7 @@ class Dash(object):
         <!DOCTYPE html>
         <html>
             <head>
-                <meta charset="UTF-8"/>
+                <meta charset="UTF-8">
                 <title>{}</title>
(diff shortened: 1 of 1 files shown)


chriddyp:
wow, i've been doing it wrong all along. Here's a helpful link from the w3 spec itself: https://www.w3.org/TR/html5/syntax.html#void-elements. These are apparently "void elements", meaning they don't need a closing tag. Others include area, base, br, col, embed, hr, img, input, keygen, link, meta, param, source, track, wbr
image

chriddyp:
thanks again @jbampton !

PR #167, opened 2017-11-29
Standard import "import io" should be placed before  "from setuptools…

jbampton:
… import setup, find_packages" (wrong-import-order). Also fix continued line indentation.  These issues are found from using Pylint https://www.pylint.org/

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -284,7 +284,7 @@ class Dash(object):
         css = self._generate_css_dist_html()
         config = self._generate_config_html()
         title = getattr(self, 'title', 'Dash')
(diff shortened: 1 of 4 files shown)


jbampton:
You can see an example of the standard import order here https://stackoverflow.com/questions/22722976/import-order-coding-standard

chriddyp:
Nice, thanks! I had thought that flake8 would catch all of these things but apparently not. It would be good to get pylint in the acceptance tests as well. Here's where we do the flake8 tests:
https://github.com/plotly/dash/blob/master/tox.ini#L15

PR #169, opened 2017-11-29
Standard imports "unittest, os, collections" should be placed before …

jbampton:
…"import plotly" (wrong-import-order).  Remove used variable "i"


chriddyp:
thanks!

Issue #144, opened 2017-10-12
pycharm autocomplete not work for dash


czhiling:
Thanks a lot for developing plotly/dash. 

I am using pycharm IDE. But the editor cannot do auto-compete for dash. I use anaconda3\python as interpreter. Since the dash in installed by pip. Pycharm provide the auto-complete for plotly which installed via anaconda. Is it possible to let the pycharm do this also for dash?

chriddyp:
thanks for reporting! we should definitely get this sorted out 👍

streamerd:
Hi, having a similar issue, using Pycharm. 

Installed dependencies still look as if they are not installed (underlined,red). Other dependencies are ok, those are the problematic ones. Does not affect the code, still looks a bit weird.

```
import plotly.plotly as py
import plotly
from plotly import figure_factory as FF
import plotly.graph_objs as go
import cufflinks as cf

import dash
from dash.dependencies import Input, Output
import dash_core_components as dcc
import dash_html_components as html
```

alysivji:
I'm looking into this issue.

Thanks.

chriddyp:
Getting autocomplete to work for the component libraries would be a quite a bit more work. I've outlined the steps that would be required to make this happen in this comment here: https://github.com/plotly/dash/issues/150#issuecomment-350872879

chriddyp:
This is getting done as part of the [sponsored](https://plot.ly/products/consulting-and-oem) initiative "A Pleasant and Productive Developer Experience" in #150

Issue #150, opened 2017-10-29
explicit Components classes to allow autocomplete


sdementen:
[metadata.zip](https://github.com/plotly/dash/files/1424805/metadata.zip)
Currently, the Dash components classes are generated only at runtime. Therefore, it is not possible to get autocomplete feature or doc/help contextual hints in IDE (e.g. pycharm).
I would propose to explicit the classes as a real python module and so convert the metadata.json in a metadata.py that could be imported in a standard way.
It could also ease the debugging when adapting these component classes to implement the different TODOs in the code.

I have slightly adapted the `component_loader.py` as well as the `__init__.py` of the dash_core_components and dash_html_components to achieve this.
You can find in attachements the resulting metadata.py for both modules.

As this requires changes in different repos (dash but also the dash_*_components), I haven't yet done a PR on this. 
If you find the approach valuable, I am ready to follow your guidelines on how to push the changes.

sdementen:
I have pushed the changes in the branches classgen-#150 in https://github.com/sdementen/dash/tree/classgen-%23150
https://github.com/sdementen/dash-core-components/tree/classgen-%23150
https://github.com/sdementen/dash-html-components/tree/classgen-%23150

sdementen:
btw, I am not yet happy with the ISUNDEFINED and MUSTBEDEFINED sentinel approach...

chriddyp:
Thanks @sdementen ! Excited to check this out.

I'm having a hard time finding the commits in `classgen-#150` - am I looking in the right place?
![image](https://user-images.githubusercontent.com/1280389/32146221-3320ebde-bcaa-11e7-9a4c-75148986de92.png)

sdementen:
sorry, it wasn't pushed correctly. You should be able to see them now.

on the approach, I wonder if the metadata.py file should be "compiled" during the development/release cycle (and so included in the source), or generated on the fly (with some caching mechanism that checks metadata.json is newer than metadata.py)

chriddyp:
> on the approach, I wonder if the metadata.py file should be "compiled" during the development/release cycle (and so included in the source)

Great idea, I think that's it exactly.

Here is a rough set of steps that would be required for creating these components at build-time vs run-time. cc @alysivji

**1. Modify `dash.development.base_component.generate_class` and `dash.development.component_loader.load_components`**

[...]

PR #174, opened 2017-12-02
Import submodules into __init__.py for IDE code completion

alysivji:
Fix for #144, but only for the dash library itself.

I was also looking to add IDE completion for [dash-core-components](https://github.com/plotly/dash-core-components), but this seems a bit more difficult as component data is loaded from `metadata.json`.

```python
print(dir(dcc))
```
We see all the core components in the namespace (and, therefore, in IPython), but they do not show up in the IDE. Hardcoding `__all__` is a workaround, but not something I want to do.

@chriddyp do you have any thoughts on this?

### DIFF
--- a/dash/__init__.py
+++ b/dash/__init__.py
@@ -1,5 +1,4 @@
 from .dash import Dash  # noqa: F401
-from . import authentication  # noqa: F401
 from . import dependencies  # noqa: F401
 from . import development  # noqa: F401
 from . import exceptions  # noqa: F401
--- a/dash/__init__.py
+++ b/dash/__init__.py
@@ -1,3 +1,7 @@
 from .dash import Dash  # noqa: F401
+from . import authentication  # noqa: F401
+from . import dependencies  # noqa: F401
 from . import development  # noqa: F401
+from . import exceptions  # noqa: F401
+from . import resources  # noqa: F401
 from .version import __version__  # noqa: F401


dash/__init__.py
```diff
@@ -1,3 +1,7 @@
 from .dash import Dash  # noqa: F401
+from . import authentication  # noqa: F401
```
chriddyp: let's leave this one out, as it isn't used anymore (`dash-auth` package is used instead)

chriddyp:
This looks good to me. Is this syntax safe for python 2 and python 3? Tests passed, so I'm assuming that it is.

chriddyp:
IDE completion for the component libraries will probably require doing code generation/builds rather than dynamic code evaulation at runtime. This was started in https://github.com/plotly/dash/issues/150, I'll continue the discussion over there.

chriddyp:
> I was also looking to add IDE completion for dash-core-components, but this seems a bit more difficult as component data is loaded from metadata.json.

I've added some comments about what would be required to do this here: https://github.com/plotly/dash/issues/150#issuecomment-350872879

dash/__init__.py
alysivji: Fixed!

alysivji:
Relative imports work in both Python 2 and 3. Looks like they were introduced in [Python2.5 via PEP 328](https://docs.python.org/2.5/whatsnew/pep-328.html).

Thanks for the detailed writeup on how to add IDE completion for components. I will take a look when I get some time.

chriddyp:
Thanks @alysivji !

Issue #182, opened 2017-12-27
Set a cookie from callback function


oryba:
It seems that setting a cookies from callback function is currently impossible. I think it would be great if this feature will appear in the future versions.

Thank you:)

ned2:
My immediate thought was that this could be done with a callback that inserts a <script> tag with some JavaScript that sets a cookie, but according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML), the html5 spec says that <script> tags inserted via innerHTML should not execute. So no go there.

There are other ways to get arbitrary JS code to execute via innerHTML, which might work (the same MDN article provides an example of how), but at that point it's definitely a hack job.

oryba:
Thank you for response!

Interesting and useful workaround. But despite the fact it's a hack job, setting `httponly` cookie couldn't be done this way.

chriddyp:
I suspect this could be a custom Dash component.

What's the use case that you have in mind?

PR #183, opened 2017-12-30
Enable custom response support for callback functions

oryba:
Based on issue #182 (Set a cookie from callback function).

With the following functionality callback function can return not only Dash components tree, but a customizable DashResponse with cookies, headers etc. 

```
from dash import DashResponse
...

@app.callback(
    dash.dependencies.Output('content-block', 'children'),
    [dash.dependencies.Input('comparison-filter', 'value')])
def on_filter_change(comparison_value):
    response = DashResponse(html.Div('foo', className="test"))
    response.set_cookie('bar', 'baz')
    return response
```

### DIFF
--- a/dash/response.py
+++ b/dash/response.py
@@ -1,0 +1,34 @@
+"""
+Extended Flask Response for using inside a callbacks
+"""
+import json
+import plotly
+from flask import Response
+
+
+class DashResponse(Response):
+    """
+    Flask Response extended with option to convert a regular response to
+    valid Dash json-encoded
+    """
+    def __init__(self, *args, **kwargs):
+        self.dash_response = args[0] if args else None
+        super(DashResponse, self).__init__(
+            mimetype='application/json', *args, **kwargs)
+
+    def jsonify_response(self, output):
+        """
+        convert response to valid Dash json-encoded
+        :param output: output element for the callback
+        :return:
+        """
+        response = {
+            'response': {
+                'props': {
+                    output.component_property: self.dash_response
+                }
+            }
+        }
+        self.set_data(
+            json.dumps(response, cls=plotly.utils.PlotlyJSONEncoder)
+        )
--- a/dash/__init__.py
+++ b/dash/__init__.py
@@ -1,4 +1,5 @@
 from .dash import Dash  # noqa: F401
+from .response import DashResponse  # noqa: F401
 from . import dependencies  # noqa: F401
 from . import development  # noqa: F401
 from . import exceptions  # noqa: F401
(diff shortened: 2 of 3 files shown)


chriddyp:
Yeah, I like this solution. An alternative way to solve #182 would be to implement a `Cookie` Dash component. However, this solution is even more flexible by allowing things like headers.

dash/response.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+import json
+import plotly
+from flask import Response
+
+
+class DashResponse(Response):
+    """
+    Flask Response extended with option to convert a regular response to
+    valid Dash json-encoded
```
chriddyp: missing "component" here? How about:
> Flask Response extended with option to convert a regular response to
> valid Dash json-encoded component. 
> Return a `DashResponse` object from a Dash callback in order to set
> other properties of the response like headers or cookies

dash/response.py
```diff
@@ -0,0 +1,34 @@
...
+        response = {
+            'response': {
+                'props': {
+                    output.component_property: self.dash_response
+                }
+            }
+        }
+        self.set_data(
+            json.dumps(response, cls=plotly.utils.PlotlyJSONEncoder)
+        )
```
chriddyp: could we make this code a little bit more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) by combining it in some way with these lines? https://github.com/plotly/dash/blob/b61f0d371239ee5a6286ce539fab24d2d3de432a/dash/dash.py#L494-L506

chriddyp:
Looks good! In order for this to be merged, I'd like to see:
- The `DRY` component addressed
- A test

I think that a full integration test would be best suited here. Most of the integration tests are in the `dash-renderer` library (https://github.com/plotly/dash-renderer/blob/cef2313478ee2c37b91b9ef773bf6991bc96240e/tests/test_render.py). We should add an integration test to this library as well.

chriddyp:
I have merged #184 , so let's rebase this PR off of that and then add an integration test.

@oryba - Do you have the time to wrap up this PR from my comments above or would you like one of the core contributors to do so?

oryba:
@chriddyp - unfortunately, now I have no time to do this, but I'm going to go back to this PR as soon as my current project will be finished

jkseppan:
Hey @oryba - I'm interested in this feature and I'm willing to work on it, so if you don't mind I could try to add the requested test and do the refactoring?

chavis1789:
Hey @jkseppan, sorry for the late answer. I don't mind, thanks for your help, glad to see someone has a chance to finalize this feature.

oryba:
Hey @jkseppan, sorry for the late answer. I don't mind, thanks for your help, glad to see someone has a chance to finalize this feature.

alexcjohnson:
Superseded by #401 - thanks @oryba and @jkseppan! We still need to get the new one merged but I believe this older PR can be closed.

Issue #189, opened 2018-01-05
favicon.ico request now causes an Exception 


ned2:
It looks like this commit #181 and specifically [this line](https://github.com/plotly/dash/commit/2c07f191bac881b37d8705aaa4374d04a796f9cb#diff-228c9aefac729977642672aa1416bacaL278) which removed *args and **kwargs from Dash's index function signature is now causing an exception on the initial page load of an app due to the `/facivon.ico` request being handled by the index method, thereby getting a `path=favicon.ico` parameter.

As mentioned in #152 this is caused by Dash serving the index route as a catchall for otherwise unmatched routes. 

```
Traceback (most recent call last):
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/app.py", line 1998, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/app.py", line 1986, in wsgi_app
    response = self.handle_exception(e)
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/app.py", line 1983, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/app.py", line 1615, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/home/nejl/.pyenv/versions/3.6.1/envs/dash/lib/python3.6/site-packages/flask/app.py", line 1613, in full_dispatch_request
    rv = self.dispatch_request()
[...]

PR #190, opened 2018-01-05
added PreventUpdate exception and attached handler to Flask server

ned2:
create a new exception `PreventUpdate` for use when a callback needs to be aborted without updating the state of the client.

This exception is registered with the Flask server using an `errorhandler` such that on catching this exception Flask will print any message included in the exception to standard error and then return an empty 204 NO CONTENT response.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import sys
 import collections
 import importlib
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1,3 +1,4 @@
+import sys
 import collections
 import importlib
 import json
@@ -52,6 +53,12 @@ class Dash(object):
 
         # gzip
         Compress(self.server)
+
+        @self.server.errorhandler(exceptions.PreventUpdate)
+        def handle_error(error):
+            """Handle a halted callback and return an empty 204 response"""
+            print(error, file=sys.stderr)
+            return ('', 204)
 
         # static files from the packages
         self.css = Css()
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -57,7 +57,7 @@ class Dash(object):
         Compress(self.server)
 
         @self.server.errorhandler(exceptions.PreventUpdate)
-        def handle_error(error):
+        def _handle_error(error):
             """Handle a halted callback and return an empty 204 response"""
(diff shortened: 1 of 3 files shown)


ned2:
Oops, I forgot about Python 2. I hope importing the print function from __future__ is ok. Without this I don't really know of an clean way to print to stderr. But now I think about it, it's not clear this message should be going to stderr. thoughts?

Also, tests are failing because of #189.

chriddyp:
Also #189 has been merged, so we can rebase this branch off master and the baseline tests should be fixed

ned2:
I rebased off master and tests are passing again :)

I also added an integration test that hopefully addresses the criteria you described. 

One thing that was unexpected is that apparently callback2, which is the one that is dependant on the one that is aborted is *never* triggered, even on the initial page load. Is this expected? My intuitions are that callback2 should have been run once on the initial page load given that output1 is present in the layout on page load. May there's also an argument for the current behaviour though...

ned2:
Yep, am happy, put it in!

rquintino:
Hi @chriddyp @ned2 , I was testing this with a live dcc.Interval and a table component. 

My issue, although the table values are not updated, as expected when raising the new exception, it seems to reset/mess up with table selected indexes/sort at each refresh.

Main scenario here was table live update, but refreshing only when data changes.  

thx! 

import datetime
import sys
import pandas as pd

app = dash.Dash()

DF_SIMPLE = pd.DataFrame({
    'x': ['A', 'B', 'C', 'D', 'E', 'F'],
    'y': [4, 3, 1, 2, 3, 6],
    'z': ['a', 'b', 'c', 'a', 'b', 'c']
})

app.layout = html.Div([
    html.H4('DataTable Sample',id="app_title"),
    dt.DataTable(
        rows=DF_SIMPLE.to_dict('records'),id='datatable-simple'
    ),
    dcc.Interval(id='my-interval', interval=5000),
], className="container")


app.css.append_css({
    'external_url': 'https://codepen.io/chriddyp/pen/bWLwgP.css'
})

@app.callback(
    Output('datatable-simple', 'rows'),
    [Input('my-interval', 'n_intervals')])
def update_table(n):
    print("Checking...")
    raise dash.exceptions.PreventUpdate("No data changed!")

PR #200, opened 2018-01-21
Improve callback decorator

ned2:
The function objects stored in a `Dash` instance's `callback_map` are currently not useful for introspecting:

```
>>> app.callback_map['dash-container.children']['callback'].__name__
'add_context'
```

The `functools.wraps` decorator passes various attributes of a decorated function onto the resulting function object, making introspection work as expected:

```
>>> app.callback_map['dash-container.children']['callback'].__name__
'add_context'
```

The `callback_map` can now be used for more useful profiling of all registered callbacks in an app.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -6,6 +6,7 @@ import importlib
 import json
 import pkgutil
 import warnings
+from functools import wraps
 
 import plotly
 import dash_renderer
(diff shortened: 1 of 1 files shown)


chriddyp:
very nice. thanks @ned2 !

Issue #202, opened 2018-01-22
allow disabling compression of Flask responses 


ned2:
Dash currently uses the flask-compress package to automatically compress every response from the Flask server. It does this on the attached Dash.server instance regardless of whether the server was supplied as a param or created in `__init__`. This means that there's currently no way to opt-out of compression.

Apps that are concerned at all with scalability should have responses compressed by the web server (eg Apache or Nginx) as this is more efficient than having the Python process do this. We should therefore provide the ability to opt out of compression.

Two possible options:
1. Have a `compress=True` param on the Dash instance, such that compression only occurs when the user creates a Dash instance as `Dash(compress=False)`
2. Only perform compression on Flask instances created in `__init__` if the user did not supply a server instance.

Both options make compression the default behaviour, which I think is what we want. Of the two, I think I like 2 more, as no extra Dash param is required, and to me it makes sense that if a user has chosen to supply a server instance, they're more likely to have more advanced needs and know what they're doing.  The only disadvantage is that people with apps currently supplying their own their own servers would silently lose compression.

ned2:
Sure, I'm happy with that then. PR created :) #203

ned2:
Fixed with #203

PR #203, opened 2018-01-22
Add a 'compress' param to Dash class's __init__()

ned2:
Adding a `compress` param provides the ability to opt out of automatically having all responses gzipped.

This is desirable because it is less efficient to be doing compression within the Python process than for the web server to handle it. Dash apps that care about scalability can now opt out.

Addresses #202


chriddyp:
this looks good to me. thanks @ned2

PR #233, opened 2018-03-25
update minimum versions for other dash packages in dev-requirements.txt

ned2:
I found out recently that tox was running the Dash tests with somewhat older versions of the other Dash packages. I had assumed that tox created fresh virtualenvs, pulling down the latest versions of the needed packages. Turns out that the default behaviour is to reuse the existing virtualenvs without updating packages if the minimum version requirements are met.

I might just get into the habit of using `tox -r` which forces re-creation of the virtualenvs. I wonder if it would be worth making this the default. Seems like not doing this leaves open the possibility of running against older Dash packages when the dev-requirements.txt is not updated.


chriddyp:
> I might just get into the habit of using tox -r  which forces re-creation of the virtualenvs. I wonder if it would be  worth making this the default. Seems like not doing this leaves open the  possibility of running against older Dash packages when the  dev-requirements.txt is not updated.

Yeah, this seems like a good default to use here. Would you mind updating this PR to include that option?

ned2:
Done

Also, just realised I forgot to create a new branch for this PR, sorry. let me know if that's an issue.

Issue #236, opened 2018-03-30
Tox tests fail on master branch


rmarren1:
With a fresh clone of the master branch, running tox hits an error for both python2.7 and python3.6
Full output:
```
ryan@galliumos:~/MSK/dash/base/dash$ tox
GLOB sdist-make: /home/ryan/MSK/dash/base/dash/setup.py
py27 create: /home/ryan/MSK/dash/base/dash/.tox/py27
py27 installdeps: -rdev-requirements.txt
py27 inst: /home/ryan/MSK/dash/base/dash/.tox/dist/dash-0.21.0.zip
py27 installed: asn1crypto==0.24.0,astroid==1.6.2,backports.functools-lru-cache==1.5,certifi==2018.1.18,cffi==1.11.5,chardet==3.0.4,click==6.7,configparser==3.5.0,cryptography==2.2.2,dash==0.21.0,dash-core-components==0.21.1,dash-flow-example==0.0.3,dash-html-components==0.9.0,dash-renderer==0.12.1,decorator==4.2.1,enum34==1.1.6,flake8==3.5.0,Flask==0.12.2,Flask-Compress==1.4.0,funcsigs==1.0.2,functools32==3.2.3.post2,futures==3.2.0,idna==2.6,ipaddress==1.0.19,ipython-genutils==0.2.0,isort==4.3.4,itsdangerous==0.24,Jinja2==2.10,jsonschema==2.6.0,jupyter-core==4.4.0,lazy-object-proxy==1.3.1,MarkupSafe==1.0,mccabe==0.6.1,mock==2.0.0,nbformat==4.4.0,pbr==4.0.0,percy==1.1.0,plotly==2.5.1,pluggy==0.6.0,py==1.5.3,pycodestyle==2.3.1,pycparser==2.18,pyflakes==1.6.0,pylint==1.8.3,pyOpenSSL==17.5.0,pytz==2018.3,requests==2.18.4,selenium==3.11.0,singledispatch==3.4.0.3,six==1.11.0,tox==2.9.1,tox-pyenv==1.1.0,traitlets==4.3.2,urllib3==1.22,virtualenv==15.2.0,Werkzeug==0.14.1,wrapt==1.10.11
py27 runtests: PYTHONHASHSEED='486919446'
py27 runtests: commands[0] | python --version
Python 2.7.12
py27 runtests: commands[1] | python -m unittest tests.test_integration
 * Running on http://127.0.0.1:8050/ (Press CTRL+C to quit)
[...]

Issue #232, opened 2018-03-25
Events are broken; base_component.parse_events needs updating


ned2:
Moving from `dash-html-components==0.9.0rc1` to `dash-html-components==0.9.0` causes any callbacks using `click` Events  to break as `Dash._validate_callback` now raises a `NonExistantEventException`. This appears to be because base components are now generated with `available_events = []`. This is due to  `base_component.parse_events` not being updated to handle the new PropType data structure used to define Component props created by `dash_html_components` and `dash_core_components`.

This is also causing the tests to fail now, as invocation of a callback with a valid event in `test_exception_event_not_in_component` is now failing. So would be good to prioritise this.

PR #237, opened 2018-03-30
Support for wildcard attributes, implement data-* attribute

rmarren1:
Adds the data-* html attributes, like so:
```
>>> import dash_html_components as html
>>> html.Div(id="my-id", **{'data-foo': 'foo', 'data-bar': 'bar'})
Div(id='my-id', data-foo='foo', data-bar='bar')
```

Tested using Tox, however I
  1. removed the `python -m unittest tests.test_react` test, since this is broken in #236 and #232 
  2. added the `python -m unittest tests.development.test_base_component` and `python -m unittest tests.development.test_component_loader` tests, since only running `tox` was not causing these tests to run (or are are these tests generally invoked in another way?)


dash/development/base_component.py
```diff
@@ -366,6 +381,27 @@ def parse_events(props):
...
+    props: dict
+        Dictionary with {propName: propMetadata} structure
+
+    Returns
+    -------
+    list
+        List of Dash valid wildcard prefixes
+    """
+    list_of_valid_wildcard_attr_prefixes = []
+    for wildcard_attr in ["data-*"]:
```
chriddyp: I suppose `"aria-*"` too

chriddyp:
This looks great, thanks @rmarren1 ! 
- Let's add an integration test as well that verifies that the attributes are actually rendered in the application.
   - In the `dash-renderer` repo, we can add a `data-test` attribute in this test: https://github.com/plotly/dash-renderer/blob/master/tests/test_render.py#L70
   - In this repo, let's create a new test like `test_wildcard_data_attributes`. This test would look roughly like:
```
app = dash.Dash()
app.layout = html.Div([
    html.Div(
        id='data-element',
        **{
            'data-string': 'multiple words',
            'data-number': 512,
            'data-none': None,
            'data-date': datetime.datetime(2012, 1, 10)
        }
    )
])

self.startServer(app)

div = self.wait_for_element_by_id('data-element')
self.assertEqual(
    div.get_attribute('innerHTML'),
    (
        '<div id="data-element" '
        'data-string="multiple words" '
        'data-number="512" '
        'data-none=""'
        'data-date="2012-01-10"></div>'
)
```
- Similarly, let's add an integration test that updates these attributes

I'll take a look at https://github.com/plotly/dash-renderer/issues/51 now

chriddyp:
I've published a prerelease version at `pip install dash==0.22.0rc1`

chriddyp:
Added some comments in https://github.com/plotly/dash-renderer/issues/51#issuecomment-378293579 that actually apply this repo

rmarren1:
https://github.com/plotly/dash-html-components/pull/40
https://github.com/plotly/dash-renderer/pull/52

tests/test_integration.py
```diff
@@ -116,6 +173,61 @@ def callback2(value):
...
 
+    def test_wildcard_data_attributes(self):
+        app = dash.Dash()
+        app.layout = html.Div([
+            html.Div(
+                id="inner-element",
+                **{
+                    'data-string': 'multiple words',
+                    'data-number': 512,
+                    'data-none': None,
```
rmarren1: This value will not be rendered in the html div.

rmarren1:
Test is dependent on changes in https://github.com/plotly/dash-html-components/pull/40

claresloggett:
This is awesome, thank you! 

I see that to implement this, you're now checking not just for valid attributes but also for attributes that match valid wildcards, i.e. 'data-*' and 'aria-*'. A question - and I'm totally open to being convinced this is a bad idea - is there a reason to not simply any attributes through, i.e. just not bother checking them? From my naive point of view that seems more future-proof. Could it be harmful somehow?

chriddyp:
> is there a reason to not simply allow any attributes, i.e. just not bother checking them

Yeah, the main issue is that higher-level React based components (i.e. components that don't just render the HTML components) might not actually end up rendering these attributes.

For example, here is the `dcc.Slider` component: https://github.com/plotly/dash-core-components/blob/d034e25f6f56423ef2de7e5905328396040cf4d8/src/components/Slider.react.js#L23. It is just a light wrapper around the 3rd party `rc-slider` component (https://react-component.github.io/slider/examples/slider.html) which might not actually end up rendering `data-*` attributes in the DOM.

So, if we don't do validation of properties on the component-level, then the end user might think that their `data-*` attributes are getting rendered in the DOM when they actually aren't. It's up to the component author to make sure that their component will render these attributes.

Re future proof - once the component author ends up supporting them, the end users will be able to upgrade their package to the new version that supports them. If they _don't_ get errors on previous versions, then the end user might not be aware that they are working on a version that _doesn't yet_ support these properties and they won't be aware that they need to upgrade.

chriddyp:
image

Very, very nice @rmarren1 🎉 

- I've published a prerelease version of `dash-html-components` at `dash-html-components==0.11.0rc1`. Let's update the `dev-requirements.txt` file with this new version so that we can make sure that circleci passes the tests. After everything is merged and released, we can move it up to the new stable version at `0.11.0`.

- Finally, let's update the `CHANGELOG.md` and `version.py` file. That way, once tests pass I can merge and release immediately.

rmarren1:
This is down by the way" https://unpkg.com/dash-html-components@0.11.0rc1/dash_html_components/bundle.js

chriddyp:
@rmarren1 - Thank you! This is really great stuff. I'm 👍 with these changes. Is there anything else that you would like to add? Otherwise, I will merge and release.

rmarren1:
I think it's good for now. go ahead!

kmader:
@chriddyp to get it working I needed to do both
```pip install dash==0.22.0rc1``` and 
```pip install dash_html_components==0.11.0rc1```

rmarren1:
I needed `dash==0.22.0rc2`

flasker:
I got an exception saying:

"Exception: Layout must be a dash component or a function that returns a dash component."

claresloggett:
I'm using `dash==0.22.0rc2` and `dash_html_components==0.11.0rc1`, but as @rmarren1 points out https://unpkg.com/dash-html-components@0.11.0rc1/dash_html_components/bundle.js appears to be down. 

So far as I can tell the fix is to use `app.scripts.config.serve_locally=True`, but this makes it hard for me to load other external JavaScript files (I am trying to also load Bootstrap). Can I currently get this PR working without `serve_locally=True`? @chriddyp will the hosting of `dash_html_components==0.11.0rc1` go back up in the near future? `dash==0.22.0rc2` seems to be fine.

rmarren1:
Try `dash-html-components==0.11.0rc5
dash==0.21.1`

claresloggett:
@rmarren1 thank you, those versions are working for me! \o/

@ned2 that's good advice thanks, it's a workaround for other issues too. I had seen it, just thought I'd check if I really needed to go down that route. For others' reference there is also an open PR to support more flexible JS and CSS inclusion: https://github.com/plotly/dash/pull/171

Issue #198, opened 2018-01-15
static_folder param does not work with default name param


ned2:
When trying to set Dash's `static_folder` param, I found that this initialisation doesn't work, resulting in 404's for my static assets:

`app = Dash(static_folder='static')`

This however does work:

`Dash(name=__name__, static_folder='static')`

The problem is the default value for `name` is `dash`, which then get's passed into Flask's `name` param. This breaks as `dash` is the name of an importable package. Changing `name` to any other installed package (eg `requests`, `pandas`) also breaks, but changing it to some random non importable name works.

Perhaps this should be fixed by having Dash create a Flask instance like this `Flask(__name__, static_folder=static_folder)`. I've been using `__name__` when passing in my own Flask server and according to the [Flask docs](http://flask.pocoo.org/docs/0.12/api/), this is always correct for a single module. When creating a package, user's can supply their own Flask server (which they'll likely be doing anyway for a complex project) and modify accordingly. Which, I think, would makes Dash's `name` redundant.  

Additionally, I think the default value for static_folder should mirror that of [Flasks](http://flask.pocoo.org/docs/0.12/api/). Currently a Flask instance created with all default params and passed into Dash's server, behaves differently (in this respect) to the server that Dash creates by default. I think that Dash's default should mirror Flask's, for the sake of minimising surprising behaviour.

ned2:
A little bit more info on this. It looks like the Flask `static_folder` param can either be an absolute path or a relative one. If a relative one is provided the final `server.static_folder` value is derived as the path of the module that the `name` param resolves to `static_folder` concatenated to the end. If the provided `name` param is not an importable module, then the base path is taken to be the current directory the script is running from  This explains the behavior whereby a name of a non importable module was working, as I was running the script from the directory that had a `static` folder in it. 

So the invocation `app = Dash(static_folder='static')` fails due to `name` in the Dash `class` defaulting to 'dash'. Ideally we would want to provide a default of the value of `__name__` from the module/file that the Dash class was created in, however I'm not really sure if that can be done other than by the user passing in __name__ explicitly each time eg `Dash(name=__name__, static_folder='static')`.

One option could be to make name a mandatory parameter of the Dash class, mirroring the Flask interface, but this would break existing apps and add complexity for those not wanting to add static assets.

[...]

PR #238, opened 2018-03-31
Enhance default support for custom static assets

ned2:
This is a proposal for adding better support for custom static assets to that addresses a couple of current friction points.

I've noticed on the Dash forums that a question that keeps arising is how to add custom CSS assets. Besides from adding them to the app, they also need to be configured to be served by Flask. Currently, the way to set this up with is not particularly intuitive and there is at least one potential gotcha. The most straightforward way to make Flask serve static assets is for users to create a folder such as 'static' in the same directory as their Dash app, and then initialize their app like this:

`app = dash.Dash(__name__, static_folder='static')`

As discussed in #198, confusingly, this will not work if users omit the initial `name` param, or supply an incorrect value for it. This is because this param becomes the `import_name` param to the `Flask` instance, which it uses to for deriving the location of your static folder. So the default value of `dash` is not useful here (however this does mean you could serve a static folder from your installed Dash package in eg site-packages).

My proposal is to firstly make the default value of `static_folder` to be `static`, which brings it inline with Flask's default and then secondly make the default value of the `name` param be `__main__`, which is the value it will take when a script is being run from the command line, which I suspect corresponds to most frequently used layout of a Dash app. Most users will then simply be able to create a `static` directory in the same directory as their app and static assets will served (although they still need to be added to the app itself). For users with more complex apps they will need to ensure they specify the name param correctly. The documentation can also be enhanced to make the significance of the `name` param to the `static_folder` clear.

Perhaps there is an argument to only change the static_folder param, however by making name optional and providing a default value, there will inevitably be some contexts where the value is incorrect for the Flask `import_name`. So I guess my suggestion is to choose a default that is more likely to be right most of the time for users.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -26,9 +26,9 @@ from ._utils import AttributeDict as _AttributeDict
 class Dash(object):
     def __init__(
             self,
-            name=None,
+            name='__main__',
             server=None,
-            static_folder=None,
+            static_folder='static',
             url_base_pathname='/',
             compress=True,
             **kwargs):


chriddyp:
Yup, I like this. I think there is some strength in having all of these things be unified rather than configurable: "it should just work".

Originally I thought that having this `static` folder was implicitly a security issue: what if people aren't aware that that everything in their static folder is being served? What if they _happen_ to have a folder that's named static full of confidential assets?

However, now I'm convinced that the benefits outweigh the security issue:
- `static` isn't a common folder name outside of web development
- we can document this in the user guide and they can easily remove it

chriddyp:
Anything else that you'd like to add @ned2 ? Otherwise I'll merge and release 🍻

ned2:
Great!

Also, oops -- this actually had two commits from another PR (an object lesson in why to not push from master). Just rebased to get rid of these. I think it's right to go in now.

PR #243, opened 2018-04-11
Updated index method

mmphego:
Add additional members on index, ie update/change html metadata


chriddyp:
Thanks for the PR @mmphego ! I'm going to close this in favor of a new, more abstract proposal #265

PR #252, opened 2018-05-02
setuptools installation fixes

moorepants:
- Added dash_renderer as a dependency in setup.py.
- Added the LICENSE to MANIFEST.in.

Came across these when packaging for conda forge: https://github.com/conda-forge/dash-feedstock


chriddyp:
💃 Thanks!

Issue #269, opened 2018-06-11
JSON serializable errors are hard to debug


chriddyp:
If the json serializer fails, the traceback doesn't indicate which object couldn't be serialized or where to find it. For large apps, this becomes really difficult to track down. For example:
```
Traceback (most recent call last):
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/chriddyp/Repos/dash-stuff/my-project/venv/lib/python3.6/site-packages/dash/dash.py", line 556, in dispatch
    return self.callback_map[target_id]['callback'](*args)
[...]

PR #273, opened 2018-06-17
Give a more informative error for JSON not serializable. (#269)

rmarren1:
Fix for #269 
when a non JSON serializable object is returned from a dash callback, the following exception is raised:
```
dash.exceptions.ReturnValueNotJSONSerializable:
Callback for property `<property>` from component id `<id>`
returned a value `<value>` of type `<type>`,
```

an example app:
```
import dash

app = dash.Dash("")

app.layout = html.Div([
    html.Div(id='output'),
    html.Button(id='button')
])


              [Input('button', 'n_clicks')])
def test(n_clicks):
    if n_clicks:
        return dash
    return "HELLO WORLD"


app.run_server()
```
results in:
```
dash.exceptions.ReturnValueNotJSONSerializable:
Callback for property `children` from component id `output`
returned a value `<module 'dash' from '/home/ryan/Dash/dash/dash/__init__.py'>` of type `module`,
```

when the button is clicked (and returns the dash module rather than a value).

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -545,7 +545,7 @@ class Dash(object):
                                 bad_val=child,
                                 outer_val=val,
                                 bad_type=type(child).__name__,
-                                path=p + "\n" + type(child).__name__,
+                                path=p + "\n-   " + type(child).__name__,
                                 index=index
                             )
 
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -617,7 +617,6 @@ class Dash(object):
             def add_context(*args, **kwargs):
 
                 output_value = func(*args, **kwargs)
-                self._validate_callback_output(output_value, output)
                 response = {
                     'response': {
                         'props': {
@@ -626,9 +625,16 @@ class Dash(object):
                     }
                 }
 
+                try:
+                    jsonResponse = json.dumps(
+                        response,
+                        cls=plotly.utils.PlotlyJSONEncoder
+                    )
+                except TypeError:
+                    self._validate_callback_output(output_value, output)
+
                 return flask.Response(
-                    json.dumps(response,
-                               cls=plotly.utils.PlotlyJSONEncoder),
+                    jsonResponse,
                     mimetype='application/json'
                 )
 
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -475,6 +475,62 @@ class Dash(object):
                 output.component_id,
                 output.component_property).replace('    ', ''))
 
+    def _validate_callback_output(self, output_value, output):
+        valid = [str, dict, int, float, type(None), Component]
+
+        def _raise_invalid(bad_val, outer_type, bad_type, path, index=None):
+            raise exceptions.ReturnValueNotJSONSerializable('''
+            The callback for property `{:s}` of component `{:s}`
+            returned a tree with one value having type `{:s}`
+            which is not JSON serializable.
+
+            The value in question is located at
+
+            `{:s}`
+
+            and has string representation
+
+            `{}`.
+
+            In general, Dash properties can only be
+            dash components, strings, dictionaries, numbers, None,
+            or lists of those.
+            '''.format(
+                output.component_property,
+                output.component_id,
+                bad_type,
+                (
+                    "outer list index {:d} ({:s}) -> "
+                    .format(index, outer_type)
+                    if index is not None
+                    else (outer_type + " -> ")
+                ) + path,
+                bad_val).replace('    ', ''))
+
+        def _value_is_valid(val):
+            return (
+                # pylint: disable=unused-variable
+                any([isinstance(val, x) for x in valid]) or
+                type(val).__name__ == 'unicode'
+            )
+
+        def _validate_value(val, index=None):
+            if isinstance(val, Component):
+                for p, j in val.traverse_with_paths():
+                    if not _value_is_valid(j):
+                        _raise_invalid(j, type(val).__name__, type(j).__name__,
+                                       p, index)
+            else:
+                if not _value_is_valid(val):
+                    _raise_invalid(val, type(val).__name__, type(val).__name__,
+                                   '', index)
+
+        if isinstance(output_value, list):
+            for i, val in enumerate(output_value):
+                _validate_value(val, index=i)
+        else:
+            _validate_value(output_value)
+
     # TODO - Update nomenclature.
     # "Parents" and "Children" should refer to the DOM tree
     # and not the dependency tree.
@@ -513,6 +569,7 @@ class Dash(object):
             def add_context(*args, **kwargs):
 
                 output_value = func(*args, **kwargs)
+                self._validate_callback_output(output_value, output)
                 response = {
                     'response': {
                         'props': {
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -478,34 +478,41 @@ class Dash(object):
     def _validate_callback_output(self, output_value, output):
         valid = [str, dict, int, float, type(None), Component]
 
-        def _raise_invalid(bad_val, outer_type, bad_type, path, index=None):
+        def _raise_invalid(bad_val, outer_val, bad_type, path, index=None,
+                           toplevel=False):
+            outer_id = "id={:s}".format(outer_val.id) \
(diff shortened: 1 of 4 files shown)


chriddyp:
Very nice! One case that I'm interested in is how we can deal with objects that have a non-JSON-serializable object deep inside an object, for example:
```
html.Div([
    html.Div(lambda: 'test')
])
```

With this current exception, it would say that "Div is not JSON serializable" which could be misleading. In an ideal case, we would be able to say something like:
```
The property "children" of a component "Div" is a "function" which is not serializable. This Div is located at:

Div.children[0].children

In general, Dash properties can only be dash components, strings, dictionaries, numbers, None, or lists of those.
```

Or, if the particular parent object was assigned an ID, like
```
html.Div([
    html.Div(lambda: 'test', id='some-div')
])
```
then our error message could be like:
```
The property "children" on the element "some-div" is not JSON serializable.
```

Now, I'm not actually sure the best way to go about doing this. One way might be to subclass the `plotly.utils.PlotlyJSONEncoder`. That class tries to encode an object a bunch of different ways. If it fails all of the ways, it raises an error:
https://github.com/plotly/plotly.py/blob/519e0ab32b2164583f78b51168ce3412c08334e5/plotly/utils.py#L213-L229

It's been a few years since I worked with that class, but I believe that `json.dumps` traverses the object that you give it and calls `cls.default` with the nested object
https://github.com/plotly/plotly.py/blob/519e0ab32b2164583f78b51168ce3412c08334e5/plotly/utils.py#L182

So perhaps we could raise a nicer exception within there somehow. I would be OK with our own JSON encoder class that uses that `plotly.utils.PlotlyJSONEncoder`.

Alternatively, we might be able to traverse the tree ourselves. If the object isn't serializable, then we could walk down the component tree until we find the particular property that isn't serializable. I actually wrote a method that traverses the tree: `traverse`:
https://github.com/plotly/dash/blob/f309cc2b8a45644004db98ffcaf0d9a72870c27d/dash/development/base_component.py#L152
which I believe you can call with something like:
```
for object in layout:
    print(object)
```

However, I don't think that really gives us the _path_ of the component, it would only give us each individual component in the tree.

_Another_ alternative would be to just raise an exception for the particular property and print the entire component. The user might be able to figure out _which_ component was having the issue if they say both the property that wasn't JSON serializable _and_ if the saw the rest of the properties that weren't serializable. For example, if the issue was something like:
```
html.Div([
   html.Div(style={'color': 'blue'}, children=lambda: 'test')
])
```

Then the error message would be like:
```
The property `children` in the following component is not serializable:
Div(children=<function <lambda> at 0x109181840>, style={'color': 'blue'})
```

I believe this would be relatively easy with the `traverse` method or even with subclassing the `plotly.utils.PlotlyJSONEncoder` and raising a `dash.exceptions` instead of the default `TypeError` 

Oddly, I can't seem to find any other solution out there.

chriddyp:
😻  that looks SO good @rmarren1 ! This seems like the winning solution.

@plotly/dash - Anyone else like to review?

nicolaskruchten:
Love this! I would recommend putting each path element on its own line rather than separating them with ` -> ` as those lines will get looong :)

We could also structure the lines a bit differently like:

```
The value in question is located at

Div
[2]: Span
[1]: Div: list index 1
[1]: P: list index 1
[0]: function
```

or something? Maybe add in the `id` of those elements if present?

rmarren1:
It turns out that `traverse` doesn't yield leaves of the tree which are not of type `Component` or `collections.MutableSequence`, which complicates things a bit since an exception will not be thrown for return values like
```
html.Div(
    html.Div(lambda: 'hi')
)
```
since the `lambda` is not wrapped in a list. (Is this the expected behavior of `traverse`? When I change it to return such elements, test cases fail.)

My quick patch was to check each value in the traversal, and if is a `Component` and has a child not of type `collections.MutableSequence` we validate the child.

Also, I changed the output format (thanks @nicolaskruchten). Outputs now read
```
The callback for property `children` of component `output`
returned a tree with one value having type `function`

The value in question is located at
-   Div (id=top-div)
-   Div (id=list-div)
[2] Span 
-   function

and has string representation
`<function change_button.<locals>.<lambda> at 0x7fc67896d6a8>`

or lists of those.

```

I also added special cases for when the bad value is the only value returned or is a list (previously, the error message would talk about trees and print out a tree where there is only one element, this could be confusing). These error messages look like this now:

```
The callback for property `children` of component `output`
returned a value having type `function`

The value in question is either the only value returned,
or is in the top level of the returned list,
and has string representation
`<function change_button.<locals>.<lambda> at 0x7fb2f3ca66a8>`

or lists of those.

```

nicolaskruchten:
Nice! I would favour the `[index]` for every element in the list, even when the id is present, so that you could mechanically just walk to the right place in the tree if you wanted do :)

rmarren1:
I think that was just a coincidence in the example I used, which was something like this:
```
html.Div(
    id='top-div',
    children=html.Div(
        id='list-div',
        children=['hi', 'hello', Span(lambda: 'hi')]
    )
)
```
If the `id` is present it should be printed no matter if the component is a single child or a member of a list: https://github.com/plotly/dash/pull/273/files#diff-228c9aefac729977642672aa1416bacaR511.

I will double check this

nicolaskruchten:
Right, so what I'm saying is that even if elements are singletons, I think we should show `[0]` rather than `-`.

nicolaskruchten:
Ah, I see what the problem is. In that case I might suggest `[*]` just for nicer visual effect scanning down :)

rmarren1:
Error paths now read like this:
```
[*] Div (id=top-div)
[*] Div (id=list-div)
[2] Span 
[*] function
```
looks better imo, thanks @nicolaskruchten

rmarren1:
Is this good to merge?

dash/dash.py
```diff
@@ -475,6 +475,110 @@ def _validate_callback(self, output, inputs, state, events):
                 output.component_id,
                 output.component_property).replace('    ', ''))
 
+    def _validate_callback_output(self, output_value, output):
+        valid = [str, dict, int, float, type(None), Component]
```
chriddyp: Since we JSON serialize with the `json.dumps(obj, cls=plotly.utils.PlotlyJSONEncoder)`, this list actually has a few other items: https://github.com/plotly/plotly.py/blob/6b3a0135977b92b3f7e0be2a1e8b418d995c70e3/plotly/utils.py#L137-L332

So, if there was a component property that accepted a list of numbers, then technically the user could return a numpy array or a dataframe.

The only example that immediately comes to mind is updating the `figure` property in a callback with the `plotly.graph_objs`. These objects get serialized because they have a `to_plotly_json` method.

So, I wonder if instead of validating up-front, we should only run this routine only if our `json.dumps(resp, plotly.utils.PlotlyJSONEncoder)` call fails.

This would have the other advantage of being faster for the non-error case. If the user returns a huge object (e.g. some of my callbacks in apps return a 2MB nested component), doing this recursive validation might be slow.

chriddyp:
I have just one more comment about the placement of the validation call. Otherwise, I'm 👍  with the code.

It would be great if another dash dev could review this so that more people are familiar with this code. @T4rk1n , could you take a look as well?

dash/dash.py
rmarren1: That makes sense. I am not sure how large of a speed up that would be since validating a large callback output is likely much faster than the network delay of sending that large output, but it definitely makes it easier than crafting a perfect validator (that would need to update every time we want to extend `PlotlyJSONEncoder`).

dash/exceptions.py
```diff
@@ -44,3 +44,7 @@ class CantHaveMultipleOutputs(CallbackException):
 
 class PreventUpdate(CallbackException):
     pass
+
+
+class ReturnValueNotJSONSerializable(CallbackException):
```
T4rk1n: Maybe just me, but I don't like that exception name, it gives too much info on the cause while not giving the true nature of the error, that is the return value of a callback is invalid. I would change it to `InvalidCallbackReturnValue`.

rmarren1: Makes sense to me, pushed those changes.

Issue #265, opened 2018-06-08
Proposal for Offline CSS and JS and Customizable `index.html`


chriddyp:
Opening this issue to propose the officially endorsed and documented method for embedding custom, local CSS and JavaScript in Dash apps.

> This development for this issue has been sponsored by an organization. Many thanks! 
> If your company or organization would like to sponsor development, [please reach out](https://plot.ly/products/consulting-and-oem).

Background: In Dash, HTML tags and higher-level components are embedded in the App by assigning the `app.layout` property to a nested hierarchy of Dash components (e.g. the components in the `dash-core-components` or the `dash-html-components` library). These components are serialized as JSON and then rendered client-side with React.js. On page load, Dash serves a very minimal HTML string which includes the blank container for rendering the app within, the component library's JavaScript and CSS, and a few meta HTML tags like the page title and the encoding (see https://github.com/plotly/dash/blob/6a1809fc619671f8bf7e1294d12ef1180e90df6b/dash/dash.py#L293-L318).

This architecture doesn't work well for embedding custom JavaScript scripts and CSS Stylesheets because these scripts and stylesheets usually need to be included in the HTML that is served _on page load_.

We will support user-supplied JavaScript and CSS through two enhancements:

**Enhancement 1 - Automatic**

- This "automatic" method will template in all stylesheets (CSS) and scripts (JavaScript) that are included in a `static` folder
- These links will be included in alphabetical order
- These links will be included _after_ the component library's stylesheets and scripts
- The server route (`/static/<path:string>`) for serving these files will be configured automatically by Dash

This method will be what _most_ Dash users will use. It's very easy to use and easy to document ("To include custom CSS, just place your CSS files in `static` folder. Dash will take care of the rest"). Since the files will be templated alphabetically, the user can control the order (if necessary) by prefixing ordered numbers to the files (e.g. `1-loading.css`, `2-app.css`).

With this method, we'll be able to add custom CSS processing middleware like minifying CSS or creating cache-busting URLs completely automatically.

If the user needs more control over the placement of these files, they can use the method in "Enhancement 2 - Manual Override".

**Enhancement 2 - Manual Override**

Currently on page load, Dash serves this HTML string:  https://github.com/plotly/dash/blob/6a1809fc619671f8bf7e1294d12ef1180e90df6b/dash/dash.py#L293-L318.

This enhancement will make this overridable:
- The user will assign a HTML string or a function that returns an HTML string to a property on the `app` object (e.g. `app.index`).
[...]

PR #275, opened 2018-06-21
Improves cross-browser compatibility

amarvin:
Adds meta tags recommended by [Bootstrap 3.3 basic template](https://getbootstrap.com/docs/3.3/getting-started/#template) for cross-browser compatibility. Relates to #170

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -300,6 +300,8 @@ class Dash(object):
         <html>
             <head>
                 <meta charset="UTF-8">
+                <meta http-equiv="X-UA-Compatible" content="IE=edge">
+                <meta name="viewport" content="width=device-width, initial-scale=1">
                 <title>{}</title>
                 {}
             </head>
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -301,7 +301,7 @@ class Dash(object):
             <head>
                 <meta charset="UTF-8">
                 <meta http-equiv="X-UA-Compatible" content="IE=edge">
-                <meta name="viewport" content="width=device-width, initial-scale=1">
+                <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
                 <title>{}</title>
                 {}
             </head>


pope1ni:
Would it be better to use those recommended for Bootstrap v4.1?

https://getbootstrap.com/docs/4.1/getting-started/introduction/#starter-template

amarvin:
I think the Bootstrap 4 template doesn't support older browsers, and assumes HTML5 compatibility. I was hoping that Dash would support old browsers like IE11.

chriddyp:
Here's a reference on this tag: https://stackoverflow.com/a/6771584/4142536

I still need to take some time to thoroughly think this one through (e.g. make sure it's the right move for Dash in the future, make sure that it won't break anything)

pope1ni:
> I was hoping that Dash would support old browsers like IE11.

IE11 does support HTML5. Bootstrap 4 support IE10+...

https://getbootstrap.com/docs/4.1/getting-started/browsers-devices/#supported-browsers

amarvin:
Ah good catch @pope1ni. I guess the real reason that I need the `X-UA-Compatible` tag in IE11 is that my Dash apps are hosted on the Intranet, and IE11 displays Intranet pages in compatibility mode by default (unless overridden by `X-UA-Compatible`. Not sure how common this specific problem is, but I bet a lot of Dash users want to build internal apps on their company's Intranet.

pope1ni:
> IE11 displays Intranet pages in compatibility mode by default

A quick search shows that using meta tags may not solve this problem for you: https://stackoverflow.com/a/9338959

amarvin:
It does work in my case. I overrode `dash.Dash.index()` with a modified version that includes `<meta http-equiv="X-UA-Compatible" content="IE=edge">`, and the Intranet Dash app now renders fine in IE11.

chriddyp:
> Would it be possible to adjust this patch to allow arbitrary of either

See the proposal in https://github.com/plotly/dash/issues/265

dash/dash.py
```diff
@@ -300,6 +300,8 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
         <html>
             <head>
                 <meta charset="UTF-8">
+                <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
billdoors: My dash app doesn't work with IE. Only after I added these two lines, it work with all browsers. It will be nice to include this in the Dash by default.

chriddyp:
I'm still interested in this but I'm afraid I don't have enough information. Could someone help me verify a few things?
- Will adding this tag negatively effect any Dash apps on any browsers or is it safe to add to our default index? For example, will this break some of our user's apps that are embedding a certain version of bootstrap?
- When we say that Dash "Doesn't work in IE11", what in particular isn't working? When I render the app in browserstack on IE11, Dash renders for me:
image
- In particular, is it that you are appending external CSS stylesheets that need the tag? i.e. bootstrap 3
- Or is there JavaScript functions that are failing? If its a JavaScript issue, is the real root of the problem that we are not embedding some polyfill that we need to be?
- For those IE users, are you able to visit the dash docs? If so, could you find another public Dash app that doesn't render correctly that I can use for testing?

pope1ni:
> I'm still interested in this but I'm afraid I don't have enough information.

So see https://github.com/plotly/dash/pull/275#issuecomment-399995235 and my response https://github.com/plotly/dash/pull/275#issuecomment-400003563 above for the reasoning. Dash works fine in IE11, but apparently in certain configurations, e.g. when hosted on an intranet, IE11 will default to one of it's compatibility modes which causes it to behave like an older version of IE. I don't know what compatibility version was used for @amarvin's intranet, but it sounds like Dash doesn't work with it (and indeed Dash should not make any efforts to support IE<11 as Microsoft no longer support it).

Setting the `X-UA-Compatible` HTTP header to `IE=edge` will ensure that each version of IE which supports the header will render and behave at the current version, i.e. disabling compatibility modes. In the case of IE11, it will be forced to behave like IE11. This is often done in-page using `<meta>` tags.

> Will adding this tag negatively effect any Dash apps on any browsers or is it safe to add to our default index?

The `X-UA-Compatible` header will only affect IE. Microsoft named it in such a way that it *could* be used by other vendors. (Once upon a time Google had Chrome Frame which was a plugin for IE that sort of embedded Chrome. It also used the header so that developers could force their users to use the Chrome Frame plugin for their site.)

> For example, will this break some of our user's apps that are embedding a certain version of bootstrap?

The `X-UA-Compatible` header is unlikely to break anything. It is more likely that things are already broken as in @amarvin's case.

The `<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">` tag is more likely to break things for existing users, but is probably quite a sensible default. There are two options here - add it by default and document the change - or don't add it and allow people to add it easily. This is now possible with #286 merged.

> Or is there JavaScript functions that are failing? If its a JavaScript issue, is the real root of the problem that we are not embedding some polyfill that we need to be?

This, I imagine, was the problem, but only because of IE11 being in compatibility mode and behaving as an older version of IE. You could actually simulate the problem by adding something like `<meta http-equiv="X-UA-Compatible" content="IE=8">` and testing again.

I hope that this clears things up for you @chriddyp.

chriddyp:
Alright, I think I'm convinced. @T4rk1n - could you wire this in? It looks like this tag needs to be the _first_ meta tag (see https://stackoverflow.com/a/9624500/4142536). And we should allow users to override this tag with meta_tags, in a similar way to how we allow users to override the default encoding.

Many thanks everyone for the feedback and discussion ❤️

T4rk1n:
Alright I'll add this.

T4rk1n:
Implemented in #316

drRussClay:
I'm using version 0.21.0, and since this fix is pretty new, I'm guessing I don't have it included.  I will update and test to see if it fixes the problem.  Thanks!

PR #286, opened 2018-07-11
Assets files & index customizations

T4rk1n:
## Assets includes & index customization

Solution for #265 Proposal for Offline CSS and JS and Customizable `index.html`

#### Assets include

Dash will now look for a folder named `assets` on the current work directory, if one is found, it will walk that directory and include js and css files on the index.

Related configs:

```python
import dash

app = dash.Dash()

# default values
app.config.assets_folder = 'assets'     # The path to the assets folder.
app.config.include_asset_files = True   # Include the files in the asset folder
app.config.assets_external_path = ''    # The external prefix if serve_locally == False
app.config.assets_url_path = '/assets'  # the local url prefix ie `/assets/*.js`
```

Supported files:

- `.js`, javascript files will be included after the components libs and before the dash-render.
- `.css`, stylesheets will be included as a `<link>` in the head.
- `favicon.ico`, include as a `<link>` in the head.

The files are included in alphabetic order.
The directories can be nested.

Advanced use:

To host your assets content externally:

- set `app.scripts.config.serve_locally = False`
- `app.config.assets_external` to your base host url, ie `http://bucket.s3.amazonaws.com/`
- `app.config.include_asset_files` must still be set to `True` for the files to be indexed by dash.
- Duplicate the file structure in your assets folder to your file hoster and the files will be loaded from there instead. 

##### Meta tags

It is now possible to add meta tags to the index of dash.

Example:
```python
import dash

metas = [
    {'name': 'description', 'content': 'My description'}
]

app = dash.Dash(meta_tags=metas)

# alternatively
app.add_meta_tag({'http-equiv': 'X-UA-Compatible', 'content': 'IE=edge'})
```

##### Customizing the index

Add an `index_string` to change the default index dash use.

```python
import dash

app = dash.Dash()
app.index_string = '''
<!DOCTYPE html>
<html>
    <head>
        {%metas%}
        <title>{%title%}</title>
        {%favicon%}
        {%css%}
    </head>
    <body>
        <div>My Custom header</div>
        {%app_entry%}
        <footer>
            {%config%}
            {%scripts%}
        </footer>
        <div>My Custom footer</div>
    </body>
</html>
'''
```

The `{%key%}`s will be formatted like in the default index.

Available keys:
```
{%metas%}        # optional - The registered meta tags.
{%favicon%}      # optional - A favicon link tag if found in `assets`.
{%css%}          # optional - link tags to css resources.
{%config%}       # required - Config generated by dash for the renderer.
{%app_entry%}    # required - The container where dash react components are rendered.
{%scripts%}      # required - Collected dependencies scripts tags.
```

Also added `interpolate_index` method on Dash, override it to get the context values of the index before rendering.

```python
import dash

class CustomDash(dash.Dash):
    def interpolate_index(self,
                          metas, title, css, config,
                          scripts, app_entry, favicon):
        return '''
        <!DOCTYPE html>
        <html>
            <head>
                <title>My App</title>
            </head>
            <body>
                
                <div id="custom-header">My custom header</div>
                {}
                {}
                <div id="custom-footer">My custom footer</div>
            </body>
        </html>
        '''.format(app_entry, scripts)
```

I added tests for the meta keys and index customization. For the assets, I made a test in an external project to test if the files are included in that project, @plotly/dash I need help to integrate that test to run with the other tests.


dash/dash.py
```diff
@@ -294,28 +357,27 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
...
-                </footer>
-            </body>
-        </html>
-        '''.format(title, css, config, scripts)
+        if self._favicon:
+            favicon = '<link rel="icon" type="image/x-icon" href="{}">'.format(
+                flask.url_for('assets.static', filename=self._favicon))
+        else:
+            favicon = ''
+        return self.interpolate_index(
```
nicolaskruchten: 👍

dash/_utils.py
```diff
@@ -1,3 +1,11 @@
+def interpolate_str(template, **data):
+    s = template
+    for k, v in data.items():
+        key = '{' + k + '}'
```
nicolaskruchten: This is probably too brittle... in modern Javascript you could easily write something like `setConfig({config})` and have someone paste that into `index`

dash/dash.py
```diff
@@ -294,28 +357,27 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+            metas, title, css, config, scripts, _app_entry, favicon)
+
+    def interpolate_index(self,
+                          metas, title, css, config,
+                          scripts, app_entry, favicon):
+        return _interpolate(self.index_string,
```
nicolaskruchten: There's probably not enough error-checking here... Shouldn't we raise a very helpful and clear message if they forget to add `{scripts}` or something?

dash/dash.py
```diff
@@ -558,11 +620,56 @@ def dispatch(self):
...
+    def add_meta_tag(self, meta):
+        self._meta_tags.append(meta)
+
     def run_server(self,
                    port=8050,
                    debug=False,
                    **flask_run_options):
+        bp = flask.Blueprint('assets', 'assets',
+                             static_folder=self.config.assets_folder,
+                             static_url_path=self.config.assets_url_path)
```
chriddyp: Do we need to ensure a minimum version of flask for this? i.e. >1.0?

T4rk1n: It was add in 0.7, we have `Flask>=0.12`.

dash/dash.py
```diff
@@ -260,6 +309,20 @@ def _generate_config_html(self):
             '</script>'
         ).format(json.dumps(self._config()))
 
+    def _generate_meta_html(self):
+        has_charset = any('charset' in x for x in self._meta_tags)
+
+        tags = []
+        if not has_charset:
+            tags.append('<meta charset="UTF-8"/>')
```
chriddyp: This is very thoughtful 👍

dash/dash.py
```diff
@@ -558,11 +620,56 @@ def dispatch(self):
...
+        for current, _, files in os.walk(walk_dir):
+            if current == walk_dir:
+                base = ''
+            else:
+                s = current.replace(walk_dir, '').lstrip('\\').lstrip('/')
+                splitted = slash_splitter.split(s)
+                if len(splitted) > 1:
+                    base = '/'.join(slash_splitter.split(s))
+                else:
+                    base = splitted[0]
```
chriddyp: Could we use `base = os.path.split(walk_dir)[0]`?

chriddyp: NVM, I get it now, we have to replace `\` with `/` for URL friendly paths.

dash/dash.py
T4rk1n: Two options I thought of:
- check  in a property setter on the index_string.
  - fast only one check, but no checking on interpolate_index.
  - raise when set.
- check after interpolate_index the string contains the ids of required elements.
  - check every time the index render.
  - raise only when browsing.

Which would be best ?

dash/dash.py
```diff
@@ -19,18 +22,51 @@
...
+        {metas}
+        <title>{title}</title>
+        {favicon}
+        {css}
+    </head>
+    <body>
+        {app_entry}
+        <footer>
+            {config}
+            {scripts}
```
chriddyp: Could we split out scripts into `dash_renderer_scripts` and `dash_components_scripts`? The main reason is that `dash_renderer` is going to become configurable in when we add in custom JS hooks. So, users will end up needing to do something like:
```
dash_renderer = DashRenderer({
    request_hook: function(...) {
         ...
    }
})
```

And so it'd be nice if they could replace this:
```
<!DOCTYPE html>
<html>
    <head>
        {metas}
        <title>{title}</title>
        {favicon}
        {css}
    </head>
    <body>
        {app_entry}
        <footer>
            {config}
            {dash_renderer}
            {dash_component_scripts}
        </footer>
    </body>
</html>
```

with something like this:
```
<!DOCTYPE html>
<html>
    <head>
        {metas}
        <title>{title}</title>
        {favicon}
        {css}
    </head>
    <body>
        {app_entry}
        <footer>
            {config}
            dash_renderer = DashRenderer({
                request_hook: function(...) {
                     ...
                }
            })
            {dash_component_scripts}
        </footer>
    </body>
</html>
```

Now, I suppose the other way they could do this would be with the interpolated index function, where they would do something like:
```
class CustomDash(dash):
    def interpolated_index(metas, title, css, config, scripts, _app_entry, favicon):
        filtered_scripts = [
            script for script in scripts if 'dash_renderer' not in script
        ]

        return '''
        <!DOCTYPE html>
        <html>
            <head>
                {metas}
                <title>{title}</title>
                {favicon}
                {css}
            </head>
            <body>
                {app_entry}
                <footer>
                    {config}
                    dash_renderer = DashRenderer({
                        request_hook: function(...) {
                             ...
                        }
                    })
                    {filtered_scripts}
                </footer>
            </body>
        </html>
        '''.format(metas, title, css, config, filtered_scripts, _app_entry, favicon)
```

Am I understanding that correctly?

chriddyp:
Overall this is looking really good! A few topics of discussion below. Would love feedback @plotly/dash

***

```
import dash

metas = [
    {'name': 'description', 'content': 'My description'}
]

app = dash.Dash(meta_tags=metas)
```

I actually like this declarative syntax. This seems like a nice pattern that we could use elsewhere, like as a way to replace `app.scripts.append_script`:
```
app = dash.Dash(
     js_script_urls=['https://cdn.google.com/google-analytics.js'],
     css_stylesheet_urls=['https://cdn.bootstrap.com/bootstrap.css']
)
```

I never really liked the `app.scripts.append_script` syntax and I'd generally prefer for the `dash.Dash()` object to have as few methods and properties as possible. If everything could just be in the `app.config` and settable through the `dash.Dash()` constructor, I'd be very happy.

***

> `app.add_meta_tag({'http-equiv': 'X-UA-Compatible', 'content': 'IE=edge'})`

I'm less fond of this syntax as I prefer if there would only be a single declarative way to do things. Now, I'm biased as I've been in the declarative-or-nothing plotly world for several years, so I'd be happy to hear other perspectives :). Anyway, here are my thoughts:
- For documentation and answering questions in the community forum, it's easier to have a single way to explain how to do things. In particular, one scenario is when users aren't using the API correctly: if there is more than one way to do things, they'll try each way (incorrectly) and get frustrated that none of them work. They aren't sure if the methods are truly equivalent or if they need to use one method instead of the others.
- I prefer to keep the number of methods on `dash.Dash()` small. That way, we only have to document the `dash.Dash` constructor and users can do everything they need by calling `help(dash.Dash)`.
- If we have `add_meta_tag`, then for consistency I'd expect there to be equivalent `add_other_thing` whenever we add other resources. For example, `scripts`, or `css`. Which we could do, but it's one more thing to keep track of.
- Similarly, for some of the attributes, the `add_*` method doesn't make sense. For example, the `title` attribute: `add_title` doesn't really make sense because there can only be a single title. However, this becomes inconsistent with `meta` because `meta` can be set with two different ways (`add_meta` and the `dash.Dash`) but other properties in `dash.Dash` can only be set one way.
- If they have `app.add_meta_tag` in a different file that maybe isn't getting imported correctly, then they might think that `app.add_meta_tag` isn't getting called. If we force users to use `app.Dash(meta_tags=[...])`, then it should always work and they don't need to worry about the location of their other function calls. This is sort of a general argument about why I prefer declarative configurations to imperative function calls (declarative attributes are order and placement independent)
- Similarly,  I think we encourage good code organization by forcing users to put the `meta_tags` inside the constructor, near where they are configuring everything else about their Dash app, rather than giving them control over where they want to place it.

Would love feedback on this @plotly/dash and community. Are these valid arguments or am I being too pedantic?

dash/dash.py
T4rk1n: I wasn't too sure about the js hooks so I left them out of this, right now the scripts are all bundled together in _generate_scripts_html that return a string with all the scripts tags, can't iterate over it but it could change. 

I thought we could have a config to disable the includes of dash-renderer. Then the user can append a custom renderer to the scripts resources or include it a custom index.

dash/dash.py
```diff
@@ -217,8 +262,12 @@ def _relative_url_path(relative_package_path='', namespace=''):
                         srcs.append(url)
             elif 'absolute_path' in resource:
                 raise Exception(
-                    'Serving files form absolute_path isn\'t supported yet'
+                    'Serving files from absolute_path isn\'t supported yet'
                 )
+            elif 'asset_path' in resource:
+                static_url = flask.url_for('assets.static',
+                                           filename=resource['asset_path'])
+                srcs.append(static_url)
```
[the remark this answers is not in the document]
T4rk1n: I did cache busting before with webpack and an extension. Was basically a template render that replaced the hash part of a filename with a new one. 

In dash, I think it would be kinda hard to do that, but we could have a watcher on the asset folder, copy those assets with a filename including the hash to a temp static folder, keep the hash in a dict with the path as key and serve the file with the hash formatted in. When a file change, copy it to the temp folder with a new hash and put the new hash as the value for the path. Could also tell the browser to reload while we're at it.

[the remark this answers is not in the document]
chriddyp: Yeah that's a good idea. I'm a little worried about introducing a temp folder, I feel like users won't know why it's there or if they should commit it, etc.

Instead of having a watcher, could we just call this function on every page load (while in `dev` mode) and get the latest timestamp of the file? And then if we're not in `dev` mode, we could store the timestamps on the first page load?

dash/dash.py
nicolaskruchten: I would actually do both :) That way it fails fast for those using `index` but it still checks for those using `interpolate_index` ... better developer experience all around

dash/_utils.py
T4rk1n: Think doubling the brackets `{{config}}` would be enough or maybe add a character before `%{config}` ?

nicolaskruchten: doubling is probably not enough... I would do something like `{%dash_config%}` or something personally just to make it super clear

T4rk1n:
> Could you include an assets folder inside the tests folder?

Thanks that worked.

valentijnnieman:
🎉 Looks very cool! Good work! :) @chriddyp I agree that a single way of doing things is probably better - I don't really see when methods like `append_*` could be useful, and I think the more we get users to see the benefits of reactive style programming (or declarative) the better it is for Dash!

dash/dash.py
nicolaskruchten: If they're customizing the `DashRenderer` don't they need to have the tags to load the uncustomized one first?

dash/dash.py
```diff
@@ -558,11 +668,56 @@ def dispatch(self):
...
+        self._meta_tags.append(meta)
+
     def run_server(self,
                    port=8050,
                    debug=False,
                    **flask_run_options):
+        bp = flask.Blueprint('assets', 'assets',
+                             static_folder=self.config.assets_folder,
+                             static_url_path=self.config.assets_url_path)
+        self.server.register_blueprint(bp)
```
ned2: I don't know much about Flask Blueprints, but what about users who are using a WSGI server like Gunicorn rather than the builtin server. Would this setup mean that they won't get the assets added?

dash/dash.py
```diff
@@ -43,19 +87,34 @@ def __init__(
                 ''', DeprecationWarning)
 
         name = name or 'dash'
+
+        self.assets_folder = assets_folder or os.path.join(
+            os.getcwd(), 'assets'
```
ned2: I'm a little worried that the use of `os.getcwd()` could be a little brittle. The `name` param to the Flask instance is mandatory so that Flask can derive the root for place the relative `static_folder`. If we use getcwd and Flask is always placing its static folder relative to the user's package/module then there is the possibility for the root of our assets location and the root of Flask's static folder locations to become out of synch (eg with `os.chdir()`). Does that make any sense?

If this is a real issue I'm wondering if it would be better to tap into however Flask derives the root location from the `name` param?

dash/dash.py
T4rk1n: It will work, wsgi just take the flask instance as the wsgi app.

dash/dash.py
T4rk1n: I investigated what flask does, it get the root directory from `flask.helpers.get_root_path`. It first try to get it from `sys.modules`, if it's not found, it will do `os.getcwd` as a fallback. That is the reason it takes a required arg and generally you do `app = Flask(__name__)`, this will give the  directory of the file that contain the app instance in every case. 

But dash do `name='__main__'` by default, which can result in unexpected behavior from what I've tested. If the app was defined on the root folder, but it uses a script in a sub folder to start the server, it will have the wrong directory. Or if you run the app with the command `python -m flask run`, it will be `/venv/lib/site-packages/flask`. I think it also applies to wsgi runned instances, the `__main__` file won't be the app file but the wsgi runner. 

We could use `flask.helpers.get_root_path` instead of `os.getcwd`, but I think we should require the `__name__` of the file like flask does to ensure consistency, but it would break compatibility with older version as now we would have a required arg in the dash constructor.

T4rk1n:
I changed the key formatting to `{%key%}` like @nicolaskruchten suggested and added checks on the index_string setter and after interpolate_index.

Also, I got a problem with the tests, it fails at pylint and it look like it's not getting the config from the `.pylintrc` file. When I run pylint locally I get 10/10. Anyone got a clue on how to fix that ?

chriddyp:
> Also, I got a problem with the tests, it fails at pylint and it look like it's not getting the config from the .pylintrc file. When I run pylint locally I get 10/10. Anyone got a clue on how to fix that ?

i’ve seen this happen in other projects as well. unfortunately, i never was able to figure out why. i ended up just make the fixes that circle reported

chriddyp:
> Also, I got a problem with the tests, it fails at pylint and it look like it's not getting the config from the .pylintrc file. 

Looks like @rmarren1 was having the same issues in https://github.com/plotly/dash/pull/294. We're not locking down the version of `pylint` in `dev-requirements.txt`, so I bet that they just published a new version and that's getting installed on circle:
https://github.com/plotly/dash/blob/f309cc2b8a45644004db98ffcaf0d9a72870c27d/dev-requirements.txt#L14-L15

chriddyp:
Yup, pylint 2.0 was released on Sunday: http://pylint.pycqa.org/en/latest/whatsnew/changelog.html

T4rk1n:
> Looks like @rmarren1 was having the same issues in #294. We're not locking down the version of pylint in dev-requirements.txt, so I bet that they just published a new version and that's getting installed on circle:

Locked pylint to 1.9.2 and it solved the issue.

dash/dash.py
T4rk1n: Oh you'r right, thanks for bringing it up, I'll move it.

chriddyp:
@ned2 - Could you give this one another review? All of the requested changes have been made, and I'd appreciate another set of eyes on it.

dash/dash.py
chriddyp: In a subsequent PR, maybe we can make `name` a keyword argument and then include usage in the docs?

dash/dash.py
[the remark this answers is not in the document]
chriddyp: Cool, sounds like there are a couple of options. So, let's create a new GitHub issue about this and tackle it in a subsequent PR

dash/dash.py
```diff
@@ -29,8 +68,13 @@ def __init__(
...
             server=None,
             static_folder='static',
+            assets_folder=None,
+            assets_url_path='/assets',
+            include_assets_files=True,
             url_base_pathname='/',
             compress=True,
+            meta_tags=None,
+            index_string=_default_index,
             **kwargs):
```
chriddyp: Let's also create an issue about adding a `css_urls` and `js_script_urls` arguments to this init argument as an alternative to our `app.css.append_css` and `app.scripts.append_script`:

> I actually like this declarative syntax. This seems like a nice pattern that we could use elsewhere, like as a way to replace app.scripts.append_script:

app = dash.Dash(
     js_script_urls=['https://cdn.google.com/google-analytics.js'],
     css_stylesheet_urls=['https://cdn.bootstrap.com/bootstrap.css']
)

> I never really liked the app.scripts.append_script syntax and I'd generally prefer for the dash.Dash() object to have as few methods and properties as possible. If everything could just be in the app.config and settable through the dash.Dash() constructor, I'd be very happy.

(from comment https://github.com/plotly/dash/pull/286/#issuecomment-404281839)

dash/dash.py
```diff
@@ -294,28 +386,47 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+                    ids=', '.join(missing),
+                    pl=plural
+                )
+            )
+
+        return index
+
+    def interpolate_index(self,
+                          metas='', title='', css='', config='',
+                          scripts='', app_entry='', favicon=''):
```
chriddyp: Let's include a quick docstring for our users. Something like:
> A function that is called to create the initial HTML string that is loaded on the page.
> Override this function to provide your own HTML string. For example,
> def ...

chriddyp:
I left a few small comments. Otherwise, this is looking good. I gave it a spin with two examples:
**Simple**
```python
# See https://github.com/plotly/dash/pull/286
# for details

import dash

app = dash.Dash()

app.layout = html.Div([
    html.H1('Hotpink'),
    html.H2('A Dash App with Local CSS'),
    html.Pre('''
    Testing https://github.com/plotly/dash/pull/286/
    ''')
])

if __name__ == '__main__':
    app.run_server(debug=True)
```

**Overriding index**
```python
# See https://github.com/plotly/dash/pull/286
# for details

import dash


class DashOverride(dash.Dash):
    def interpolate_index(self, **kwargs):
        print(kwargs)
        return '''
        <!DOCTYPE html>
        <html>
            <head>
                <title>My App</title>
                {css}
            </head>
            <body>
                <div id="custom-header">My custom header</div>
                {app_entry}
                <div id="custom-footer">My custom footer</div>
            </body>
            <footer>
                {config}
                {scripts}
            </footer>
        </html>
        '''.format(
            css=kwargs['css'],
            app_entry=kwargs['app_entry'],
            scripts=kwargs['scripts'],
            config=kwargs['config']
        )

app = DashOverride(static_folder='assets')

app.layout = html.Div([
    html.H1('Hotpink'),
    html.H2('A Dash App with Local CSS'),
    html.Pre('''
    Testing https://github.com/plotly/dash/pull/286/
    ''')
])

if __name__ == '__main__':
    app.run_server(debug=True)
```

With files in an `assets` folder as:
`simple-script.js`
```
alert('Testing')
```
`typography.css`
```
html, body {
    margin: 0;
    font-family: sans-serif;
}

h1 {
    color: hotpink;
}
```

and it worked as expected 🎉 

image

So, 💃  from me. I'll let @ned2 and/or another Dash team member give a second 💃

T4rk1n:
@chriddyp There is an error with your example.

> app = DashOverride(static_folder='assets')

That will create two static handlers for the files in `/assets`, one with a url prefix of `assets` and one with `/static`. The `static_folder` keyword is to keep compatibility with the old system, to specify the assets folder there is the `assets_folder` keyword.

dash/dash.py
```diff
@@ -558,9 +705,47 @@ def dispatch(self):
...
+                base = ''
+            else:
+                s = current.replace(walk_dir, '').lstrip('\\').lstrip('/')
+                splitted = slash_splitter.split(s)
+                if len(splitted) > 1:
+                    base = '/'.join(slash_splitter.split(s))
+                else:
+                    base = splitted[0]
+
+            for f in files:
```
rmarren1: `os.walk` calls `os.listdir` internally and the `os.listdir` docs [state](https://docs.python.org/3.6/library/os.html#os.listdir) "The [returned] list is in arbitrary order" since the user's OS ultimately decides what ordering this returns. Definitely not desirable as apps will behave differently on different machines.

I had a hunch the test case was just getting lucky so I added more files and got this ordering: `['load_first', 'load_after6', 'load_after', 'load_after1', 'load_after2', 'load_after5', 'load_after11', 'load_after4', 'load_after3', 'load_after10', 'load_
after7']`

This can be fixed by just adding
```
for current, _, files in os.walk(walk_dir):
            if current == walk_dir:
                base = ''
            else:
                s = current.replace(walk_dir, '').lstrip('\\').lstrip('/')
                splitted = slash_splitter.split(s)
                if len(splitted) > 1:
                    base = '/'.join(slash_splitter.split(s))
                else:
                    base = splitted[0]
            files.sort()  # ADDED LINE: Sort the files!
            for f in files:
                if base:
                    path = '/'.join([base, f])
                else:
                    path = f

```

This sorts the files as: `['load_first', 'load_after', 'load_after1', 'load_after10', 'load_after11', 'load_after2', 'load_after3', 'load_after4', 'load_after5', 'load_after6', 'load_
after7']`

I personally would prefer if 'load_after10' and 'load_after11' came last, which can be done by changing
`files.sort() -> files.sort(key=lambda f: int('0' + ''.join(filter(str.isdigit, f))))`

The former is what is expected, but the latter is what a programmer usually wants when sorting files. Either way I like the way everything else looks and am :dancer: once the sorting works.

ned2:
I can take a look at this shortly, if you hold off merging for a couple of
hours

dash/dash.py
```diff
@@ -419,6 +419,42 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+                            {}
+                            {}
+                            <div id="custom-footer">My custom footer</div>
+                        </body>
+                    </html>
+                    '''.format(
+                        kwargs.get('app_entry'),
+                        kwargs.get('config'),
+                        kwargs.get('scripts'))
+
```
ned2: I'd have a preference for using named interpolations here, as I think this is good practice for such a long string being interpolated. eg `'{scripts}'` ... `.format(scripts=kwargs['scripts'])` But it's not that much of an issue.

dash/dash.py
```diff
@@ -149,12 +208,26 @@ def layout(self, value):
...
+    @index_string.setter
+    def index_string(self, value):
+        checks = (
+            (_re_index_entry.search(value), 'app_entry'),
+            (_re_index_config.search(value), 'config',),
+            (_re_index_scripts.search(value), 'scripts'),
         )
+        missing = [missing for check, missing in checks if not check]
+        if missing:
+            raise Exception(
```
ned2: This should be a Dash-specific exception from the `dash.exceptions` module, but there's already a couple of other plain `Exception`s in `dash.py`, so it could just be fixed in another PR.

dash/dash.py
ned2: Ah that's interesting. Changing `name` to default to `__main__` was my change. I didn't test it as thoroughly as that, so missed those issues. So while it was an improvement on the previous default of `dash`, it's still got issues. Well spotted!

Not sure what you mean @chriddyp?  `name` currently is a keyword argument of `Dash`.

I don't think we can hope to intelligently infer the correct value of `name`, as the challenges around that are precisely why Flask makes this param mandatory, as @T4rk1n points out. So perhaps documentation is the best way around those edge cases.

bhuffman-usgs:
Chris, is this now functioning?  I've made an attempt to replicate your "Simple" and "Override Index" example however, the .css formatting isn't being applied.

chriddyp:
> Chris, is this now functioning? I've made an attempt to replicate your "Simple" and "Override Index" example however, the .css formatting isn't being applied.

Yeah this is. See https://community.plot.ly/t/dash-version-0-22-0-released/12098 for some official notes. I'm working on some official docs in https://github.com/plotly/dash-docs/pull/126

jonboone1:
I've been reading through documentation and am having a really hard time understanding how to accomplish what I want. I am trying to include additional tags in the header of the page, not just meta tags. I have both **link** and **script** tags I want to include, which doesn't seem to work with the `meta_tags` functionality. 

Any thoughts on how to accomplish this? I've also tried to set the index_string but received the following error:
`dash.exceptions.InvalidIndexException: Missing item new DashRenderer in index.`

alexcjohnson:
@jonboone1 by link and script tags do you mean you want to include your own .css and .js files on the page? Those are both included in the page automatically if you put them in the `assets` directory, no need to explicitly create the tags - see https://dash.plotly.com/external-resources

If the resources you want to add are elsewhere on the internet, you can use `external_stylesheets` and `external_scripts` - that's farther down on the same page https://dash.plotly.com/external-resources#adding-external-css/javascript

joshsmith2:
@jonboone1 - recognise this is now 18 months old - but if anyone else stumbles on this with requirements for custom `<head>` - in my case, the need to add custom attributes to the `<script>` tags I was embedding,  I found that a missing `{%renderer%}` tag in the body was causing that `InvalidIndexException`.

The docs for this are under 'Option 1 - Index string' [here](https://dash.plotly.com/external-resources).

Issue #302, opened 2018-07-24
Add js and css urls arguments to Dash.__init__.


T4rk1n:
We should add keywords arguments to add javascript urls and stylesheets to the `Dash.__init__` instead of relying on `app.scripts.append_scripts` and `app.css.append_css`.

Example:
```python
import dash

app = dash.Dash(
    js_scripts_urls=['https://cdn.google.com/google-analytics.js'],
    css_stylesheet_urls=['https://cdn.bootstrap.com/bootstrap.css']
)
```

Basically just lists of urls that are added as `external_url` to the appropriate resources (app.css or app.scripts).

PR #305, opened 2018-07-26
Add `external_js/css_urls` to dash init.

T4rk1n:
Solution for #302.

Add two keywords arguments to `Dash.__init__`:

- `external_scripts_urls` to include javascript files.  
- `external_css_urls` to include stylesheets.

They are both list of urls:

```python
import dash

external_js = [
    'https://www.google-analytics.com/analytics.js',
    'https://cdn.polyfill.io/v2/polyfill.min.js'
]

external_css = [
    'https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css',
    'https://codepen.io/chriddyp/pen/bWLwgP.css'
]

app = dash.Dash(
    external_script_urls=external_js, 
    external_css_urls=external_css)
```


chriddyp:
I personally love this. One more step towards getting rid of `app.scripts.append_script` and `app.css.append_css`.

Now, what should the name be? I guess technically it doesn't even need to be "external". And should we say "script" or "js" or "javascript"?

The perm options:
- `javascript_srcs=[...]`
- `javascript_urls=[...]`
- `js_srcs=[...]`
- `js_urls=[...]`
- `css_hrefs=[...]`
- `css_urls=[...]`

cc @plotly/dash as well

T4rk1n:
I put `external` because they are always loaded externally, I added a prop to serve it from external even if `serve_locally=True`, otherwise they wouldn't be included with that option and we already have the assets for loading from disk so it's really just for external urls.

rmarren1:
I like having the word `external` in it to make it super clear what is intended, and I think beginners who are still googling most things and copy-pasting code would be more likely to catch an error before it occurs if they intended a local script.

dash/dash.py
```diff
@@ -128,6 +131,14 @@ def _handle_error(error):
...
         self.css = Css()
         self.scripts = Scripts()
+
+        for method, resource in itertools.chain(
+                ((self.scripts.append_script, x)
+                 for x in (external_script_urls or [])),
+                ((self.css.append_css, x)
+                 for x in (external_css_urls or []))):
+            method({'external_url': resource, 'external_only': True})
+
```
chriddyp: I think that we should just add these URLs no matter what. In other words, reserve the `serve_locally` argument _just_ for the component CSS where there will always be a local and a remote version. Slowly, I'd like to move ourselves away from the `app.scripts.method` interface as well.

Users may want to serve the assets locally because `unpkg` is unreliable for them or its blocked for them but load the JS or CSS urls from some other CDN. Or, they might have an internal CDN where they host assets but they can't load external assets from unpkg.

So instead, let's keep this as a private variable or something and then just embed it directly in the `_generate_scirpts_html` function: https://github.com/plotly/dash/blob/54ca4a376ec349639da5140f1e2c42a739815626/dash/dash.py#L312-L332 (or however else you want to organize it) rather than embedding it through `append_script`

chriddyp: Ah, nevermind I see that `external_only` will embed it no matter what. OK, then that weakens my argument for not using `append_script`. I still don't really like the `app.scripts...` and `app.css...` namespace and I'd like for us to move away from it. However, I guess I don't have strong feelings about whether we should do this incrementally (i.e. not use it in this PR) or remove it all at once in a major version upgrade down the line.

chriddyp:
Alright, this looks good to me. I'm curious about your thoughts w.r.t. https://github.com/plotly/dash/pull/305#discussion_r206679837 but I won't let it block this PR. You're welcome to keep it as is or begin migrating away from `append_script` and add it directly to `_generate_scripts_html`. Up to you!

💃

T4rk1n:
I'd like us to move from the css and js resources as I don't really like that system but it works.

I hadn't thought about just putting them in `_generate_scripts_html` and `_generate_css_dist_html`, I think it would work too, I'll try that.

pope1ni:
I have a few comments here:

1. What control is there over ordering? Are internal scripts/styles always output before external?
    - *This should be documented, and probably mention #286 for more complex configuration.*
2. Why not allow setting attributes, e.g. `integrity`, `crossorigin`, `media`, `type`, etc.
    - *See my previous comment about this: https://github.com/plotly/dash/pull/171#issuecomment-375687736*
    - *Can we allow url strings or dicts of attributes for `<link rel="stylesheet">` and `<script>`.*
    - *In this case it would make more sense to drop the `_urls` suffix from the new arguments.*

As an aside, Dash also serves up it's internal resources from unpkg.com and, as this is an untrusted third-party, these should also use [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity). This would mean adjusting the code as in my comment mentioned above and generating hashes automatically on build of `bundle.js` in component bundles.

T4rk1n:
> What control is there over ordering? Are internal scripts/styles always output before external?

External scripts/styles are output after the assets. The renderer script is always loaded last.

> Why not allow setting attributes, e.g. integrity, crossorigin, media, type, etc.

I thought about it and I opted for a list of string for simplicity of usage, could put a check if it's string or a dict then format the tag accordingly.

T4rk1n:
I added support for tags with custom attributes, they can be mixed with simple urls. Also removed `_urls` prefix from the keywords.

Example:
```python
app = dash.Dash(external_scripts=[
    'https://www.google-analytics.com/analytics.js',
    {
        'src': 'https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.10/lodash.core.js',
        'integrity': 'sha256-Qqd/EfdABZUcAxjOkMi8eGEivtdTkh3b65xCZL4qAQA=',
        'crossorigin': 'anonymous'
    }
])
```
With that in place, we could add SRI to the bundles resources on unpkg as @pope1ni suggested in a future PR.

chriddyp:
Nicely done @T4rk1n , this looks great to me. Many thanks for the feedback @pope1ni !

💃

rmarren1:
Doesn't look like this is explained here https://dash.plot.ly/external-resources

PR #307, opened 2018-07-29
[WIP] Dash Dev Tools -- backend

rmarren1:
This solution uses the flask `got_request_exception` method to call a block of code before an exception is raised in the Dash backend when `debug=True`. In this block of code, we use socket.io (Flask-SocketIO) to emit an 'error' event from the namespace `/{prefix}_dash-errors` endpoint which includes error information like the Exception name, message, and traceback.

Still need to figure out:
* socket.io prints logs every couple of seconds which is annoying, can't figure out how to turn them off.
* does wrapping the server in a SocketIO instance in this line: `socketio = SocketIO(self.server)` break anything in servers which use Dash plus some other Flask plugins?

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -8,6 +8,7 @@ import json
 import pkgutil
 import warnings
 import re
+import traceback
 
 from functools import wraps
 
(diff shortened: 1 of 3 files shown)


rmarren1:
Closing in favor of catching exceptions in the front-end. In this solution, the error data transferred over to the front-end was the input to the flask exception handler rather than the output, so we did not get the nice werkzueg error pages.

PR #309, opened 2018-07-31
Bust the assets cache on modification.

### LINKED ISSUE #303: Cache-busting URLs for CSS included in the assets folder
See https://github.com/plotly/dash/pull/286#discussion_r201812562 for context

T4rk1n:
Append time modified as query string to assets urls to trick the cache to refresh the file.

cc @plotly/dash

Resolves #303


rmarren1:
I like this, annoying to always have dev tools open and the cache probably trips up beginners.

T4rk1n:
@chriddyp I changed this PR to use time modified query string.

T4rk1n:
@chriddyp  Ok, I'll release this along with #305 tomorrow.

chriddyp:
@T4rk1n - Could we get this one merged and released?

T4rk1n:
@chriddyp 
Sorry I thought @bpostlethwaite wanted to chime in, we discussed that in local, we shouldn't add the modified query string for the assets to be mapped to the right local file and instead use the no-cache while on local and use the query string only in a prod settings.

chriddyp:
> we discussed that in local, we shouldn't add the modified query string for the assets to be mapped to the right local file and instead use the no-cache while on local and use the query string only in a prod settings

OK. Why? It seems to complicate things to have two different ways that caches are invalidated in local vs deployed. Also, we don't have the config abstraction around local vs production yet.

T4rk1n:
That is also my feeling, I'll merge & release.

PR #316, opened 2018-08-01
Add meta http-equiv for ie compatibility to the index by default.

T4rk1n:
As discussed in #275, add `<meta http-equiv="X-UA-Compatible" content="IE=8">` to the index if it's not specified in the `meta_tags`.

Should the meta tags be closed with '`/>` or just `>`  ? The HTML5 specification says it doesn't need it but it's needed by `xhtml`. We have `<!DOCTYPE html>` by default, but if someone changes it to something else it might needs it.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -339,16 +339,21 @@ class Dash(object):
         ).format(json.dumps(self._config()))
 
     def _generate_meta_html(self):
+        has_ie_compat = any(
+            x.get('http-equiv', '') == 'X-UA-Compatible'
+            for x in self._meta_tags)
         has_charset = any('charset' in x for x in self._meta_tags)
 
         tags = []
+        if not has_ie_compat:
+            tags.append('<meta equiv="X-UA-Compatible" content="IE=edge">')
         if not has_charset:
-            tags.append('<meta charset="UTF-8"/>')
+            tags.append('<meta charset="UTF-8">')
         for meta in self._meta_tags:
             attributes = []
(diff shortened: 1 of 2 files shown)


chriddyp:
Let's follow the lead on this project:
https://github.com/h5bp/html5-boilerplate/blob/2e32b433f77ea693d69f1728e6796e102fd84df0/dist/index.html#L1-L6

chriddyp:
@T4rk1n - Could we add this item to the changelog?

T4rk1n:
@chriddyp - It is in the changelogs:

https://github.com/plotly/dash/blob/master/CHANGELOG.md#0231---2018-08-02

chriddyp:
🙈 sorry about that, totally missed it. thank you! 🙇

dash/dash.py
```diff
@@ -339,16 +339,21 @@ def _generate_config_html(self):
...
 
     def _generate_meta_html(self):
+        has_ie_compat = any(
+            x.get('http-equiv', '') == 'X-UA-Compatible'
+            for x in self._meta_tags)
         has_charset = any('charset' in x for x in self._meta_tags)
 
         tags = []
+        if not has_ie_compat:
+            tags.append('<meta equiv="X-UA-Compatible" content="IE=edge">')
```
pope1ni: Hmm. Should this not be `http-equiv="..."`?

T4rk1n: Yes, it should.

PR #318, opened 2018-08-02
Add `assets_ignore` init keyword, regex filter for the assets files.

### LINKED ISSUE #306: Allow for defining ignored files from `assets/` that should not be automatically served
I have couple of available color pallets on my site, the `app.callback` changes the `href` value of corresponding `html.Link` when user picks another option. I would like to keep all my styles (default ones + colors defining) in single directory, it would be nice to have some way to define ignored files, of best regexp that will define ignored filename from the assets directory as in case like this including all styles in the `<head>` will not work.

T4rk1n:
Resolves #306, add a regex ignore filter to the assets auto include on the index. 

```python
import dash

app = dash.Dash(assets_ignore='.*ignored.*')
```

The ignored assets are still available from a request context (a callback):

`flask.url_for('assets', filename='ignored.css')`

They won't work with #309 .


chriddyp:
> They won't work with #309 .

Let's get #309 first, then revisit this one

radekwlsk:
@chriddyp @T4rk1n As #309 is released in 0.24.0 as I believe, I assume this one is no longer ready for release and has to be updated?

T4rk1n:
> They won't work with #309 .

I meant that they won't have the query string with the modified time when you get them from somewhere else than the assets includes.

chriddyp:
💃

Issue #312, opened 2018-07-31
Dash's config system


chriddyp:
(placeholder)

chriddyp:
This is a discussion about Dash's config system. As we add features that can be turned on and off or configured, like those in our dash dev tools system, we should make sure that we're building them on a sound configuration / settings system.

In principle, I'm interested in Dash having a small, declarative API. That is, when you learn Dash, you only need to know:
- Config settings
- `app.callback`
- `app.layout`
- `app.run_server`
- The component's declarative input arguments

The config settings are currently exposed in an ad-hoc way in `app.config` and through the `dash.Dash` constructor. I propose a few changes/guidelines:

***

**config in constructor**
Expose all config arguments through `dash.Dash` constructor. One day, years from now, this may mean that we'll have hundreds of items in here

***

**config in `os.environ`**
Provide an alternative way of setting the variables through `os.environ`. We'll prefix `os.environ` settings with `dash_` so that there aren't conflicts with other services.
For example, the following two settings are equivalent:
```python
dash.Dash(port=8050)
```

```python
os.environ['dash_port'] = '8050'
dash.Dash()
```

The arguments in the constructor will take precedence over `os.environ`

Some other notes:
- lowercase or UPPERCASE would be accepted
- environ variables are strings, so we'd convert them back into numbers

***

**immutable config**
Right now, config is accessible in `app.config` which sort of implies that its mutable. but many of the settings aren't actually mutable. We want to do things in our `__init__` with the arguments that are passed in or available in `os.environ` (like set up our `routes` with `requests_pathname_prefix`). We don't want to have to maintain multiple pathways for setting and resetting aspects of the object that depend on `app.config`.

So, we'd deprecate `app.config`

We'd also deprecate the ad-hoc `app.<whatever>` settings like `app.title`

***

**defaults, overrides, specificity**
Some settings will override other settings. For example, originally I had
`url_base_pathname` but found that this wasn't specific enough so I introduced `requests_pathname_prefix` and `routes_pathname_prefix`

So, setting `url_base_pathname` would set the same variable for `requests_pathname_prefix` and `routes_pathname_prefix`

***

**`debug`, `dev`, `production`**
as we introduce different debugging features as part of https://github.com/orgs/plotly/projects/3, we will want users to turn them on and off.

Perhaps we have a `dev` variable: `True` or `False`

Each individual feature should have its own variable as well. `dev=True` turns them all on, `dev=False` turns them all off, and individual features could configured separately:
[...]

PR #322, opened 2018-08-03
Pathname prefix from environ variables.

T4rk1n:
Fixes:

- Take `requests_pathname_prefix` config when creating scripts tags.

Added:

- configs from init or from environ.
  - If init is supplied, take that value, if not take from environ or default.
  -  Dash environ variables starts with `DASH_` (eg: `DASH_ROUTES_PATHNAME_PREFIX`) 
- `routes/requests_pathname_prefix` evaluation and checks.
  - defaults to `'/'`
  - `routes_pathname_prefix` must start/end with `/`.
  - Cannot supply both `url_base_pathname` and `routes/requests_pathname_prefix`.
  - `requests_pathname_prefix` must ends with `routes_pathname_prefix`
  - If no other options is supplied and `DASH_APP_NAME` is in `os.environ`, format `requests_pathname_prefix` as `/{DASH_APP_NAME}{routes_pathname_prefix}`


utyf:
IMHO, taking config implicitly from environment sounds like not a very good idea, at least for Dash.

dash/dash.py
```diff
@@ -105,7 +105,10 @@ def __init__(
         self.config = _AttributeDict({
             'suppress_callback_exceptions': False,
             'routes_pathname_prefix': url_base_pathname,
-            'requests_pathname_prefix': url_base_pathname,
+            'requests_pathname_prefix': os.getenv(
+                'DASH_REQUESTS_PATHNAME_PREFIX',
+                '/{}/'.format(os.environ['DASH_APP_NAME'])
+                if 'DASH_APP_NAME' in os.environ else url_base_pathname),
```
ned2: It looks like this makes the os.environ parameter take precedence over the value of `url_base_pathname` set in the constructor, which might go against @chriddyp's suggestion in #312 of having config values set in the constructor take precedence. (Although the fallback logic here is a bit more subtle than simple parameter shadowing). But then there's the challenge of detecting whether a user specified a value through the constructor or if the value is just the default. I'm not sure what the best of doing that is though.

I wonder if it might make sense to come up with a systematic abstraction for getting config values from the constructor, falling back to os.environ (and handling either upper or lower case DASH_ vars), and then falling back to a final default value? This is probably taking the discussion further afield of this PR (sorry!) but one possibility is to switch to using `**kwargs` in the constructor. eg:

`some_param = kwargs.get('some_param', os.environ.get('dash_some_param', os.environ.get('DASH_SOME_PARAM', defaults.some_param)))`

Although I suppose that breaks IDE autocomplete `for Dash.__init__`...

(apologies if I'm missing some additional context to this PR also!)

chriddyp:
> IMHO, taking config implicitly from environment sounds like not a very good idea, at least for Dash.

Why is that? All the variables will be prefixed by `dash_`, so there shouldn't be any inadvertent readings. There is also an alternative to set them explicitly. 

For some more context, I laid out some ideas here: https://github.com/plotly/dash/issues/312

T4rk1n:
@chriddyp 

`requests_pathname_prefix` needs to contains the `routes_pathname_prefix` to work properly.

Example

requests='/dash-app-name'
routes='/dash'
custom-flask-index='/'

The working url for that app would be `host/dash-app-name/` which leads to custom-flask-index that contains a link to the dash app in `/dash`.

The requests thus needs to come from `/dash-app-name/dash` in that case.

Should we issue a warning and append the routes_pathname if it's not included in the requests_pathname or raises an error in that case ?

T4rk1n:
@chriddyp I updated this PR, please review.

dash/_configs.py
```diff
@@ -0,0 +1,104 @@
...
+    the prefix for the API routes on the backend (this flask server).
+    `url_base_pathname` will set `requests_pathname_prefix` and
+    `routes_pathname_prefix` to the same value.
+    If you need these to be different values then you should set
+    `requests_pathname_prefix` and `routes_pathname_prefix`,
+    not `url_base_pathname`.
+    '''
+    environ_configs = environ_configs or env_configs()
+
+    url_base_pathname = choose_config('url_base_pathname',
```
chriddyp: Instead of "`choose_config`", maybe "`get_config`"?

dash/_configs.py
```diff
@@ -0,0 +1,104 @@
...
+        raise exceptions.InvalidConfig(
+            '`routes_pathname_prefix` needs to start with `/`')
+    if not routes_pathname_prefix.endswith('/'):
+        raise exceptions.InvalidConfig(
+            '`routes_pathname_prefix` needs to end with `/`')
+
+    app_name = environ_configs.DASH_APP_NAME
+
+    if not requests_pathname_prefix and app_name:
+        requests_pathname_prefix = '/' + app_name + routes_pathname_prefix
```
chriddyp: It took me a while to understand this but I think I finally get it. This is the case for users that are deploying on DDS that want a different URL base name. i.e. instead of just `/my-app-name` they want it to be `/some-base-url/another-app`. I think this will be uncommon but it makes sense and it seems complete 👍

tests/test_configs.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_pathname_prefix_environ_routes(self):
+        os.environ['DASH_ROUTES_PATHNAME_PREFIX'] = '/routes/'
+        _, routes, req = _configs.pathname_configs()
+        self.assertEqual('/routes/', routes)
+
+    def test_pathname_prefix_environ_requests(self):
+        os.environ['DASH_REQUESTS_PATHNAME_PREFIX'] = '/requests/'
+        _, routes, req = _configs.pathname_configs()
+        self.assertEqual('/requests/', req)
+
```
chriddyp: 🙇 thanks for the tests!

dash/dash.py
```diff
@@ -101,13 +107,30 @@ def __init__(
...
+            ),
+            'routes_pathname_prefix': routes_pathname_prefix,
+            'requests_pathname_prefix': requests_pathname_prefix,
+            'include_assets_files': _configs.choose_config(
+                'include_assets_files',
+                include_assets_files,
+                env_configs,
+                True),
+            'assets_external_path': _configs.choose_config(
+                'assets_external_path', assets_external_path, env_configs, ''),
```
chriddyp: 👍 this looks great. Thanks for making these changes!

chriddyp:
> I wonder if it might make sense to come up with a systematic abstraction for getting config values from the constructor, falling back to os.environ (and handling either upper or lower case DASH_ vars), and then falling back to a final default value? This is probably taking the discussion further afield of this PR (sorry!) but one possibility is to switch to using **kwargs in the constructor. eg:

@ned2 - Check it out now. We're now taking precedence in the constructor with `None` checks. This will break down in the future if we end up having config values that are and truly should be `None`, but it works for now.

chriddyp:
Looks good to me @T4rk1n ! 💃

T4rk1n:
One thing that bothers me with the None checks is that there is no more checking of the default value in the IDE while you type, but this could be countered with good docstring.

> This will break down in the future if we end up having config values that are and truly should be None, but it works for now.

Could put a `accept_none_init` as parameter and have the init be '' instead.

PR #325, opened 2018-08-08
Fix bugs in 0.23.1

### LINKED ISSUE #321: Generate_class issue with plotly dahs 0.23.1 and dangerously_set_inner_html
After updating to 0.23.1, the following code importing "dangerously_set_inner_html" does not work anymore. It seems that some field may be missing / not set correctly and that this is field is used in an exec statement. The same import is working with previous release.

Error:
```
In [1]: import dash_dangerously_set_inner_html

  File "/miniconda/envs/.../lib/python3.4/site-packages/IPython/core/interactiveshell.py", line 2963, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  File "<ipython-input-1-8721236f3c51>", line 1, in <module>
    import dash_dangerously_set_inner_html

  File "/miniconda/envs/////lib/python3.4/site-packages/dash_dangerously_set_inner_html/__init__.py", line 10, in <module>
    'dash_dangerously_set_inner_html'

  File "/miniconda/envs/.../lib/python3.4/site-packages/dash/development/component_loader.py", line 51, in load_components
    namespace

  File "/miniconda/envs/...lib/python3.4/site-packages/dash/development/base_component.py", line 441, in generate_class
    exec(string, scope)

  File "<string>", line 27
    def __init__(self, children=None, , **kwargs):
```

rmarren1:
* Fix #321
* Add test case to make sure Dash can import prop-less components, using dash_dangerously_set_inner_html.DangerouslySetInnerHTML as the test component.

* Fix https://github.com/plotly/dash-html-components/issues/54
* Add test case to make sure Dash can import components which use arguments that are Python keywords


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.23.2 - 2018-08-07
+## Fixed
+- Fixed bug in 0.23.1 where importing Dash components with no props would result in an error. (Fixes [#321](https://github.com/plotly/dash/issues/321)).
+- Fixed bug in 0.23.1 where importing components with arguments that are python keywords could cause an error.
```
chriddyp: Could you add a note here like "In particular, this fixes importing the modules with Python 3.7"?

Issue #264, opened 2018-06-08
Validation on component properties


chriddyp:
This is part of the "A Pleasant and Productive Developer Experience" objective ❤️ 

Currently, we don't do any validation on property values of components. So, if you pass in 
```
html.Div([[]])
```
instead of 
```
html.Div([])
```
Dash will _not_ raise an error but instead silently fail in the front-end, sometimes but not always showing the notorious "Error loading dependencies" error message.

We have all of the prop-type values available in the component's `metadata.json` files, extracted from the component's `propTypes`.
For example, here is the `Dropdown.react.js` prop types:
https://github.com/plotly/dash-core-components/blob/7c2b102c680d9fe0cb9df61f1682feccfa3ddc05/src/components/Dropdown.react.js#L101-L180
and here is it's associated JSON types:
https://github.com/plotly/dash-core-components/blob/7c2b102c680d9fe0cb9df61f1682feccfa3ddc05/dash_core_components/metadata.json#L810-L946

So, implementing prop type checking would involve:
- Making sure that all component props are sufficiently described. I'm sure that flexible properties like `children` aren't sufficiently described. For example, `children` is most explicitly:
```
propTypes.oneOfType([
    PropTypes.string,
    PropTypes.null,
    PropTypes.number,
    PropTypes.node,
    PropTypes.arrayOf(
        propTypes.oneOfType([
            PropTypes.string,
            PropTypes.null,
            PropTypes.number,
            PropTypes.node,
        ])
    )
])
````
- Use the `metadata.json` property types to raise helpful error messages. I used the [Cerberus package](http://docs.python-cerberus.org/en/stable/api.html#error-codes) recently for type-checking and I think it's a really good fit for this project. It has a declarative structure for describing nested properties that is very similar to the structure in `metadata.json`
- Type validation will exist as part of the component base class: https://github.com/plotly/dash/blob/master/dash/development/base_component.py

zogzog:
Well, ok. However the experience of working dash apps suddenly not starting was not _that_ pleasant (I know, I should have pinned a precise dash version).
Also I had to stuff `children={}` all over the place. And replace a few Nones with ''. That gets the validator satisfied. And then "Error loading dependencies".

</rant>

rmarren1:
[...]

PR #330, opened 2018-08-14
Check to make sure children are not nested lists. Fix plotly/dash-renderer#23

rmarren1:
Proposed fix to https://github.com/plotly/dash-renderer/issues/23

This solution borrows heavily from https://github.com/plotly/dash/pull/273. Basically the same except instead of checking if values in the tree are a member of a valid set, we just check if the value is not a nested list.

A major difference is that this check if performed before _each_ callback rather than being called only if `json.dumps` fails. In this case, `json.dumps` succeeds since nested arrays are valid JSON.

Examples:
```
dash.exceptions.InvalidCallbackReturnValue: 
The callback for property `children` of component `output`
returned a tree with one value having type `list`

The value in question is located at
[*] list 


and has string representation
`['nested']`

or un-nested lists of those.

```
```
dash.exceptions.InvalidCallbackReturnValue: 
The callback for property `children` of component `output`
returned a tree with one value having type `list`
which can not be serialized by Dash.

The value in question is located at
[*] Div 


and has string representation
`[['nested']]`

or un-nested lists of those.
```

### DIFF
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -1,3 +1,8 @@
+import collections
+from . import exceptions
+from .development.base_component import Component
+
+
 def interpolate_str(template, **data):
     s = template
     for k, v in data.items():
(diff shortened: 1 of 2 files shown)


chriddyp:
> A major difference is that this check if performed before each callback rather than being called only if json.dumps fails. In this case, json.dumps succeeds since nested arrays are valid JSON.

This is an interesting approach. I'm a little worried about:
- Performance issues in crawling the tree, especially with really large apps
- Keeping this in check with our json serializer. I believe that `children=pandas.Series(...)` is technically valid but it would fail in this case.

***

Architecturally, I originally imagined this working on the component-level. That way, we could verify:
- All properties, not just the `children` prop
- Not just primative types but also _shapes_ and nested types

In the most general solution, I imagined this as being part of the base `Component` and doing type-checking with the metadata that we are provided through `propTypes`. That way, the component author could specify the shapes of their props and the validation would automatically be done. I have some more details about this here: https://github.com/plotly/dash/issues/264

If we ended up implementing #264, then I think that we would get this issue (children as nested children) for free. Of course, there are still some issues:
- The higher-level types. Technically a `pandas.Series` should be considered valid just like a `list`. Perhaps for each propType we implement a list of valid Python types or validators
```
{
    'array': [pandas.Series, list, numpy.array]
}
```
- We might want to make it configurable (similar to `suppress_callback_exceptions`, something like `suppress_component_validation_exceptions`), in case the component author didn't fully annotate their proptypes

rmarren1:
closed in favor of more general https://github.com/plotly/dash/pull/340

PR #335, opened 2018-08-14
CSS/JS external resources loaded before the assets

### LINKED ISSUE #331: Can't override external CSS without !important
Not really a bug, just something I noticed that we should be aware of.

In the new `assets` directory approach to including CSS, [external stylesheets always appear beneath local stylesheets.](https://github.com/plotly/dash/blob/7361db52690810891c5f68501eda1d71043cb777/dash/dash.py#L333-L336)

There doesn't seem to be a way to switch this (at least that I'm aware of), so a user can't override CSS from external libraries without !important since external scripts will always be imported last and thus take precedence. I think this is a very common use case (e.g. import bootstrap, then add your own CSS to customize a small part of bootstrap) and users shouldn't have to manually add !important to all their stylesheets or change the specificity of selectors to get it to work.

T4rk1n:
Ensure CSS/JS external resources are loaded before the assets, Fix #331.


rmarren1:
Looks good! :dancer:

eddy-ojb:
Is this feature optionally enforced?

I am looking to apply Bootstrap popovers in order to provide descriptions to new users on a per graph basis since some of my users may be new to some graph types. In order to use a popover on a regular html page, a bit of jquery os required after the document is ready as described here:

https://getbootstrap.com/docs/4.0/components/popovers/

Before I go down this road, would you expect this to be feasible in Dash? I see the html components facilitate this, so i dont see why not..

rmarren1:
@eddy-ojb I've attempted to use JQuery before and it did not work too well. Dash is built on react, and mixing react and jQuery gets tricky at times. Best bet is probably to use React to do popovers (e.g. make a custom component library). I think the easiest route would be to build on the `dash-core-components` `Graph` component and add popover functionality.

Some links:
'Official' info on using JQuery with react - https://reactjs.org/docs/integrating-with-other-libraries.html
Reasons to not use JQuery and just use react - https://discuss.reactjs.org/t/jquery-with-react/683
React-Strap -- boostrap-like components in React - https://reactstrap.github.io/components/popovers/

eddy-ojb:
Thanks for the tip @rmarren1. Rather than waste my time with jquery I will explore making my own component. I am a javascript noob and so I expect this to take a while but will share afterwards.

Issue #334, opened 2018-08-14
Functions as layout no longer work (_validate_layout throws error)


phifre:
Hello,

since the `_validate_layout` method was introduces in dash.py setting functions as layout no longer works, because calling `traverse()` on an function fails. 
Setting a function as layout is necessary to make the app reload the layout on page reload as described in the [documentation](https://dash.plot.ly/live-updates).

Currently the fastest option is to remove the `_validate_layout` call in the `_setup_server` method.

phifre:
Checking if the layout is a function and then calling it once to validated the returned layout would be an appropriate fix.

rmarren1:
Will have a fix for this soon, this was introduced in 0.22.0 via https://github.com/plotly/dash/pull/294 so you can downgrade to 0.21.1 for now or edit as above.

phifre:
Actually it was introduced with version 0.24.2 in commit [https://github.com/plotly/dash/commit/e190768d7ae3db400e744d6d1d594969aacd2bf9](e190768d7ae3db400e744d6d1d594969aacd2bf9) so 0.24.1 works fine.

T4rk1n:
Yes, it's from the disallow duplicate ids of 0.24.2.

@rmarren1  Can fix this with `self._layout_value().traverse()` instead of `self.layout.traverse()` or taking the `self._cached_layout` value.

rmarren1:
Ah right, forgot more was added to validation. Thanks for the detailed report @phifre !

rmarren1:
Fix for this released in `dash==0.26.1` [see changelog](https://github.com/plotly/dash/blob/master/CHANGELOG.md#0261---2018-08-26)

PR #336, opened 2018-08-15
Change _validate_layout to accept functions that return layouts.

rmarren1:
Fix https://github.com/plotly/dash/issues/334


dash/dash.py
```diff
@@ -875,10 +875,15 @@ def _validate_layout(self):
...
+        if callable(self.layout):
+            to_validate = self.layout()  # pylint: disable=not-callable
+        else:
+            to_validate = self.layout
+
         layout_id = getattr(self.layout, 'id', None)
 
         component_ids = {layout_id} if layout_id else set()
-        for component in self.layout.traverse():
+        for component in to_validate.traverse():
```
T4rk1n: Why not use `self._layout_value()` ? it basically does the same check.

tests/test_integration.py
```diff
@@ -12,7 +12,7 @@
 import time
 
 from dash.dependencies import Input, Output
-from dash.exceptions import PreventUpdate
+from dash.exceptions import PreventUpdate, NoLayoutException
```
T4rk1n: I don't see NoLayoutException used anywhere.

rmarren1:
Fixed those, I'll rebase and see if the new CI passes

chriddyp:
Would this PR fix https://community.plot.ly/t/attribute-error-layout-traverse/13031?

rmarren1:
^ Yes it should.

Can someone check out whats up with Percy, those tests worked out before bumping changelog and version

rmarren1:
Okay, it looks like https://github.com/plotly/dash-core-components/pull/280 and the release of `dash-core-components==0.28.0` resulted in some changes in percy. I'll release this since it is working with `dash-core-components==0.27.2`.

Can someone check to make sure those are okay? I do not currently have permissions to view

PR #338, opened 2018-08-16
Update circle 2

T4rk1n:
- Update the circle configs to use version 2.
- Removed tox.
- Add python 3.7 test environment.

It's mostly the same as #222 but with python 3.7 tests. Pylint for python 3.7 has changed some rules naming so I created another .pylintrc for that one.


.circleci/config.yml
```diff
@@ -0,0 +1,70 @@
...
+  "python-3.6":
+    <<: *test-template
+    docker:
+      - image: circleci/python:3.6-stretch-browsers
+        environment:
+          REQUIREMENTS_FILE: dev-requirements.txt
+          PYLINTRC: .pylintrc
+
+  "python-3.7":
+    <<: *test-template
```
chriddyp: nice work getting this test-template stuff in. code is super DRY 🐫

chriddyp:
Looks great @T4rk1n ! There might be some CONTRIBUTING.md files or something that need updating as well. Otherwise, 💃

T4rk1n:
@chriddyp There is no CONTRIBUTING.md for dash, maybe we should add one ?

PR #340, opened 2018-08-17
Validate component properties #264

rmarren1:
PR for #264 
#### Current Prerelease
```
pip install dash==0.29.0rc8
pip install dash-renderer==0.15.0rc1
pip install dash-core-components==0.31.0rc2
pip install dash-html-components==0.14.0rc3
```

Validation is on by default. To turn it off, you can set `app.config.disable_component_validation = True`

#### Test Cases to run for demo
```
import dash

app = dash.Dash()
app.scripts.config.serve_locally=True

app.layout = html.Div([
    html.Button(id='click1', children='click to return bad Div children'),
    html.Div(id='output1', **{'data-cb': 'foo'}),
    html.Button(id='click2', children='click to return a bad figure'),
    dcc.Graph(id='output2', figure={'data': [], 'layout': {}}),
    html.Button(id='click3', children='click to return a bad radio'),
    dcc.RadioItems(id='output3', options=[{'value': 'okay', 'label': 'okay'}]),
    html.Button(id='click4', children='click to make a figure with no id'),
    html.Div(id='output4'),
])



              [Input('click1', 'n_clicks')])
def crash_it1(clicks):
    if clicks:
        return [[]]
    return clicks

@app.callback(Output('output2', 'figure'),
              [Input('click2', 'n_clicks')])
def crash_it2(clicks):
    if clicks:
        return {'data': {'x': [1, 2, 3], 'y': [1, 2, 3], 'type': 'scatter'}, 'layout': {}}
    return go.Figure(data=[go.Scatter(x=[1,2,3], y=[1,2,3])], layout=go.Layout()) 

@app.callback(Output('output3', 'options'),
              [Input('click3', 'n_clicks')])
def crash_it3(clicks):
    if clicks:
        return [{'value': {'not okay': True}, 'labl': 'not okay'}]
    return [{'value': 'okay', 'label': 'okay'}]

@app.callback(Output('output4', 'children'),
              [Input('click4', 'n_clicks')])
def crash_it4(clicks):
    if clicks:
        return dcc.Graph()
    return dcc.Graph(id='hi')

app.run_server(debug=True, port=8050)
```

#### Example Error Messages
```
CallbackOutputValidationError: 


A Dash Callback produced an invalid value!

Dash tried to update the `figure` prop of the
`Graph` with id `output2` by calling the
`crash_it2` function with `(1)` as arguments.

This function call returned `{'layout': {}, 'data': {'y': [1, 2, 3], 'x': [1, 2, 3], 'type': 'scatter'}}`, which did not pass
validation tests for the `Graph` component.

The expected schema for the `figure` prop of the
`Graph` component is:

***************************************************************
{'validator': 'plotly_figure'}
***************************************************************

The errors in validation are as follows:

* figure	<- Invalid Plotly Figure:


    Invalid value of type '__builtin__.dict' received for the 'data' property of 
        Received value: {'y': [1, 2, 3], 'x': [1, 2, 3], 'type': 'scatter'}

    The 'data' property is a tuple of trace instances
    that may be specified as:
      - A list or tuple of trace instances
        (e.g. [Scatter(...), Bar(...)])
      - A list or tuple of dicts of string/value properties where:
        - The 'type' property specifies the trace type
            One of: ['mesh3d', 'splom', 'scattercarpet',
                     'scattergl', 'scatterternary', 'pie',
                     'surface', 'histogram', 'ohlc', 'heatmapgl',
                     'cone', 'scatterpolar', 'table',
                     'scatterpolargl', 'histogram2d', 'contour',
                     'carpet', 'box', 'violin', 'bar',
                     'contourcarpet', 'area', 'choropleth',
                     'candlestick', 'streamtube', 'parcoords',
                     'heatmap', 'barpolar', 'scattermapbox',
                     'scatter3d', 'pointcloud',
                     'histogram2dcontour', 'scatter', 'scattergeo',
                     'sankey']

        - All remaining properties are passed to the constructor of
          the specified trace type

        (e.g. [{'type': 'scatter', ...}, {'type': 'bar, ...}])

```
```
CallbackOutputValidationError: 


A Dash Callback produced an invalid value!

Dash tried to update the `options` prop of the
`RadioItems` with id `output3` by calling the
`crash_it3` function with `(1)` as arguments.

This function call returned `[{'value': {'not okay': True}, 'labl': 'not okay'}]`, which did not pass
validation tests for the `RadioItems` component.

The expected schema for the `options` prop of the
`RadioItems` component is:

***************************************************************
{'schema': {'allow_unknown': False,
            'nullable': False,
            'schema': {'disabled': {'type': 'boolean'},
                       'label': {'type': 'string'},
                       'value': {'type': 'string'}},
            'type': 'dict'},
 'type': 'list'}
***************************************************************

The errors in validation are as follows:

* options
 * 0
  * value	<- must be of string type
  * labl	<- unknown field

```
```
ComponentInitializationValidationError: 


A Dash Component was initialized with invalid properties!

Dash tried to create a `RadioItems` component with the
following arguments, which caused a validation failure:

***************************************************************
{'id': 'output3', 'options': [{'label': 'okay', 'value': {}}]}
***************************************************************

The expected schema for the `RadioItems` component is:

***************************************************************
{'className': {'type': 'string'},
 'dashEvents': {'allowed': ['change'], 'type': ('string', 'number')},
 'fireEvent': {},
 'id': {'type': 'string'},
 'inputClassName': {'type': 'string'},
 'inputStyle': {'type': 'dict'},
 'labelClassName': {'type': 'string'},
 'labelStyle': {'type': 'dict'},
 'options': {'schema': {'allow_unknown': False,
                        'nullable': False,
                        'schema': {'disabled': {'type': 'boolean'},
                                   'label': {'type': 'string'},
                                   'value': {'type': 'string'}},
                        'type': 'dict'},
             'type': 'list'},
 'setProps': {},
 'style': {'type': 'dict'},
 'value': {'type': 'string'}}
***************************************************************

The errors in validation are as follows:


* options
 * 0
  * value	<- must be of string type

```

#### PropTypes to Cerberus reference
| PropType  | Cerberus Schema Validated Against | Known Current Limitations |
| ---------- | ------------------------------------ | -------------------------- |
| `array`     | `{'type': 'list'}` | |
| `bool`      | `{'type': 'boolean'}` | |
| `func`      | `{}` | No validation occurs. |
| `object`      | `{'type': 'dict'}` | Validates that input is explicitly a `dict` object. We cannot be more general (e.g. `collections.abc.Mapping`) since the core `Component` is an instance of that.|
| `string`    | `{'type': 'string'}`  | |
| `symbol`    | `{}`  | No validation occurs |
| `node`    | `{'anyof': [{'type': 'component'}, {'type': 'boolean'}, {'type': 'number'}, {'type': 'string'}, {'type': 'list', 'schema': {'anyof': [{'type': 'component'}, {'type': 'boolean'}, {'type': 'number'}, {'type': 'string'}]}}]}`  | |
| `instanceOf(Object)`    | `{}`  | No validation occurs |
| `oneOf(['val1', 2])`    | `{'allowed': [None, 'val1', 2]}`  | Strings will have `'` characters stripped off each end. This is because `metadata.json` generation serializes the literal values as json, so for example `PropTypes.oneOf(['News', 'Photos'])` serializes to `["'News'", "'Photos'"]`.  https://github.com/reactjs/react-docgen/issues/57 |
| `oneOfType( [ PropTypes.string, PropTypes.bool ] )`    | `{'anyof': [{'type': 'string', 'type': 'boolean'}]}`  |  If one of the types is a `PropType` that cannot be validated individually (e.g. `PropType.func`), no validation will occur and the schema will effectively be `{}` |
| `arrayOf( PropTypes.number )`    | `{'type': 'list', 'schema': {'type': 'number'}}`  |  |
| `objectOf( PropTypes.number )`    | `{'type': 'dict', 'valueschema': {'type': 'number'}}`  |  |
| `shape( { k1: PropTypes.string, k2: PropTypes.number } )`    | `{'type': 'dict', 'allow_unknown': False, 'schema': {'k1': {'type': 'string}, 'k2': {'type': 'number'}}}`  |  |
| `any`    | `{'type': ('boolean', 'number', 'string', 'dict', 'list')}`  |  |


dash/dash.py
```diff
@@ -234,6 +234,12 @@ def layout(self, value):
 
         self._layout = value
 
+        self._components = {}
```
rmarren1: We keep a dictionary of all the component instances which can have their props changed (e.g. components with the `id` field. Dash did not have a way to reference the class or the instance of the component associated with an (`id`, `prop`) combo before.

dash/dash.py
```diff
@@ -805,6 +811,23 @@ def wrap_func(func):
             def add_context(*args, **kwargs):
 
                 output_value = func(*args, **kwargs)
+                setattr(
```
rmarren1: For each callback, we can actually update the component object with the correct props so that the layout in the backend matches the layout in the front-end. This is not really necessary for this PR, but may be useful in the future.

dash/dash.py
```diff
@@ -805,6 +811,23 @@ def wrap_func(func):
...
+                    self._components[output.component_id],
+                    output.component_property,
+                    output_value
+                )
+                self._components[output.component_id]._validator.validate({
```
rmarren1: Validate that the callback made a valid prop

dash/dash.py
```diff
@@ -805,6 +811,23 @@ def wrap_func(func):
...
+                    output.component_property: output_value
+                })
+                if output.component_property == 'children':
```
rmarren1: If the component prop is `children`, we must also add any new components with `id` fields which may have been added to the layout as a result of this callback to the `self._components` dictionary.

dash/dash.py
```diff
@@ -851,18 +874,34 @@ def dispatch(self):
...
-            ][0])
+            matched_state = {}
+            for c in state:
+                if (c['property'] == component_registration['property'] and
+                   c['id'] == component_registration['id']):
+                    matched_state = c
+                    break
+            matched_state_value = matched_state.get('value', None)
+            args.append(matched_state_value)
+            setattr(
```
rmarren1: Update component objects with the correct state when new callbacks come in.

dash/development/base_component.py
```diff
@@ -2,7 +2,19 @@
...
 import os
 import inspect
+import json
 import keyword
+import cerberus
+
+
+
+# Forward declated Component class, see below.
+class Component:
```
rmarren1: We must forward declare to add the `Component` type to our cerberus validator

dash/development/base_component.py
```diff
@@ -2,7 +2,19 @@
...
+    pass
+
+
+cerberus.Validator.types_mapping['component'] =\
```
rmarren1: Let us validate `Component` objects with cerberus (useful for `children` prop)

dash/development/base_component.py
```diff
@@ -87,6 +100,40 @@ def __init__(self, **kwargs):
                         ', '.join(sorted(self._prop_names))
                     )
                 )
+
+        # Make sure arguments have valid values
+        self._validator = cerberus.Validator(self._schema, allow_unknown=True)
```
rmarren1: Validate the initialization props of a `Component`

dash/development/base_component.py
```diff
@@ -265,6 +312,64 @@ def __len__(self):
         return length
 
 
+def js_to_cerberus_type(type_object):
```
rmarren1: Convert the `metadata.json` type objects to a cerberus schema

dash/development/base_component.py
```diff
@@ -382,7 +488,8 @@ def __repr__(self):
          p not in keyword.kwlist and
          p not in ['dashEvents', 'fireEvent', 'setProps']] + ['**kwargs']
     )
-
+    schema = {str(k): generate_property_schema(v)
```
rmarren1: Make cerberus schema for this dash component type.

dash/development/base_component.py
```diff
@@ -87,6 +100,40 @@ def __init__(self, **kwargs):
...
+                Errors:
+
+                {}
+            """.format(
+                self.__class__.__name__,
+                "\n\n".join([(
+                    "Invalid value {} for property {}.\n"
+                    "This broke the rule {} which should be {}\n"
+                    "Here is the expected schema for the {} prop:\n\n"
+                    "{}"
```
chriddyp: 🎉  this is so awesome

dash/development/base_component.py
```diff
@@ -265,6 +312,64 @@ def __len__(self):
...
+        'objectOf': lambda x: {
+            'type': 'dict',
+            'allow_unknown': False,
+            'schema': js_to_cerberus_type(x['value'])
+        },
+        'array': lambda x: {'type': 'list'},
+        'arrayOf': lambda x: {
+            'type': 'list',
+            'allow_unknown': False,
+            'schema': js_to_cerberus_type(x['value'])
```
chriddyp: I'm assuming that this is recursive in a sense? That is, that this could contain it's own `arrayOf` which could be itself an `arrayOf`, etc?

dash/development/base_component.py
```diff
@@ -87,6 +100,40 @@ def __init__(self, **kwargs):
...
+                 .format(
+                     e.value,
+                     e.schema_path[0],
+                     e.schema_path,
+                     e.constraint,
+                     e.schema_path[0],
+                     json.dumps(
+                         self._schema[e.schema_path[0]],
+                         indent=1
+                     ))
```
chriddyp: I've had a really hard time getting useful information from deeply nested errors with cerberus. Have you had better luck?

dash/development/base_component.py
```diff
@@ -265,6 +312,64 @@ def __len__(self):
         return length
 
 
+def js_to_cerberus_type(type_object):
+    converters = {
+        'None': lambda x: {},
+        'func': lambda x: {},
+        'custom': lambda x: {},
+        'node': lambda x: {'type': 'component'},
```
chriddyp: what should we do about `children`? Should component authors say that it's 
```
oneOfType([
     PropTypes.string, PropTypes.number, PropTypes.node, PropTypes.None,
     PropTypes.oneOf([PropTypes.string, PropTypes.number, PropTypes.node, PropTypes.None])
])
```

or something more complex?

dash/development/base_component.py
rmarren1: Errors coming from nested schemas seem to only give error information about the top schema (e.g. the second example given above), I think we may be able to get something better by digging into the `ValidationError` class (http://docs.python-cerberus.org/en/stable/errors.html#python-interfaces). I'll take a look and see what we can do.

dash/development/base_component.py
rmarren1: I think it would make sense to leave it up to authors to give more fine grained control. For example if there is a component that should only take a single dash component as a child (not strings or numbers) a user could use `PropTypes.element`.

For the more complicated things / things we already have validators for (like `figure`, we could use use the function validators pattern [here](http://docs.python-cerberus.org/en/stable/validation-rules.html#validator)).

To denote these in react, the `instanceOf` prop type would be useful. For example, something like `PropTypes.instanceOf(PlotlyFigure)` could trigger the figure validator to be used.

dash/development/base_component.py
rmarren1: Yes, it seems to be working well with recursive definitions.

Bachibouzouk:
Hi, I tried to `pip install` this branch locally in a venv to test this PR, but I the way I did it doesn't work as I get an error by running the example code you provide above : 
`AttributeError: 'RadioItems' object has no attribute '_schema'`

I typed the command `pip install git+https://github.com/rmarren1/dash.git@validate`, maybe it is not the way to proceed?

dash/development/base_component.py
```diff
@@ -2,7 +2,19 @@
 import copy
 import os
 import inspect
+import json
 import keyword
+import cerberus
+
+
+
+# Forward declated Component class, see below.
```
Bachibouzouk: did you mean 'declared'? PEP8 requires one space less :)

PS : this is my first review on dash code and I am mainly reading the code you wrote to learn more, my suggestion will mostly be about styling and spelling, I apologize in advance for that!

chriddyp: > my suggestion will mostly be about styling and spelling, I apologize in advance for that!

Note that we enforce PEP8 in our tests here: https://github.com/plotly/dash/blob/4d93652e40b55eb94c75359f5d47013051839ecf/.circleci/config.yml#L38-L39

So, it's not necessary to review PEP8 issues in our PRs as they will get fixed before it gets merged into master.

dash/development/base_component.py
chriddyp: Yeah, that sounds right. Looking at the `propTypes` docs, it seems like `children` should just always be a `node` and our `node` validation should handle all of the things that Dash can render
```
  // Anything that can be rendered: numbers, strings, elements or an array
  // (or fragment) containing these types.
  optionalNode: PropTypes.node,
```

chriddyp:
To validate the components in the callbacks, I think we'll need to pass the component name back in the `_dash-update-components` API call. With Dash's stateless backend, I think this will be the most reliable way to ensure that the front-end component that is getting updated is the same one that we are validating against.

So, that is, we could modify the Dash API to include `type` and `namespace` in the payload: https://github.com/plotly/dash-renderer/blob/0dc6f331eda404ee848145db71a4dc3bc35a759e/src/actions/index.js#L406-L408 and then do a component lookup in the callback decorator

rmarren1:
^ Definitely, I was going to add this to the next iteration. You don't need this to validate against components in the initial layout, but I don't think there is any other way to validate against dynamically generated components without updating state in the backend after initialization.

rmarren1:
https://github.com/plotly/dash-core-components/issues/257

rmarren1:
https://github.com/plotly/dash-renderer/pull/69

rmarren1:
For users to use validation they are going to have to upgrade `dash-renderer`. Anyone have thoughts on how we should address version incompatibilities. Two possibilities:
1. Users who update `dash` and not `dash-renderer` will have a fatal error thrown telling them to update `dash-renderer`
2. Users who update `dash` will have validation shut off until they update `dash-renderer`, and a warning will be printed when the app is started.

valentijnnieman:
@rmarren1 Very cool! If you can get solution 2 working, I'd say that that's a nice friendly way of letting users know that they should update. I'm guessing a fatal error would be the source of many new community forum posts!

chriddyp:
> Users who update dash and not dash-renderer will have a fatal error thrown telling them to update dash-renderer

I'd say that we just add a minimum version requirement to `dash-renderer`, similar to how we have a minimum version requirement to `Flask`:
https://github.com/plotly/dash/blob/7219b3d9f827e95bbbace801aa955b8c21ff8ffe/setup.py#L18-L23

rmarren1:
^ That sounds like a better solution

chriddyp:
This is starting to look really good! Thanks for providing the table. A few scattered comments:

**`null`**
We should think about how to handle `null`. Unfortunately, React prop-types hasn't made `null` a first class property. There is a long PR here: https://github.com/facebook/prop-types/pull/90 and here https://github.com/facebook/prop-types/pull/90 but it doesn't look promising. 

This will be important as some components are OK with this (e.g. `children`) and others might not be (e.g. `Dropdown.options`). It's also a pretty common pattern for some dash developers to have callbacks with a control statement that doesn't return anything, implicitly returning `None`.

Ideally, we'd have something like `oneOfType([PropTypes.null, PropTypes.arrayOf(PropTypes.string])`. I found this workaround: https://github.com/facebook/react/issues/2166#issuecomment-63986511, but I haven't tested it.

If that doesn't work, then we _might_ have to create our own prop-type library that, to start, would have support for `null` in some way. We could even expand on it in the future with things like `min`, `max`, and more. Other component authors would import this custom prop-type library to write their own components and we'd keep it compatible with our `cerberus` backend.

Hopefully there would be an easier intermediate step though.

**`oneOf`**
For that `oneOf` issue with the double strings, I found this issue: https://github.com/reactjs/react-docgen/issues/57. Doesn't look like they are going to fix it.

**component tests**
We'll need to start locking down our prop types in `dash-core-components`. As we do so, I'd like to create a set of tests that test all of the permutations of particular properties through initial renders _and_ through callback updates.

That is, if we say that `options: propTypes.oneOfType([propTypes.null, propTypes.object])`, then having 6 tests: test setting `options` to `[]`, `None`, and `[{'label': 'test', 'value': 'test'}]` on both initial render and on callback update. I can help write out the framework for these.

So psyched about this!

valentijnnieman:
@chriddyp TypeScript has null I believe, wink wink. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html

chriddyp:
> typeScript has null I believe, wink wink

It seems like typescript isn't supported by react-docgen though, which is what we use to serialize the props as JSON: https://github.com/reactjs/react-docgen/issues/67#issuecomment-203501239

`flow` is able to serialize some properties with `{"nullable": true}`

In any case, I think we'll still need to support propTypes for the greater dash community of component authors, who we shouldn't require to use typescript or flow.

rmarren1:
That workaround for `PropTypes.null` looks promising. Working pretty well except for this issue: https://github.com/pyeve/cerberus/issues/437, which would break components which want to accept nested lists and `null` values together (probably a rare case). I'll see if I can get a PR into Cerberus to fix.

rmarren1:
Looks like this is the winning schema for `children`
``` 
   children: PropTypes.oneOfType([
      PropTypes.string,
      PropTypes.number,
      PropTypes.bool,
      PropTypes.element,
      PropTypes.oneOf([null]),
      PropTypes.arrayOf(
        PropTypes.oneOfType([
          PropTypes.string,
          PropTypes.number,
          PropTypes.bool,
        PropTypes.element,
          PropTypes.oneOf([null])
        ])
      )
    ]),
```

rmarren1:
@plotly/dash Ready to review

dev-requirements.txt
```diff
@@ -14,3 +14,4 @@ plotly>=2.0.8
 requests[security]
 flake8
 pylint==1.9.2
+Cerberus==1.2
```
T4rk1n: You also need to add it in setup.py `install_requires` since it is directly used by the code.

dash/dash.py
```diff
@@ -890,7 +892,40 @@ def dispatch(self):
...
+            output['type'] = str(output['type'])
+            if output['namespace'] not in self.namespaces:
+                self.namespaces[output['namespace']] =\
+                    importlib.import_module(output['namespace'])
+            namespace = self.namespaces[output['namespace']]
+            component = getattr(namespace, output['type'])
+            # pylint: disable=protected-access
+            validator = DashValidator({
+                output['property']: component._schema.get(output['property'],
+                                                          {})
```
T4rk1n: Not pretty... put output['property'], {} on the line under.

rmarren1:
Need some way to address this: https://github.com/numpy/numpy/issues/2951... This caused that error on `Td`.

I'm in the process of working through the `dash-docs` example as a reference test. I'll make a PR to `dash-docs` with edits to things that failed validation, and I think I figured out how to speed up the initial layout validation so I'll update here when that is working.

chriddyp:
> Need some way to address this: numpy/numpy#2951... This caused that error on Td.

For numpy validation, you might want to check what @jonmmease did in plotly.py. plotly.py does validation on these types and dash and plotly.py share the same json encoder (`plotly.utils.PlotlyJSONEncoder`)

rmarren1:
Updates to the dash-docs that gets rid of the errors: https://github.com/plotly/dash-docs/pull/209. Still has those Plotly warnings, I'll figure out if we can suppress those

rmarren1:
Bumped to `dash==0.27.0rc9` with support for numpy / pandas

rmarren1:
https://github.com/plotly/dash/issues/413

rmarren1:
Some updates:
* I re-worked when the validation for the initial layout, so now it takes less time and happens within the `_validate_layout` function.
* You can now shut off component validation with `app.config.disable_component_validation = True`
* Released `dash-renderer==0.15.0rc1` which contains just the pre-reqs for component validation without dev tools, so you can use component libraries with react 15 (e.g. `dash-table-experiments`)
* Bumped `dash`, `dash-core-components`, and `dash-html-components`, posted new versions at top level.
* Now, deep validation is done for callback outputs to a `children` property. Before, an error would only be thrown if the output was not a number, string, None, boolean, component, or list of those. Now, the new components that appear in this output are also validated as if they were just initialized.

vanbenschoten:
Want to give a big +1 for this change. This validation works fantastically well for our use cases. Thanks all!

chubukov:
On `0.29.0-rc6` I get a validation error if I try to create an input component with numeric type.

e.g. ```app.layout = html.Div([
    dcc.Input(id='input1',type='number',value=5) 
    ]
)```
also happens for min and max.

Sorry if this is a known issue -- haven't kept track of this development, just wanted to try it out. Love the idea.

rmarren1:
@chubukov Thanks for the report! I can reproduce this, looks like the `Input` component needs its PropTypes updated, as right now it says `value`, `min`, and `max` can only be strings. I'll let you know when that is fixed in the `dash-core-components` pre-release.

T4rk1n:
@rmarren1 Please add cerberus in setup.py install_requires.

setup.py
```diff
@@ -21,6 +21,7 @@
         'flask-compress',
         'plotly',
         'dash_renderer',
+        'Cerberus'
```
rmarren1: It's there

T4rk1n: ah just not in dash==0.27.0rc9

T4rk1n:
`dash-table` sorting prop did does not validate properly. 

Here's the definition:
`sorting: PropTypes.oneOf(['fe', 'be', true, false]),`

my value was `True`.

dash/dash.py
```diff
@@ -914,7 +924,85 @@ def dispatch(self):
                 c['id'] == component_registration['id']
             ][0])
 
-        return self.callback_map[target_id]['callback'](*args)
+        output_value = self.callback_map[target_id]['func'](*args)
+
+        # Only validate if we get required information from renderer
+        # and validation is not turned off by user
+        if (
+                (not self.config.disable_component_validation) and
```
T4rk1n:
```suggestion
                not self.config.disable_component_validation and
```

dash/dash.py
```diff
@@ -931,6 +1019,11 @@ def _validate_layout(self):
 
         component_ids = {layout_id} if layout_id else set()
         for component in to_validate.traverse():
+            if (
```
T4rk1n: Normally there is no `()` after if in python.

dash/development/base_component.py
```diff
@@ -223,17 +232,66 @@ def traverse_with_paths(self):
...
+                The errors in validation are as follows:
+
+
+            """.replace('    ', '').format(
+                component_name=self.__class__.__name__,
+                component_args=pprint.pformat(args),
+                component_schema=pprint.pformat(self.__class__._schema)
+            )
+
+            # pylint: disable=protected-access
```
T4rk1n: :zap:

dash/development/base_component.py
```diff
@@ -265,6 +323,137 @@ def __len__(self):
...
+    if type_object:
+        if type_object.get('name', None) == 'enum':
+            values = type_object['value']
+            for v in values:
+                value = v['value']
+                if value == 'null':
+                    return True
+        if type_object.get('name', None) == 'union':
+            values = type_object['value']
+            if any([schema_is_nullable(v) for v in values]):
```
T4rk1n:
```suggestion
            if any(schema_is_nullable(v) for v in values):
```

Don't need the duplicate list, generator is better for any since it returns as one is true.

dash/development/base_component.py
```diff
@@ -265,6 +323,137 @@ def __len__(self):
...
 
+def schema_is_nullable(type_object):
+    if type_object:
+        if type_object.get('name', None) == 'enum':
+            values = type_object['value']
+            for v in values:
+                value = v['value']
+                if value == 'null':
+                    return True
+        if type_object.get('name', None) == 'union':
```
T4rk1n:
```suggestion
        if type_object.get('name') == 'union':
```

Default to get is already None.

dash/development/base_component.py
```diff
@@ -265,6 +323,137 @@ def __len__(self):
...
+
+def generate_property_schema(jsonSchema):
+    schema = {}
+    type_object = jsonSchema.get('type', None)
+    required = jsonSchema.get('required', None)
+    propType = js_to_cerberus_type(type_object)
+    if propType:
+        schema.update(propType)
+    if schema_is_nullable(type_object):
+        schema.update({'nullable': True})
```
T4rk1n:
```suggestion
        schema['nullable'] = schema_is_nullable(type_object)
```

dash/development/base_component.py
```diff
@@ -265,6 +323,137 @@ def __len__(self):
...
+    if required:
+        schema.update({'required': True})
```
T4rk1n:
```suggestion
        schema['required'] = required
```

rmarren1:
I added support for Booleans in `enum`, I'll see if I can get the `dash-table` to work.

T4rk1n:
:dancer: once the toggle is integrated with the dev tools.

rmarren1:
I think we can keep this separate from dev-tools, and treat it like the current callback exceptions. Actually, I think we should change the name match, e.g.
```
app.config.suppress_callback_exceptions = True
app.config.suppress_validation_exceptions = True
```
dev-tools would just make the exceptions appear on the front-end.

Also, I don't think we have discussed whether we want this feature to be enabled by default. Even if this code is 100% correct, it can still break people's custom components if they are not using them properly (e.g. using `PropTypes.number` for a string property wouldn't cause an error currently in Dash, but this PR would make it error).

I think we can enable by default add a note to all the error messages it throws explaining how to turn validation off if it is breaking an app (kind of like how we currently do that with the callback exceptions).

T4rk1n:
Changing that config name is a breaking change.

T4rk1n:
Integrating the validation with the dev tools is the way to go for this, we don't want validation on prod servers, it should be enabled by default when debug=True and False otherwise.

rmarren1:
`suppress_callback_exceptions` is the current name, I would just be changing the name of the config to disable validation from  `disable_component_validation` to `suppress_validation_exceptions`.

T4rk1n:
Oh ok, I would still integrate with the dev tools so just setting debug=True is required for it.

rmarren1:
Oh I thought you meant something different, yeah it should be disabled by default in production.

T4rk1n:
@rmarren1 How would the component as prop validate ? proptype is `PropTypes.node`.

rmarren1:
@T4rk1n There is a custom type called `component` which makes sure objects are an instance of `dash.development.base_component.Component`, `PropTypes.node` works with this custom type:
https://github.com/plotly/dash/blob/588ce3df16bfea5fa917ea49656d14587901fe72/dash/development/base_component.py#L380-L397

dash/dash.py
```diff
@@ -130,9 +130,9 @@ def __init__(
                 'suppress_callback_exceptions',
                 suppress_callback_exceptions, env_configs, False
             ),
-            'disable_component_validation': _configs.get_config(
-                'disable_component_validation',
-                disable_component_validation, env_configs, False
+            'suppress_validation_exceptions': _configs.get_config(
```
T4rk1n: 😸

dash/dash.py
```diff
@@ -930,7 +930,7 @@ def dispatch(self):
         # Only validate if we get required information from renderer
         # and validation is not turned off by user
         if (
-                (not self.config.disable_component_validation) and
+                (not self.config.suppress_validation_exceptions) and
```
T4rk1n: Remove the () on a single condition.

dash/dash.py
```diff
@@ -950,11 +950,8 @@ def dispatch(self):
     def _validate_callback_output(self, namespace, component_type,
                                   component_id, component_property,
                                   callback_func_name, args, value):
-        if namespace not in self.namespaces:
-            self.namespaces[namespace] =\
-                importlib.import_module(namespace)
-        namespace = self.namespaces[namespace]
-        component = getattr(namespace, component_type)
+        module = sys.modules[namespace]
```
T4rk1n: 😸

dash/dash.py
```diff
@@ -988,12 +990,17 @@ def _validate_callback_output(self, namespace, component_type,
...
-            raise exceptions.CallbackOutputValidationError(
-                generate_validation_error_message(
-                    validator.errors, 0, error_message))
+            error_message = generate_validation_error_message(
+                    validator.errors,
+                    0,
+                    error_message
+            ) + dedent("""
+                You can turn off these validation exceptions by setting
+                `app.config.suppress_validation_exceptions=True`
```
T4rk1n: 😸

dash/development/base_component.py
```diff
@@ -280,18 +280,23 @@ def validate(self):
...
                 component_schema=pprint.pformat(self.__class__._schema)
             )
 
+            error_message = generate_validation_error_message(
+                    validator.errors,
+                    0,
+                    error_message
+            ) + dedent("""
+                You can turn off these validation exceptions by setting
+                `app.config.suppress_validation_exceptions=True`
```
T4rk1n: It the same message as in dash, can we make that :camel:

dash/dash.py
rmarren1: Not sure how else to style since the if statement is > 80 characters. PEP isn't super specific about what to do here: https://www.python.org/dev/peps/pep-0008/#multiline-if-statements.

dash/development/base_component.py
rmarren1: :cat:

dash/development/base_component.py
rmarren1: :cat:

rmarren1:
Closing in favor of #452, which is the same PR but with a more readable commit history. Will link back to here for the relevant discussions.

PR #346, opened 2018-08-24
Make `Dash` instances use `name` attr of Flask server if one is provided

### LINKED ISSUE #345: Assets don't work for Dash apps created with an existing Flask server
This is because the `name` param is taken to be the default value from the constructor (namely `__main__`), even if a Flask instance is provided. In this case, the `name` used to locate the base path of the package should be updated to be the `name` attribute of the Flask instance.

ned2:
Fixes #345

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -96,6 +96,7 @@ class Dash(object):
                 See https://github.com/plotly/dash/issues/141 for details.
                 ''', DeprecationWarning)
 
+        name = name if server is None else server.name
         self._assets_folder = assets_folder or os.path.join(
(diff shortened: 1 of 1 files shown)


alistairwelchJD:
This makes sense, but caused an issue for me when packaging the flask server outside of the root folder. The fix was to set assets_folder='assets', setting the value to the default to change behaviour isn't very intuitive, perhaps could be fixed just with some extra documentation on dash.Dash API reference, explaining what's been done here. Also I think this causes issues for favicon too, even after applying this assets_folder='assets' fix, the app is still using the flask root to try find the favicon

PR #350, opened 2018-08-24
Use HTML5 syntax for the meta tag

jbampton:
Examples of the HTML5 meta tags here -> https://www.w3schools.com/tags/tag_meta.asp
shows that it does not have an ending slash.


T4rk1n:
I missed that one. We had agreed that the meta tags should not have a slash.

PR #351, opened 2018-08-24
Prefix assets with `requests_pathname_prefix`.

### LINKED ISSUE #349: Assets are not prefixed with `requests_pathname_prefix`.
The assets files should be prefixed with the `requests_pathname_prefix` like other resources.

T4rk1n:
Fix #349, the assets were not prefixed with the `requests_pathname_prefix` resulting in 404 on the dash deployment server.

- Add `get_asset_url` method to dash, retrieve the url of an asset by it's filename (eg: `app.get_asset_url('style.css')` will return `/assets/style.css`), it is automatically prefixed with `requests_pathname_prefix`.


chriddyp:
Can we add a test? Otherwise 💃

T4rk1n:
@chriddyp 

I added a test and found out that it was not working as expected in some cases, thanks.

ned2:
Looks good to me :dancer:

PR #362, opened 2018-08-31
Hot reload

### LINKED ISSUE #66: Hot reloading
I love that the Python module reloads automatically on save but I still need to refresh my browser when I make a change to the Python code, which means I lose the state I was in and need to fiddle with the UI to get the view back to how it was before I made the change. It would be great if we could leverage a system like [React Hot Loader](https://github.com/gaearon/react-hot-loader) or something similar for totally seamless multi-monitor development. Just type code on one monitor, save, and turn your head to to see the changes on the other monitor ;)

### LINKED ISSUE #341: Assets not included when added when the app is running
PR Reference: #286 

At the moment, the only way to refresh the changes to the assets is to restart the app, and open up a new browser.

T4rk1n:
**Available with the following versions:**
- `dash==0.29.0rc4`
- `dash-renderer==0.15.0rc2`

Reload the browser when files are changed.

- Use the flask debug reloader to reset the hash on python file change.
- Watch the assets files and change the hash on assets modified.
- Add a route to serve the reload hash.

Dev tools params (in `app.run_server` or `app.enable_dev_tools`

- `dev_tools_hot_reload`, bool set to true to enable hot reload (default=False).
- `dev_tools_hot_reload_interval`, int, interval at which the renderer will request the reload hash (default=3000).
- `dev_tools_hot_reload_watch_interval`, float, The delay between each walk of the assets folder to detect file changes.
- `dev_tools_silence_routes_logging`, disable the route logging.

Hot reload is activated by default when `debug=True` !!

Need plotly/dash-renderer#73.

Resolves #66.
Fix #341.


T4rk1n:
@plotly/dash Ready for reviews!

dash/dash.py
```diff
@@ -217,6 +225,24 @@ def add_url(name, view_func, methods=('GET',)):
...
+            if hot_reload else ''
+        self._reload_interval = hot_reload_interval
+        self._hard_reload = False
+        self._lock = threading.RLock()
+        self._watch_thread = None
+        if hot_reload:
+            self._watch_thread = threading.Thread(
+                target=lambda: _watch.watch([self._assets_folder],
+                                            self._on_assets_change,
+                                            sleep_time=0.5))
```
rmarren1: Shouldn't `sleep_time` be `hot_reload_interval / 1000`? I see that `hot_reload_interval` is passed to the front-end via `config`, but I'm not sure if it is intended to be the same on the back-end. If not (e.g. back-end should have a fixed constant refresh interval regardless of `hot_reload_interval` value) I think you should make this a constant set in the `_watch` file.

dash/dash.py
```diff
@@ -217,6 +225,24 @@ def add_url(name, view_func, methods=('GET',)):
...
+            self._watch_thread.daemon = True
+            self._watch_thread.start()
+
+            if not self.server.debug:
+                self.server.debug = True
```
rmarren1: I think this would get over-ridden when somebody calls `app.run_server(debug=False)`. Should that happen? I think in `run_server` you can just say `debug = debug or self.hot_reload`

dash/dash.py
```diff
@@ -276,10 +302,21 @@ def serve_layout(self):
...
+        if self._reload_hash:
+            config['hot_reload'] = {
+                'interval': self._reload_interval
+            }
+        return config
+
+    def serve_reload_hash(self):
+        return flask.jsonify({
+            'reloadHash': self._reload_hash,
+            'hard': self._hard_reload
```
rmarren1: It doesn't look like the 'hard' variable is used at all in the front-end. Can we just remove it here, in `_on_assets_change` and in `__init__`?

dash/dash.py
```diff
@@ -949,6 +986,12 @@ def get_asset_url(self, path):
             self.config.routes_pathname_prefix,
             path)
 
+    def _on_assets_change(self, _):
```
rmarren1: Why pass the argument here? Can't we just call `on_change()`

dash/dash.py
T4rk1n: It is used to trigger a soft reload.
https://github.com/plotly/dash-renderer/blob/71a42e2f63880ec8ce2cbc01a8e18e582857f97d/src/components/core/Reloader.react.js#L35

dash/dash.py
T4rk1n: The sleep time is the poll time for the assets watch thread and `hot_reload_interval` is for the client requests interval. I don't think the user will need to change the sleep time so it can be a constant.

dash/dash.py
T4rk1n: The watch function I got here is from another project of mine that I adapted to run on threads instead of asyncio. When a file changed in the other project, I would log the filename thus giving the changed filename as argument in the callback. But we don't have a logging system in-place so I just put a pass param. I guess I could remove the options we're not gonna use here.

dash/dash.py
rmarren1: Oh okay, I see. So is this variable just set to true on the first asset change? I see it gets set to True in `_on_assets_change` but don't see it getting set back anywhere

T4rk1n: Good catch, need to reset it after serving the response.

dash/dash.py
T4rk1n: I don't think it matter much, it just the timer for when to run the check in the back end and should not be linked in any way to the request timer that can be changed. It should be a constant that work goods in the conditions. We don't want run that check too often because of the `GIL` and it should still be low so you can get the reload a short while after making changes. So I think a check every half a second is good with this in mind.

In my `asyncio` based implementation it was a really low number like `yield from asyncio.sleep(0.000001)` and it didn't make a difference in the performance of the app.

dash/_watch.py
```diff
@@ -0,0 +1,26 @@
...
+    watched = collections.defaultdict(lambda: -1)
+
+    def walk():
+        for folder in folders:
+            for current, _, files, in os.walk(folder):
+                for f in files:
+                    if pattern and not pattern.search(f):
+                        continue
+                    path = os.path.join(current, f)
+                    info = os.stat(path)
```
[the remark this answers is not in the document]
T4rk1n: Yea, some methods in that module seems to be wrapper around `os.stat`. I'm used to `os.stat` so I just do that.

dash/dash.py
T4rk1n: I think I will remove that from here and add it in the enable method from #369.

T4rk1n:
If I add a new component library in layout, the JS bundle doesn't get served.

That's a hard one. Since it's a soft reload, it won't reload the index and the new resources doesn't get served.

The solution I am thinking of:

1. Keep a list of the dependencies in a list in the store, available in a API.requestComponentsLib.
2. Check the components libs scripts tags are all included in the requestComponentLib response, if not append it to the footer and wait for it to be loaded before triggering a soft reload. 

>  if I created a new CSS file, I had to restart the app before I saw it.

I think that is related to #341. The assets folder is walked only once during initialization, my initial fix was to walk every time the index is served. But then I thought it would be better to integrate the watch of `hot-reload` and add it to the assets resources since that fix is not needed for deployments and incur a performance cost.

> It seems like the developer-facing API is small (just the hot_reload tag),

There's also `hot_reload_interval`, and maybe `watch_interval` that can be added.

> An approach where CSS files are reloaded separately without a full page reload.

Could do that by disabling the old stylesheets and append the new one with the modified timestamp in the head.

chriddyp:
> > If I add a new component library in layout, the JS bundle doesn't get served.

> That's a hard one. Since it's a soft reload, it won't reload the index and the new resources doesn't get served.

Another approach might be to solve it by solving https://github.com/plotly/dash-renderer/issues/46. I think that issue will be a lot more challenging to solve though.

T4rk1n:
@chriddyp 

I solved it by including the keys of `dash.registered_paths` in the reload request and checking that the length and content is the same, if not hard reload. Was easier than I thought.

For plotly/dash-renderer#46, I think we would have to add a method to register the packages externally from the layout collection so they can be included even if not in the initial layout.

T4rk1n:
> An approach where CSS files are reloaded separately without a full page reload

@chriddyp I added that functionality.

T4rk1n:
Updated the rc versions:

```
dash==0.29.0rc3
dash-renderer==0.15.0rc2
```

chriddyp:
This latest version is looking really great. The new CSS pathway is especially nice 🎉 
hot-reload-demo

mbkupfer:
I'm having trouble with running this. When I start my app, it goes into an endless loop with the following error messages:

```

  File "C:\Users\mkupfer\Desktop\MYDOCU~1\01_VIS~1\PAY-RA~1\venv\lib\site-packag
es\flask\app.py", line 1718, in handle_user_exception
  File "C:\Users\mkupfer\Desktop\MYDOCU~1\01_VIS~1\PAY-RA~1\venv\lib\site-packag
es\flask\_compat.py", line 35, in reraise
    raise value
  File "C:\Users\mkupfer\Desktop\MYDOCU~1\01_VIS~1\PAY-RA~1\venv\lib\site-packag
es\flask\app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "C:\Users\mkupfer\Desktop\MYDOCU~1\01_VIS~1\PAY-RA~1\venv\lib\site-packag
es\flask\app.py", line 1799, in dispatch_request
  File "C:\Users\mkupfer\Desktop\MYDOCU~1\01_VIS~1\PAY-RA~1\venv\lib\site-packag
es\dash\dash.py", line 336, in serve_reload_hash
    'files': list(changed)
  File "C:\Users\mkupfer\Desktop\MYDOCU~1\01_VIS~1\PAY-RA~1\venv\lib\site-packag
es\flask\json\__init__.py", line 321, in jsonify
    dumps(data, indent=indent, separators=separators) + '\n',
  File "C:\Users\mkupfer\Desktop\MYDOCU~1\01_VIS~1\PAY-RA~1\venv\lib\site-packag
es\flask\json\__init__.py", line 179, in dumps
    rv = _json.dumps(obj, **kwargs)
  File "C:\Users\mkupfer\AppData\Local\Programs\Python\Python36-32\Lib\json\__in
it__.py", line 238, in dumps
    **kw).encode(obj)
  File "C:\Users\mkupfer\AppData\Local\Programs\Python\Python36-32\Lib\json\enco
der.py", line 201, in encode
    chunks = list(chunks)
  File "C:\Users\mkupfer\AppData\Local\Programs\Python\Python36-32\Lib\json\enco
der.py", line 430, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "C:\Users\mkupfer\AppData\Local\Programs\Python\Python36-32\Lib\json\enco
der.py", line 404, in _iterencode_dict
    yield from chunks
  File "C:\Users\mkupfer\AppData\Local\Programs\Python\Python36-32\Lib\json\enco
der.py", line 437, in _iterencode
    o = _default(o)
  File "C:\Users\mkupfer\Desktop\MYDOCU~1\01_VIS~1\PAY-RA~1\venv\lib\site-packag
es\flask\json\__init__.py", line 81, in default
    return _json.JSONEncoder.default(self, o)
  File "C:\Users\mkupfer\AppData\Local\Programs\Python\Python36-32\Lib\json\enco
der.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'dict_keys' is not JSON serializable
 
```

T4rk1n:
@mbkupfer thanks for reporting, I think I know what's wrong, will post update soon.

T4rk1n:
@mbkupfer Released dash==0.29.0rc4, should fix the json issue on py3.

T4rk1n:
@mbkupfer I added an option to disable the logger output (#382) since the reload requests spam the console. You can enable it back with `dev_tools_silence_route_logging`, I'll look in making the debugger pin available with that option enabled.

mbkupfer:
Got it. Thanks!

T4rk1n:
@plotly/dash need approval.

dash/_watch.py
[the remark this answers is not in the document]
rmarren1: :+1:, either way is good, just seemed a bit strange, seems to break `"There should be one-- and preferably only one --obvious way to do it".`

dash/dash.py
```diff
@@ -947,19 +994,20 @@ def _setup_server(self):
         self._generate_scripts_html()
         self._generate_css_dist_html()
 
+    def _add_assets_resource(self, url_path, file_path):
```
rmarren1: :cat:

dash/dash.py
```diff
@@ -1014,17 +1062,45 @@ def get_asset_url(self, path):
...
+                         dev_tools_hot_reload_watch_interval=None,
+                         dev_tools_hot_reload_max_retry=None,
+                         dev_tools_silence_routes_logging=None):
         """
         Activate the dev tools, called by `run_server`. If your application is
         served by wsgi and you want to activate the dev tools, you can call
         this method out of `__main__`.
 
-        :param debug: If True, then activate all the tools unless specified.
+        If an argument is not provided, it can be set with environ variables.
```
rmarren1: Why 'environ' versus 'environment'?

dash/dash.py
```diff
@@ -1014,17 +1062,45 @@ def get_asset_url(self, path):
...
+        :type dev_tools_hot_reload: bool
+        :param dev_tools_hot_reload_interval: Interval at which the client will
+            request the reload hash. Available as `DASH_HOT_RELOAD_INTERVAL`
+            environ var.
+        :type dev_tools_hot_reload_interval: int
+        :param dev_tools_hot_reload_watch_interval: Interval at which the
+            assets folder are walked for changes. Available as
+            `DASH_HOT_RELOAD_WATCH_INTERVAL` environ var.
+        :type dev_tools_hot_reload_watch_interval: float
+        :param dev_tools_hot_reload_max_retry: Maximum amount of retries before
```
rmarren1: Were these two (dev_tools_hot_reload_max_retry, dev_tools_silence_routes_logging) purposefully omitted from the variables you can set using the environment? It would be nice to have the full functionality via either method.

dash/dash.py
```diff
@@ -1035,12 +1111,111 @@ def enable_dev_tools(self,
...
+
         return debug
 
+    # noinspection PyProtectedMember
+    def _on_assets_change(self, filename, modified, deleted):
+        self._lock.acquire()
+        self._hard_reload = True
+        self._reload_hash = _generate_hash()
+
+        asset_path = filename.replace(self._assets_folder, '')\
```
rmarren1: Can we use the proper `os.path` manipulations here?

```
>>> import os
>>> first_path = os.path.join('project', 'assets')
>>> second_path = os.path.join('project', 'assets', 'pictures', 'dog_pictures')
>>> os.path.commonpath([first_path, second_path])
'project/assets'
>>> os.path.relpath(second_path, os.path.commonpath([first_path, second_path]))
'pictures/dog_pictures'
```

dash/dash.py
```diff
@@ -1035,12 +1111,111 @@ def enable_dev_tools(self,
...
+            if filename.endswith('js'):
+                self.scripts.append_script(res)
+            elif filename.endswith('css'):
+                self.css.append_css(res)
+
+        if deleted:
+            if filename in self._assets_files:
+                self._assets_files.remove(filename)
+
+            def delete_resource(resources):
```
rmarren1: Can this be a bit more pythonic? It was a bit hard to read through the first time.

>>> to_delete = next(filter(lambda r: r.get('asset_path') == asset_path, resources), None)
>>> if to_delete:
...     resources.remove(to_delete)

rmarren1:
Looks great from a code standpoint :dancer:. I tried it out for a bit with a small app and it worked well!

I made a few more comments in addition to my old ones. They are all style and shouldn't change functionality.

dash/dash.py
T4rk1n: `os.environ`, tho environment variables sounds better.

dash/dash.py
T4rk1n: I must admit it looks weird, but `filter` is not pythonic according to Guido. It would also iterate over all the array while the `for` loop stops once found.

T4rk1n:
Had to remove the resources cache, a proper cache bust check was more expansive than remapping them.

T4rk1n:
@chriddyp There is an issue for the dev tools https://github.com/plotly/dash-docs/issues/196

PR #365, opened 2018-09-04
Fix url_base_pathname init configs

### LINKED ISSUE #364: Dash's config validation prevents `requests_pathname_prefix` and `requests_pathname_prefix` from being set
Trying to supply values for either of `requests_pathname_prefix` and `requests_pathname_prefix` results in an error:

```
dash.exceptions.InvalidConfig:
    You supplied `url_base_pathname` and `requests_pathname_prefix`. This is ambiguous.
    To fix this, set `routes_pathname_prefix` instead of `url_base_pathname`.
```

However I did not supply a value for `url_base_pathname`. Looks like this is because `_configs.pathname_configs` is checking to see if  `url_base_pathname` is not None to indicate that a user supplied their own value, but Dash sets a default value of '/', so this logic doesn't work.

T4rk1n:
- Set `url_base_pathname` to None in `dash.__init__` Fix #364 .
- Fix circleci bad cache key (was the same key for both python 3.6 and python 2.7 causing issues).


ned2:
Cool, didn't realise the fix was that simple. I can't comment on the config.yml changes, but that fix to url_base_pathname default looks good. :dancer:

PR #367, opened 2018-09-05
Standalone dash renderer (for custom hooks) - March 1

valentijnnieman:
This enables the `dash-renderer` to be called as a standalone class, so that we can pass config objects to it. 

Depends on [this PR that enables dash-renderer to be a standalone class](https://github.com/plotly/dash-renderer/pull/75) to be merged (also see this PR for an example of how to use a custom `DashRenderer` instance).


dash/dash.py
```diff
@@ -462,7 +474,7 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
 
     def interpolate_index(self,
                           metas='', title='', css='', config='',
-                          scripts='', app_entry='', favicon=''):
+                          scripts='', app_entry='', favicon='', renderer=''):
```
chriddyp: we need to add this to the docstring below as well

tests/test_integration.py
```diff
@@ -532,3 +535,82 @@ def create_layout():
...
+                {%favicon%}
+                {%css%}
+            </head>
+            <body>
+                <div>Testing custom DashRenderer</div>
+                {%app_entry%}
+                <footer>
+                    {%config%}
+                    {%scripts%}
+                    <script id="_dash-renderer" type"application/json">
```
chriddyp: `type="application/javascript">`

tests/test_integration.py
```diff
@@ -532,3 +535,82 @@ def create_layout():
 
         self.startServer(app)
         time.sleep(0.5)
+
+    def test_with_custom_renderer(self):
+        app = dash.Dash(__name__)
+
+        app.index_string = '''
```
chriddyp: Can we add another test that does the same thing but uses the `interpolate_index` method?

tests/test_integration.py
```diff
@@ -532,3 +535,82 @@ def create_layout():
...
+            )
+        ])
+
+        call_count = Value('i', 0)
+
+        @app.callback(Output('output-1', 'children'), [Input('input', 'value')])
+        def update_output(value):
+            call_count.value = call_count.value + 1
+            return value
+        call_count = Value('i', 0)
```
chriddyp: ⚡️ duplicate call count

chriddyp: There actually aren't any assertions for `call_count` below and I don't think it would be necessary to do a `call_count` assertion in here, so I think we can just remove `call_count` all together

dash/dash.py
```diff
@@ -378,6 +381,13 @@ def _generate_config_html(self):
             '</script>'
         ).format(json.dumps(self._config()))
 
+    def _generate_renderer(self):
+        return (
+            '<script id="_dash-renderer" type"application/json">'
```
chriddyp: `type="application/javascript"`

chriddyp:
This side of the code is looking good to me! @T4rk1n could you take a look as well?

dash/dash.py
```diff
@@ -440,7 +451,8 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
 
         index = self.interpolate_index(
             metas=metas, title=title, css=css, config=config,
-            scripts=scripts, app_entry=_app_entry, favicon=favicon)
+            scripts=scripts, app_entry=_app_entry, favicon=favicon,
+            renderer=renderer)
 
         checks = (
             (_re_index_entry_id.search(index), '#react-entry-point'),
```
T4rk1n: Can we add a check for the new renderer, it should check that `new DashRenderer` is somewhere in the renderer tag.

dash/dash.py
```diff
@@ -217,6 +217,9 @@ def add_url(name, view_func, methods=('GET',)):
         self._cached_layout = None
         self.routes = []
 
+        # default renderer string
+        self.renderer = 'const renderer = new DashRenderer();'
```
T4rk1n: Should use `var` since it's not transpiled.

tests/test_integration.py
valentijnnieman: Good point. I'm having a bit of trouble writing a test for this - could you take a look at the included test `test_with_custom_renderer_interpolated`? I'm not sure what's going on there, I'm getting JS errors about `window.push('load_after1')` not working for some reason.

dash/dash.py
```diff
@@ -378,6 +381,13 @@ def _generate_config_html(self):
             '</script>'
         ).format(json.dumps(self._config()))
 
+    def _generate_renderer(self):
+        return (
+            '<script id="_dash-renderer" type"application/javascript">'
```
chriddyp: `type=`

tests/test_integration.py
```diff
@@ -532,3 +535,153 @@ def create_layout():
...
+                {%favicon%}
+                {%css%}
+            </head>
+            <body>
+                <div>Testing custom DashRenderer</div>
+                {%app_entry%}
+                <footer>
+                    {%config%}
+                    {%scripts%}
+                    <script id="_dash-renderer" type"application/javascript">
```
chriddyp: `type=`

tests/test_integration.py
```diff
@@ -532,3 +535,153 @@ def create_layout():
...
+        favicon = ''
+        _app_entry = '''
+        <div id="react-entry-point">
+            <div class="_dash-loading">
+                Loading...
+            </div>
+        </div>
+        '''
+        renderer = '''
+        <script id="_dash-renderer" type"application/javascript">
```
chriddyp: `type=`

tests/test_integration.py
```diff
@@ -532,3 +535,153 @@ def create_layout():
...
+            renderer=renderer)
+
+        self.startServer(app)
+
+        input1 = self.wait_for_element_by_id('input')
+        input1.clear()
+
+        input1.send_keys('fire request hooks')
+
+        output1 = self.wait_for_element_by_id('output-1')
```
chriddyp: these items should already be on the page at this point, so I don't think that there will be any waiting that will happen. However, `send_keys` will fire off several async callback functions, so you will likely need to _wait_ for `output1.text` text to equal "fire request hooks". In some other dash projects, we had a utility `wait_for_text_to_equal` that would be good here, or there might be something native in selenium

dash/dash.py
valentijnnieman: `const` is already supported in all major browsers I believe.

T4rk1n: It may be supported in current versions, but if someone has an old version it won't work. The default should use `var` for maximum compatibility.

Akronix:
Hi. Can I ask how is the status of this? Is it in your roadmap or is there any tentative date for its release?

dash/dash.py
```diff
@@ -464,6 +467,13 @@ def _generate_config_html(self):
             '</script>'
         ).format(json.dumps(self._config()))
 
+    def _generate_renderer(self):
+        return (
+            '<script id="_dash-renderer" type="application/javascript">'
+            '{}'
+            '</script'
```
alexcjohnson: missing `>`?

dash/dash.py
alexcjohnson: I see at least one reference to us testing in IE10 https://community.plot.ly/t/browsers-compatibility/5047/2 so yeah, better drop it to `var`.

dash/dash.py
```diff
@@ -589,19 +601,22 @@ def interpolate_index(self, **kwargs):
...
-                        scripts=kwargs.get('scripts'))
+                        scripts=kwargs.get('scripts'),
+                        renderer=kwargs.get('renderer'))
 
         :param metas: Collected & formatted meta tags.
         :param title: The title of the app.
         :param css: Collected & formatted css dependencies as <link> tags.
         :param config: Configs needed by dash-renderer.
         :param scripts: Collected & formatted scripts tags.
+        :param renderer: The DashRenderer instance.
```
alexcjohnson: For clarity, perhaps:
`:param renderer: A script tag that instantiates the DashRenderer.`

requirements.txt
```diff
@@ -0,0 +1,47 @@
+appnope==0.1.0
```
alexcjohnson: Are this and the two new dev-requirements files needed for something?

tests/test_integration.py
```diff
@@ -26,6 +26,32 @@ def wait_for_element_by_id(id):
             return self.driver.find_element_by_id(id)
         self.wait_for_element_by_id = wait_for_element_by_id
 
+
+    def wait_for_element_by_css_selector(self, selector):
```
alexcjohnson: We've got these two methods in the base class already, do the versions there not suffice?

https://github.com/plotly/dash/blob/1249ffbd051bfb5fdbe439612cbec7fa8fff5ab5/tests/IntegrationTests.py#L24-L33

alexcjohnson: Oh I see - these were actually added to the base class after this PR was opened. Presumably the versions you added here can just be taken out now.

requirements.txt
valentijnnieman: Hmm nope - I'm guessing these files were probably moved and my branch still had the old files here. Merging with master was quite an ordeal since it was such an old PR - seems like I still made some mistakes merging here and there! I'll clean this up.

alexcjohnson: OK great! yes, we will do our best avoid such long-festering PRs in the future!

valentijnnieman: Yeah no worries! :)

PR #369, opened 2018-09-06
Add unminified components bundle support.

### LINKED ISSUE #313: Serve unminified JS bundles in dev mode
In Dash Dev Tools (https://github.com/plotly/dash/issues/292), developers will see the JS tracebacks when they develop their app, they won't need to dive into the browser console.

These tracebacks are pretty hard to debug with our minified bundles. We should serve the unminified bundles and/or source maps in some kind of dev-mode

T4rk1n:
Add support for serving unminified component bundles. Resolves #313.

Enable with:

```python
app = dash.Dash(__name__, serve_dev_bundles=True)
```

The components libs needs a new key in the their `_js_dist`: `dev_package_path`.

- [x] dash-html-components
- [x] dash-core-components
- [x] dash-renderer


rmarren1:
This looks great! I think we should actually turn this feature on when a user is in a global "dev" mode, where they would also turn on the [Dash Dev Tools](https://github.com/plotly/dash-renderer/pull/64) features. I think an argument to `dash.Dash` is the most intuitive way to turn such a thing on.

I don't have a method in Dev Tools just yet. Maybe we should change `serve_dash_bundles` to something more general like `use_dev_tools`, then I can piggy back off this flag in the Dev Tools PR and turn those on when it is true.

T4rk1n:
@rmarren1 So you want to combine the flag to serve the dev bundle and activate the devtools ? It make sense because the stacktraces will be much better with dev bundles.

I think we should call the keyword argument `enable_dev_tools` or just `dev_tools`.

rmarren1:
Yeah either of those sound good. Yeah the stack traces are generally unreadable for the production bundles, so Dev tools should definitely depend on them

chriddyp:
> I think an argument to dash.Dash is the most intuitive way to turn such a thing on.

One issue with this is that users will accidentally deploy with `dev_tools` turned on. I'd also like for this to become the default behaviour for dash. So, I think we should introduce a second set of config parameters in the `run_server` call that turn on all `dev_tools`-like things. Users shouldn't be doing `run_server` in production as they should be running it with `gunicorn`. We could even move towards renaming `run_server` to like `run_dev_server` or something. 

For some more context, see the `debug`, `dev`, `production` section in this issue: https://github.com/plotly/dash/issues/312

T4rk1n:
For the same reason @chriddyp mentionned, I don't really like the argument in the constructor for this (and hot-reload). I think these arguments could very well be in `run_server` since that the most likely place it gonna be used and have them default to `true`.

T4rk1n:
@chriddyp I made the change to `run_server`, please review.

rmarren1:
:dancer: looks good to me!

ned2:
I like the approach of pushing dev mode activation away from Dash's constructor. I also think it's important that we don't bake dev-mode specific setup logic into `run_server` because users should still be able to conveniently run in dev mode using a WSGI server like gunicorn etc. This enables development/debugging in an environment that is closer to production for those people not using the built-in development server. 

What about having a separate Dash method like `activate_dev_mode` or something that does all the relevant setup (I guess currently just setting `_serve_dev_bundle` to `True`, but I can imagine the setup might grow over time) and which `run_server` can call? Then this could be manually invoked for those wanting dev mode on gunicorn etc?

If we wanted to do that, this could be a separate PR though.

T4rk1n:
@chriddyp @rmarren1 @ned2 

I changed the naming of the dev tools variables, please review. 

Also I used `_config.get_config` in lieu of `with_defaults`,  this allow for setting the dev tools variables from environ, can use them to have different run configs without changing the code.

T4rk1n:
I released the component suites bundles, when I tried to load them from unpkg it failed because of the name change from `bundle.js` to `lib_name.min.js`

Fix:
- [ ] dash-core-components
- [ ] dash-html-components
- [ ] dash-renderer

dash/dash.py
```diff
@@ -979,8 +985,49 @@ def get_asset_url(self, path):
...
 
+    def enable_dev_tools(self,
+                         debug=False,
+                         dev_tools_serve_dev_bundles=None):
+        """
+        Activate the dev tools, called by `run_server`. If your application is
+        served by wsgi and you want to activate the dev tools, you can call
+        this method out of `__main__`.
+
+        :param debug: If false no tools will be activated.
```
chriddyp: Should we add `If True, then all of the other dev tool settings will be True unless specifically overridden.`

Issue #370, opened 2018-09-07
Multi percy tests 


T4rk1n:
When we upgraded to circle 2, we now have 3 containers that runs the tests separately and there is also 3 percy builds. In the percy, it's like the screenshots are all new since the builds race each other.

Verify and fix for:

- [x] dash
- [x] dash-core-components
- [x] dash-html-components
- [x] dash-renderer
- [x] dash-docs

PR #372, opened 2018-09-07
Only run percy on 2.7.

T4rk1n:
Only enable percy build on a single test container.

#370

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,9 +7,14 @@ jobs:
         environment:
           REQUIREMENTS_FILE: dev-requirements.txt
           PYLINTRC: .pylintrc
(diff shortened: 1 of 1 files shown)


rmarren1:
What is the reason to set `PERCY_PARALLEL_TOTAL` and the `NONCE`?

T4rk1n:
@rmarren1 That was my first try at combining the output of each container, didn't work. I've removed them thanks.

rmarren1:
Okay thanks. I'm :dancer: to this, I can't think of a situation where a different python version would cause all CI tests to pass but visual regression to fail. If this ever actually happens, we can just add the python version as a prefix to all screenshot names.

Issue #214, opened 2018-03-02
A simple working example for embedding dash in flask under a path


Aso1977:
Dash documentation lacks a simple but full working example of how to embed dash layout as a path in a flask app, so that the dash layout becomes a div in the flask and follows the base template. For example  path '/plot&type=plot1' return dash layout for plot type 1.
I know that one can get the underlying flask app in dash, but how to get rendered dash layout in flask is not well documented.

jkgenser:
I'm also very interested in this. I want to use a Flask app for things like authentication, and use Flask for setting up the base html. But then use Dash for  setting up dashboard pages. Anyway figured out how to do this?

fordanic:
Yes, looking for a similar example.

jacohn16:
dito

shaunvxc:
Likewise-- this would be super helpful!

lazyspark:
Okay. So I tried this out and created a gist - 

[https://gist.github.com/lazyspark/f6e74ebc756c19bc3a06e10117333145
](Link to app.py gist)

The main part of the code is creating the Flask Server, creating routes and passing the server to the dash app.

moonlight16:
@lazyspark 
I think you also want to add url_base_pathname to Dash to give it a route:

eg.
app = dash.Dash(__name__, server=server, url_base_pathname='/dash')

This would leave the rest of your Flask app the same, i.e. route '/' could stay the same.  And you would specify where the URL path of the dash app.  I think that would be more useful in the example.

jacohn16:
I think I figured out the answer to the original question.  Your able to use the native Flask route decorator, and pass back the results from index() - which is a Dash instance method that 
generates the HTML and returns.  

For example:

```
server = Flask(__name__)
app = dash.Dash(__name__, server=server, url_base_pathname='/dummypath')
app.layout = <setup your Dash app here>

@server.route("/dash")
def MyDashApp():
    return app.index()

```
When you go to the URL:

`http://<ip>/dash`

This will render the HTML output as shown in the Dash source code for the index() method - it contains a header, body, css, scripts, etc... which basically runs your app.layout inside a script tag.  

At this point you can modify the MyDashApp to return any HTML as you desire.  You can Jinja with templating.  OR you can create your own customer Dash class and override the index() method and create whatever you want.

For example:

```
class CustomIndexDash(Dash):
    """Custom Dash class overriding index() method for local CSS support"""
    def _generate_css_custom_html(self):
[...]

PR #377, opened 2018-09-08
Move add_url function definition out of Dash.__init__()

hinnefe2:
First of all, thanks for a great tool!

This PR moves the `add_url` utility function out of the `__init__` method of the `Dash` class. This change makes it easier to subclass `Dash`. 

My use case is adding authentication via [flask_login](https://flask-login.readthedocs.io/en/latest/). Pulling `add_url` out lets me subclass like so:

```
  class MyDash(Dash):                                                                                                                                                     
                                                               
      def add_url(self, name, view_func, methods=('GET',)):                                                                                                               
          self.server.add_url_rule(                                                                                                                                       
              name,                                                                                                                                                       
              view_func=login_required(view_func),                                                                                                                        
              endpoint=name,                                                                                                                                              
              methods=list(methods)                                                                                                                                       
          )  
```
Note that each `view_func` is wrapped by `login_required` so all my dash pages now require authentication.

I think this addresses some of the comments in #214.


dash/dash.py
```diff
@@ -217,6 +209,14 @@ def add_url(name, view_func, methods=('GET',)):
         self._cached_layout = None
         self.routes = []
 
+    def add_url(self, name, view_func, methods=('GET',)):
```
T4rk1n: Make that "private" by prefixing with `_`.

hinnefe2:
@T4rk1n Sorry for the delay, I think this is ready for another look. Thanks!

T4rk1n:
@hinnefe2 I think your merge with master went bad, there is a line that should be in `__init__` but is in the `add_url` method. Move back line 225-226 in `__init__`.

T4rk1n:
Thanks, looks good now. :dancer:

Since this is for wrapping the routes in `flask_login`, you could add in the add_url method a way to save the route names. You can append the url name to `self.routes`, it is not used presently. Just make sure it is instanciated before the `_add_url calls` since it is at line 215. I would use that to solve plotly/dash-auth#59 if you do that.

Then instead of subclassing Dash, you can wrap the routes externally:

```python
for view_name, view_method in (
        (x, app.server.view_functions[x])
        for x in app.routes
        if x in app.server.view_functions
    ):
    app.server.view_functions[view_name] = login_required(view_method)
```

@rmarren1 can you have a look too ?

rmarren1:
I like this change. I don't see a drawback to defining `add_url` as a method of `Dash` rather than an internal function of `__init__`, and should give advanced users more flexibility.

Not sure why `routes` doesn't do anything right now,  I think we should populate that in `_add_url` then we can merge this
:dancer:

hinnefe2:
@T4rk1n Ok, I think this is ready for another look. Thanks again for your patience!

T4rk1n:
@hinnefe2 Great.

We got a new [CONTRIBUTING guide](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md#pre-merge-checklist), make sure to you have everything on the pre-merge checklist (update version.py and changelogs) to accelerate the merge/release process.

hinnefe2:
Alright, changelog is updated and version is bumped, anything else for me to do?

rmarren1:
Looks great, thanks! :confetti_ball:

rmarren1:
Released `dash==0.28.2`

PR #387, opened 2018-09-13
Add cache control header and cache busting urls to components suites.

### LINKED ISSUE #380: Dash components suites doesn't have a cache headers.
When `app.scripts.config.serve_locally=True`, the components bundles are served without a `Cache-Control` header. They are loaded from the server everytime. The same is true for the css served that way.

T4rk1n:
Fix #380.

- Add `Cache-Control` header to files served by `serve_component_suites`.
- Append a time modified query parameter to the components suites urls.


ned2:
I don't have much experience with cache busting, but seems ok to me :dancer:

Issue #397, opened 2018-09-20
Removing the catch all route.


T4rk1n:
There is a catch all route in dash, it catch any route that doesn't exist and return a 200 with the index. I think this is bad and should be removed as it swallow errors.

https://github.com/plotly/dash/blob/a6388bebe88c5aaba3c549832da398b21a16db24/dash/dash.py#L214-L217

Example of unexpected behavior coming from this:
The browser tries to get the `/favicon.ico` from the server and dash replies OK here's the index. That means the browser will try to get the favicon everytime you load the index since it returns a 200 instead of a 404.

T4rk1n:
cc @plotly/dash

rmarren1:
Will this break multi-page apps that rely on the `Location`? (e.g. [this](https://dash.plot.ly/urls))

I always thought this catch-all was to support this functionality, especially in the case where a user navigates directly to a page rather than manipulating the `Location` via a callback.

T4rk1n:
I see, it's probably because of that, it's a good reason to keep it then.

chriddyp:
Yeah, that's why we have it

ned2:
I wonder if it might be nice to at least fix the special case of the `/favicon.ico` route (as discussed in #152). As it can be a little confusing to see the index being returned as the body of the favicon response.

PR #398, opened 2018-09-20
Add used by dcc.Location to catch-all route comment.

T4rk1n:
Add more detail to the catch-all routes comment so it is more evident that it is useful. #397.


rmarren1:
lgtm 💃

PR #418, opened 2018-10-04
Fix http-equiv typo.

T4rk1n:
Fix missing `http` in default `http-equiv` meta tag include.


T4rk1n:
@rmarren1 I don't understand the failure of py 3.6 and 3.7, there was no modifications to that part of code.

rmarren1:
Strange, the test passes locally for me. I started getting the same error a few days ago and thought it was something in the code.

I'll look into it, looks like a python path issue with Python3 was introduced somehow, perhaps in the CircleCI docker images? Perhaps we can just run the same test in a different way to that it works again.

rmarren1:
https://github.com/plotly/dash/issues/419

T4rk1n:
@rmarren1 I skipped the test, please review.

Issue #419, opened 2018-10-05
Path issue in Python3 Circleci


rmarren1:
There is an error when attempting to create then dynamically import a module in the test cases: https://circleci.com/gh/plotly/dash/1251?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link.

T4rk1n:
All new branch fail on 3.6 and 3.7 can we skip the test while you work on a fix ?

rmarren1:
Yeah skip it! Sorry, overlooked this in my inbox.

There is a similar test in `test_base_component.py` which makes sure the the string created and then imported by the dynamic loader is the same as the string written to the python class files, so skipping this shouldn't leave a huge gap.

My only concern is that sometime in the future the `_imports_.py` file might get messed up and you would not be able to import the class files even though they are structured perfectly. I will add a test for this rather than the module import that is breaking.

PR #429, opened 2018-10-18
Fix ci path issue

rmarren1:
Fixes https://github.com/plotly/dash/issues/419

This one was tricky to figure out.

For some reason, the Python 3 `venv` package behaves differently from the `virtualenv` package. The current working directory was not included in the `PYTHONPATH` and needed to be manually added back using `sys.path.append`. (Not sure why the `dash` module is available, I think this is because the new `default_namespace` module is created during execution rather than before).

Just getting rid of `python -m venv` and using `virtualenv` works.

Is there a good reason to be using `venv` rather than `virtualenv`?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
           name: Install dependencies
           command: |
               sudo pip install virtualenv
-              python -m venv venv || virtualenv venv
+              virtualenv venv
               . venv/bin/activate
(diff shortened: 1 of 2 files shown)


T4rk1n:
That venv thing was from the circleci example, but we install virtualenv anyway so we don't need it.

venv is stdlib for python >= 3.3

:dancer:

PR #431, opened 2018-10-18
Replace docstring `\r\n` with `\n` when generating classes from windows.

### LINKED ISSUE #366: Generate class endlines.
When generating classes for the components libs, if you are on windows, the endlines of the docstring will be `\r\n` instead of `\n`. This gives `^m` diff on github when there shouldn't be a difference.

T4rk1n:
Resolves #366


rmarren1:
Would just changing this part work? https://github.com/plotly/dash/blob/1469268cbfa7142e83bf46ee85beb850ef16b76a/dash/development/base_component.py#L489-L495

I think git will auto-magically put `\r\n` instead of `\n` in this code block, which we can avoid by explicitly putting `\n`
e.g.     `return ("A {name} component.\n{description}\n\nKeyword arguments:\n{args}\n\nAvailable events: {events}").format(...)`

T4rk1n:
No it wouldn't work, it's not those that are explicit the problems (`\n`), it's the whitespace newlines that get auto `\r\n` by python not git. By replacing the output of the docstring, we can keep a sane docstring.

rmarren1:
:dancer:

Do we also need to do that replacement here? https://github.com/plotly/dash/blob/1469268cbfa7142e83bf46ee85beb850ef16b76a/dash/development/base_component.py#L271-L352

T4rk1n:
No, I tested it and it works with just that change.

PR #430, opened 2018-10-18
Allow Dash `children` to be a tuple instead of a list

rmarren1:
Fix https://github.com/plotly/dash/issues/358.

Now, `Component.traverse` will work correctly when Dash component `children` are `tuple`s rather than only with `list`s.

PR #432, opened 2018-10-18
Updates to Contributing Checklists

rmarren1:
I went through the Checklists here https://github.com/plotly/dash/pull/430 and noticed some small areas that can be improved.

Also, I found it really helpful to copy-paste the checklist and click the checkboxes as I went down the list. I put a small note to encourage new contributors to do the same.


CONTRIBUTING.md
```diff
@@ -66,6 +67,8 @@ _For larger features, your contribution will have a higher likelihood of getting
 
 ### Post-Merge checklist
 - [ ] You have tagged the release using `git tag v<version_number>`.
+- [ ] You have pushed this tag using `git push <tag_name>`
```
T4rk1n: This push to origin, only contributors can push to origin. So it should state that the merger have to tag.

Issue #234, opened 2018-03-27
Turn off undo/redo by default in Dash 1.0


Moisan:
The current solution to remove the undo-redo button is to [change the css of the page to hide it](https://community.plot.ly/t/is-it-possible-to-hide-the-floating-toolbar/4911/11) while [using a hosted css file](https://stackoverflow.com/questions/45137459/how-to-remove-the-undo-button-in-plotly-dash-after-a-dropdown-update). 

It would be much simpler to have an option that toggles this button, a bit like the graph toolbar.

charleyferrari:
@Moisan Thanks for writing in. This seems like a good feature to have within Dash. Adding some tags to encourage some discussion and collaboration.

ebber:
Would very much appreciate this feature as well. I am having trouble using the workaround. While I will be able to do it, it seems like a useful feature to be able to config, rather than making messy code.

Myijin:
The undo button does not change anything for my dash app, which has input box and button.
My function looks like this:
```
@app.callback(Output('Docs', 'children'),
              [Input('button', 'n_clicks')],
              [State('input', 'value')])
def clean_data(n_clicks, input):
     if n_clicks is not None and input is not None:
            return ...
```
For example, I enter “A” in the input box and click the button, then enter “B” in the input box and click the button. Then clicking undo nothing is changed. I am hoping that the undo can go back to the result for “A”.

rmarren1:
@Myijin I think this issue describes the behavior you are talking about https://github.com/plotly/dash-renderer/issues/66, does that sound right?

Myijin:
@rmarren1 That's right. When I only use the input box, the undo can go back to previous input. But it's not working after using button.

rmarren1:
@Myijin Okay, I have a PR to fix this that is ready but have not merged, I'll see if I can expedite that

vanbenschoten:
Is there any update on an easy option to remove the Undo-Redo button? I'll add my support for an easy toggle option as the button currently gets in the way of our dashboards.

T4rk1n:
Include a css file in the assets folder with this:

```
._dash-undo-redo {
  display: none;
}
```

vanbenschoten:
@T4rk1n thanks! That took care of the issue (though I still think a toggle feature would be helpful as well)

eddy-ojb:
[...]

PR #435, opened 2018-10-24
[WIP] Add internal_styles argument to Dash.__init__() for passing custom CSS

hinnefe2:
This PR adds an optional kwarg to the `Dash` constructor to allow passing CSS as a nested dictionary. This CSS gets included in a `<style>` block in the document head.

The use case I have in mind is hiding the Undo button, as in #234. With these changes you can do something like this
```
hide_undo_css = {'._dash-undo-redo': {'display': 'none'}}

app = dash.Dash('app', internal_styles=hide_undo_css)
``` 
instead of requiring an external stylesheet or messing around with `serve_locally`.

If this seems like a reasonable addition I'll take a stab at writing some tests. If not, I'm happy to try another approach, or if you're not interested in adding something like this then just let me know and I can close the PR. Also if there's a better place to put this configuration option than kwargs in the constructor please let me know.

Thanks!

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -8,7 +8,6 @@ import json
 import pkgutil
 import warnings
 import re
-import six
 
 from functools import wraps
 
@@ -17,6 +16,7 @@ import dash_renderer
 import flask
 from flask import Flask, Response
 from flask_compress import Compress
+import six
 
 from .dependencies import Event, Input, Output, State
 from .resources import Scripts, Css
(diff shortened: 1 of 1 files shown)


alexcjohnson:
Hi @hinnefe2 - my sincere apologies for not responding to your PR earlier! The consensus here is that the existing solution - a `.css` file in the `assets/` folder, which is now served by default - is at this point nearly as easy as this in the simple case but is more scalable and maintainable as people grow their apps. So we'd like to encourage people to use the `assets/` folder rather than learning one thing for simple cases and then having to learn (and switch to) another method later.

If the `assets/` solution has shortcomings or could be made easier to learn and use we'd love to hear your thoughts!

hinnefe2:
Hey, no problem! I haven't been keeping up with the CSS serving changes, but being able to put some CSS in a local file which is served by default works for me. Back when I made this it was only possible to source external CSS files and it seemed like a hassle to set up some hosted CSS just for this small change.

Thanks!

PR #436, opened 2018-10-25
Multi output callbacks support.

T4rk1n:
**Release candidates**
```
pip install dash==0.38.0rc1
```

**Add multi output callbacks support.**

- Use a list/tuple of `Output` as output in callbacks.
- Return a tuple/list of value from the callbacks
  - The returned list must have the same length as the list of output
  - The output props are applied in the order they were declared in the output list.

Usage example:

```python
import dash

from dash.dependencies import Output, Input, State

app = dash.Dash(__name__)

app.layout = html.Div([
    html.Button('OUTPUT', id='output-btn'),

    html.Table([
        html.Thead([
            html.Tr([
                html.Th('Output 1'),
                html.Th('Output 2')
            ])
        ]),
        html.Tbody([
            html.Tr([html.Td(id='output1'), html.Td(id='output2')]),
        ])
    ]),
])


@app.callback([Output('output1', 'children'), Output('output2', 'children')],
              [Input('output-btn', 'n_clicks')],
              [State('output-btn', 'n_clicks_timestamp')])
def on_click(n_clicks, n_clicks_timestamp):
    if n_clicks is None:
        raise PreventUpdate

    return n_clicks, n_clicks_timestamp


if __name__ == '__main__':
    app.run_server(debug=True, port=6060)
```

Needs plotly/dash-renderer#91

Tests will fails until I release a rc version of dash-renderer, this is kind of breaking changes as no apps will run if they don't have the right version of dash-renderer (The json payload changed).

Todos:

- [x] rc versions for tests
- [x] re-enable callback validation
- [x] add integration test


sdementen:
hello,
I just saw the 1.0.0 announcement #469 but could not find this topic in it (multiple output).
Is it too early for it to be included ?

T4rk1n:
@sdementen This is a standalone update. You can try it right now with:

```
pip install dash==0.36.0rc1
pip install dash-renderer==0.17.0rc1
```

T4rk1n:
@rmarren1 Please review.

rmarren1:
This allows you to define multiple callbacks to the same output.

For example, adding
```
              [Input('output-btn', 'n_clicks')])
def on_click(n_clicks):
    if n_clicks is None:
        raise PreventUpdate

    return 'something else'
```
results in a race condition (sometimes Output 1 is 'something else', sometimes it is the number of clicks).

I think this just needs logic changed in the condition throwing `exceptions.CantHaveMultipleOutputs`

rmarren1:
I think we should _eventually_ allow multiple callbacks to be defined for an output component on the condition that those input sets are disjoint (e.g. `I1, I2 -> O1` and `I3, I4 -> O1` can be simultaneously defined but `I1, I2 -> O1` and `I2, I3 -> O1` ). Definitely the scope of this PR, but something to keep in mind as a future possibility when doing above ^

alexcjohnson:
> I think we should _eventually_ allow multiple callbacks to be defined for an output component on the condition that those input sets are disjoint

This discussion might belong in #475, but I'm skeptical of both the idea of multiple callbacks and the disjoint exception. I guess you're imagining cases where one set of inputs disappears and another appears? But you could still have some other global setting that factors in to the final output... so seems to me the cleaner way to handle it is to make the output depend on *all* of the inputs, but the conditionally-available ones are wildcarded somehow - eg @chriddyp's https://github.com/plotly/dash/issues/475#issuecomment-448100006

dash/dash.py
```diff
@@ -320,6 +321,7 @@ def serve_layout(self):
 
     def _config(self):
         config = {
+            'multi_output': True,
```
rmarren1: Is the idea to make this configurable by the user?

rmarren1: Or is this for backward compatibility?

dash/dash.py
```diff
@@ -660,6 +662,13 @@ def _validate_callback(self, output, inputs, state, events):
                             name.lower(), str(arg), name
                         ))
 
+                if '.' in arg.component_id:
+                    raise exceptions.IDsCantContainPeriods('''The element
+                    `{}` contains a period in its ID.
+                    Periods are not allowed in IDs right now.'''.format(
+                        arg.component_id
+                    ))
```
alexcjohnson: Do we need to exclude `:` now too?

dash/dash.py
T4rk1n: It's for backward compatibility.

T4rk1n:
@rmarren1 

> I think this just needs logic changed in the condition throwing exceptions.CantHaveMultipleOutputs

I added validation for multi output and changed `CantHaveMultipleOutputs` to `DuplicateCallbackOutput`. There's two test for the logic, think of any other test cases ?

dash/dash.py
```diff
@@ -662,11 +662,13 @@ def _validate_callback(self, output, inputs, state, events):
                             name.lower(), str(arg), name
                         ))
 
-                if '.' in arg.component_id:
+                invalid_characters = ['.', ':', '[', ']']
+                if any(x in arg.component_id for x in invalid_characters):
                     raise exceptions.IDsCantContainPeriods('''The element
-                    `{}` contains a period in its ID.
+                    `{}` contains {} in its ID.
                     Periods are not allowed in IDs right now.'''.format(
```
rmarren1: It says `Periods` regardless of the invalid character, also this will now break any apps that used `[, ], :` as an ID character for a single output. I think that we would instead need to check if `.` is in the ID for all callbacks, then check if `[`. `]`, `:` is in the ID for all callbacks where `isinstance(output, (list, tuple))`.

dash/dash.py
```diff
@@ -754,15 +755,47 @@ def _validate_callback(self, output, inputs, state, events):
             ).replace('    ', ''))
 
         callback_id = _create_callback_id(output)
-        if callback_id in self.callback_map:
-            raise exceptions.CantHaveMultipleOutputs('''
+        is_multi = isinstance(output, (list, tuple))
+        callbacks = set(itertools.chain(*(
+            x[1:-1].split(':')
+            if x.startswith('[')
```
rmarren1: Backwards incompatible with apps that use a callback output ID starting with '['

dash/dash.py
```diff
@@ -754,15 +755,47 @@ def _validate_callback(self, output, inputs, state, events):
...
+            else [x]
+            for x in self.callback_map
+        )))
+        ns = {
+            'duplicates': callback_id
```
rmarren1: Why is this initialized to `callback_id`? (It does not look to be used below, same result if it was initialized to `'duplicates': None`)

dash/dash.py
```diff
@@ -754,15 +755,47 @@ def _validate_callback(self, output, inputs, state, events):
...
+        }
+        if is_multi:
+            def duplicate_check():
+                ns['duplicates'] = intersection = callbacks.intersection(
```
rmarren1: Can you just return `ns['duplicates'`? Why `intersection` local variable?

dash/dash.py
alexcjohnson: If we're going to make any breaking change we might as well give ourselves freedom for future expandability - thinking about #475 in particular but who knows what else may come up.

What if we required IDs to be valid Python identifiers? ie letters, numbers, digits, and underscore, first character not a digit. That would enable other possibilities like collecting all components into a namespace somewhere - eg `app.components_by_id.myDiv3`. Dunno what that would be useful for, but the point is "valid Python identifier" is a broad enough class that people shouldn't feel constricted by it, and it's instantly recognizable to Python devs so folks won't need to hunt around for the rule as to which characters are allowed or not.

rmarren1:
I think there will be an issue encoding the IDs of multi-output callbacks as an ASCII-delimited string, since there will be a backwards compatibility issue with apps that use the ASCII characters we choose as delimiters.

One idea: We already blacklisted `.`, so we can transform `[output1.prop1:output2.prop2]` to `...output1.prop1..output2.prop2...`, e.g. replace `[` and `]` with `...` and replace `:` with `..`, pretty hacky but it should work.

Another idea is to use a different key for `self.callback_map`. The retrieval of callbacks is an internally facing API and it appears the only reason we use the output ID to identify these is that it is naturally unique per callback. We could use `hash(frozenset(['output1.prop1', 'output2.prop2']))`, and then just pass a list of outputs back-and-forth between the front-end and back-end like we do with the output id, re-computing the hash when needed. I think this may be more natural, e.g. to test if a callback is multi-output on the front-end you just see if the output is a list rather than if it starts with `[`.

T4rk1n:
>  We could use hash(frozenset(['output1.prop1', 'output2.prop2'])), and then just pass a list of outputs back-and-forth between the front-end and back-end like we do with the output id, re-computing the hash when needed. I think this may be more natural, e.g. to test if a callback is multi-output on the front-end you just see if the output is a list rather than if it starts with [.

:smiley_cat: I like that.

T4rk1n:
The dot notation could work well too.

rmarren1:
Oh right, I've seen that pattern before.

Perhaps we could add a backwards compatibility entry to `callback_map` with the output ID for single output callbacks. E.g., `callback_map[output_id]` and `callback_map[hash(frozenset([output_id]))]` point to the same function in memory, then code that accesses and decorates using the `output_id` will still work

T4rk1n:
I think the dot notation would be best, doesn't require that much change to the current implementation.

Yoann-Brenet:
Hi,
Was wondering if that is part of the latest 0.36 release or if it will come back later ?

T4rk1n:
> Was wondering if that is part of the latest 0.36 release or if it will come back later ?

No, you can try the rc versions listed in opening post in the meantime.

Yoann-Brenet:
> > Was wondering if that is part of the latest 0.36 release or if it will come back later ?
> 
> No, you can try the rc versions listed in opening post in the meantime.

Ok.
Thanks a lot for that feature, this is really useful.

Yoann-Brenet:
Hi,

Have just tried the latest 0.38.0rc1.

I have 2 callbacks with the same list of multi outputs, but different inputs/states:

**Callback 1:**
_@app.callback(
    output=[Output('datetime_RangeSlider', 'min'),
            Output('datetime_RangeSlider', 'max'),
            Output('datetime_RangeSlider', 'marks'),
            Output('datetime_RangeSlider', 'value'),
            ],
    inputs=[
            Input(component_id='country_name_Dropdown', component_property='value'),
            ],
    state=[
        State(component_id='refinery_name_Dropdown', component_property='value'),
        State(component_id='data_set_Dropdown', component_property='value'),
        ],
)
def update_date_slider_country_name_Dropdown(country_name, refineries_names, data_set_name):_

**Callback 2:**
_@app.callback(
    output=[Output('datetime_RangeSlider', 'min'),
            Output('datetime_RangeSlider', 'max'),
            Output('datetime_RangeSlider', 'marks'),
            Output('datetime_RangeSlider', 'value'),
            ],
    inputs=[
            Input(component_id='refinery_name_Dropdown', component_property='value'),
            ],
    state=[
        State(component_id='country_name_Dropdown', component_property='value'),
        State(component_id='data_set_Dropdown', component_property='value'),
    ],
)
def update_date_slider_refinery_name_Dropdown(refineries_names, country_name, data_set_name):_

Am getting the following error, looks like it needs unicity of Outputs between callbacks:
image

T4rk1n:
@sdementen Thanks for trying. We do not allow duplicate output in single and multi-output callbacks, you can combine the callbacks inputs and with #608 you'll be able to know which input has fired.

Yoann-Brenet:
Hi,

Thanks. Sorry not sure I understand: the callbacks I am using do not have duplicates outputs in the same callback (so for instance callback 1 has 4 outputs which are different, same for callback 2).
But they have the same outputs when you compare outputs from callback 1 and 2. They do have different inputs though, so they are not perfect callbacks duplicates.

Yoann-Brenet:
Or do you mean that we just cannot have several callbacks outputting to the same outputs at all ?

tests/test_integration.py
```diff
@@ -530,6 +529,85 @@ def create_layout():
...
+                          [Input('output-btn', 'n_clicks')])
+            def on_click_duplicate(n_clicks):
+                if n_clicks is None:
+                    raise PreventUpdate
+
+                return 'something else'
+
+        self.assertTrue('output1' in context.exception.args[0])
+
+        # Test a multi output cannot contain a used single output
```
alexcjohnson: Two more cases I don't see covered (unless I'm missing it somewhere):
- you can't use the same output twice in one callback
- you can't use the same output in two overlapping multi-output sets (eg `#o1.value,#o2.value` and `#o1.value,#o3.value`

dash/dash.py
```diff
@@ -170,6 +170,9 @@ def __init__(
         self._meta_tags = meta_tags or []
         self._favicon = None
 
+        # default renderer string
+        self.renderer = 'var renderer = new DashRenderer();'
```
alexcjohnson: FYI @valentijnnieman - the hazards of a 6-month-old PR, this default was being set again and again in `_add_url`... no reason for that, right?

valentijnnieman: Oh weird - no I doubt that that would be useful! Sorry about that.

.circleci/config.yml
```diff
@@ -28,7 +28,7 @@ jobs:
               sudo pip install virtualenv
               virtualenv venv
               . venv/bin/activate
-              pip install -r $REQUIREMENTS_FILE
+              pip install -r $REQUIREMENTS_FILE --force-reinstall
```
alexcjohnson: @T4rk1n I've already merged this but... was `--force-reinstall` just for debugging or do we need it going forward if we're using git branches in requirements files?

PR #439, opened 2018-10-29
Refactor requirements.txt

### LINKED ISSUE #434: Rename `requirements.txt` in this repo to prevent confusion from users vs package devs
Several users have thought that installing  the`requirements.txt` in this project is the correct way to install the necsesary dependencies for `dash` (rather than visiting dash.plot.ly/installation).  See https://github.com/plotly/dash/issues/433#issuecomment-432810843 and some of the issues in https://github.com/plotly/dash/issues?utf8=%E2%9C%93&q=is%3Aissue+requirements.txt. 

Is there any way that we can help prevent these issues? Thoughts:
- Rename `requirements.txt` to something like `contributing_requirements.txt` or place it in a folder like `contributing/requirements.txt`
- Perhaps we can add a "Quickstart" link at the very top of the README.md with a link to https://dash.plot.ly/installation

T4rk1n:
- Deleted `requirements.txt`
- Move dev-requirements* to `.circleci/requirements/`

Closes #434

cc @plotly/dash


T4rk1n:
I don't understand the percy diffs, running locally the input gets cleared. There's also an assert for the text and it would fail if the input was not cleared.

PR #442, opened 2018-10-31
Generate integrity hash for scripts/link tags on the index.

### LINKED ISSUE #422: Generate integrity hash for the components bundles and use them when serving.
Currently we don't include the integrity hash in the bundles we serve.

I think we could generate them at the same time as we generate the bundles and include a key in the `_js_dist` with the hash. Then we can add that key when generating the html tags in dash index.

Repo to add integrity hashes support:

- [ ] dash-core-components
- [ ] dash-html-components
- [ ] dash-renderer
- [ ] dash-component-boilerplate

Dash:
- [ ] include the integrity hash in scripts/css tags.

T4rk1n:
- Generate a sha384 hash from the local copy of the file.
- Format the tags with `integrity` and `crossorigin` attributes.
- Flattened the resources collection from the dist.
- Map external unpkg to local host file.

As the version of plotly.js hosted locally and externally was not the same, this breaks dash-core-components Graph component.

The files hosted locally must be downloaded from the external source so they match 1:1. Same for the component libs bundles, do `npm publish` first so you have the builds with the `prepublish` hook, then do `python setup.py sdist` and `twine upload dist/*` without rebuilding the bundles so they have the same hash, all version of dcc I tested had the right bundles, but it needs to be said somewhere.

Closes #422

### DIFF
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -70,23 +70,23 @@ def first_key(data, *keys):
     return None, None
 
 
-@memoize
-def integrity_hash_from_file(filename):
-    with open(filename, 'rb') as f:
-        h = hashlib.sha384(f.read())
+def _integrity_hash(bytes):
+    h = hashlib.sha384(bytes)
 
     return 'sha384-{}'.format(
         base64.b64encode(h.digest()).decode()
     )
+
+
+@memoize
+def integrity_hash_from_file(filename):
+    with open(filename, 'rb') as f:
+        return _integrity_hash(f.read())
 
 
 @memoize
 def integrity_hash_from_package(namespace, path):
-    h = hashlib.sha384(pkgutil.get_data(namespace, path))
-
-    return 'sha384-{}'.format(
-        base64.b64encode(h.digest()).decode()
-    )
+    return _integrity_hash(pkgutil.get_data(namespace, path))
 
 
 class AttributeDict(dict):
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -1,3 +1,9 @@
+import functools
+import hashlib
+import base64
+import pkgutil
+
+
 def interpolate_str(template, **data):
     s = template
     for k, v in data.items():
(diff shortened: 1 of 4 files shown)


chriddyp:
This is very cool. We should note in our documentation and the changelog the browser support https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#Browser_compatibility (not supported in IE. partial support in Edge, not sure what that means)

T4rk1n:
@chriddyp I tested from ie11 and edge, it loaded, I think it just ignore the attribute.

T4rk1n:
Tagging this as dash 2.0 as it breaks old versions of dcc that did not have the same plotly.js files. It's a nice feature, but not essential.

dash/_utils.py
```diff
@@ -1,3 +1,24 @@
...
+import base64
+import pkgutil
+
+
+def memoize(func):
+    results = {}
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        key = hash((args, frozenset(kwargs)))
```
rmarren1: should this be `hash((args, frozenset(kwargs.keys()), frozenset(kwargs.values()))`?

`frozenset(kwargs)` gets a set of the keys, so `function(1, 2, a=1)` and `function(1, 2, a=2)` would produce the same hash.

T4rk1n: Good catch, the function it wraps doesn't have kwargs so it didn't impact it. I think frozenset(kwargs.items()) so the values and keys are grouped together in a tuple?

rmarren1:
Yeah either one, it shouldn't matter for the hash

3. Nov 2018 11:06 by notifications@github.com <mailto:notifications@github.com>:

alexcjohnson:
This has gotten stale - closing. Will likely need to be reimplemented if we pick this feature up again.

mkhorton:
> Tagging this as dash 2.0 as it breaks old versions of dcc that did not have the same plotly.js files. It's a nice feature, but not essential.

With Dash 2.0 on the horizon, has this feature been re-evaluated? It would give a lot more confidence if using third-party CDNs like unpkg.

alexcjohnson:
It's not on our roadmap ATM, but if anyone wants to tackle it we'd be happy to review a PR!

mkhorton:
@alexcjohnson Thanks for the response. Is the approach taken by this PR no longer feasible? It looks like a sensible approach.

PR #443, opened 2018-10-31
Change registered_paths to a `collections.defaultdict(set)`.

### LINKED ISSUE #441: Dash.registered_paths appending package path on every index access.
This code append the relative_package_path every time the index is accessed, taking up memory:
https://github.com/plotly/dash/blob/61b05450ed562923f7d5b67842c9b04c899c4928/dash/dash.py#L316-L319

Should be a `collections.defaultdict(set)` instead.

T4rk1n:
Replace registered_paths with a `collections.defaultdict(set)` so it doesn't waste memory on every index render if  `app.scripts.config.serve_locally = True` or `app.css.config.serve_locally = True`.

Fix #441.


rmarren1:
LGTM, nice catch. :dancer:

PR #444, opened 2018-10-31
Register component suites by importing.

### LINKED ISSUE #424: Figure doesn't load from callback unless another dash_core_component is in the layout
I was trying to create a workable example to solve this [community question](https://community.plot.ly/t/waiting-long-time-after-chart-load/14347), but I could only make it work with another dcc component inside. Here is the example:

```
import time

import dash


loading_component = html.H1(
    style={},
    children='Loading...'
)

app = dash.Dash(__name__)
app.layout = html.Div([
    dcc.Input(), # callback doesn't work without this line

    html.Div(id='live-graph-container', children=[
        loading_component
    ]),
    # Arbitrary element to fire the callback below
    html.P(
        id='trigger-live-chart',
        children=''),
])

@app.callback(Output('live-graph-container', 'children'),
    [Input('trigger-live-chart','children')])
def generate_live_chart(_):
    time.sleep(3) # simmulate 3 seconds of data processing
    return dcc.Graph(
        id='live-graph',
        figure={
            'data': [
                {'x': [1, 2, 3], 'y': [4, 1, 2], 'type': 'bar', 'name': 'SF'},
                {'x': [1, 2, 3], 'y': [2, 4, 5], 'type': 'bar', 'name': u'Montréal'},
            ],
            'layout': {
                'title': 'Dash Data Visualization'
            }
        }
    )


if __name__ == '__main__':
    app.run_server(debug=True)

```

If you remove that `dcc.Input()` line, the program still runs, but you are just stuck with the loading text. Is this happening for others when they try to use it, or just me? Any thoughts on why this is occurring?

For reference, here are my plotly related packages version:

```
dash                   0.28.2
dash-core-components   0.33.0
dash-html-components   0.13.2
dash-renderer          0.14.1
plotly                 3.3.0
```

T4rk1n:
Refactored the component suites to register without crawling the layout by using a ComponentRegistry metaclass.

Closes #424, Closes plotly/dash-renderer#46


T4rk1n:
@rmarren1 I don't know what the matter with the component in `dash_flow_example` and `dash_dangerously_set_inner_html`. I think they don't have a generated python Component class ?

chriddyp:
Interesting, `ComponentRegistry` is like a global, shared import? This is a very cool solution

T4rk1n:
@chriddyp It's a metaclass,  `__new__` get called every time a new subclass of Component get created and add the namespace to the class instance registry so it's like a global but it's not.

chriddyp:
The other way of solving this would be by requesting components on-the-fly during rendering. That is, in here: https://github.com/plotly/dash-renderer/blob/b1cfc996563bd0b57f9487b15212dd5b782e5b3a/src/TreeContainer.js#L73
We add something like:
```
if (!window[component.namespace]) {
   await fetch(`/_dash-component-registry?library=${component.namespace}`).then(
      res => {
          const resources = res.scripts_and_stylesheets;
          return Promise.all(resourcesmap(lib => fetch(`/_dash-component-suites/${lib}`))
      }
   )
}
```
which might make it easier for other dash in other languages to do this (languages that don't have the benefit of metaclasses).

But for now, I'm 👍  with this solution as it's much simpler

chriddyp:
This is also one of the gnarliest bugs that users in encounter, so 💯 💯 💯 for taking this on 🥇

chriddyp:
> you don't have the actual bundle name that need to be loaded and the dist that comes with the component library.

Right, we'd need a new endpoint for discovery (as above `/_dash-component-registry?library=${component.namespace}`)

But in any case, this looks like a great solution 👍

rmarren1:
@T4rk1n Yeah those repos don't have the auto-generated classes

T4rk1n:
@rmarren1 I register those namespace in the load_component, works good.

T4rk1n:
@plotly/dash ready for reviews.

T4rk1n:
@rmarren1 Thanks for the review. I'll incorporate some of the suggestions but I fear using the gh suggestion will make a bunch of small commits that might not fit together so I'll do them together in a real commit. 

> "The class of Component is ComponentRegistry"

A metaclass is the type of a class.

```python
>>> type(Component)
<class 'abc.ABCMeta'>
```

> Also I'm not sure if the name ComponentRegistry makes sense, maybe ComponentMeta

I called it ComponentMeta first, then renamed to ComponentRegistry to reflect it's purpose. I'll go further and keep the ComponentRegistry class and 
refactor the metaclass part to ComponentMeta again as it should be the type of the Component class.

T4rk1n:
@rmarren1 I have incorporated the changes and removed the len cache busting, will integrate with hot-reload when the time come. Please review.

chriddyp:
Once we merge this, we'll want to remove some of the references to this issue in our docs: https://github.com/plotly/dash-docs/search?q=https%3A%2F%2Fgithub.com%2Fplotly%2Fdash-renderer%2Fissues%2F46&unscoped_q=https%3A%2F%2Fgithub.com%2Fplotly%2Fdash-renderer%2Fissues%2F46 🥇

Issue #311, opened 2018-07-31
Make Generated Python Classes more minimal.


rmarren1:
A lot of the logic in the dash component python class files is repeated and could (and should) be moved into the base `Component` class to be inherited.

gvwilson:
closing as stale

PR #449, opened 2018-11-05
WIP Initial version of updated code for R transpiler

rpkyle:


dash/development/base_component.py
```diff
@@ -387,6 +387,120 @@ def __repr__(self):
     required_args = required_props(props)
     return c.format(**locals())
 
+def generate_class_string_r(typename, props, description, namespace):
+    """
+    Dynamically generate class strings to have nicely formatted docstrings,
+    keyword arguments, and repr
```
T4rk1n: I think you can update this docstring to reflect that it generate R components. `docstring`, `keywords arguments`, `repr` are python things that may not relate to R.

dash/development/base_component.py
```diff
@@ -387,6 +387,120 @@ def __repr__(self):
...
+    #
+    # TODO - Handle "required"
+    #
+    # TODO - How to handle user-given `null` values? I want to include
+    # an expanded docstring like Dropdown(value=None, id=None)
+    # but by templating in those None values, I have no way of knowing
+    # whether a property is None because the user explicitly wanted
+    # it to be `null` or whether that was just the default value.
+    # The solution might be to deal with default values better although
+    # not all component authors will supply those.
```
T4rk1n: Remove those comments, they contains todo's that are specifics to python classes.

dash/development/base_component.py
```diff
@@ -387,6 +387,120 @@ def __repr__(self):
...
+    # pylint: disable=unused-variable
+    list_of_valid_keys = repr(list(map(str, filtered_props.keys())))
+    list_of_valid_keys = list_of_valid_keys[1:-1]
+    package_name = make_package_name(namespace)
+    # pylint: disable=unused-variable
+    docstring = create_docstring_r(
+        component_name=typename,
+        props=filtered_props,
+        events=parse_events(props),
+        description=description).replace('\r\n', '\n')
```
T4rk1n: I added that `.replace('\r\n', '\n')` for generating the docstring on windows because git doesn't auto replace them when they are in docstring. Not sure this problem is the same with R.

dash/development/base_component.py
```diff
@@ -387,6 +387,120 @@ def __repr__(self):
...
+
+    # pylint: disable=unused-variable
```
T4rk1n: :zap:

dash/development/base_component.py
```diff
@@ -387,6 +387,120 @@ def __repr__(self):
...
+         for p in prop_keys
+         if '*' in p]
+    )
+    default_argtext += ", ".join(
+        [('{:s}={}'.format(p, props[p]['defaultValue']['value'])
+          if 'defaultValue' in props[p] else
+          '{:s}=NULL'.format(p))
+         for p in prop_keys
+         if not p.endswith("-*") and
+         p not in keyword.kwlist and
```
T4rk1n: :zap: This keyword list filter is for python keywords, won't be the same in R.

dash/development/base_component.py
```diff
@@ -387,6 +387,120 @@ def __repr__(self):
...
+         p not in ['setProps']]
+    )
+    default_paramtext += ", ".join(
+            [('{:s}={:s}'.format(p, p)
+               if p != "children" else
+               '{:s}=c(children, assert_valid_children(..., wildcards = c({:s})))'.format(p, default_wildcards))
+             for p in props.keys()
+             if not p.endswith("-*") and
+             p not in keyword.kwlist and
```
T4rk1n: :zap:

dash/development/base_component.py
```diff
@@ -387,6 +387,120 @@ def __repr__(self):
...
+             p not in ['setProps']]
+    )
+    required_args = required_props(props)
+    return c.format(**locals())
```
T4rk1n: I'd like them to be explicitly stated instead of locals(), then we can remove all those `# pylint: disable=unused-variable` and easily see what variables are not used.

dash/development/base_component.py
```diff
@@ -445,6 +589,27 @@ def generate_class(typename, props, description, namespace):
     result = scope[typename]
     return result
 
+def generate_class_r(typename, props, description, namespace):
```
T4rk1n: :zap: the generate_class for python is for dynamically loaded components, we don't do that in R.

dash/development/base_component.py
```diff
@@ -507,6 +672,49 @@ def create_docstring(component_name, props, events, description):
...
+    Returns
+    -------
+    str
+        Dash component docstring
+    """
+    # Ensure props are ordered with children first
+    props = reorder_props(props=props)
+
+    desctext = ''
+    desctext += "\n".join(
```
T4rk1n: No need to append, just do `desctext = "\n".join(`

dash/development/component_loader.py
```diff
@@ -1,9 +1,12 @@
 import collections
 import json
 import os
+#from dash.development.base_component import generate_class
```
T4rk1n: :zap:

dash/development/component_loader.py
```diff
@@ -1,9 +1,12 @@
...
 from .base_component import generate_class
 from .base_component import generate_class_file
-
+from .base_component import generate_class_r
+from .base_component import generate_class_file_r
+#from dash.development.base_component import generate_class_file
```
T4rk1n: :zap:

dash/development/component_loader.py
```diff
@@ -94,8 +97,15 @@ def generate_classes(namespace, metadata_path='lib/metadata.json'):
...
         )
+        generate_class_file_r(
+            name,
+            componentData['props'],
+            componentData['description'],
+            namespace
+        )
 
         # Add an import statement for this component
+        # RK: need to add import statements in R namespace file also
```
T4rk1n: Is that still todo ?

dash/development/base_component.py
rpkyle: Right, this makes sense. When I first started editing, I think I was trying too hard to maintain parity between the existing code and the R-related bits I introduced. I'll revise so that my additions reflect R conventions when appropriate.

dash/development/base_component.py
rpkyle: Will do. I'll also finish commenting my additions where appropriate.

dash/development/base_component.py
rpkyle: Actually, it is a consideration in R as well, though I think I should remove the `\n` and replace with a single space, since R's help viewer does its own line wrapping without adding linefeeds.

dash/development/base_component.py
rpkyle: Right, I think we talked about this when you explained list comprehensions to me a while ago. Will remove the shorthand and explicitly declare the variables.

dash/development/component_loader.py
rpkyle: I've added `export()` statements to the `NAMESPACE` file for the R package, don't think we'll need to provide an import statement. Will remove the comment.

dash/development/base_component.py
rpkyle: P.S. Thanks for having a look!

dash/development/base_component.py
rmarren1: Actually now we manually maintain a file of python keywords (https://github.com/plotly/dash/blob/master/dash/development/_all_keywords.py). Perhaps we can re-name that to python_keywords and make a new one called R_keywords.

That is, if this problem even exists in R. Can you define a function with argument names that are the same as a reserved keyword in R?

dash/development/base_component.py
rmarren1: I agree with this, and we should do it in the Python version as well. I've added a note here: https://github.com/plotly/dash/issues/311

dash/development/base_component.py
```diff
@@ -507,6 +672,49 @@ def create_docstring(component_name, props, events, description):
...
+        [('#\' @param {:s} {:s}'.format(p, props[p]['description'].replace('\n', ' ')))
+         for p in props.keys()
+         if not p.endswith("-*") and
+         p not in keyword.kwlist and
```
rmarren1: :zap:

dash/development/base_component.py
```diff
@@ -550,6 +758,25 @@ def parse_wildcards(props):
             list_of_valid_wildcard_attr_prefixes.append(wildcard_attr[:-1])
     return list_of_valid_wildcard_attr_prefixes
 
+def parse_wildcards_r(props):
```
rmarren1: :camel: I do not think this function is necessary, it is the same as `parse_wildcards` currently. We can keep this as one function used by the R and Python generation until we have a reason to re-write it.

dash/development/base_component.py
rpkyle: > That is, if this problem even exists in R. Can you define a function with argument names that are the same as a reserved keyword in R?

If the function argument matches a reserved word, the R parser will throw an error:

`Error: unexpected 'if' in "myfun <- function(if"`

I've gone ahead and produced a GitHub gist including all the reserved words in R, similar to the keyword list already available for Python. I can submit it as a pull request if it would be helpful.

https://gist.github.com/rpkyle/622155e24f603fc54a664b06f8036f3d

T4rk1n: You can include those keywords in https://github.com/plotly/dash/blob/master/dash/development/_all_keywords.py as a new variable `r_keywords` and use it in this PR.

rpkyle: Great, will do.

dash/development/_all_keywords.py
```diff
@@ -0,0 +1,72 @@
+# This is a set of Python keywords that cannot be used as prop names.
+# Keywords for a particular version are obtained as follows:
+# >>> import keyword
+# >>> keyword.kwlist
+
+kwlist = set([
```
rmarren1: We should change this to `python_keywords` then to be consistent. I think we should do it here, it's only 4 lines and I don't think it needs its own PR.

dash/development/base_component.py
```diff
@@ -380,17 +376,25 @@ def __repr__(self):
...
          for p in prop_keys
          if not p.endswith("-*") and
-         p not in keyword.kwlist and
+         p not in kwlist and
          p not in ['dashEvents', 'fireEvent', 'setProps']] + ['**kwargs']
     )
 
     required_args = required_props(props)
-    return c.format(**locals())
+    return c.format(typename=typename,
```
rmarren1: :cat:

dash/development/base_component.py
rpkyle: I'm in the process of updating the code I've written, so please hold off on reviewing the updates to this PR for a day or so. Nicolas suggested remerging my new branch with R, so that we can keep the current PR alive.

rpkyle:
Closing this PR in favour of #483.

Issue #361, opened 2018-08-30
Un-usable Dash 'prop' names


rmarren1:
Since we are now pre-compiling Dash component python classes, we need to make sure those pre-compiled files are compatible with each version of Python we care about.

One source of incompatibility is that we can't use a Python keyword as a class initialization argument. Python keywords change from version to version, and no one version has a superset of the keywords of other versions (e.g. `async` is a keyword in 3.7 and not 2.7, while `exec` is a keyword in 2.7 and not 3.7), so this is difficult.

The only reported issue of a Dash component having a prop name that is also a keyword is `dash_html_components.Script.async` sharing the Python3.7 keyword. This caused a bug which was patched. For now we can release packages that were built with 3.7; however, moving forward we need a way to address this systematically so that people can make components which will be compatible with any version of Python.

I'm open to solutions on how to address this, here are a few of my own ideas:
**1. Release versioned PyPi packages**
By building classes for each version, we can use the versions `keyword.kwlist` list to avoid naming class initialization arguments a Python keyword.
This would require us to build more tooling, but I do not think it would be more work on from the users perspective (probably more build time to generate 3+ class files instead of one, but the process is pretty quick).
_Prop names that are Python keywords for the particular version the user has installed will not appear in type-ahead/auto-complete_

**2. Keep a master list of all Python keywords**
Avoid naming class initialization arguments values that were ever a Python keyword, so a single class file is compatible with every version of Python
_Prop names that were ever a Python keyword will not appear in type-ahead/auto-complete_

**3. If loading a class file fails, build a new working one at run-time**
We still have the logic to generate class files on the fly, which was used previously for every class. We can just use this as a fallback if loading a component class fails.
_Prop names that are Python keywords for the particular version the user has installed will not appear in type-ahead/auto-complete, and some IDEs will not support type-ahead for the components which have to use this fallback_

T4rk1n:
Thanks, now I understand what was the problem with the tests.

I think option 2 would be good, we could generate a keyword file for the python versions we support and merge them and use that list. If a new python version come out we just have to generate a new file for that and merge it.

rmarren1:
I think 2 is the simplest. I don't think keywords change that frequently so this should not be too annoying to maintain.

PR #450, opened 2018-11-06
Use all python keywords when checking if prop name is valid

rmarren1:
Solves https://github.com/plotly/dash/issues/361

Previously, we did not allow a prop name to be written as a keyword argument in a Python Component class if it was in `keyword.kwlist`. Since `keyword.kwlist` is version specific, class files were dependent on the Python version used to create them.

This PR removes this problem, Python Component class files built with any of our supported versions (2.7, 3.3, 3.4, 3.5, 3.6, 3.7) should work for all the other ones too.

Note that these props are still usable, they are just passed through `**kwargs` rather than a proper keyword argument, so IDEs will not autocomplete for these props. `async` in `html.Script` is the only component in the core libraries affected by this so far.


T4rk1n:
I thought those tests failures were resolved by https://github.com/plotly/dash/pull/439/commits/9e3ca4d0e05576ba4ed15edbee305f64cc6b22e0. Or are they new failures ?

rmarren1:
Locally I get this... strange
```
======================================================================
FAIL: test_serving_scripts (tests.test_react.IntegrationTest)
----------------------------------------------------------------------
  File "/home/ryan/Plotly/dash/tests/test_react.py", line 331, in test_serving_scripts
    self.assertEqual(response.status_code, 200)
AssertionError: 404 != 200

----------------------------------------------------------------------
```

T4rk1n:
Fails for me too, that file is not tested on ci, maybe it's old ?

rmarren1:
Oh ok I'm using `nose2` so it picked up that file. Looks like an old un-used test that was broken by a recent PR. Without it everything else works.

T4rk1n:
Try locking `dash_core_components==0.35.1`.

rmarren1:
Cool. Looks like `dcc==0.37.0` was breaking

T4rk1n:
Yes, there is this bug https://github.com/plotly/dash-core-components/issues/362, thought it was fixed in https://github.com/plotly/dash-core-components/pull/356 but I guess it just made it worse. I have a test that fails in the conditions, I tried on one version of the commits and it passed. but after merge it failed again.

https://circleci.com/gh/plotly/dash-core-components/2022?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #451, opened 2018-11-06
Combine extract-meta and component generation in a cli.

T4rk1n:
- Add `extract-meta.js`
- Combine `extract-meta.js` call and component generation.
- Wrap as a cli under the command `dash-generate-components`

Usage:
`$ dash-generate-components src/lib/components my_components_namespace`

cc @plotly/dash 

Closes plotly/dash-component-boilerplate#17


nicolaskruchten:
This seems like a good step forward. A few questions:

1. Is this backwards-compatible with existing components written using the boilerplate?
2. Can we see the matching PR in the boilerplate repo that calls this new script?
3. Should we wait until #449 is merged and refactor the two together, perhaps with an `--experimental-r-components` flag?

T4rk1n:
> Is this backwards-compatible with existing components written using the boilerplate?

Yes, the old code was not modified.

> Can we see the matching PR in the boilerplate repo that calls this new script?

Coming up soon.

> Should we wait until #449 is merged and refactor the two together, perhaps with an --experimental-r-components flag?

Just need to add `generate_class_file_r` call in the loop next to `generate_class_file`. Should be easy to refactor, can add the argument with argparse.

dash/development/component_generator.py
```diff
@@ -0,0 +1,94 @@
...
+import subprocess
+import shlex
+import os
+import textwrap
+
+from .base_component import generate_class_file
+
+
+# pylint: disable=too-many-locals
+def generate_components(component_src, output_dir):
```
nicolaskruchten: I would rename `output_dir` to `project_shortname` and not treat it as a path.

dash/development/component_generator.py
```diff
@@ -0,0 +1,94 @@
...
+        os.path.dirname(__file__),
+        '..',
+        'extract-meta.js'
+    ))
+
+    os.environ['NODE_PATH'] = 'node_modules'
+    cmd = shlex.split('node {} {}'.format(extract_path, component_src),
+                      posix=not is_windows)
+
+    namespace = os.path.basename(output_dir)
```
nicolaskruchten: I would just remove `namespace` as a variable here.

dash/development/component_generator.py
```diff
@@ -0,0 +1,94 @@
...
+    if len(sys.argv) != 3:
+        print(
+            'Invalid number of arguments'
+            ' expected 2 but got {}\n\n'
+            'Arguments: src output_directory'.format(len(sys.argv) - 1),
+            file=sys.stderr
+        )
+        sys.exit(1)
+    # pylint: disable=unbalanced-tuple-unpacking
+    src, out = sys.argv[1:]
```
nicolaskruchten: again, I would rename this `project_shortname` instead of `out`... because the R code won't output there.

dash/development/component_generator.py
```diff
@@ -0,0 +1,94 @@
...
+    for component_path, component_data in metadata.items():
+        name = component_path.split('/')[-1].split('.')[0]
+        components.append(name)
+        generate_class_file(
+            name,
+            component_data['props'],
+            component_data['description'],
+            namespace
+        )
+        print('Generated {}/{}.py'.format(namespace, name))
```
nicolaskruchten: for the Python piece, if it turns out that we do need `package.json` in the Python module directory next to `__init__.py` then I would do the copying here

nicolaskruchten:
:dancer: once the tests pass and the internal logic of the program are little bit more language-agnostic :)

dash/development/component_generator.py
nicolaskruchten: See also https://github.com/plotly/dash-component-boilerplate/issues/38

dash/development/component_generator.py
```diff
@@ -0,0 +1,93 @@
...
+    for component_path, component_data in metadata.items():
+        name = component_path.split('/')[-1].split('.')[0]
+        components.append(name)
+        generate_class_file(
+            name,
+            component_data['props'],
+            component_data['description'],
+            project_shortname
+        )
+        print('Generated {}/{}.py'.format(project_shortname, name))
```
nicolaskruchten: can we move this line into the `generate_class_file` function? It's Python-specific and there will likely be an R-specific version as well and i don't think it's good to duplicate the path-as-function-of-project_shortname logic here as well :)

dash/development/component_generator.py
```diff
@@ -0,0 +1,93 @@
...
+                project_shortname, status),
+            file=sys.stderr)
+        sys.exit(1)
+
+    metadata = json.loads(out.decode())
+
+    components = []
+
+    for component_path, component_data in metadata.items():
+        name = component_path.split('/')[-1].split('.')[0]
```
nicolaskruchten: maybe name this `component_name` just for extra clarity/readability

dash/development/component_generator.py
```diff
@@ -0,0 +1,93 @@
...
+            component_data['props'],
+            component_data['description'],
+            project_shortname
+        )
+        print('Generated {}/{}.py'.format(project_shortname, name))
+
+    with open(os.path.join(project_shortname, 'metadata.json'), 'w') as f:
+        json.dump(metadata, f)
+
+    with open(os.path.join(project_shortname, '_imports_.py'), 'w') as f:
```
nicolaskruchten: can we move this Python-specific stuff into a Python-specific place? doesn't have to be in this PR but just in general :)

dash/development/component_generator.py
T4rk1n: >  i don't think it's good to duplicate the path-as-function-of-project_shortname logic here as well

I don't understand what you mean by that, just print the component name ?

dash/development/component_generator.py
T4rk1n: It's actually duplicate code so let's refactor that for the old method too.

dash/development/component_generator.py
```diff
@@ -5,8 +5,8 @@
 import subprocess
 import shlex
 import os
-import textwrap
 
+from dash.development.component_loader import generate_imports
```
nicolaskruchten: should this also be a relative import? :)

dash/development/component_generator.py
```diff
@@ -0,0 +1,79 @@
...
+            'Error generating metadata in {} (status={})'.format(
+                project_shortname, status),
+            file=sys.stderr)
+        sys.exit(1)
+
+    metadata = json.loads(out.decode())
+
+    components = []
+
+    for component_path, component_data in metadata.items():
```
nicolaskruchten: as a general comment, this seems to be repeated code with https://github.com/plotly/dash/blob/d892b9e8d1c5a617eb1d9e5c2a23ea47bce558c2/dash/development/component_loader.py#L68 ... any way to DRY it up?

T4rk1n:
@nicolaskruchten  Tests are in https://github.com/plotly/dash-component-boilerplate/pull/40

@rmarren1 Please review

dash/development/component_loader.py
```diff
@@ -81,32 +84,7 @@ def generate_classes(namespace, metadata_path='lib/metadata.json'):
...
-            name,
-            componentData['props'],
-            componentData['description'],
-            namespace
-        )
-
-        # Add an import statement for this component
-        with open(imports_path, 'a') as f:
-            f.write('from .{0:s} import {0:s}\n'.format(name))
+    components = generate_classes_files(namespace, data, generate_class_file)
```
rmarren1: Four lines up is not relevant anymore, write append not used.

dash/development/base_component.py
```diff
@@ -882,3 +886,37 @@ def js_to_py_type(type_object, is_flow_type=False, indent_num=0):
...
+            ]
+            '''.format(
+                '\n'.join(
+                    'from .{0} import {0}'.format(x) for x in components),
+                ',\n'.join('    "{}"'.format(x) for x in components)
+            )
+        ).lstrip())
+
+
+def generate_classes_files(project_shortname, metadata, *component_generators):
```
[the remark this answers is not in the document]
rmarren1: Actually, perhaps we can just move these into `components_generator.py`

dash/development/component_generator.py
```diff
@@ -0,0 +1,95 @@
...
+    is_windows = sys.platform == 'win32'
+
+    extract_path = os.path.abspath(os.path.join(
+        os.path.dirname(__file__),
+        '..',
+        'extract-meta.js'
+    ))
+
+    os.environ['NODE_PATH'] = 'node_modules'
+    cmd = shlex.split('node {} {}'.format(extract_path, components_source),
```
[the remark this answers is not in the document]
T4rk1n: I use `shlex` to make sure it works across platforms, had problems on windows passing arguments without the `posix=False` argument.

nicolaskruchten:
@rmarren1 I think we need to keep supporting both methods for backwards-compatibility, otherwise older components may break, no?

T4rk1n:
> In generate_components, we can write metadata.json first, then use a call to dash.development.component_loader.generate_classes from within generate_components to generate the components.

Writing just to read back seems like a bad pattern, I initially did not write the `metadata.json` and just generate the components but figured it should still be there for backward compatibility.

> Just delete the dash.development.component_loader.generate_classes function and rely on the CLI.

I think they should stay there for backward compatibility. Someone working on a component library should be able to update dash without breaking their build system.

T4rk1n:
@rmarren1 

I moved all python class generation code to `_py_components_generation`, only kept the component, meta and registry in `base_component`. `component_generator` should be only code needed for the cli generation. So we'll add the code for R generation in `_r_components_generation` and use it in `component_generator` cli under a experimental flag at first. Maybe I should change the name of the cli command to `dash-component-generator` instead of `dash-generate-components` to match the name of the module ?

dash/development/component_generator.py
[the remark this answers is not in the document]
rmarren1: Sounds good :+1:

rmarren1:
:dancer:
I like this organizational change, should help to keep our python / r code separate and keep the `base_component` file slim.

Deleting the non-cli method of generating component class files would only break build systems for component authors and shouldn't affect things for end users (as long as we keep the `load_components` method), and this should require a one line change to `build:py`. I didn't think this was a 'breaking change' since it didn't affect end users, but I agree developers should be able to upgrade without having their build systems break and now am in favor of keeping it. We can just make a note of this and remove it during the `1.0.0` release.

nicolaskruchten:
@rmarren1 FWIW here's the terminology I use when thinking about our stakeholders:

* *Component Developers*: makers of components, users of boilerplate and component-generation.
* *App Developers*: makers of Dash apps, installers of components, users of Dash API
* *End Users*: users of Dash apps

In general our commitment should be not to break things for any of these folks :)

nicolaskruchten:
@T4rk1n I'm a bit confused about what the `component_loader`'s responsibility here is/when it's invoked... can you enlighten me?

PR #452, opened 2018-11-08
Validate component properties #264 - March 1

rmarren1:
**Note**: This PR is a continuation of https://github.com/plotly/dash/pull/340. That PR had ~80 comments (most were outdated and no longer relevant), and 96 commits. This PR is the same code, just more readable.
___
PR for #264 
#### Current Prerelease
```
pip install dash==0.35.0rc1
pip install dash-renderer==0.15.0rc1
pip install dash-core-components==0.42.0rc1
pip install dash-html-components==0.14.0rc6
```

Validation is on by default. To turn it off, you can set `app.config.suppress_validation_exceptions = True`

#### Test Cases to run for demo
```
import dash

app = dash.Dash()
app.scripts.config.serve_locally=True

app.layout = html.Div([
    html.Button(id='click1', children='click to return bad Div children'),
    html.Div(id='output1', **{'data-cb': 'foo'}),
    html.Button(id='click2', children='click to return a bad figure'),
    dcc.Graph(id='output2', figure={'data': [], 'layout': {}}),
    html.Button(id='click3', children='click to return a bad radio'),
    dcc.RadioItems(id='output3', options=[{'value': 'okay', 'label': 'okay'}]),
    html.Button(id='click4', children='click to make a figure with no id'),
    html.Div(id='output4'),
])



              [Input('click1', 'n_clicks')])
def crash_it1(clicks):
    if clicks:
        return [[]]
    return clicks

@app.callback(Output('output2', 'figure'),
              [Input('click2', 'n_clicks')])
def crash_it2(clicks):
    if clicks:
        return {'data': {'x': [1, 2, 3], 'y': [1, 2, 3], 'type': 'scatter'}, 'layout': {}}
    return go.Figure(data=[go.Scatter(x=[1,2,3], y=[1,2,3])], layout=go.Layout()) 

@app.callback(Output('output3', 'options'),
              [Input('click3', 'n_clicks')])
def crash_it3(clicks):
    if clicks:
        return [{'value': {'not okay': True}, 'labl': 'not okay'}]
    return [{'value': 'okay', 'label': 'okay'}]

@app.callback(Output('output4', 'children'),
              [Input('click4', 'n_clicks')])
def crash_it4(clicks):
    if clicks:
        return dcc.Graph()
    return dcc.Graph(id='hi')

app.run_server(debug=True, port=8050)
```

#### Example Error Messages
```
CallbackOutputValidationError: 


A Dash Callback produced an invalid value!

Dash tried to update the `figure` prop of the
`Graph` with id `output2` by calling the
`crash_it2` function with `(1)` as arguments.

This function call returned `{'layout': {}, 'data': {'y': [1, 2, 3], 'x': [1, 2, 3], 'type': 'scatter'}}`, which did not pass
validation tests for the `Graph` component.

The expected schema for the `figure` prop of the
`Graph` component is:

***************************************************************
{'validator': 'plotly_figure'}
***************************************************************

The errors in validation are as follows:

* figure	<- Invalid Plotly Figure:


    Invalid value of type '__builtin__.dict' received for the 'data' property of 
        Received value: {'y': [1, 2, 3], 'x': [1, 2, 3], 'type': 'scatter'}

    The 'data' property is a tuple of trace instances
    that may be specified as:
      - A list or tuple of trace instances
        (e.g. [Scatter(...), Bar(...)])
      - A list or tuple of dicts of string/value properties where:
        - The 'type' property specifies the trace type
            One of: ['mesh3d', 'splom', 'scattercarpet',
                     'scattergl', 'scatterternary', 'pie',
                     'surface', 'histogram', 'ohlc', 'heatmapgl',
                     'cone', 'scatterpolar', 'table',
                     'scatterpolargl', 'histogram2d', 'contour',
                     'carpet', 'box', 'violin', 'bar',
                     'contourcarpet', 'area', 'choropleth',
                     'candlestick', 'streamtube', 'parcoords',
                     'heatmap', 'barpolar', 'scattermapbox',
                     'scatter3d', 'pointcloud',
                     'histogram2dcontour', 'scatter', 'scattergeo',
                     'sankey']

        - All remaining properties are passed to the constructor of
          the specified trace type

        (e.g. [{'type': 'scatter', ...}, {'type': 'bar, ...}])

```
```
CallbackOutputValidationError: 


A Dash Callback produced an invalid value!

Dash tried to update the `options` prop of the
`RadioItems` with id `output3` by calling the
`crash_it3` function with `(1)` as arguments.

This function call returned `[{'value': {'not okay': True}, 'labl': 'not okay'}]`, which did not pass
validation tests for the `RadioItems` component.

The expected schema for the `options` prop of the
`RadioItems` component is:

***************************************************************
{'schema': {'allow_unknown': False,
            'nullable': False,
            'schema': {'disabled': {'type': 'boolean'},
                       'label': {'type': 'string'},
                       'value': {'type': 'string'}},
            'type': 'dict'},
 'type': 'list'}
***************************************************************

The errors in validation are as follows:

* options
 * 0
  * value	<- must be of string type
  * labl	<- unknown field

```
```
ComponentInitializationValidationError: 


A Dash Component was initialized with invalid properties!

Dash tried to create a `RadioItems` component with the
following arguments, which caused a validation failure:

***************************************************************
{'id': 'output3', 'options': [{'label': 'okay', 'value': {}}]}
***************************************************************

The expected schema for the `RadioItems` component is:

***************************************************************
{'className': {'type': 'string'},
 'dashEvents': {'allowed': ['change'], 'type': ('string', 'number')},
 'fireEvent': {},
 'id': {'type': 'string'},
 'inputClassName': {'type': 'string'},
 'inputStyle': {'type': 'dict'},
 'labelClassName': {'type': 'string'},
 'labelStyle': {'type': 'dict'},
 'options': {'schema': {'allow_unknown': False,
                        'nullable': False,
                        'schema': {'disabled': {'type': 'boolean'},
                                   'label': {'type': 'string'},
                                   'value': {'type': 'string'}},
                        'type': 'dict'},
             'type': 'list'},
 'setProps': {},
 'style': {'type': 'dict'},
 'value': {'type': 'string'}}
***************************************************************

The errors in validation are as follows:


* options
 * 0
  * value	<- must be of string type

```

#### PropTypes to Cerberus reference
| PropType  | Cerberus Schema Validated Against | Known Current Limitations |
| ---------- | ------------------------------------ | -------------------------- |
| `array`     | `{'type': 'list'}` | |
| `bool`      | `{'type': 'boolean'}` | |
| `func`      | `{}` | No validation occurs. |
| `object`      | `{'type': 'dict'}` | Validates that input is explicitly a `dict` object. We cannot be more general (e.g. `collections.abc.Mapping`) since the core `Component` is an instance of that.|
| `string`    | `{'type': 'string'}`  | |
| `symbol`    | `{}`  | No validation occurs |
| `node`    | `{'anyof': [{'type': 'component'}, {'type': 'boolean'}, {'type': 'number'}, {'type': 'string'}, {'type': 'list', 'schema': {'anyof': [{'type': 'component'}, {'type': 'boolean'}, {'type': 'number'}, {'type': 'string'}]}}]}`  | |
| `instanceOf(Object)`    | `{}`  | No validation occurs |
| `oneOf(['val1', 2])`    | `{'allowed': [None, 'val1', 2]}`  | Strings will have `'` characters stripped off each end. This is because `metadata.json` generation serializes the literal values as json, so for example `PropTypes.oneOf(['News', 'Photos'])` serializes to `["'News'", "'Photos'"]`.  https://github.com/reactjs/react-docgen/issues/57 |
| `oneOfType( [ PropTypes.string, PropTypes.bool ] )`    | `{'anyof': [{'type': 'string', 'type': 'boolean'}]}`  |  If one of the types is a `PropType` that cannot be validated individually (e.g. `PropType.func`), no validation will occur and the schema will effectively be `{}` |
| `arrayOf( PropTypes.number )`    | `{'type': 'list', 'schema': {'type': 'number'}}`  |  |
| `objectOf( PropTypes.number )`    | `{'type': 'dict', 'valueschema': {'type': 'number'}}`  |  |
| `shape( { k1: PropTypes.string, k2: PropTypes.number } )`    | `{'type': 'dict', 'allow_unknown': False, 'schema': {'k1': {'type': 'string}, 'k2': {'type': 'number'}}}`  |  |
| `any`    | `{'type': ('boolean', 'number', 'string', 'dict', 'list')}`  |  |


rmarren1:
@T4rk1n I made the commit history more readable, hopefully this is easier to review. This includes fixes to address the review you made in https://github.com/plotly/dash/pull/340

T4rk1n:
Thanks, I'll have a good look on friday review party.

dash/dash.py
```diff
@@ -884,6 +894,7 @@ def add_context(*args, **kwargs):
                     mimetype='application/json'
                 )
 
+            self.callback_map[callback_id]['func'] = func
```
T4rk1n: Why do that in two difference part ? Couldn't your additions in dispatch be in `add_context` ?

dash/dash.py
```diff
@@ -912,7 +923,85 @@ def dispatch(self):
                 c['id'] == component_registration['id']
             ][0])
 
-        return self.callback_map[target_id]['callback'](*args)
+        output_value = self.callback_map[target_id]['func'](*args)
```
T4rk1n: The old call to the functions also had `**kwargs`, is that handled somewhere else or just not needed ?

dash/dash.py
```diff
@@ -912,7 +923,85 @@ def dispatch(self):
...
+
+        # Only validate if we get required information from renderer
+        # and validation is not turned off by user
+        if (
+                not self.config.suppress_validation_exceptions and
+                'namespace' in output and
+                'type' in output
+        ):
```
T4rk1n: This would fit without the ()
```
if not self.config.suppress_validation_exceptions and \
                'namespace' in output and 'type' in output:
```

dash/dash.py
```diff
@@ -912,7 +923,85 @@ def dispatch(self):
...
+                ***************************************************************
+
+                The errors in validation are as follows:
+
+            """).format(
+                component_property=component_property,
+                component_name=component.__name__,
+                component_id=component_id,
+                callback_func_name=callback_func_name,
+                args='({})'.format(", ".join(map(repr, args))),
```
T4rk1n: 😸

dash/development/base_component.py
```diff
@@ -4,8 +4,13 @@
 import inspect
 import abc
 import sys
+import pprint
+from textwrap import dedent
+
 import six
 
+import dash.exceptions
```
T4rk1n: Shouldn't it be `import ..exceptions` to keep consistent with the relative imports ?

dash/development/base_component.py
```diff
@@ -280,6 +287,55 @@ def traverse_with_paths(self):
...
+        DashValidator.set_component_class(Component)
+        validator = DashValidator(
+            self._schema,
+            allow_unknown=True,
+        )
+        args = {
+            k: self.__dict__[k]
+            for k in self.__dict__['_prop_names']
+            if k in self.__dict__.keys()
+        }
```
T4rk1n: I don't like this, shouldn't need to get from `__dict__`, `getattr` would be cleaner I think.


```python
args = {
        k: v
        for k, v in ((x, getattr(self, x, None)) for x in self._prop_names)
        if v
}
```

dash/development/base_component.py
```diff
@@ -280,6 +287,55 @@ def traverse_with_paths(self):
...
+                The expected schema for the `{component_name}` component is:
+
+                ***************************************************************
+                {component_schema}
+                ***************************************************************
+
+                The errors in validation are as follows:
+
+
+            """).format(
```
T4rk1n: For those big templates string, I like to put them as global as constant at the top of the file instead of polluting the function. Then the format can be on a single line. You also don't need to dedent when it's at the top level.

dash/development/base_component.py
```diff
@@ -311,6 +367,139 @@ def __len__(self):
...
+            if any(schema_is_nullable(v) for v in values):
+                return True
+    return False
+
+
+def js_to_cerberus_type(type_object):
+    def _merge(x, y):
+        z = x.copy()
+        z.update(y)
+        return z
```
T4rk1n: I'd use merge elsewhere, it could be a utils method.

dash/development/base_component.py
```diff
@@ -311,6 +367,139 @@ def __len__(self):
...
+                            'string')
+                    }
+                }
+            ]
+        },
+        'element': lambda x: {'type': 'component'},
+        'enum': _enum,
+        'union': lambda x: {
+            'anyof': [js_to_cerberus_type(v) for v in x['value']],
+        },
```
T4rk1n: 😸

dash/development/base_component.py
```diff
@@ -311,6 +367,139 @@ def __len__(self):
...
+        'any': lambda x: {
+            'type': ('boolean',
+                     'number',
+                     'string',
+                     'dict',
+                     'list')
+        },
```
T4rk1n: Shouldn't any just return `True` validated for any prop type.

dash/development/base_component.py
```diff
@@ -412,23 +607,26 @@ def __repr__(self):
...
-        [('{:s}=Component.REQUIRED'.format(p)
-          if props[p]['required'] else
-          '{:s}=Component.UNDEFINED'.format(p))
-         for p in prop_keys
-         if not p.endswith("-*") and
-         p not in kwlist and
-         p not in ['dashEvents', 'fireEvent', 'setProps']] + ['**kwargs']
-    )
-
+    for p in list(props.keys()):
```
T4rk1n: props.keys is already iterable.

dash/development/validator.py
```diff
@@ -0,0 +1,112 @@
...
+class DashValidator(cerberus.Validator):
+    types_mapping = cerberus.Validator.types_mapping.copy()
+    types_mapping.pop('list')  # To be replaced by our custom method
+    types_mapping.pop('number')  # To be replaced by our custom method
+
+    def _validator_plotly_figure(self, field, value):
+        if not isinstance(value, (dict, plotly.graph_objs.Figure)):
+            self._error(
+                field,
+                "Invalid Plotly Figure: Not a dict")
```
T4rk1n: It says Not a dict but it also accepts `plotly.graph_objs.Figure`

dash/dash.py
rmarren1: I don't think it did, [this](https://github.com/plotly/dash/commit/033084416c334d2dbd4e7888b3e45196abecd112#diff-228c9aefac729977642672aa1416bacaR448) was the last change to that line and it had `*args`

dash/dash.py
rmarren1: We would have to pass `namespace`, `component_id`, `component_property`, to this since that info is available in `dispatch`. I originally thought this might clash with the other arguments (e.g. if you had a property called `namespace`) although I now think this would be fine, but it might make it difficult if we want to support passing props with keyword arguments in the future.

dash/development/base_component.py
rmarren1: Yeah I'll change, that looks better

dash/development/base_component.py
rmarren1: Yeah that looks better, not sure what I was thinking there..

dash/development/base_component.py
rmarren1: Sounds good

dash/development/base_component.py
```diff
@@ -311,6 +367,139 @@ def __len__(self):
...
+                try:
+                    int_value = int(string_value)
+                    schema['allowed'].append(int_value)
+                except ValueError:
+                    pass
+                try:
+                    float_value = float(string_value)
+                    schema['allowed'].append(float_value)
+                except ValueError:
+                    pass
```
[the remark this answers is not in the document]
rmarren1: I wanted to do along the lines of
```
if string_value can be cast as an integer:
    schema['allowed'].append(int(string_value))
if string_value can be cast as a float:
    schema['allowed'].append(float(string_value))
```
Not sure if there is a cleaner way to do this... this is what worked

dash/development/base_component.py
rmarren1: Yeah, I think I misread the definition of `PropTypes.any`. Will change

dash/development/base_component.py
rmarren1: oops, forgot to change that back

dash/development/base_component.py
```diff
@@ -311,6 +367,139 @@ def __len__(self):
...
+        return schema
+
+    converters = {
+        'None': lambda x: {},
+        'func': lambda x: {},
+        'symbol': lambda x: {},
+        'custom': lambda x: {},
```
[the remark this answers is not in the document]
rmarren1: I've included them in case they are used in 3rd party dash component libraries

dash/development/validator.py
```diff
@@ -0,0 +1,112 @@
...
+        except (ValueError, TypeError):
+            return False
+        return True
+
+    # pylint: disable=no-self-use
+    def _validate_type_number(self, value):
+        if isinstance(value, (int, float)):
+            return True
+        if isinstance(value, str):  # Since int('3') works
+            return False
```
T4rk1n: The lines after will cast to int/float if it's a string, but that check before that.

dash/dash.py
```diff
@@ -1054,5 +1148,9 @@ def run_server(self,
         :return:
         """
         debug = self.enable_dev_tools(debug, dev_tools_serve_dev_bundles)
+        if not debug:
+            # Do not throw debugging exceptions in production.
+            self.config.suppress_validation_exceptions = True
+            self.config.suppress_callback_exceptions = True
```
T4rk1n: :zap: Only suppress the props validation, we still want to validate the callbacks and it's not an expensive check performance wise.

dash/development/validator.py
rmarren1: This logic is actually to handle numpy ints and floats without having to add `numpy` as a run-time dependency just for type checking.
```
>>> isinstance(np.int(7), int)
True
>>> isinstance(np.float(7), float)
True
>>> isinstance(np.uint8(7), int)
False
>>> isinstance(np.int64(7), int)
True
>>> isinstance(np.int32(7), int)
False
>>> isinstance(np.float64(7), float)
True
>>> isinstance(np.float32(7), float)
False
```
But all numpy int types are castable to an `int`, and same for floats.

rmarren1: This would mean that `'7'` or `'0.7` would not be validated as `propType.number` even though they would have been automatically cast to an integer in the past. Not sure if we should also accept these? I think with `propTypes.number`, you should be using numbers in Python just to stay consistent, but this is more of a stylistic opinion rather than a requirement for correct functionality.

T4rk1n: We should not accept a string for a number, even if it's castable. Maybe add a comment that it is for the numpy types, because reading that is not so obvious and it look like a logic error.

rmarren1:
In response to those other alternatives...
1) I considered this but the main drawback was, as you mention, that it does not validate validate at runtime. For example, if you are creating components dynamically and setting their props via a variable, I do not think this solution would get you very far.
2-3) Honestly did not know these solutions existed. I think that both would have a similar code footprint, since a majority of the code in this PR is just converting a component's metadata.json to something Python understands, which would need to be done for both of these as well.
4) This actually looks very similar to Cerberus, the package used in this PR. Not sure if one is 'better', but they seem to have the same set of features.

T4rk1n:
I couldn't generate any validation errors when I tried, the schema of the component was always `{}`, do I need a version of dcc and html components ?

T4rk1n:
Same for:

```python
class MyOtherType:
    def __init__(self):
        pass


app.layout = html.Div([
    html.Div(MyOtherType())
])
```
Got `AttributeError: MyOtherType instance has no attribute '__trunc__'`

And:
`app.layout = html.Div(MyOtherType())` -> `TypeError: <__main__.MyOtherType instance at 0x00000000189B5708> is not JSON serializable`


`app.layout = html.Div(id={'hello': 'world'})` -> `TypeError: unhashable type: 'dict'`

This validated true:

`app.layout = html.Div(className={'hello': 'world'})`

rmarren1:
Updates:
- [X] Discovered and fixed an issue where `required` falsy values were not detected (e.g. on `dcc.Checklist`, if `values=[]` it would not properly detect it and say `values` is required during validation).
- [X] Removed old method of required component validation (e.g. a check in the `__init__` function of a component). This is now handled by validation done in this PR.
- [ ] Discovered a small issue in figure validation. The plan was to run a figure value for `go.Figure` for validation, but there is one issue in `dash-docs` where `go.Figure` will error when a proper plot would have been produced. (https://github.com/plotly/dash-core-components/issues/257#issuecomment-443344500) Not sure how to address this.

Bumped all release candidates.

rmarren1:
Tested the RCs with dash-docs, and a few Dash apps I have built.
@plotly/dash if you have a large dash app and some time, please try to run it with these RCs and let me know how it goes!

T4rk1n:
```
Dash tried to update the `figure` prop of the
`Graph` with id `output2` by calling the
`crash_it2` function with `(None)` as arguments.

This function call returned `Figure({
    'data': [{'type': 'scatter', 'uid': '91f6c03c-17bb-42a3-a502-4d9441696413', 'x': [1, 2, 3], 'y': [1, 2, 3]}],
    'layout': {}
})`, which did not pass
validation tests for the `Graph` component.

The expected schema for the `figure` prop of the
`Graph` component is:

***************************************************************
{'nullable': False, 'required': False, 'type': 'dict'}
***************************************************************

The errors in validation are as follows:

* figure	<- must be of dict type

You can turn off these validation exceptions by setting
`app.config.suppress_validation_exceptions=True`
```

Shouldn't `Figure` be valid ?

T4rk1n:
> `html.Button(id='click4', children='click to make a figure with no id'),`

I think the graph now put a random id if there's none.

chubukov:
Wanted to note that the issue I raised in the other thread still exists. Numerical types in checklist values and dropdown values (and probably other components too) still raise errors with the latest release.

rmarren1:
Updated top level comment, the correct way to turn off validation is by setting `app.config.suppress_validation_exceptions`, not `app.config.disable_component_validation`. This was changed to have a parallel config to the current `app.config.suppress_callback_exceptions`

chriddyp:
This was a really great effort and was in line to be a PR of the Year 🏆 . In the end, with the addition of alternative language backends for Dash ([DashR](https://github.com/plotly/dashr)), we decided to double down on the [Dash Dev Tools UI](https://community.plot.ly/t/announcing-dash-dev-tools/22481) and invest & maintain the validation code in the front-end.

PR #454, opened 2018-11-09
Enable more pylint rules for the tests folder

jbampton:
Here is a list of Pylint codes

http://pylint-messages.wikidot.com/all-codes


alexcjohnson:
@jbampton Thanks for the PR, and my apologies, we seem to have let this PR slip through the cracks and now we've changed the test structure underneath it. Anyway I'd like to see if we can just use the same linter settings for tests as for source - made a new issue for this, #604.

PR #461, opened 2018-11-19
Add assets css to percy snapshots

T4rk1n:


T4rk1n:
Those snapshot changed because of the reset.css changing the body to 0 padding.

PR #473, opened 2018-11-29
Fix `_imports_.py` generation

T4rk1n:
The generation for `_imports_.py` made a strange indentation for the `__all__` key.  This fix the indentation.

@nicolaskruchten  @rmarren1 Please review


rmarren1:
Looks good, I got it to print
```
from .App import App
from .Bar import Bar
from .Foo import Foo

__all__ = [
    "App",
    "Bar",
    "Foo"
]
```

Not sure what is wrong with the other one... was this causing errors? Should be tested here: https://github.com/plotly/dash/blob/951c8909a25ff14de7165fced9a1dfb7cb2b8849/tests/development/test_component_loader.py#L188-L190

T4rk1n:
I don't know what caused the errors, it's tested on the boilerplate and it was generating right but when I went to dcc it failed the tests with indentation before the `__all__`.

PR #474, opened 2018-11-29
push py-specific stuff down, pull generic stuff up

nicolaskruchten:
@T4rk1n please review :)

### DIFF
--- a/dash/development/_py_components_generation.py
+++ b/dash/development/_py_components_generation.py
@@ -1,6 +1,7 @@
 import collections
 import copy
 import os
+import json
 
 from dash.development.base_component import _explicitize_args
 from ._all_keywords import kwlist
@@ -164,7 +165,7 @@ def generate_class_file(typename, props, description, namespace):
     print('Generated {}'.format(file_name))
 
 
-def generate_imports(project_shortname, components):
+def generate_imports(project_shortname, components, metadata=None):
     with open(os.path.join(project_shortname, '_imports_.py'), 'w') as f:
         imports_string = '{}\n\n{}'.format(
             '\n'.join(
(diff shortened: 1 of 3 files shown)


dash/development/component_generator.py
```diff
@@ -10,16 +10,34 @@
...
+            generator(
+                component_name,
+                component_data['props'],
+                component_data['description'],
+                project_shortname
+            )
+
+    return components
+
+def generate_suite_files(project_shortname, components, metadata, *suite_generators):
```
nicolaskruchten: I'm pushing the writing of `metadata.json` into this function here, as the path will depend on the language.

nicolaskruchten:
@T4rk1n I can't get the autoformatting rules to work... can you format this as required please and add a commit? :)

T4rk1n:
Look like it worked.

dash/development/_py_components_generation.py
```diff
@@ -164,7 +165,7 @@ def generate_class_file(typename, props, description, namespace):
     print('Generated {}'.format(file_name))
 
 
-def generate_imports(project_shortname, components):
+def generate_imports(project_shortname, components, metadata=None):
```
T4rk1n: I think R will need the package.json info too, we should also pass that argument too.

nicolaskruchten: Can we add that in at the same time as the R stuff instead?

T4rk1n: Actually I think it's just a confusion, `metadata` should be `package_info`. We already mapped the `metadata.json` in generate_component_files, it's the `package.json` mapping that is needed for R. So just rename the argument to `package_info`.

nicolaskruchten: That doesn’t make sense to me. The file dump line I moved writes to metadata.json

nicolaskruchten: Is it wrong on master now?

T4rk1n: No, I forgot about the move of the metadata write, let's keep this as is and we'll modify the functions as needed in the R generation PR.

alexcjohnson:
@T4rk1n looks like there's still some useful stuff here, but might be easier to just recreate it with a new PR? Or would you prefer to ignore this for now and just take care of it in oomph?

T4rk1n:
> Or would you prefer to ignore this for now and just take care of it in oomph?

Ignore for now, this is just a refactor and the underlying code has changed.

Issue #96, opened 2017-07-24
is there a way to load static file like png or ttf in a custom dash component plugin package?


pingf:
I'm writting my own dash components on top of dash & react-semantic-ui.
just find that it's hard to use the icon in dash, cause most react lib pack png files(and many others like ttf) for the icon.

chriddyp:
Hm good question. Does the solution in #71 work for you?

chriddyp:
Er, it wouldn't really if you're writing your own bundle. The Dash component bundles only include CSS and JS bundles (see e.g. https://github.com/plotly/dash-core-components/blob/master/dash_core_components/__init__.py#L15-L43).

We could also include a way to specify the image assets of the bundle / package as well.

pingf:
yep, i've find that there are only two kinds of dist config(js and css) which makes hard to write components like icon on top of other react library.
for now , I'm writting my own components using react-semantic-ui, for the icon part, I have to use icons in awesome font(just include the awesome css and put the fa-xxx in the class).

yunake:
@chriddyp are there any plans to look into this? would you accept a PR that implements it? i need to serve map files for my components and an assortment of static resources such as fonts and icons.

chriddyp:
@yunake - I'm not planning on looking into this in the immediate future. I would accept a PR that would add this. This will likely be added to the https://github.com/plotly/dash-components-archetype repository. If possible, a little write up (in a github issue) on the general approach of how this would be accomplished would be really helpful for me and others before the PR gets submitted.

yunake:
Great!
I was thinking of creating a mechanism similar to _css_dist/Resources(), but special-cased for directories. For the sake of clean user interface, I think it would be preferable to have user specify his static artefact directory in his `__init__.py`, and have us hook it into `flask.send_from_directory`, under a component-specific url just like current resources. The downside is that the client code now needs to know the component name to figure out which URIs to generate to reach the resources, but I think this is a reasonable requirement, since these components would be tied to Dash anyway. What do you think?
I have no idea when I will work on this, for now I have a work-around with a custom flask route that handles everything for me, but as we expand our collection of modules, we certainly want to establish clear relationships between the resources.

yunake:
actually, since packages can be egg files and even pex, send_from_directory won't cut it - something like this might be more appropriate, with a pregenerated set of files+urls (for security) - this is what i currently use in my workaround:
```
from pkg_resources import resource_stream
flask.send_file(
[...]

PR #478, opened 2018-12-05
Issue 480 - Add .map file extension support to dash

### LINKED ISSUE #480: Add support for '.map' file extension
As a Dash developer, I find myself wanting (needing) to debug the inner workings of dash components and dash renderer -- since we bundle and uglify with webpack, that is difficult.

Adding support for '.map' extension support (with return type application/json) would allow dash components developers to expose source maps and have access to them at runtime.

Marc-Andre-Rivet:
Fixes #480.

Adding support for .js.map files in Dash so source map file is accessible from browser when available

Also, adding basic support for dynamic file loading. Otherwise the map gets injected as a script in the page, killing perf and just being generally useless!


dash/dash.py
```diff
@@ -492,7 +492,8 @@ def serve_component_suites(self, package_name, path_in_package_dist):
 
         mimetype = ({
             'js': 'application/JavaScript',
-            'css': 'text/css'
+            'css': 'text/css',
+            'map': 'application/json'
```
T4rk1n: What is the actual extension of those maps ? I remember seeing `.map.js`. In that case it would evaluate to `.js` since we take the last element of the split.

There's a stdlib package named `mimetypes` I would prefer if it would use that instead so we can support more than css/js but also images and other file types #96. It's missing a few types like javascript is guessed to `text/plain`, but we can add them with `mimetypes.add_type`.

dash/dash.py
```diff
@@ -492,7 +492,8 @@ def serve_component_suites(self, package_name, path_in_package_dist):
...
         })[path_in_package_dist.split('.')[-1]]
```
T4rk1n: I don't like this split, if someone send a filename with an extension that is not in the dict it will throw. We can go around that by using `.get` instead and return a 400 instead of a 500 if the content_type is None. But I would still prefer if we used `mimetypes.guess_type`.

dash/dash.py
Marc-Andre-Rivet: > What is the actual extension of those maps ? I remember seeing .map.js. In that case it would evaluate to .js since we take the last element of the split.

Correct source map extension is '.js.map'.

> There's a stdlib package named mimetypes I would prefer if it would use that instead so we can support more than css/js but also images and other file types #96. It's missing a few types like javascript is guessed to text/plain, but we can add them with mimetypes.add_type.

Using a library to manage mimetypes is a good idea in itself but is scope creep on what is a simple change.

dash/dash.py
Marc-Andre-Rivet: Same comment as above. This is to get base source map functionality in. This brings other concerns in that should be able to stand on their own and be their own issue or PR and feels out of scope.

T4rk1n: What I'm saying here is before adding the map option it would raise an error on requesting the map because of the direct access. We should put the dict in a variable and add a default: 

```python
content_types = {
    'map': 'application/json',
    ...
}
content_type = content_types.get(path_in_package_dist, 'text/plain')
```

I don't think guessing the content types with `mimetypes` would be enough to solve #96 anyway, the issue is more complex than that.

Marc-Andre-Rivet:
@T4rk1n I've created an issue to track this PR in a more normal manner and linking it above in the description.  Also updated the changelog.  As it should have been done from the start :)

dash/dash.py
Marc-Andre-Rivet: Problem with setting a default is that it prevents content type sniffing from browsers capable of doing so -- the relevant entry from the norm:

https://tools.ietf.org/html/rfc7231#section-3.1.1.5
> If a Content-Type header field is not present, the recipient
   MAY either assume a media type of "application/octet-stream"
   [...] or examine the data to determine its type.

Better off not applying defaults and letting the client do its thing if we do not have an opinion.

If we want a fast way to support unknown file types without providing the content-type, we're better off letting it pass through and just not providing a content type. As you pointed above for the ```mimetypes``` the issue issue is more complex than that.

I will create an issue for general file type support and include both the norm entry here and the ```mimetypes``` package solution in it. A combination of both is probably what we need: Mimetypes for 90%+ of cases, overrides for known cases (.map is not in the list), pass through / no content-type for unknown types.

Marc-Andre-Rivet:
@T4rk1n Created #481 for followup discussion on supporting arbitrary file extensions.

dash/dash.py
T4rk1n: Yes, what I want is the content type sniffing from the browser, so no default but still a `.get` so it doesn't throw. Thanks for doing the research.

dash/dash.py
```diff
@@ -378,15 +378,27 @@ def _relative_url_path(relative_package_path='', namespace=''):
 
         srcs = []
         for resource in resources:
+            is_dynamic_resource = (
+                'dynamic' in resource and
+                resource['dynamic'] is True
+            )
```
Marc-Andre-Rivet: Not sure what the preferred indentation pattern is for Python -- went with something that seemed natural with the syntax

dash/dash.py
```diff
@@ -378,15 +378,27 @@ def _relative_url_path(relative_package_path='', namespace=''):
...
+
             if 'relative_package_path' in resource:
                 if isinstance(resource['relative_package_path'], str):
-                    srcs.append(_relative_url_path(**resource))
+                    path = _relative_url_path(
+                        resource['relative_package_path'],
+                        resource['namespace']
+                    )
+                    if not is_dynamic_resource:
```
Marc-Andre-Rivet: We need to figure out what the relative path is because that's also where we register the file -- but only add it to srcs (will be used below for ```<script />``` entries if it's *not* dynamic

dash/dash.py
```diff
@@ -378,15 +378,27 @@ def _relative_url_path(relative_package_path='', namespace=''):
...
                 else:
                     for rel_path in resource['relative_package_path']:
-                        srcs.append(_relative_url_path(
+                        path = _relative_url_path(
                             relative_package_path=rel_path,
                             namespace=resource['namespace']
-                        ))
+                        )
+                        if not is_dynamic_resource:
+                            srcs.append(path)
```
Marc-Andre-Rivet: Same here..

dash/dash.py
```diff
@@ -1028,6 +1041,7 @@ def _walk_assets_directory(self):
 
                 full = os.path.join(current, f)
 
+                print(f)
```
Marc-Andre-Rivet: Oups! Will remove

dash/resources.py
```diff
@@ -19,6 +19,8 @@ def _filter_resources(self, all_resources, dev_bundles=False):
         filtered_resources = []
         for s in all_resources:
             filtered_resource = {}
+            if 'dynamic' in s:
+                filtered_resource['dynamic'] = s['dynamic']
```
Marc-Andre-Rivet: ```dynamic``` property needs to survive the sanitation process -- adding it to the list

dash/dash.py
T4rk1n: No `()` and single line, we prefer explicit:  `if 'dynamic' in resources and resource.get('dynamic')`. Also `.get` over `[key]` to avoid unnecessary KeyErrors. Only use `is` for comparing pointer references, ie `is None` to assert null and not just falsy.

Marc-Andre-Rivet: Did this because the line is too long for flake8's taste.

Marc-Andre-Rivet: How do I one line this and make flake8 happy? Apart from an actual function, etc.

T4rk1n: Break them using `\` at the end of the line.

Marc-Andre-Rivet: With '== True' I get this from flak8
dash/dash.py:382:41: E712 comparison to True should be 'if cond is True:' or 'if cond:'

Marc-Andre-Rivet: Getting the same error from pylint

Marc-Andre-Rivet: And now this.. dash/dash.py:355:4: R0912: Too many branches (13/12) (too-many-branches

I guess the additional `if` triggered this rule.

T4rk1n: I prefer `if cond` to `is True`, `is True` is when you need to assert that it the actual `True` and not another type that evaluate to True in a condition like a string.

Marc-Andre-Rivet: Ah! Misread! No need for the explicit comparison...

T4rk1n: Add `# pylint: disable=too-many-branches` to the start of the function.

dash/dash.py
```diff
@@ -378,15 +379,25 @@ def _relative_url_path(relative_package_path='', namespace=''):
 
         srcs = []
         for resource in resources:
+            is_dynamic_resource = 'dynamic' in resource and \
+                resource.get('dynamic')
+
             if 'relative_package_path' in resource:
                 if isinstance(resource['relative_package_path'], str):
```
rmarren1: An alternate way might solve the too-many-branches :camel:
```
paths = resource['relative_package_path']
paths = [paths] if isinstance(paths, str) else paths
for rel_path in paths:
   ... the logic that is repeated twice ...

```

dash/dash.py
```diff
@@ -378,15 +379,25 @@ def _relative_url_path(relative_package_path='', namespace=''):
 
         srcs = []
         for resource in resources:
+            is_dynamic_resource = 'dynamic' in resource and \
```
rmarren1: perhaps `resource.get('dynamic', False)` so it is not split

dash/dash.py
rmarren1: Can we instead just move the `self.registered_paths[namespace].add(relative_package_path)` out? I'm not sure why that logic is there in the first place, but now it is confusing to create an un-used local variable `path` in the case that `not is_dynamic_resource` only trigger the side effect of `_relative_url_path`. 

I think that would make this portion easier to understand, e.g.
```
self.registered_paths[namespace].add(relative_package_path)
if not is_dynamic_resource:
    srcs.append(_relative_url_path(
        resource['relative_package_path'],
        resource['namespace']
    ))
```

dash/dash.py
rmarren1: Agree that we should use mimetypes, but in a different PR.

dash/dash.py
rmarren1: I don't think this has to be changed here, that is outside of the scope of what this PR does -- just adding the `map` type by using the same pattern we have used for `js` and `css`, and that old pattern would throw an error. If we want to change the pattern that should be done in a different PR, perhaps in the same one we add the `mimetypes` module.

dash/dash.py
```diff
@@ -378,15 +376,29 @@ def _relative_url_path(relative_package_path='', namespace=''):
...
+                    if not is_dynamic_resource:
                         srcs.append(_relative_url_path(
-                            relative_package_path=rel_path,
-                            namespace=resource['namespace']
+                            resource['relative_package_path'],
+                            resource['namespace']
                         ))
+                else:
+                    for rel_path in resource['relative_package_path']:
+                        self.registered_paths[resource['namespace']]\
```
Marc-Andre-Rivet: @rmarren1 Not sure this is considered correct (escaping with '\' in the middle of a chain (as opposed to an 'if' as I've done elsewhere)

dash/dash.py
```diff
@@ -352,16 +352,14 @@ def serve_routes(self):
...
         # now needs the app context.
         # template in the necessary component suite JS bundles
         # add the version number of the package as a query parameter
         # for cache busting
         def _relative_url_path(relative_package_path='', namespace=''):
 
-            # track the registered packages
-            self.registered_paths[namespace].add(relative_package_path)
-
             module_path = os.path.join(
```
Marc-Andre-Rivet: As suggested, moved the side-effect outside of the path calculator and calling it directly below

dash/dash.py
Marc-Andre-Rivet: dictionary entry with default I guess?

dash/dash.py
Marc-Andre-Rivet: Sure.. I'm in, if only because it makes me learn the language haha!

Marc-Andre-Rivet:
@rmarren1 

> I'm also curious what the motivation behind 'dynamic' is? Do the .map files change at run-time independently of the bundles? I always thought they were static content, just static content that shouldn't be added to the <script> tag of a Dash app. Although, I am not sure what a good short word for that is which can be used as a variable name.

Good question. This is 'dynamic' in the same sense as Babel and Webpack defines it -- these files are loaded on a "per need" basis, or dynamically -- not to be confused with the other meaning of dynamic where the files would change at runtime. In that sense 'dynamic' really is the correct word vs. static, as opposed to say 'async'.

This is useful for source maps but also when splitting code into chunks that are only needed if certain features / behaviors are activated. Imagine a hypothetical component that could connect to provider A or B based on some settings -- if these libraries are big, it makes sense to delay the loading of these dependencies to when (and if!) they are needed. This need is already foreseen in the dash-table where certain features would require us to load libraries that are bigger than the table's bundle itself :)

dash/dash.py
rmarren1: Do you mean using the `\`? That is standard to do

Marc-Andre-Rivet:
Since this evolved from being a small modification (.map support) to adding an entirely new feature (dynamic support) I think we should make this a minor version bump instead of a patch.

rmarren1:
^ Agreed

Issue #491, opened 2018-12-14
Base component __repr__ logic is wrong.


rmarren1:
Base component `__repr__` is broken. It assumed that `_prop_names[0]` is always `children`, which is not true.

This breaks some things, for example if a component does not have `children` and thus its first prop is `id`, it will print the `id` as the first positional argument rather than a proper keyword argument. This is mostly relevant for `dash-html-components`.

For example:
```
Python 2.7.15rc1 (default, Nov 12 2018, 14:31:15) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import dash_core_components as dcc
>>> dcc.RadioItems(id='hello') 
RadioItems('hello')
>>> import dash_html_components as html
>>> html.Div(id='hello')
Div(id='hello')
```

rmarren1:
will fix in https://github.com/plotly/dash/issues/311

gvwilson:
Hi - we are tidying up stale issues and PRs in Plotly's public repositories so that we can focus on things that are most important to our community. If this issue is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. (Please note that we will give priority to reports that include a short reproducible example.) If you'd like to submit a PR, we'd be happy to prioritize a review, and if it's a request for tech support, please post in our community forum. Thank you - @gvwilson

PR #492, opened 2018-12-14
Fix __repr__, move it to the base component

rmarren1:
This does two things at once:
1. Fixes some [issues](https://github.com/plotly/dash/issues/491) in `__repr__`
2. (proposed) Move `__repr__` to the base component, as there is really no reason it needs to be part of the generated Python classes.

I realize now that 2 has the slight issue that a new component using an old version of `dash` will not have its own `__repr__` function, and it will not inherit one from the base component, so the resulting `__repr__` will be the default.

We have required updated component libraries to update `dash` before, e.g. [here](https://github.com/plotly/dash-html-components/blob/fec42728198d58a9d5153245c4ee525694ee4130/dash_html_components/__init__.py#L18-L23), but I think these occurrences should be rare, so I'm not really sure if we should also do 2 here and perhaps I should delete it and just do 1.

@T4rk1n please review 1, also do you have an opinion on whether we should do 2? (I would also have to re-build core component libraries with and add version checking to core libraries)


T4rk1n:
`__repr__` should be defined in the base class.

rmarren1:
Okay, then I'll add something along the lines of
```
if not _dash.version.__version__ > <version_with_repr>:
    print("Please update the `dash` module to >= <version_with_repr> to use this "
          "version of dash_html_components.\n"
          "You are using version {:s}".format(_dash.version.__version__),
          file=_sys.stderr)
```

dash/development/_py_components_generation.py
```diff
@@ -122,11 +93,21 @@ def __repr__(self):
...
+        namespace=namespace,
+        filtered_props=filtered_props,
+        list_of_valid_wildcard_attr_prefixes=wildcard_prefixes,
+        list_of_valid_keys=list_of_valid_keys,
+        docstring=docstring,
+        events=events,
+        default_argtext=default_argtext,
+        argtext=argtext,
+        required_props=required_args
+    )
```
T4rk1n: 😸

alexcjohnson:
@T4rk1n if you still want this PR included, please resolve conflicts and merge it.

PR #524, opened 2019-01-04
Remove ComponentRegistry dist cache.

T4rk1n:
The cache was preventing the dist from changing after a "reset" of an application with different component libraries requirements running on the same process.

Fixes T4rk1n/pytest-dash#15


T4rk1n:
@alexcjohnson Please review.

PR #535, opened 2019-01-13
Add two tests testing for embedding Dash in Flask and arbitrary WSGI apps

ned2:
These two tests cover the examples included in plotly/dash-docs#357

 - test_flask_integration: Test two Dash apps being mounted at routes within an
   existing Flask app
 - test_wsgi_integration: Test two Dash apps being mounted at routes alongside
   a Flask app using DispatcherMiddleware, which can support routing to any WSGI
   app

### DIFF
--- a/tests/IntegrationTests.py
+++ b/tests/IntegrationTests.py
@@ -39,7 +39,7 @@ class IntegrationTests(unittest.TestCase):
             return _error.apply(console, arguments);
         };
         '''
-    
+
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Closing as stale. Would be nice to include tests like this, but these ones are not compatible with the new `dash.testing`, also I'm not sure we want to explicitly invoke `multiprocessing` due to potential platform (ie windows) issues.

Issue #91, opened 2017-07-19
Dropdown 'x' button throws error


thomasantony:
The following code implements a simple Dropdown that updates a Div element. 

```
import dash
import dash_core_components as dcc
import dash_html_components as html

from dash.dependencies import Input, Output

app = dash.Dash()
app.layout = html.Div([
    html.H1(children='Dropdown test'),
    dcc.Dropdown(
                id='test-selector',
                placeholder='Select a value',
                options=[{'label':'foo', 'value':0},{'label':'bar', 'value':1},{'label':'baz', 'value':2}],
    ),
    html.Div(
        id='foobar'
    )

])

@app.callback(
    Output(component_id='foobar', component_property='children'),
    [Input(component_id='test-selector', component_property='value')]
)
def test_interaction(value):
    print(value)
    return value

if __name__ == '__main__':
    app.run_server(debug=True)
```

If you click the second or third options, a small "x" button appears on the right end of the component.
![dropdown-button](https://user-images.githubusercontent.com/1804570/28390673-afb423c8-6ca0-11e7-9972-44466a20de17.png)

I expected the dropdown to be cleared of any selection I clicked the 'x'. Instead, I get the following error in the console:

```
Uncaught TypeError: Cannot read property 'value' of null
    at Object.onChange (https://unpkg.com/dash-core-components@0.5.3/dash_core_components/bundle.js:13:31574)
    at Object.setValue (https://unpkg.com/dash-core-components@0.5.3/dash_core_components/bundle.js:11:11319)
    at Object.clearValue (https://unpkg.com/dash-core-components@0.5.3/dash_core_components/bundle.js:11:12224)
    at Object.r (https://unpkg.com/react-dom@15.4.2/dist/react-dom.min.js:14:5395)
    at a (https://unpkg.com/react-dom@15.4.2/dist/react-dom.min.js:12:22306)
    at Object.s [as executeDispatchesInOrder] (https://unpkg.com/react-dom@15.4.2/dist/react-dom.min.js:12:22496)
    at f (https://unpkg.com/react-dom@15.4.2/dist/react-dom.min.js:12:18835)
    at m (https://unpkg.com/react-dom@15.4.2/dist/react-dom.min.js:12:18961)
    at Array.forEach (native)
    at r (https://unpkg.com/react-dom@15.4.2/dist/react-dom.min.js:15:12116)
    at Object.processEventQueue (https://unpkg.com/react-dom@15.4.2/dist/react-dom.min.js:12:20039)
```

[...]

PR #545, opened 2019-01-18
Fixes to address unintended props reshuffling behaviour

rpkyle:
Proposed changes introduced by this PR:

- The ordering of `props` in Dash components is maintained by providing `object_pairs_hook=OrderedDict` to `json.loads` .
- Handling of `prop_keys` exclusions has been refactored to account for removal of events, and to properly omit wildcards.
- R components now require `children` to be passed as a list or function, as in Dash for Python, and ellipsis is used for wildcards only, as the final formal argument for component functions.
- Component library dependencies are now copied to `inst/deps` rather than `inst/lib` to address an issue with `install_github`.
- R documentation now has additional whitespace to improve readability.
- R (overall) package help files are now available.
- R component handling of `defaultValue` fields is now similar to Dash for Python, fixing issues with rendering of components due to passing incorrect defaults.
- Removed dependency of JavaScript metadata generation for DashR on `version.py`.
- Use modified version of Mark Amery's `byteify` function to avoid storing strings from JSON as Unicode in Python 2.x, so that lists of strings in components should be identical in appearance across versions.
- A new function `assert_valid_wildcards` is called when wildcards are available for a given component, to validate any wildcard input passed in via `...` in the corresponding function formals in R.


.gitignore
```diff
@@ -20,3 +20,4 @@ dist
 npm-debug*
 /.tox
 .idea
+package-lock.json
```
alexcjohnson: It shouldn't be necessary to ignore this, though there are occasional issues.

rpkyle: Sorry about that, will edit to remove!

dash/development/_r_components_generation.py
```diff
@@ -180,15 +159,13 @@ def generate_class_string(name, props, project_shortname, prefix):
...
     # Filter props to remove those we don't want to expose
-    for p in prop_keys:
-        if p.endswith("-*") \
-                or p in r_keywords \
-                or p in ['setProps', 'dashEvents', 'fireEvent']:
-            prop_keys.remove(p)
+    for item in prop_keys[:]:
+        if item.endswith('-*') \
+                or item in r_keywords \
+                or item == 'setProps':
```
T4rk1n: could be just:

```python
or item in r_keywords + ['setProps']
```

dash/development/_r_components_generation.py
```diff
@@ -163,7 +142,7 @@ def generate_class_string(name, props, project_shortname, prefix):
         '\'{}\''.format(p)
         for p in prop_keys
         if '*' not in p and
-        p not in ['setProps', 'dashEvents', 'fireEvent']
+        p != 'setProps' + ['**kwargs']
```
T4rk1n: That doesn't work.
```python
>>> 'setProps' + ['**kwargs']
  File "<input>", line 1, in <module>
TypeError: must be str, not list
```

```suggestion
        p not in ['setProps', '**kwargs']
```

alexcjohnson: Oh that's not quite it either... the `p not in [...]` or `p != ...` clause is a filter in an iterator - in the python generator code: 
https://github.com/plotly/dash/blob/e72f483c85aaa91e4316a548ed83ab40a6819452/dash/development/_py_components_generation.py#L118-L126

note the extra `[...]` from immediately after `.join(` to right before `+ ['**kwargs']` that turns this into a list that we can then append `'**kwargs'` to by adding a second list. I'm assuming in R `**kwargs` doesn't mean anything, but there's some other feature to play the role of allowing arbitrary extra named arguments?

alexcjohnson:
> I don't understand the `package-lock.json`, there's no `package.json` in dash. 🔪

Hah too funny! Right, `dash` is the one repo that doesn't use `npm` at all. Oh well, sounds like @rpkyle sorted this out :)

dash/development/_r_components_generation.py
rpkyle: > I'm assuming in R **kwargs doesn't mean anything, but there's some other feature to play the role of allowing arbitrary extra named arguments?

@alexcjohnson Correct. For the wildcards, I've used ellipsis (aka `\dots`, `...`) for this purpose. It seems to be working to pass in an arbitrary number (including zero) of wildcard arguments.

rpkyle:
> > I don't understand the `package-lock.json`, there's no `package.json` in dash. hocho
> 
> Hah too funny! Right, `dash` is the one repo that doesn't use `npm` at all. Oh well, sounds like @rpkyle sorted this out :)

I'm glad I've somehow fallen into the component management pipeline on the Python side as well, because for all the bumps and bruises, I'm certainly learning a few things.

dash/development/_r_components_generation.py
rpkyle: This didn't work:

```ts_generation.py", line 164, in generate_class_string
    or item in r_keywords + ['setProps']:
TypeError: unsupported operand type(s) for +: 'set' and 'list'```

**Comment by alexcjohnson** [comment]
> I attempted to address this by wrapping `props.keys()` with `list`;

Yeah, `.keys()` and various other things return iterators instead of lists in py3, that one continually trips me up / annoys me... you've got the correct solution.

**Comment by T4rk1n** [comment]
> File "dash/dash/development/_r_components_generation.py", line 282, in write_help_file for item in prop_keys[:]: TypeError: 'odict_keys' object is not subscriptable

Yes, `dict.keys()` returns a list in python 2 and a generator in python 3 so you need to cast to list.

File: dash/development/component_generator.py [hunk 4]
```diff
@@ -82,13 +89,17 @@ def generate_components(components_source, project_shortname,
     )
 
     with open(os.path.join(project_shortname, 'metadata.json'), 'w') as f:
-        json.dump(metadata, f)
+        json.dump(metadata, f, sort_keys=True)
```
T4rk1n: :rewind: Sorting the key reorder the keys, the ordered dict will keep the right order alone.

rpkyle:
@T4rk1n If PRs #439 and #91 look ready to go, mind quickly checking this one also? It should be somewhat straightforward.

PR #547, opened 2019-01-20
Fix assets blueprint path management

### LINKED ISSUE #529: Don't strip asset paths
Currently `routes_pathname` is stripped from `requests_pathname` for [asset paths only](https://github.com/plotly/dash/blob/2c182b14b070228efc9a69cfa96c57578e801e93/dash/_utils.py#L30-L43), causing these to be incorrect.

This was noted in https://github.com/plotly/streambed/issues/12111. We're working around this issue in DDS 3.0.0, but it would be good to fix the underlying bug.

@chriddyp @T4rk1n FYI

T4rk1n:
## Fixed
- Asset blueprint takes routes prefix into it's static path.
- Asset url path no longer strip routes from requests.

## Changed
- `assets_folder` argument now default to 'assets'
- The assets folder is now always relative to the given root path of `name` argument, the default of `__main__` will get the `cwd`.
- No longer coerce the name argument from the server if the server argument is provided.

Fixes #529

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -79,7 +79,6 @@ class Dash(object):
             assets_folder='assets',
             assets_url_path='/assets',
             assets_ignore='',
-            assets_blueprint_name='assets',
             include_assets_files=True,
             url_base_pathname=None,
             assets_external_path=None,
@@ -144,7 +143,7 @@ class Dash(object):
 
         assets_blueprint_name = '{}{}'.format(
             self.config.routes_pathname_prefix.replace('/', '_'),
-            assets_blueprint_name
+            'dash_assets'
         )
 
         self.server.register_blueprint(
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -76,9 +76,10 @@ class Dash(object):
             name='__main__',
             server=None,
             static_folder='static',
-            assets_folder=None,
+            assets_folder='assets',
             assets_url_path='/assets',
             assets_ignore='',
+            assets_blueprint_name='assets',
             include_assets_files=True,
             url_base_pathname=None,
             assets_external_path=None,
@@ -103,19 +104,14 @@ class Dash(object):
                 ''', DeprecationWarning)
 
         name = name if server is None else server.name
-        self._assets_folder = assets_folder or os.path.join(
-            flask.helpers.get_root_path(name), 'assets'
+        self._assets_folder = os.path.join(
+            flask.helpers.get_root_path(name),
+            assets_folder,
         )
         self._assets_url_path = assets_url_path
 
         # allow users to supply their own flask server
         self.server = server or Flask(name, static_folder=static_folder)
-
-        if 'assets' not in self.server.blueprints:
-            self.server.register_blueprint(
-                flask.Blueprint('assets', 'assets',
-                                static_folder=self._assets_folder,
-                                static_url_path=assets_url_path))
 
         env_configs = _configs.env_configs()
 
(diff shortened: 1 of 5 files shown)


dash/dash.py
```diff
@@ -76,9 +76,10 @@ def __init__(
             name='__main__',
             server=None,
             static_folder='static',
-            assets_folder=None,
+            assets_folder='assets',
             assets_url_path='/assets',
             assets_ignore='',
+            assets_blueprint_name='assets',
```
ned2: How important is this to be a user-configurable param? If we made the initial value `dash-assets` (which would still be prefixed by custom route values -- which was a good move btw) then it seems unlikely that a user would need to modify this.

Just thinking about slowing the rate of kwarg growth for the constructor where possible, as it will get a but unwieldy at some point.

ned2:
Here's the example app I was using to check this was working and find that issue with the static_folder value of the Blueplrint: https://github.com/ned2/dash-embed-recipes/tree/master/flask_multi

It's has the same layout as in the example I mentioned above.

dash/dash.py
```diff
@@ -146,6 +142,22 @@ def __init__(
...
 
+        assets_blueprint_name = '{}{}'.format(
+            self.config.routes_pathname_prefix.replace('/', '_'),
+            assets_blueprint_name
+        )
+
+        self.server.register_blueprint(
+            flask.Blueprint(
+                assets_blueprint_name, name,
+                static_folder='assets',
```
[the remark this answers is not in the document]
T4rk1n: Slipped in after a rebase thanks.

T4rk1n:
@ned2 Your examples are missing the `__name__` , but still it won't work because when we provide the server, the name is taken from the server instead so it won't be relative to the app but instead will be relative to where the server was instantiated.

Should we remove that or improve the logic here ?

https://github.com/plotly/dash/blob/e72f483c85aaa91e4316a548ed83ab40a6819452/dash/dash.py#L105

dash/dash.py
T4rk1n: Good point let's remove that, there's already much clutter in the kwargs. :ok_hand:

ned2:
> @ned2 Your examples are missing the `__name__` , but still it won't work because when we provide the server, the name is taken from the server instead so it won't be relative to the app but instead will be relative to where the server was instantiated.
> 
> Should we remove that or improve the logic here ?
> 
> [dash/dash/dash.py](https://github.com/plotly/dash/blob/e72f483c85aaa91e4316a548ed83ab40a6819452/dash/dash.py#L105)
> 
> Line 105 in [e72f483](/plotly/dash/commit/e72f483c85aaa91e4316a548ed83ab40a6819452)
> 
>  name = name if server is None else server.name

Yeah, I didn't include the `name` argument precisely because I knew that it would be changed to be the value of `server.name`. 

yep, so a potential change could be to remove that line overriding `name`, then `flask.helpers.get_root_path(name)` would resolve to the path the file with the Dash instance is found in, but `server.name` would still be whatever the original Flask instance was set to. 

Are there any other places that would be affected by decoupling the Dash `name` param from the Flask `name` param? I can't see any scanning through `dash.py`

ned2:
I just tested the effect of removing that line that overwrites the `name` variable with `server.name` on my example when both Dash instances are updated to pass the `name` parameter. Looks like it's all working exactly as it should now!

If you think this won't have any other unwanted repercussions, then I'm happy to make that change. I'm also happy with how everything else is looking.

:dancer:

T4rk1n:
Thanks @ned2, let's remove the line, I don't think it's useful.

PR #548, opened 2019-01-20
Remove print statement from PreventUpdate handler.

### LINKED ISSUE #516: raising PreventUpdate should not print to stderr, and PreventUpdate should not inherit Exception
Hi, I think the default implementation should not print to `stderr` when `PreventUpdate` is raised. This is a minor point, because the following is a work-around (and as the docs improve both `PreventUpdate` and example of a custom error handler could be documented!)

```
# Example of replacing the default error handler for PreventUpdate
app = dash.Dash(__name__)

@app.server.errorhandler(dash.exceptions.PreventUpdate)
def _handle_error(error):
    """Handle a halted callback and return an empty 204 response"""
    #print(error, file=sys.stderr)
    return ('', 204)
```

Also, as a minor point, PreventUpdate should probably inherit `BaseException` _without_ inheriting Exception - the reason is the same as why `StopIteration` does not inherit `Exception` (because its not an error).

T4rk1n:
Closes #516

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ## Fixed
 - Asset blueprint takes routes prefix into it's static path. [#547](https://github.com/plotly/dash/pull/547)
 - Asset url path no longer strip routes from requests. [#547](https://github.com/plotly/dash/pull/547)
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Alright, I'm convinced - https://github.com/plotly/dash/issues/516#issuecomment-456656483
💃

PR #550, opened 2019-01-20
Remove events

### LINKED ISSUE #531: Remove Event system
Expanding on https://github.com/plotly/dash/issues/469#issuecomment-449485114:
> Perhaps we can remove events here (in dash 1.0) as well?

We've had some offline discussions about this, but this deserves a more public issue both to allow any further input folks might have, and to tie to forthcoming PRs on the topic.

### Two parallel callback trigger systems: `Input` and `Event`

Most interactions in dash are handled by properties, referenced by either `Input` or `State` as callback parameters, or by `Output` as the result of a callback. But there's also an `Event` system, which can trigger a callback without a value, just based on some event occurring. This system was originally made to handle user interactions with no inherent visible consequences (so they don't have an obvious mapping to stateful properties), like button clicks. But for some time now we've had another way to handle this kind of scenario in a stateful way, which has all the same capabilities and more without needing a parallel system, and a lot of our newer components have no events defined at all. Dash 1.0 is the right time to remove `Event` entirely and only accept the property-based alternative.

### Transition from `Event` to `Input`

The key is to create a property that the component modifies every time the event occurs, so that the appropriate callback will fire on every event. We've generally done this by incrementing a counter. The most widely used right now in our codebase is the click event, where instead of emitting a `'click'` event we create a `'n_clicks'` property. One complication of this is that in the property version, the callback will be fired on initial render with `n_clicks=0`. To get around this, you can use `PreventUpdate`:

```py
@app.callback(Output('output', 'children'), [Input('button', 'n_clicks')])
def set_output(n_clicks):
    if not n_clicks:
        raise PreventUpdate
    return 'You clicked the button!'
```

An issue that we never had a simple solution for with `Event` was multiple events triggering the same callback, for example two buttons that should cause the same output component to behave differently. We have another set of properties you can use to handle this, based on timestamps. By convention these properties get the same name as the counter, suffixed with `_timestamp` - `n_clicks_timestamp` for example, and this would contain the timestamp (milliseconds since 1970) when the latest click occurred:

```py
              [Input('button1', 'n_clicks_timestamp'), Input('button2', 'n_clicks_timestamp')])
def set_output(ts1, ts2):
    if ts1 < 0 and ts2 < 0:
        # before any click, timestamp is -1 by convention
        raise PreventUpdate
    if ts1 > ts2:
        return 'You clicked button 1 most recently'
    else:
        return 'You clicked button 2 most recently'
```

I'll make corresponding issues in the other core repos to track any specific aspects of this relevant there and to link to PRs.

alexcjohnson:
Closes #531 

Dash complement to https://github.com/plotly/dash-renderer/pull/114 - note that the `no-events` branch of `dash-renderer` is needed to get tests to pass here, but not vice versa.

In the other repos I've been making a single `npm` command to run all the tests, and using that command in ci. Here we don't use `npm`, so I made a shell script `test.sh` for the same purpose.

The component generation routines, as modified here, are what I used to rebuild the html components in the already-merged https://github.com/plotly/dash-html-components/pull/89, and in a forthcoming PR in `dcc`


dash/development/_py_components_generation.py
```diff
@@ -122,7 +120,7 @@ def __repr__(self):
          for p in prop_keys
          if not p.endswith("-*") and
          p not in python_keywords and
-         p not in ['dashEvents', 'fireEvent', 'setProps']] + ['**kwargs']
+         p != 'setProps'] + ['**kwargs']
```
Marc-Andre-Rivet: Not sure if that was intentional but still present for the [R generation](https://github.com/plotly/dash/blob/32277157bd2874dcca63436959c43e4b1c7f227f/dash/development/_r_components_generation.py#L166)

tests/development/test_base_component.py
```diff
@@ -687,8 +691,8 @@ def test_docstring(self):
 
     def test_events(self):
         self.assertEqual(
-            self.ComponentClass().available_events,
-            ['restyle', 'relayout', 'click']
+            hasattr(self.ComponentClass(), 'available_events'),
+            False
```
Marc-Andre-Rivet: Maybe this test should be renamed to make clearer what it's testing now

Marc-Andre-Rivet:
This looks good to me
- changelog entry missing
- obviously need to point to master of dash-renderer after [this](https://github.com/plotly/dash-renderer/pull/114)

Would appreciate @T4rk1n or @rmarren1 input.

dash/development/_py_components_generation.py
rpkyle: Sorry, will modify my code to match.

alexcjohnson:
> obviously need to point to master of dash-renderer after [this](https://github.com/plotly/dash-renderer/pull/114)

That's been merged - at this point I believe we can simply make a release of `dash-renderer` off master, and it will stop processing events even if used with older versions of everything else that still supports them; then we can make new releases of everything else (the only PR I haven't made yet is `dcc`, coming soon...) acknowledging and locking in the lack of event support. I believe the only incompatible combination will be new `dash` with old `dash-renderer`.

tests/development/test_base_component.py
alexcjohnson: > Maybe this test should be renamed

good call -> https://github.com/plotly/dash/pull/550/commits/ec9aaeee4bd109febbfd8f86e626f36f1ae6af58

dash/development/_py_components_generation.py
```diff
@@ -274,30 +268,24 @@ def create_docstring(component_name, props, events, description):
...
+    ?
     """
-    if 'dashEvents' in props and props['dashEvents']['type']['name'] == 'enum':
-        events = [v['value'] for v in props['dashEvents']['type']['value']]
-    else:
-        events = []
-
-    return events
+    if 'dashEvents' in props or 'fireEvents' in props:
+        raise AttributeError('Events are no longer supported by dash')
```
T4rk1n: Can we add a custom exception or raise a DeprecationWarning instead.

test.sh
```diff
@@ -0,0 +1,15 @@
...
+python -m unittest tests.test_resources || EXIT_STATE=$?
+python -m unittest tests.test_configs || EXIT_STATE=$?
+
+if [ $EXIT_STATE -ne 0 ]; then
+    echo "One or more tests failed"
+else
+    echo "All tests passed!"
+fi
+
+exit $EXIT_STATE
```
T4rk1n: Can we include the linting also ?

test.sh
```diff
@@ -0,0 +1,15 @@
+EXIT_STATE=0
+
+python -m unittest tests.development.test_base_component || EXIT_STATE=$?
+python -m unittest tests.development.test_component_loader || EXIT_STATE=$?
+python -m unittest tests.test_integration || EXIT_STATE=$?
+python -m unittest tests.test_resources || EXIT_STATE=$?
+python -m unittest tests.test_configs || EXIT_STATE=$?
+
+if [ $EXIT_STATE -ne 0 ]; then
```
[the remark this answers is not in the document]
alexcjohnson: The `||` means `EXIT_STATE` only updates if the preceding command fails, so they'll all run but only the failing ones pass that failure along to `EXIT_STATE`. (taken from @etpinard, used in https://github.com/plotly/plotly.js/blob/master/.circleci/test.sh)

https://www.tecmint.com/chaining-operators-in-linux-with-practical-examples/

test.sh
alexcjohnson: Sure! I'll move linting in here too.

dash/development/_py_components_generation.py
alexcjohnson: Well, it wasn't deprecated, it was outright removed 😅 but I'm happy to make a custom exception - `ObsoleteEventsError`?

T4rk1n: I think `exceptions.NonExistantPropException` with a event removal message would fit the bill.

test.sh
alexcjohnson: linting into test.sh -> https://github.com/plotly/dash/pull/550/commits/2c9a3c892da5d1cf61435da5e79ff675bdff093a

dash/development/_py_components_generation.py
alexcjohnson: I used `NonExistentEventException`, which is the same I use for corresponding [obsolete callback usage](https://github.com/plotly/dash/pull/550/files#diff-228c9aefac729977642672aa1416bacaR724).

PR #561, opened 2019-02-01
use OrderedDict in byteify

alexcjohnson:
Quick followup to #545 - We maintain order in py3, but in py2 where we run the objects through `byteify` they were reverting to regular `dict`. This keeps them as `OrderedDict`.

Tested against `dcc` locally.


dash/development/_r_components_generation.py
```diff
@@ -25,8 +25,8 @@
         )
 
     component$props <- filter_null(component$props)
-    
-    structure(component, class = c('dash_component', 'list'))
```
alexcjohnson: Not sure why this trailing whitespace wasn't caught by CI earlier... it's going to generate diffs in all the component packages that use it.

PR #565, opened 2019-02-01
Version lock for the core components libraries

T4rk1n:
Added the core libraries to setup.py `install_requires`, installing dash will now also install:
- dash-core-components
- dash-html-components
- dash-table


setup.py
```diff
@@ -20,7 +20,10 @@
         'Flask>=0.12',
         'flask-compress',
         'plotly',
-        'dash_renderer',
+        'dash_renderer==0.17.0',
+        'dash-core-components==0.43.0',
+        'dash-html-components==0.13.5',
+        'dash-table==3.3.0'
```
alexcjohnson: Until we implement the tip-o-master cross-package testing protocol, do you want to update `.circleci/requirements/*` to match these versions?

.circleci/requirements/dev-requirements-py37.txt
```diff
@@ -1,5 +1,5 @@
-dash_core_components>=0.40.2
-dash_html_components==0.12.0rc3
+dash_core_components>=0.43.0
+dash_html_components==0.13.4
 dash-flow-example==0.0.3
 dash-dangerously-set-inner-html
 git+git://github.com/plotly/dash-renderer@master#egg=dash_renderer
```
alexcjohnson: I think renderer can go back to the published version now... this was supposed to be temporary during the no-events push (sorry about that!)

T4rk1n:
There's a circular dependency issue when we update the version on the core packages with these lock.

```
pkg_resources.DistributionNotFound: The 'dash-core-components==0.43.0' distribution was not found and is required by dash
```

T4rk1n:
Ok, so after testing, no need to install the local package with develop mode on circle, it will take the local folder for the package.

PR #605, opened 2019-02-12
Raise exception when same input & output are used in a callback

### LINKED ISSUE #323: Raise an exception when initializing callbacks with circular dependencies
```
@app.callback(Output('my-graph', 'figure'), [Input('my-graph', 'figure')])
```

Our users do this pretty often. For example, see here: https://community.plot.ly/t/object-as-input-output-error-loading-dependencies/12492

T4rk1n:
Closes #323

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -638,6 +638,12 @@ class Dash(object):
         # pylint: disable=too-many-branches
         layout = self._cached_layout or self._layout_value()
 
+        for i in inputs:
+            if output == i:
+                raise exceptions.SameInputOutputException(
+                    'Same output and input: {}'.format(output)
+                )
+
         if (layout is None and
                 not self.config.first('suppress_callback_exceptions',
                                       'supress_callback_exceptions')):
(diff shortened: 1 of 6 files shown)


dash/dash.py
```diff
@@ -638,6 +638,12 @@ def _validate_callback(self, output, inputs, state):
         # pylint: disable=too-many-branches
         layout = self._cached_layout or self._layout_value()
 
+        for i in inputs:
+            if output == i:
```
alexcjohnson: Going to need some changes for multi-output...

T4rk1n: Yep, I'll update once merged.

dash/dependencies.py
```diff
@@ -1,19 +1,35 @@
...
-# pylint: disable=old-style-class, too-few-public-methods
-class Input:
-    def __init__(self, component_id, component_property):
-        self.component_id = component_id
-        self.component_property = component_property
+    def __repr__(self):
+        return '<{} `{}`>'.format(self.__class__.__name__, self)
 
+    def __eq__(self, other):
+        return str(self) == str(other)
```
alexcjohnson: Do we want to restrict this to `isinstance(other, DashDependency)`? As is, `Input('id', 'value') == 'id.value'` which feels weird to me, but within our own classes it should be fine.

PR #606, opened 2019-02-13
[DO NOT MERGE] Add references to RCs for loading states / component

valentijnnieman:
Added references to the release candidates of all dash-* libraries involving loading states and the loading component. Let's not merge this yet - this will just be to have a RC version of `dash` up with the correct Loading states RCs!


valentijnnieman:
Released pre-release as `0.37.0rc1`

Marc-Andre-Rivet:
The Dash RC is pointing to dcc 0.44.0rc3, dcc was released on pypi as 0.43.0rc3 and as 0.44.0rc3 on NPM. Also, dash-renderer is trying to access `0.18.0rc4` instead of `0.18.0-rc4` when not run in local mode.

For now, I've updated dcc to 0.44.0rc3 and released that on pypi so the current Dash RC works.

When releasing an RC, please take the same care as when releasing an official version update. Make sure it works outside the development environment and without `serve_locally`.

PR #608, opened 2019-02-14
Callback context

### LINKED ISSUE #59: determine which input is triggering in callback
Since only a single Output is allowed in a callback, and all Inputs must feed into it, how can we determine which input is being triggered during a callback?  I have three time-series line plots that I want to cross-filter, and update, any time a range is selected in any of the other plots.   For example, the range of Plot A should update when either the range of Plot B or Plot C is updated.  But the callback collects the ranges from all plots simultaneously.  Is there a way to get the id of which input was triggered during a callback?  

Here is my code but it only updates with data from plotB, when it is triggered.  When I update Plot C the datac variable is indeed updated but the data from plotB is also passed in again, so the callback function has no knowledge of which plot is actually triggering new input.     

```
@app.callback(
    Output('plotA', 'figure'),
    [Input('plotB', 'relayoutData'),
     Input('plotC', 'relayoutData')])
def display_selected_data(data, datac):

    print('data', data)
    print('datac', datac)

    startx = 'xaxis.range[0]' in data if data else None
    endx = 'xaxis.range[1]' in data if data else None

   # define the new xrange
    if startx and endx:
        xrange = [data['xaxis.range[0]'], data['xaxis.range[1]']]
    elif startx and not endx:
        xrange = [data['xaxis.range[0]'], thedates.max()]
    elif not startx and endx:
        xrange = [thedates.min(), data['xaxis.range[1]']]
    else:
        xrange = None

    traces = [go.Scatter(
        x=thedates,
        y=mdf['uniqvisits'])]
    return {
        'data': traces,
        'layout': get_layout('Unique Visits', 'Date', 'Unique Visits', xrange=xrange)
    }  
```

### LINKED ISSUE #291: Determining Which Input Has Been Fired
We need a way to determine which input has changed. We've provided a temporary hack with `n_clicks_timestamp` but we need something that is more general.

I'm not worried about the implementation, it fits into the `dash-renderer` architecture nicely.

What's not as clear to me is what the `app.callback` decorated functions should look like. How does this interface scale when we want to add more things like:
- Previous state
- Multiple outputs
- Keyword-style arguments instead of list-style arguments
- Which input changed
- Timestamps when inputs changed
- Nested properties (e.g. subscribing to `figure.layout.title`)
- Error handling (e.g. updating an error div container when an exception is thrown)

So, let's use this thread to just propose different interfaces to the callbacks. Of course, doing this in a way that would be backwards compatible is preferred.

***

cc @plotly/dash

T4rk1n:
Add a callback context object for the dash package.

- `dash.callback.inputs/states` a dict with the id and prop as key for the inputs/states value of a callback.
- `dash.callback.triggered` which inputs has fired (called `setProps`)

These properties are only available inside a callback.

Basic example:

```python
import dash

from dash.dependencies import Output, Input

app = dash.Dash(__name__)

BUTTONS = ['btn-{}'.format(x) for x in range(1, 6)]

app.layout = html.Div([
    html.Div([
        html.Button(x, id=x) for x in BUTTONS
    ]),
    html.Div(id='output'),
])


              [Input(x, 'n_clicks') for x in BUTTONS])
def on_click(*args):
    if not dash.callback.triggered:
        raise PreventUpdate
    trigger = dash.callback.triggered[0]
    input_value = dash.callback.inputs.get(trigger)
    return 'Just clicked {} for the {} time!'.format(
        trigger.split('.')[0], input_value)


if __name__ == '__main__':
    app.run_server(debug=True, port=9091)

```

Closes #291, Closes #59

### DIFF
--- a/.circleci/requirements/dev-requirements-py37.txt
+++ b/.circleci/requirements/dev-requirements-py37.txt
@@ -2,7 +2,7 @@ dash_core_components==0.43.1
 dash_html_components==0.13.5
 dash-flow-example==0.0.5
 dash-dangerously-set-inner-html
--e git://github.com/plotly/dash-renderer.git@callback-context#egg=dash_renderer
+-e git://github.com/plotly/dash-renderer.git@master#egg=dash_renderer
 percy
 selenium
 mock
--- a/.circleci/requirements/dev-requirements-py37.txt
+++ b/.circleci/requirements/dev-requirements-py37.txt
@@ -2,7 +2,7 @@ dash_core_components==0.43.1
 dash_html_components==0.13.5
 dash-flow-example==0.0.5
 dash-dangerously-set-inner-html
-dash_renderer==0.18.0
+-e git://github.com/plotly/dash-renderer.git@callback-context#egg=dash_renderer
 percy
 selenium
 mock
--- a/.circleci/requirements/dev-requirements.txt
+++ b/.circleci/requirements/dev-requirements.txt
@@ -2,7 +2,7 @@ dash_core_components==0.43.1
 dash_html_components==0.13.5
 dash_flow_example==0.0.5
 dash-dangerously-set-inner-html
-dash_renderer==0.18.0
+-e git://github.com/plotly/dash-renderer.git@callback-context#egg=dash_renderer
 percy
 selenium
 mock
--- a/.circleci/requirements/dev-requirements.txt
+++ b/.circleci/requirements/dev-requirements.txt
@@ -2,7 +2,7 @@ dash_core_components==0.43.1
 dash_html_components==0.13.5
 dash_flow_example==0.0.5
 dash-dangerously-set-inner-html
--e git://github.com/plotly/dash-renderer.git@callback-context#egg=dash_renderer
+-e git://github.com/plotly/dash-renderer.git@master#egg=dash_renderer
 percy
 selenium
 mock
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 
 ## Added
 - Added components libraries js/css distribution to hot reload watch. [#603](https://github.com/plotly/dash/pull/603)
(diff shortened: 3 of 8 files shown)


alexcjohnson:
@T4rk1n great solution, and very nice architecture - I didn't know about `flask.g` but that certainly seems the ideal place to attach this kind of thing. And seems like this will extend nicely to handle things like previous state.

Only question is about the name - I feel like `dash.callback` is apt to get confused with `app.callback` - can we call it `request` or `context`?

T4rk1n:
Yes, flask.g is alive for the duration of the request only, so each callback get it's own "scope" managed.

> Only question is about the name - I feel like `dash.callback` is apt to get confused with `app.callback` - can we call it `request` or `context`?

That was also one of my concern, but I dismissed `request` because our users knows they are in a callback but maybe not a request, so it more specific and not to be confused with `flask.request` which handle much more. `context` just feel too generic.

T4rk1n:
- [ ] triggered only contains triggered inputs (now all props by setProps)

alexcjohnson:
> I dismissed `request` because our users knows they are in a callback but maybe not a request, so it more specific and not to be confused with `flask.request` which handle much more. `context` just feel too generic.

Fair points. I'm still not excited about `callback` though... in addition to the potential confusion, it implies that this *is* the callback, whereas it's really just extra info being provided *to* the callback, detailing how it was invoked. Perhaps `callback_info` or `callback_data` then? Or even `callback_context`, as you use internally. The way you set it up, this object is persistent, so users can import it up front with an alias `from dash import callback_context as ctx` or something - so I'm not super worried about a long name.

T4rk1n:
From discussion with @Bachibouzouk , the triggered inputs should also contains the value for a cleaner api.

```
triggered = [{'id': 'input.value', 'value': 'foo']
```

matthijsramlab:
This ia awesome, when will it be realsed?

T4rk1n:
@matthijsramlab Just merged, going into next release.

PR #609, opened 2019-02-14
Require the head of master for dash core repos

Marc-Andre-Rivet:
Same as for renderer, table, etc. require the head of master when running the dev environment for dash.

### DIFF
--- a/.circleci/requirements/dev-requirements-py37.txt
+++ b/.circleci/requirements/dev-requirements-py37.txt
@@ -1,8 +1,8 @@
-dash_core_components==0.43.0
-dash_html_components==0.13.5
+git+git://github.com/plotly/dash-core-components@master#egg=dash_core_components
+git+git://github.com/plotly/dash-html-components@master#egg=dash_html_components
+git+git://github.com/plotly/dash-renderer@master#egg=dash_renderer
 dash-flow-example==0.0.5
 dash-dangerously-set-inner-html
-dash_renderer==0.17.0
 percy
 selenium
 mock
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Seems like this will work here, since we're not actually installing `dash` itself. But in the other repos where we *are* installing `dash`, this would fight with the `install_requires` there which specifies specific versions of the other repos.

What I was imagining we'd do, to be more robust and to work across all the repos, is explicitly clone and `install -e` each one. Then we know for sure that the branch version wins, and it also gives us the option to rebuild them all - either to ensure we have the right build, or to create the build in case we do end up removing it from the repos.

Marc-Andre-Rivet:
> But in the other repos where we _are_ installing `dash`, this would fight with the `install_requires` there which specifies specific versions of the other repos.

I don't think it would. As we have to explicitly run `pip install -r <file>` for the deps to be installed. The way I understand these files up to now is that they are used to create the dev environment for our tests but have nothing to do with the distributable or its dependencies.

alexcjohnson:
Right, but in the dev environment for tests we'll be installing all the related repos - listing them in the `dev-requirements.txt` file as using the master branch from github - and one of those will be `dash` itself whose `install_requires` lists exact numbered (PyPI) versions of all the other repos. So which one will win, the master branch or the numbered release?

Anyway, that's not an issue in the `dash` repo and the build question is still open so not yet important. I'm happy to approve this as is, just curious about your thoughts longer term.

Marc-Andre-Rivet:
dash is actually not listed in the requirements list and the other repos only require dash as an environment requirement. afaik they don't have it in `install_requires` -- I don't see the loop.

alexcjohnson:
Superseded by https://github.com/plotly/dash/pull/627

PR #611, opened 2019-02-14
Fix hot reload adding the changed component dist to the assets.

T4rk1n:
Fixes plotly/dash-renderer#125


alexcjohnson:
Is #603 still accomplishing its goal after this change?

T4rk1n:
> Is #603 still accomplishing its goal after this change?

The reload hash is changed so it will initiate a hard reload.

PR #623, opened 2019-03-02
Custom response support via callback_context

### LINKED ISSUE #182: Set a cookie from callback function

alexcjohnson:
Closes #182 using an alternative approach to #401 as mentioned in https://github.com/plotly/dash/pull/401#issuecomment-468877351 - instead of making your own response object to return in place of the callback result, this PR makes a global object available that you can modify inside the callback (setting cookies or headers, for example)

I put this inside the `dash.callback_context` object, rather than making a new `dash.callback_response`, mainly to take advantage of the nice protection @T4rk1n built in there against using these outside of the callback, or accidentally clobbering them.

Per the example in the test, usage is like:
```py
@app.callback(Output('output', 'children'), [Input('input', 'value')])
def update_output(value):
    dash.callback_context.response.set_cookie(
        'dash cookie', value + ' - cookie')
    return value + ' - output'
```

cc @jkseppan @plotly/dash-core

### DIFF
--- a/dash/_callback_context.py
+++ b/dash/_callback_context.py
@@ -9,9 +9,10 @@ def has_context(func):
     def assert_context(*args, **kwargs):
         if not flask.has_request_context():
             raise exceptions.MissingCallbackContextException(
-                'dash.callback.{} is only available from a callback!'.format(
-                    getattr(func, '__name__')
-                )
+                (
+                    'dash.callback_context.{} '
+                    'is only available from a callback!'
+                ).format(getattr(func, '__name__'))
             )
         return func(*args, **kwargs)
     return assert_context
@@ -33,3 +34,8 @@ class CallbackContext:
     @has_context
     def triggered(self):
         return getattr(flask.g, 'triggered_inputs', [])
+
+    @property
+    @has_context
+    def response(self):
+        return getattr(flask.g, 'dash_response')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [Unreleased]
+## Added
+- Allow multiple outputs from a single callback. [#436](https://github.com/plotly/dash/pull/436)
+- Support for custom javascript hooks to modify callback payloads and responses. [#367](https://github.com/plotly/dash/pull/367)
+- Modify the flask response with custom cookies or headers, using `dash.callback_context.response`. [#623](https://github.com/plotly/dash/pull/623)
+
 ## [0.38.0] - 2019-02-25
 ## Fixed
 - Fix missing indentation for generated metadata.json [#600](https://github.com/plotly/dash/issues/600)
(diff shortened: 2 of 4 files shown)


dash/_callback_context.py
```diff
@@ -9,9 +9,10 @@ def has_context(func):
     def assert_context(*args, **kwargs):
         if not flask.has_request_context():
             raise exceptions.MissingCallbackContextException(
-                'dash.callback.{} is only available from a callback!'.format(
-                    getattr(func, '__name__')
-                )
+                (
+                    'dash.callback_context.{} '
+                    'is only available from a callback!'
+                ).format(getattr(func, '__name__'))
```
T4rk1n: Why change? I find that less readable. The parens are unnecessary.

alexcjohnson: The previous one no longer fit in 80 characters when I fixed `dash.callback` -> `dash.callback_context`.

Also, not sure if this is a pylint version issue or what, but the previous way it was written I was getting mutually exclusive errors about indentation: CI required it to be one way, my local pylint required something else. This way they both agree is acceptable.

alexcjohnson: oh no sorry, the pylint issue was somewhere else. This was just >80 chars.

RodrigoMejias:
How can you check the value of a cookie without changing the value?

hivemall:
i landed here from SO question [https://stackoverflow.com/a/48766639/4933053](https://stackoverflow.com/a/48766639/4933053). how do i redirect within a callback, eg after some validation user should go from '/' to '/start' url

jouvin:
I have problem very similar to @hivemall, found the SO post that redirected to this issue indirectly. I don't see how this PR helps with the problem of redirection. Is it possible to get more details or a pointer to a page giving some explanations on how to implement a "login redirect" in a callback. Is it necessary to do this by talking to Falsk directly with an approach like suggested by https://community.plotly.com/t/how-protect-dash-routes-with-login-after-dash-1-0-release/25114/3?

Tasselmi:
> I have problem very similar to @hivemall, found the SO post that redirected to this issue indirectly. I don't see how this PR helps with the problem of redirection. Is it possible to get more details or a pointer to a page giving some explanations on how to implement a "login redirect" in a callback. Is it necessary to do this by talking to Falsk directly with an approach like suggested by https://community.plotly.com/t/how-protect-dash-routes-with-login-after-dash-1-0-release/25114/3?

hi brother,  I have the same issue to integrate my app with google auth. Do you solve the problem?

lindorffs:
Anyone landing here after this post, for the love of god, this isn't a help form. Don't post here.

This is like, necroposting x10

PR #627, opened 2019-03-04
Improve build - test against real latest

Marc-Andre-Rivet:


.circleci/config.yml
```diff
@@ -30,10 +27,19 @@ jobs:
...
-              - "venv"
+      - run:
+          name: Install dependencies (dash)
+          command: |
+              git clone git@github.com:plotly/dash-renderer.git
+              git clone git@github.com:plotly/dash-core-components.git
+              git clone git@github.com:plotly/dash-html-components.git
+              git clone git@github.com:plotly/dash-table.git
+              . venv/bin/activate
+              pip install -e . --quiet
```
alexcjohnson: possibly more efficient to put this to after all the other `install -e` calls, so it doesn't try to install release versions of the rest of core only to override with the `-e` source versions later?

Marc-Andre-Rivet: will test installing it with `--no-deps`. The builds `generate_class` require dash to be in the environment.

Marc-Andre-Rivet: if it works, will apply to the other prs as they would have the same issue

Marc-Andre-Rivet: Causes this error https://circleci.com/gh/plotly/dash/3285 will building renderer

Obtaining file:///home/circleci/project/dash-renderer
dash 0.38.0 requires dash-core-components==0.43.1, which is not installed.
dash 0.38.0 requires dash-html-components==0.13.5, which is not installed.
dash 0.38.0 requires dash-table==3.5.0, which is not installed.
dash 0.38.0 requires Flask>=0.12, which is not installed.
dash 0.38.0 requires flask-compress, which is not installed.

PR #638, opened 2019-03-08
Fix documentation bug which omits children from R help pages

rpkyle:
The arguments and usage for components in `dashHtmlComponents` and `dashCoreComponents` did not properly list `children` in the list of supported arguments. 

This PR will update the 'transpiler' to address this issue, while removing references to `htmltools`, and an unnecessary repeated call to `reorder_props`. A dependency on the `dashR` package is introduced for 'transpiled' component libraries.


dash/development/_r_components_generation.py
```diff
@@ -78,7 +78,8 @@
 Version: {package_version}
 Authors @R: as.person(c({package_author}))
 Description: {package_description}
-Depends: R (>= 3.5.0)
+Depends: R (>= 3.0.2)
+Imports: dashR
```
alexcjohnson: We talked on Friday about separating dependency loading from the `Imports: dashR` search, in part by extending the function name to `.{package_name}_dashR_js_metadata()`. That said it does seem like `Imports: dashR` is appropriate for its regular purpose (ensuring the right packages are installed), particularly for packages outside the core.

Issue #519, opened 2018-12-30
Immediate callback validation makes combining layout and callbacks in same module difficult 


ned2:
Callback validation is great, especially for larger apps with more callbacks. In order to apply callback validation for multi-page apps, the trick described in the [Dash Docs](https://dash.plot.ly/urls) under _Dynamically Create a Layout for Multi-Page App Validation_, needs to be used. This involves creating a layout containing the concatenated layouts of all pages, which is only served if not within a request context (ie for for collback validation, but not for handling a web request). Doing this however makes it tricky to put layouts and callbacks of different pages within the same module.

Consider the following way of defining a layout function for multi-page apps that doesn't require setting `app.config.suppress_callback_exceptions = True`:

```python

from .pages import page1, page2, page3

pages = [page1, page2, page3]

def layout_with_validation():
    this_layout = my_layout_func
    if flask.has_request_context():
         return my_layout_func
    return [my_layout_func] + [page.layout for page in pages]
```
If the callbacks for each page are defined along with the layouts within each page module (as they are done in the example multi-page app layout in the [Dash Docs](https://dash.plot.ly/urls)), then this will raise a `dash.exceptions.LayoutIsNotDefined` exception as the `callback` method calls `_validate_callback` immediately, before the `app.layout` attribute has been assigned. This means `layout` attributes and direct uses of `app.callback()` can't be combined within the same module while supporting callback validation.

One workaround would be to defer evaluation of the callbacks by wrapping all callbacks in a function that then must be manually called after. This is a bit cumbersome, and makes swapping these pages between multi-page Dash apps and single page ones harder. Are there any more sensible workarounds that this?

Another possibility is changing Dash to defer calling `_validate_callback` on each callback until app load, such as in the `enable_dev_tools` method.

chriddyp:
> Another possibility is changing Dash to defer calling _validate_callback on each callback until app load, such as in the enable_dev_tools method.

This seems like a decent way to go. 

From that chapter in the user guide:
```
def serve_layout():
    if flask.has_request_context():
        return url_bar_and_content_div
    return html.Div([
        url_bar_and_content_div,
        layout_index,
        layout_page_1,
        layout_page_2,
    ])
```

This feels too complicated to me (checking for request context)

ned2:
> This seems like a decent way to go.

I'd be up for putting together a PR for this. I think this is an important part of providing a good development experience for building making Dash apps.

> This feels too complicated to me (checking for request context)

[...]

PR #644, opened 2019-03-11
[WIP] Delay callback validation

ned2:
This is a proposed approach for handling delayed callback validation. As discussed in #519, immediate callback validation requires the Dash layout to be already set, imposing rigid constraints on where callbacks can be defined within a Dash app. In particular, this making arrangement of pages within multi-page apps that contain both callbacks and layout fragments impossible. 

This change was more extensive than expected, so at the moment, this is perhaps more of way to get the conversation going.

* this approach involves adding to a list of registered callbacks in `Dash._callback_list` (distinct from the `Dash.callback_map`) when each callback is defined. The list is then used  to perform validation when the app starts
* I think it's critical that this callback validation be applied for WSGI apps as well as apps run through `run_server`, so I did not put this in the enable_dev_tools_folder, where I think there is a risk of users not running it. So instead I've created a `_setup()` method that should both be run by `run_server` and also `get_wsgi_application`, which would become the required way to get the WSGI callable for when using a WSGI server. This is modelled off [Django's approach](https://github.com/django/django/blob/master/django/core/wsgi.py). So this is also a proposed solution to the general problem of keeping the `run_server` and WSGI app entry points synchronised (which perhaps should go in a different PR though)
* I had to split the current callback validation tests into two, as the current code assumes that buggy callbacks can be caught through exceptions and continue on to use the already defined Dash app, which doesn't work that way now
* A slightly unsatisfying property of this solution is that for an app with invalid callbacks, before validation is applied (and an error thrown), the `Dash.callback_map` will be in a somewhat dirty state, with eg duplicate callbacks clobbering  existing ones.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -12,7 +12,6 @@ import threading
 import warnings
 import re
 import logging
-import pprint
 
 from functools import wraps
 
@@ -164,6 +163,9 @@ class Dash(object):
 
         # list of dependencies
         self.callback_map = {}
+
+        # for delayed callback validation
+        self._callback_list = []
 
         self._index_string = ''
         self.index_string = index_string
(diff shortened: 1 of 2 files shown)


ned2:
Here is the output that would be produced by the below invalid Dash app:
```
The following outputs have been assigned to multiple callbacks.

<Output `target1.children`> in callback functions: 'callback1', 'callback2'
<Output `target2.children`> in callback functions: 'callback3', 'callback4'

An output can only be associated with a single callback function.
Try combining your inputs and callback functions together into one
function.

The following outputs occur multiple times in a single multi-output
callback:

<Output `target2.children`> in callback function: 'callback4'

Multi-output callbacks cannot target the same output more than
once.
```

```python
import dash
from dash.dependencies import Input, State, Output

app = dash.Dash()
app.layout = html.Div([
    html.Div(id='target1'),
    html.Div(id='target2'),
    html.Div(id='target3'),
    dcc.Input(id='my-input', type='text', value=''),
    html.Button(id='submit', n_clicks=0, children='Save')
])

@app.callback(Output('target1', 'children'), [Input('submit', 'n_clicks')],
              [State('my-input', 'value')])
def callback1(n_clicks, state):
    return "callback received value: {}".format(state)

@app.callback(Output('target1', 'children'), [Input('submit', 'n_clicks')],
              [State('my-input', 'value')])
def callback2(n_clicks, state):
    return "callback received value: {}".format(state)

@app.callback(Output('target2', 'children'), [Input('submit', 'n_clicks')],
              [State('my-input', 'value')])
def callback3(n_clicks, state):
    return "callback received value: {}".format(state)

@app.callback([Output('target2', 'children'),
               Output('target2', 'children')],
              [Input('submit', 'n_clicks')],
              [State('my-input', 'value')])
def callback4(n_clicks, state):
    return "callback received value: {}".format(state)

if __name__ == '__main__':
    app.run_server(debug=True)
```

ned2:
Keen to get feedback from @T4rk1n, @alexcjohnson, and @chriddyp

alexcjohnson:
Thanks for pushing this forward @ned2 - I think I'll comment over at #519, then we can come back here if this ends up being the preferred approach.

alexcjohnson:
Closing - #1103 moved callback validation to the front end.

PR #669, opened 2019-03-28
Fix DeprecationWarning when importing ABCs.

mkcor:
Hi,

I'm addressing the warning I get whenever I run the tests locally:
```bash
<path>/dash/dash/development/base_component.py:265: DeprecationWarning:

Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```
I have just transposed https://github.com/plotly/plotly.py/pull/1417 to this repo.

### DIFF
--- a/dash/development/base_component.py
+++ b/dash/development/base_component.py
@@ -1,14 +1,14 @@
 import abc
 import inspect
 import sys
-try:
-    from collections.abc import MutableSequence
-except ImportError:
-    from collections import MutableSequence
-
 import six
 
 from .._utils import patch_collections_abc
+
+if six.PY2:
+    from collections import MutableSequence
+else:
+    from collections.abc import MutableSequence
 
 
 # pylint: disable=no-init,too-few-public-methods
--- a/dash/development/base_component.py
+++ b/dash/development/base_component.py
@@ -5,10 +5,11 @@ import six
 
 from .._utils import patch_collections_abc
 
+# pylint: disable=no-name-in-module, import-error
 if six.PY2:
-    from collections import MutableSequence  # pylint: disable=E0611
+    from collections import MutableSequence
 else:
-    from collections.abc import MutableSequence  # pylint: disable=E0611, E0401
+    from collections.abc import MutableSequence
 
 
 # pylint: disable=no-init,too-few-public-methods
--- a/dash/development/base_component.py
+++ b/dash/development/base_component.py
@@ -1,7 +1,10 @@
-import collections
 import abc
 import inspect
 import sys
+try:
+    from collections.abc import MutableSequence
+except ImportError:
+    from collections import MutableSequence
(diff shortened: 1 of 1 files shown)


mkcor:
@alexcjohnson I agree that the `if six.PY2` syntax is nicer. I don't have read access to https://github.com/plotly/dash-component-system but I guess the diff looks like 75785c7.

mkcor:
Ah, of course, now `pylint` errors as it parses module names which are not available in Python 2.7 and vice versa. I'm curious to know how @T4rk1n handled this: via `pylint` config?

alexcjohnson:
Oh that must be why he used `getattr` - ugh, I’d rather just silence that error (ideally only on that line)

T4rk1n:
I used https://github.com/plotly/dash/blob/0812b1ac0f3e74f3809aee3a1521ccc5528cc6a1/dash/_utils.py#L46-L49

mkcor:
@T4rk1n Ok. I'm surprised I'm getting this:

line_length

since https://github.com/plotly/dash/blob/0812b1ac0f3e74f3809aee3a1521ccc5528cc6a1/.pylintrc#L225

mkcor:
I'm also curious because I'm not able to run `pylint` on this file locally: I'm getting error
```bash
RuntimeError: generator raised StopIteration
************* Module dash.development.base_component
F:  1, 0: <class 'RuntimeError'>: generator raised StopIteration (astroid-error)
```

mkcor:
In the end, I silenced these `pylint` errors at the block level, because I noticed this was done throughout the file.
@alexcjohnson may I merge? Happy to rebase if five commits seem too much for such a small change.

alexcjohnson:
@mkcor I hadn't noticed this before, but @T4rk1n's comment points to our util function `patch_collections_abc`, which is already used in this module as:

https://github.com/plotly/dash/blob/b19c2c46be1eeba38f473a833d424da528b50212/dash/development/base_component.py#L66

So rather than duplicating, you could just use it again:
```py
MutableSequence = patch_collections_abc('MutableSequence')
```

@T4rk1n OK I understand what you did, but would you agree it's nicer to avoid `getattr` if you're just passing in a constant, and simply silence the linter error with a comment? (yes, block-level is fine, I just didn't want it silenced globally in the rc file)

T4rk1n:
> would you agree it's nicer to avoid getattr if you're just passing in a constant, and simply silence the linter error with a comment?

Yes, avoid if you want auto-completion and more clarity. What I wanted to avoid was the exception based solution.

mkcor:
@alexcjohnson oh, yes, sorry, somehow I didn't read @T4rk1n's [comment](https://github.com/plotly/dash/pull/669#issuecomment-478018959) carefully either!

PR #672, opened 2019-03-30
clientside callback interface

chriddyp:
see associated PR & examples in
https://github.com/plotly/dash-renderer/pull/143


dash/dash.py
```diff
@@ -631,6 +631,7 @@ def dependencies(self):
                 'output': k,
                 'inputs': v['inputs'],
                 'state': v['state'],
+                'clientside_function': v.get('clientside_function', {})
```
alexcjohnson: Personally I'd default to `None` rather than `{}`. It seems clearer and the logic is simpler (no need for the `isEmpty` clause)

dash/dependencies.py
```diff
@@ -32,3 +32,16 @@ class Input(DashDependency):
 # pylint: disable=too-few-public-methods
 class State(DashDependency):
     """Use the value of a state in a callback but don't trigger updates."""
+
+
+# pylint: disable=too-few-public-methods
```
alexcjohnson: Do you actually need this pylint directive? You have 2 public methods, I think that's what it's looking for.

That said, I think only the first of these directives is necessary anyway, and applies to the rest of the file https://github.com/plotly/dash/pull/679#discussion_r272357329 - we should see if perhaps putting the directives inside the relevant classes rather than before would scope them as desired.

chriddyp: Looks like `__init__` & `__repr__` aren't considered public 😿 

You're totally right that these comments were being applied to the entire file! Had to mix and match the location of these comments in order to get them applied, see https://github.com/plotly/dash/commit/b9886d9ccbe36b8e7e9f830e68cdca56d3fc797f

chriddyp:
Thanks @alexcjohnson . I have addressed your comments

betizad:
Can you please direct to some examples on how to use this client-side callbak?

PR #675, opened 2019-04-01
Also ignore directories when generating components.

T4rk1n:


dash/extract-meta.js
```diff
@@ -86,7 +86,8 @@ function collectMetadataRecursively(componentPath) {
         }
         dirs.forEach(filename => {
             const filepath = path.join(componentPath, filename);
-            if (fs.lstatSync(filepath).isDirectory()) {
+            if (fs.lstatSync(filepath).isDirectory()
+                && ignorePattern.test(filename)) {
```
alexcjohnson: Should this be inverted `!ignorePattern.test(filename)`?

T4rk1n: Yes

PR #676, opened 2019-04-01
:sparkles: add dev_tools_ui in config

### LINKED ISSUE #674: Add enable dev tools config
this is a follow up task of [dash-renderer #100](https://github.com/plotly/dash-renderer/pull/100)

it adds an extra dev-tools config in `<script id="_dash-config" type="application/json">` section, the logic is simple, the value equals to the `debug` option in `run_server()`.

user set app.run_server(debug=True) will implicitly enable the dev-tools icon in FE.

byronz:
this fixes #674 

This PR introduces a new config `dev_tools_ui` to let FE know about if user has set debug=True. this implicitly enables the dev tools UI 

image

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -237,7 +237,8 @@ class Dash(object):
             'hot_reload': False,
             'hot_reload_interval': 3000,
             'hot_reload_watch_interval': 0.5,
-            'hot_reload_max_retry': 8
+            'hot_reload_max_retry': 8,
+            'dev_tools_ui': False,
         })
 
         # add a handler for components suites errors to return 404
@@ -328,8 +329,9 @@ class Dash(object):
     def _config(self):
         config = {
             'url_base_pathname': self.url_base_pathname,
-            'requests_pathname_prefix': self.config['requests_pathname_prefix']
-        }
+            'requests_pathname_prefix': self.config['requests_pathname_prefix'],
+            'dev_tools_ui': self._dev_tools.dev_tools_ui,
+        }  # noqa: E501
         if self._dev_tools.hot_reload:
             config['hot_reload'] = {
                 'interval': self._dev_tools.hot_reload_interval,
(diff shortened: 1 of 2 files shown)


dash/dash.py
```diff
@@ -328,7 +329,8 @@ def serve_layout(self):
     def _config(self):
         config = {
             'url_base_pathname': self.url_base_pathname,
-            'requests_pathname_prefix': self.config['requests_pathname_prefix']
+            'requests_pathname_prefix': self.config['requests_pathname_prefix'],  # noqa: E501
```
T4rk1n: Could also use the dot notation for less characters: `self.config.requests_pathname_prefix`.

alexcjohnson: Oh interesting, didn't realize we had a js-like dict class `AttributeDict`. FWIW that came in 18 months ago: https://github.com/plotly/dash/commit/885340256a92d2242c622a8e57fe7064890e80cb for backward compatibility with an old class-based config. Honestly I find that a bit confusing and un-pythonic, has it been long enough that we could get rid of the dot version?

Would also be nice to remove the spelling mistake fallback `supress_callback_exceptions` and just throw an error if someone tries to use the misspelled version. Maybe even list the allowed keys explicitly and throw a helpful error if any other key is provided.

byronz: I used this library `attrdict` before, and mostly the use case is for config.  yes we should create an issue for supress typo if there is no one yet
https://pypi.org/project/attrdict/

nicolaskruchten:
Does this PR enable setting debug=True and dev_tools_ui=False, which would hide the UI?

byronz:
> Does this PR enable setting debug=True and dev_tools_ui=False, which would hide the UI?

this PR is necessary if we wanna explicitly set the switch in the renderer

nicolaskruchten:
I'm not sure you answered my question... Is it possible with this PR to set debug mode to True *except for* the dev-tools UI?

byronz:
@nicolaskruchten there is no such `dev_tools_ui=False` in python code, it just adds a config field (redux state will parse the same config) which binds directly the value of `debug`.

Issue #677, opened 2019-04-02
Use inspect for automatic callback linking


kmader:
One of the really useful parts of Shiny in R is how easy it is to get started for beginners. I think the callbacks could be greatly simplified in dash by automatically creating the inputs and outputs by using the callback function name and arguments name to fill in the field (as an additional, but not necessary annotation option). It would also remove a number of the common bugs when arguments are out of order and the hassle of having to come up with meaningful, consistent names for callbacks.

I apologize if this is a duplicate issue, but a brief perusal didn't show anything that looked too closely related.

kmader:
Something like
```
app_easydash = dash.Dash()

app_easydash.layout = html.Div([
    html.H1([html.H2([])]),
    dcc.Input(id='simple_input', value='initial value', type="text"),
    html.Div(id='simple_output')
])

@app_easydash.easy_callback()
def update_simple_output_children(simple_input_value):
    return 'You\'ve entered "{}"'.format(simple_input_value)
```

kmader:
Using a wrapper this can easily be done
```
import inspect
from dash.dependencies import Input, Output
def easy_callback(dash_app, debug=False):
    valid_prefix_names = ['update_', 'callback_']
    def check_prefix(in_name):
        for prefix_name in valid_prefix_names:
            if prefix_name in in_name:
                return in_name.find(prefix_name)+len(prefix_name)
        
        raise ValueError('EasyDash requires callback name to start with {}'.format(valid_prefix_names))
    def process_output(callback_func):
        callback_name = getattr(callback_func, '__name__', '')
        comp_prop_start_idx = check_prefix(callback_name)
        comp_prop_name = callback_name[comp_prop_start_idx:]
        property_start = comp_prop_name.rfind('_')
        if property_start>=1:
            prop_name = comp_prop_name[property_start+1:]
            comp_name = comp_prop_name[:property_start]
        else:
            # assume children if no underscore
            comp_name = comp_prop_name
            prop_name = 'children'
        return Output(component_id=comp_name, component_property=prop_name)
    def process_input(callback_func):
        spec = inspect.getfullargspec(callback_func)
        input_list = []
        for c_comp_prop_arg in spec.args:
            property_start = c_comp_prop_arg.rfind('_')
            if property_start>=1:
                comp_name = c_comp_prop_arg[:property_start]
                prop_name = c_comp_prop_arg[property_start+1:]
            else:
                # assume value if no underscore
                comp_name = c_comp_prop_arg
[...]

PR #678, opened 2019-04-02
Adding a smart `auto_callback` functions to populate callback information

kmader:
Basically a response for the issue #677 

makes a simple program like this possible by automatically determining inputs and outputs
```
app = dash.Dash()

app.layout = html.Div([
    html.H1([html.H2([])]),
    dcc.Input(id='simple_input', value='initial value', type="text"),
    html.Div(id='simple_output')
])

@app.auto_callback()
def update_simple_output_children(simple_input_value):
    return 'You\'ve entered "{}"'.format(simple_input_value)
```

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1042,8 +1042,10 @@ class Dash(object):
                 if prefix_name in in_name:
                     return in_name.find(prefix_name) + len(prefix_name)
 
-            raise ValueError('auto_callback requires callback name to start with {}'.format(
-                valid_prefix_names))
+            raise ValueError(
+                'auto_callback requires name to start with {}'.format(
+                valid_prefix_names)
+            )
 
         def process_output(callback_func):
             callback_name = getattr(callback_func, '__name__', '')
@@ -1057,7 +1059,8 @@ class Dash(object):
                 # assume children if no underscore
                 comp_name = comp_prop_name
                 prop_name = 'children'
-            return Output(component_id=comp_name, component_property=prop_name)
+            return Output(component_id=comp_name,
+                          component_property=prop_name)
 
         def process_input(callback_func):
             if sys.version_info[0] == 3:
@@ -1067,7 +1070,10 @@ class Dash(object):
                 # pylint: disable=maybe-no-member
                 spec_args = inspect.getargspec(callback_func).args
             else:
-                raise ValueError("Requires Python 2 or 3, {}".format(sys.version_info))
+                raise ValueError(
+                    "Requires Python 2 or 3, {}".format(
+                    sys.version_info)
+                )
 
             input_list = []
             for c_comp_prop_arg in spec_args:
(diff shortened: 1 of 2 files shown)


chriddyp:
This is really cool. I've never seen `inspect` used in this way - it's very neat, my wheels are turning. It's compact and the symmetry between the IDs and the variable is really easy on the eyes.

However, I'd like to keep a single interface for `@app.callback`. My main concern is that there should be "one way of doing things". Although this solution is more convenient, it has a set of nuances (State isn't supported, variable names matter in certain scenarios but not others) that I think would end up complicating our pedagogy & documentation.

In more detail, here are some scenarios/guiding philosophies that come to mind:

1. A smaller API helps users when faced with an uncertain error: In helping users in the [community forum](https://community.plot.ly/c/dash), I've found that when certain users encounter problems, they will try many different permutations of all of the available interfaces. This comes up frequently with `dcc.Graph(figure=` - if a chart isn't displaying right, they might switch between e.g. `go.Scatter(...)` and `{'type': 'scatter', ...}` even though they represent the same object under the hood. A smaller API with fewer interfaces implicitly reduces the number of permutations of code that users will try when faced with a bug or a problem.
2. A smaller API keeps 3rd party documentation aligned and relevant. We can keep our own dash documentation up-to-date with the different interfaces & their nuances, but we can't control 3rd party tutorials, articles, and videos. In this case, I'd be worried about 3rd party documentation not accurately describing the differences between `@app.callback` and `@app.auto_callback` (or not documenting `@app.callback` at all!). Users who learn Dash through 3rd party sites may see a mixture of `app.callback` vs `app.auto_callback` and may not learn the differences from first principles and may become confused. We've seen this ourselves when we had to deprecate `Events` - certain tutorials on the web from last year used `Events` and users run into issues when running that code (https://community.plot.ly/t/events-equivalent-in-0-39/21664). I wish that I didn't introduce Events at all in the first place!
3. Declarative, descriptive interfaces help the reader. In workshops, when teaching Dash to new users, I’ve found the explicit “Output” and “Input” naming to be helpful: “Describe the inputs and outputs of your function.“. I’ve found that there is some confusion around variable names, I usually have to say “The variable names don’t matter, it’ll just match the order of Inputs”. If we had this second API, then the variable names _would_ matter but only in _certain contexts_. Users might have less to learn upfront (they could ignore `dash.dependencies`) but in the end they'd have to learn about `Input` and `Output` _and_ the nuances between these two interfaces. In the Dash pedagogy, I'd rather spend some more time upfront laying the foundations.

I really appreciate you taking some time to think about how to improve Dash! We won't accept this PR now but I hope this won't discourage you from contributing again some time in the future.

kmader:
@chriddyp no problem, I can definitely understand your concerns. Having taught used dash and plotly in teaching in a number of different environments, I am keenly aware at the issues of confusing, mismatched documentation. I was debating on whether to do a PR or make a new repo easydash or litedash or something like that, so I guess I'll go with the later option. I was mainly just impressed in shiny at how it does the wiring automagically itself without the  ugly (sorry) annotations.

PR #679, opened 2019-04-04
:hocho: Dash.react

alexcjohnson:
It's been long enough... I suspect at this point this causes more confusion than it fixes, like if someone finds it in `dir(app)`

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -825,7 +825,8 @@ class Dash(object):
                 ).replace('    ', '')
             raise exceptions.DuplicateCallbackOutput(msg)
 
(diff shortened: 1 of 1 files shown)


dash/dash.py
```diff
@@ -832,7 +825,8 @@ def duplicate_check():
                 ).replace('    ', '')
             raise exceptions.DuplicateCallbackOutput(msg)
 
-    def _validate_callback_output(self, output_value, output):
+    @staticmethod
+    def _validate_callback_output(output_value, output):
```
alexcjohnson: I guess `#pylint: disable=no-self-use` was masking this for the whole rest of the class...

PR #680, opened 2019-04-05
Partial updates

### LINKED ISSUE #668: need a way to control which output i don’t want to update
the multiple outputs is great but i face some issue.
if i have a lots of outputs, it need a way to control which output i don’t want to update .
the code maybe like as follows:
```

@app.callback([
    Output('graph', 'figure'),
    Output('data-table', 'data'),
    Output('data-table', 'columns'),
    Output('container', 'style')
], [Input('data-dropdown', 'value')])
def multi_output(value):
    if condition_1:
        ...
        return [preventupdate, b, c, d]
    elif condition_2:
       ...
       return [a, preventupdate, c, d]
    else:
       ...
       return [a, b, c, preventupdate]
```

alexcjohnson:
Fixes #668 

Provides a new `dash.no_update` - a singleton that you can use in a callback return to prevent a single output from updating. Two reasons you might want to do this:
1. Most importantly, for updating only *some* of a multi-output callback, leaving the rest unchanged
2. Even for a single-output callback, it might not be ideal to raise an exception. Maybe there's cleanup you want to do after determining the return value, or further logic that might reverse the decision to not update.

I'm not super excited about the name, any suggestions? `PreventUpdate` makes sense for an exception as it's based on a verb, this seems like it should be a noun but related to `PreventUpdate`, so I went with `no_update`.

Also its location: it's not an exception so doesn't seem to belong as `dash.exceptions.no_update`. For now I just put it at the top level as `dash.no_update`. OK?

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -74,6 +74,15 @@ _re_index_scripts_id = re.compile(r'src=".*dash[-_]renderer.*"')
 _re_renderer_scripts_id = re.compile(r'id="_dash-renderer')
 
 
+class _NoUpdate(object):
+    # pylint: disable=too-few-public-methods
+    pass
+
+
+# Singleton signal to not update an output, alternative to PreventUpdate
+no_update = _NoUpdate()
+
+
 # pylint: disable=too-many-instance-attributes
 # pylint: disable=too-many-arguments, too-many-locals
 class Dash(object):
@@ -990,15 +999,25 @@ class Dash(object):
                         )
 
                     component_ids = collections.defaultdict(dict)
+                    has_update = False
                     for i, o in enumerate(output):
-                        component_ids[o.component_id][o.component_property] =\
-                            output_value[i]
+                        val = output_value[i]
+                        if val is not no_update:
+                            has_update = True
+                            o_id, o_prop = o.component_id, o.component_property
+                            component_ids[o_id][o_prop] = val
+
+                    if not has_update:
+                        raise exceptions.PreventUpdate
 
                     response = {
                         'response': component_ids,
                         'multi': True
                     }
                 else:
+                    if output_value is no_update:
+                        raise exceptions.PreventUpdate
+
                     response = {
                         'response': {
                             'props': {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Support for "Clientside Callbacks" - an escape hatch to execute your callbacks in JavaScript instead of Python [#672](https://github.com/plotly/dash/pull/672)
 - Added `dev_tools_ui` config flag in `app.run_server` (serialized in `<script id="_dash-config" type="application/json">`)
   to display or hide the forthcoming Dev Tools UI in Dash's front-end (dash-renderer). [#676](https://github.com/plotly/dash/pull/676)
+- Partial updates: leave some multi-output updates unchanged while updating others [#680](https://github.com/plotly/dash/pull/680)
 
 ## [0.40.0] - 2019-03-25
 ### Changed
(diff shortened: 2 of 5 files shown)


dash/dash.py
```diff
@@ -990,15 +999,25 @@ def add_context(*args, **kwargs):
...
-                        component_ids[o.component_id][o.component_property] =\
-                            output_value[i]
+                        val = output_value[i]
+                        if val is not no_update:
+                            has_update = True
+                            o_id, o_prop = o.component_id, o.component_property
+                            component_ids[o_id][o_prop] = val
+
+                    if not has_update:
+                        raise exceptions.PreventUpdate
```
T4rk1n: Could use `flask.abort` instead of going the slow exception way.

alexcjohnson: I don't know what `flask.abort` does, but I wanted to trigger the `PreventDefault` handler, in case the renderer does any optimizations with that specific response to maximize performance.

https://github.com/plotly/dash/blob/bb5ed706f4da8ff55a56dad49c90f2f07fce3651/dash/dash.py#L180-L183

T4rk1n: It's an empty 204 which would be the same as `flask.abort(204)`. Raising the exception takes longer because it has to be caught and then the handler is called whereas abort is immediate.

alexcjohnson: Actually seems like `flask.abort` is doing more than that... if I try to do `flask.abort(204)` it fails with `LookupError: no exception for 204`. The [docs](http://flask.pocoo.org/docs/0.12/api/#flask.abort) say:
> If a status code is given it’s looked up in the list of exceptions and will raise that exception

Which anyway sounds at least as complicated as what we're doing with `raise PreventUpdate`.

T4rk1n: Right, 204 is not an error. But then you are at the dispatch level, think you can just do `return '', 204` and avoid any error handling.

alexcjohnson: We're in the weeds now - the return from this function feeds into `response.set_data()` in the actual `dispatch`, which I don't believe accepts a status code.
https://github.com/plotly/dash/blob/b2a97b41d376baf5122e59eb7d64641608cdffa6/dash/dash.py#L1094

There may be a way to improve on this but I'd like to defer it for this PR, if that's alright.

T4rk1n: Ah, that's new. Anyway, this works well for partial updates.

notatallshaw:
With this partial update logic if a callback outputs to 2 props "foo" and "bar" where "foo" receives an update but "bar" doesn't then I assume if "bar" is the input to another callback then that callback doesn't get triggered?

alexcjohnson:
@notatallshaw I believe that's how it works - I should probably add another test to 🔒 this down though. (we do have tests confirming that this is what happens with `PreventUpdate`)

dash/dash.py
```diff
@@ -990,15 +999,25 @@ def add_context(*args, **kwargs):
...
                     component_ids = collections.defaultdict(dict)
+                    has_update = False
                     for i, o in enumerate(output):
-                        component_ids[o.component_id][o.component_property] =\
-                            output_value[i]
+                        val = output_value[i]
+                        if val is not no_update:
+                            has_update = True
+                            o_id, o_prop = o.component_id, o.component_property
+                            component_ids[o_id][o_prop] = val
```
chriddyp: Glad that we're returning a dict instead of a list here 😸 I was sure that we would've had to change the renderer to do this.

chriddyp:
- `return [3, dash.no_update, 10]`
- `return [3, dash.without_update, 10]`
- `return [3, dash.existing_value, 10]`

At first glance, I sort of like `existing_value` the best. 

But I also like the symmetry of `PreventUpdate` / `no_update` is great & `PreventUpdate` might be more common than `no_update`.

And now that I think about it, with `existing_value`, folks might try to actually _access_ the existing value (instead of using `State`): "I need the current value of a property. I tried to access `dash.existing_value` with no avail."

So, 👍 `no_update`. 

💃 💃 💃

PR #687, opened 2019-04-13
Flag for props check

byronz:
- [x] add test case once confirmed with design


dash/_configs.py
```diff
@@ -4,47 +4,50 @@
...
-    )})
-
-
-def get_config(config_name, init, env, default=None, is_bool=False):
-    if init is not None:
-        return init
-
-    env_value = env.get('DASH_{}'.format(config_name.upper()))
-    if env_value is None:
+def get_combined_config(name, val, default=None):
```
byronz: simplified the get_config by removing unnecessary parameters passing of env, and removed is_bool

byronz: simplified the get_config by removing unnecessary parameters passing of env, and removed is_bool

dash/dash.py
```diff
@@ -340,6 +336,7 @@ def _config(self):
             'url_base_pathname': self.url_base_pathname,
             'requests_pathname_prefix': self.config.requests_pathname_prefix,
             'dev_tools_ui': self._dev_tools.dev_tools_ui,
+            'dev_tools_props_check': self._dev_tools.dev_tools_props_check,
```
byronz: with this, the props-check will only do when `bool(dev_tools_ui && dev_tools_props_check) == True`

byronz:
`app.run_server(debug=True, dev_tools_ui=False, dev_tools_props_check=True)`
image
`app.run_server(port=8070, debug=True, dev_tools_props_check=False)`  
image

byronz:
@chriddyp  I saw your branch about the dev_tools_ui, adding one commit here together with props_check. note that I changed the `get_config` as I found the API design made the code hard to read and awkward. the current change is still not optimal but one step cleaner.

dash/dash.py
```diff
@@ -1302,46 +1308,40 @@ def enable_dev_tools(self,
...
-
-        self._dev_tools.dev_tools_ui = debug
+        debug = debug or get_combined_config('debug', None, debug)
 
-        self._dev_tools['serve_dev_bundles'] = _configs.get_config(
-            'serve_dev_bundles', dev_tools_serve_dev_bundles, env,
-            default=debug,
-            is_bool=True
+        self._dev_tools['dev_tools_ui'] = get_combined_config(
+            'dev_tools_ui', dev_tools_ui, default=True
```
chriddyp: I think these should be `default=debug` like the rest of the properties. That way, it won't accidently appear on production.

For users working in development, we popularize `debug=True` by including it in all of our docs:
```python
if __name__ == '__main__':
     app.run_server(debug=True)
```

So, in effect, most users will get all of these features turned on by default since `debug=True`. However, since `run_server` isn't used in production, it'll default to `False`.

dash/dash.py
```diff
@@ -1440,6 +1442,10 @@ def run_server(self,
         :type port: int
         :param debug: Set the debug mode of flask and enable the dev tools.
         :type debug: bool
+        :param dev_tools_ui: Switch the dev tools UI in debugger mode
+        :type dev_tools_ui: bool
+        :param dev_tools_props_check: Switch the props check of dash components
```
chriddyp: Maybe:
```
:param dev_tools_props_check: Validate the properties of the Dash components
```
?

tests/test_configs.py
```diff
@@ -102,6 +106,37 @@ def test_pathname_prefix_assets(self):
...
+
+    def test_load_dash_env_vars_refects_to_os_environ(self):
+        for var in DASH_ENV_VARS.keys():
+            os.environ[var] = 'true'
+            vars = load_dash_env_vars()
+            self.assertEqual(vars[var], 'true')
+            os.environ[var] = 'false'
+            vars = load_dash_env_vars()
+            self.assertEqual(vars[var], 'false')
+
```
chriddyp: good idea to include this run-time check 👍

PR #692, opened 2019-04-17
support `exact` just like `shape`

chriddyp:
in python component generation


chriddyp:
As discovered in https://github.com/plotly/dash-core-components/pull/523

dash/development/_py_components_generation.py
```diff
@@ -445,6 +445,22 @@ def create_prop_docstring(prop_name, type_object, required, description,
 
 def map_js_to_py_types_prop_types(type_object):
     """Mapping from the PropTypes js type object to the Python type"""
+
+    shape_or_exact = lambda: 'dict containing keys {}.\n{}'.format(
```
byronz: :cow: pylint doesn't like lambda as var assignment

chriddyp: PYLIIINNNNNTTTTT😩😩😩

PR #714, opened 2019-05-11
Modify 'transpiler' to permit blank strings as rprefix, address #712

### LINKED ISSUE #712: Modify 'transpiler' behaviour when empty string passed as R component prefix
Currently, R components may be generated for Dash using the syntax

`dash-generate-components ./src/components dash_core_components --r-prefix="dcc"`

When `--r-prefix` is blank, the 'transpiler' assumes `dash` is the prefix. This is undesirable for third party components, or libraries such as `dash-canvas`, which already include "dash" in their names.

Alternatively, a better strategy for the 'transpiler' behaviour might be to accept a blank string, e.g.

`dash-generate-components ./src/lib/components dash_canvas --r-prefix=""`

Then, then the 'transpiler' could generate a properly camelCase function name, i.e. `dashCanvas` instead of `DashCanvas`. 

This would maintain parity with existing Plotly component libraries, and further standardize function naming. These details seem minor, but as R is case-sensitive, they remain important.

@alexcjohnson

rpkyle:
The modifications proposed in this PR will enable passing a blank string as the `--rprefix` argument in `dash-generate-components`, e.g.

`dash-generate-components ./src/lib/components dash_canvas --r-prefix=""`

will generate the function `dashCanvas.R` and the help file `dashCanvas.Rd`.

In effect, the changes will permit component generation without a prefix, but with a properly camelCased function name. R online help and exports within the `NAMESPACE` file are adjusted accordingly.

Closes #712.


dash/development/_r_components_generation.py
```diff
@@ -368,7 +371,11 @@ def write_class_file(name,
         project_shortname,
         prefix
     )
-    file_name = "{}{}.R".format(prefix, name)
+
+    if prefix:
+        file_name = "{}{}.R".format(prefix, name)
+    else:
+        file_name = "{}.R".format(name[0].lower() + name[1:])
```
alexcjohnson: Is it intentional that this is not camelCased but the function name is? Also would be nice to 🌴 use the same code for `.R` and `.Rd`

dash/development/_r_components_generation.py
```diff
@@ -204,18 +204,16 @@ def generate_class_string(name, props, project_shortname, prefix):
...
-        default_argtext=default_argtext,
-        wildcards=wildcards,
-        wildcard_declaration=wildcard_declaration,
-        default_paramtext=default_paramtext,
-        project_shortname=project_shortname,
-        prop_names=prop_names,
-        wildcard_names=wildcard_names,
-        package_name=package_name,
-    )
+    return r_component_string.format(funcname=format_fn_name(prefix, name),
```
byronz: this indent alignment is not pythonic, is this by `black`?

rpkyle: @byronz Looks plenty Pythonic to me, that's four spaces 🙂 

The indentation is not from `black`; it dates back to the initial commit of this method from six months ago.

dash/development/_r_components_generation.py
rpkyle: fixed in f3b1888c7783be39d9ff4e307a92437177208a4b

dash/development/_r_components_generation.py
alexcjohnson: Are you talking about the lines below? That's one of the standard patterns, right? `name` and everything below it aligns with `funcname`.

Sounds like we should just use black and stop discussing this stuff. But let's wait until after 1.0 to do this.

dash/development/_r_components_generation.py
alexcjohnson: To summarize a conversation @rpkyle and I had on slack, there's still something to do here:
- filenames will match function names exactly, plus the extension.
- we will convert snake_case to camelCase no matter what (even if, in principle, component names should already be CamelCase because they're js class names).
- we will NOT lowercase `prefix` - if someone wants an uppercase or mixed case prefix, they can have it.

rpkyle: > * filenames will match function names exactly, plus the extension.

fixed in a031129

> * we will convert snake_case to camelCase no matter what (even if, in principle, component names should already be CamelCase because they're js class names).

fixed in f639cd2

> * we will NOT lowercase prefix - if someone wants an uppercase or mixed case prefix, they can have it.

fixed in 7ff6730

Issue #469, opened 2018-11-26
Dash 1.0.0 Series - Proposed Set of *Breaking Changes*


chriddyp:
##Roadmap Items
### React 16 support
- [x] [dash 1.0.0 alpha release / React 16](https://github.com/plotly/dash-core/issues/3)
- [x] [dash-renderer 1.0.0 alpha release / React 16](https://github.com/plotly/dash-core/issues/4)
- [x] [dash-html-components 1.0.0 alpha release / React 16](https://github.com/plotly/dash-core/issues/5)
- [x] [dash-core-components 1.0.0 alpha release / React 16](https://github.com/plotly/dash-core/issues/6)
- [x] [dash-table mixed React 15 and 16 support](https://github.com/plotly/dash-core/issues/7)
- [x] [dash-docs working under React 16](https://github.com/plotly/dash-core/issues/8)
- [x] [dash-dangerously-set-inner-html mixed React 15 and 16 support](https://github.com/plotly/dash-core/issues/21)
### New features and changes
- [x] [serve files locally by default](https://github.com/plotly/dash-core/issues/10)
- [x] [hide undo/redo buttons by default](https://github.com/plotly/dash-core/issues/11)
- [x] [dcc react-date update](https://github.com/plotly/dash-core/issues/13)
- [x] [dcc react-markdown update](https://github.com/plotly/dash-core/issues/14)
- [x] [dcc react-syntax-highlighter update](https://github.com/plotly/dash-core/issues/15) REMOVED `SyntaxHighlighter` component entirely
- [x] [Version lock Dash repos](https://github.com/plotly/dash-core/issues/24)
### Sign-off as Release Candidate
- [x] [validation against real world applications](https://github.com/plotly/dash-core/issues/22)

### Deferred
- [CDN resource integrity](https://github.com/plotly/dash-core/issues/18) - can be added without breaking, anyway `serve_locally=True` by default now so this is somewhat less important.
- [dcc rc-slider update](https://github.com/plotly/dash-core/issues/12) - this too can be done without breaking

> This is the "Proposed Breaking Changes" issue as per the "Breaking Changes" process outlined in https://github.com/plotly/dash/issues/458.
> The version 1.0.0 of `dash`, `dash-renderer`, `dash-core-components`, and `dash-html-components` will be compatible with each other and will contain at least one breaking change across the libraries.
> We will publish the breaking changes no earlier than 2 weeks after we have publicly committed to the proposal in this issue.

## Proposed list of breaking changes

[...]

PR #722, opened 2019-05-15
serve_locally by default

alexcjohnson:
Flipping one of the remaining switches for https://github.com/plotly/dash/issues/469

Closes https://github.com/plotly/dash/issues/284
Closes https://github.com/plotly/dash-core/issues/10

### DIFF
--- a/dash/resources.py
+++ b/dash/resources.py
@@ -66,10 +66,17 @@ class Resources:
         return self._filter_resources(all_resources, dev_bundles)
 
 
+# pylint: disable=too-few-public-methods
+class _Config:
+    def __init__(self, infer_from_layout, serve_locally):
+        self.infer_from_layout = infer_from_layout
+        self.serve_locally = serve_locally
+
+
 class Css:
     def __init__(self, layout=None):
         self._resources = Resources('_css_dist', layout)
-        self._resources.config = self.config
+        self._resources.config = self.config = _Config(True, True)
 
     def _update_layout(self, layout):
         self._resources.layout = layout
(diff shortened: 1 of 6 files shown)


dash/resources.py
```diff
@@ -99,8 +101,3 @@ def append_script(self, script):
 
     def get_all_scripts(self, dev_bundles=False):
         return self._resources.get_all_resources(dev_bundles)
-
-    # pylint: disable=no-init, too-few-public-methods
-    class config:
-        infer_from_layout = True
-        serve_locally = False
```
alexcjohnson: The way we did this before, despite the fact that `Scripts` was instantiated, `Scripts.config` was a singleton. Mostly this is OK, dunno if there will ever be a reason to put two `Scripts` instances in one Python session, but it caused our test cases to be coupled to each other. I had to change this to an instance when I took out all the otherwise superfluous overrides in our test cases.

tests/test_resources.py
```diff
@@ -48,6 +48,7 @@ def test_external(self):
         )
         app.layout = dcc.Markdown()
         app.scripts.config.serve_locally = False
+        app.scripts.config.xxx = 123
```
Marc-Andre-Rivet: Hun?

alexcjohnson: Senator, I do not recall.
https://github.com/plotly/dash/pull/722/commits/c096aa1c445dada79ef4d81c146328253a742390

nicolaskruchten: I saw it at the time and had assumed it was to test for acceptance of randomly-named config keys :)

Marc-Andre-Rivet:
diffs look like noise.. 🙄

PR #725, opened 2019-05-21
Implement basic support for bundling prewritten R functions in transpiled libraries

### LINKED ISSUE #711: Modify 'transpiler' such that R components may import non-transpiled functions
Currently, the Python-based 'transpiler' invoked using `dash-generate-components` does not support generating R code which is either not based on entries in `metadata.json`, or expressly declared within the 'transpiler'.

A basic, modular system to seamlessly include both internal ("hidden") and exported (i.e. available in the package `NAMESPACE`) R functions would be beneficial to component library developers. 

Use cases have already been discussed, both for `dash-bio` and `dash-table`.

We should implement changes to the 'transpiler' that will enable
- [x] inspection of the R package directory for a pre-existing `utils.R` script, whose functions will not be exported via `NAMESPACE`
- [x] inspection of the R package directory for _any other_ R scripts, whose R functions will be collected and exposed via `export` statements in `NAMESPACE`
- [x] supplying a list of additional R package dependencies to `dash-generate-components`, which will be inserted into the package `DESCRIPTION` file

@alexcjohnson @chriddyp @nicolaskruchten @Marc-Andre-Rivet @shammamah

rpkyle:
This PR proposes initial support for including user-written R functions in 'transpiled' DashR component libraries. 

The parsing performed is very basic in nature, but should detect any functions declared on one line, which follow the form

`somefunction <- function(...` **or** `somefunction = function(...`

Changes to `dash-generate-components` also allow bundling in additional R package dependencies, in case these are required for the user-written functions to work properly.

- [x] inspection of the R package directory for a pre-existing `utils.R` script, whose functions will not be exported via `NAMESPACE`
- [x] inspection of the R package directory for _any other_ R scripts, whose R functions will be collected and exposed via `export` statements in `NAMESPACE`
- [x] supplying a list of additional R package dependencies to `dash-generate-components`, which will be inserted into the package `DESCRIPTION` file

Closes #711.


rpkyle:
@alexcjohnson Aside from the CircleCI error (which I believe is not caused by these commits), I **hope** this one is close to being ready. The inspection for user-contributed functions could be a bit more sophisticated, but I resisted the urge to implement any kind of syntax tree-based parsing 😆

I've been using these new 'transpiler' features to repeatedly rebuild the R version of `dash-bio` while testing and it's been a breeze so far.

dash/development/component_generator.py
```diff
@@ -75,87 +85,118 @@ def generate_components(components_source, project_shortname,
...
 # pylint: disable=undefined-variable
 def byteify(input_object):
     if isinstance(input_object, dict):
-        return OrderedDict([
-            (byteify(key), byteify(value))
-            for key, value in input_object.iteritems()
-        ])
+        return OrderedDict(
+            [(byteify(key), byteify(value)) for key,
+             value in input_object.iteritems()]
```
alexcjohnson: huh, was it black's choice to break between `key, value`? Kind of a subtle case, but I'd say it's objectively wrong - ie more confusing - that way. Maybe it would make a better choice if parentheses were put around `(key, value)`?

Totally non-blocking, I'm just trying to wrap my head around what black does as it seems the push to use it more places is inevitable...

rpkyle: Yeah, to be entirely honest, so far I'm not yet a fan of `black`. To be fair, `pylint` and `flake8` took a little getting used to as well, so I'm giving it time.

For one thing, it seems that the reformatting occasionally requires another round of linting after running `black`, which is frustrating if everything seemed OK beforehand. 

I don't like the "new" formatting ☝️(in this case) at all. I thought it was just me!

byronz: had some experience using black with numpy heavy code, some of the formatting result is ugly, we just block the unhappy chunk with # fmt: on/off. but the most attractive point for me is no more argument :)

alexcjohnson: Right, no more style arguments is worth a LOT 💰 - probably even worth a little bit of reduced readability

rpkyle: Yep, I completely agree.

A major role of guidelines is to force consensus where it was not easily achieved otherwise. It doesn't mean the consensus is optimal or even objectively correct, but often the common ground offered by a third party is more attractive (particularly in the face of argument).

byronz: FYI I will also apply `black` for pytest fixture porting

PR #726, opened 2019-05-21
Provide details on component type in R online help for transpiled libraries

### LINKED ISSUE #647: DashR transpiler: type and default information in help strings
Let's make sure that the R equivalent of docstrings has the same type/default information as the Python docstrings for components.

rpkyle:
This PR proposes to add support for including typing information in R's online help for arguments in 'transpiled' DashR components. The following methods are added, based on their Python counterparts:
- `get_r_prop_types`
- `get_r_type`
- `print_r_type`
- `create_prop_docstring_r`

The changes suggested here provide better parity with the Dash for Python documentation.

Closes #647. Partly addresses https://github.com/plotly/dash-docs/issues/469.


rpkyle:
When we spoke earlier this week, we discussed the list vs. dictionary typing issue in R, and agreed that replacing `dict` with `named list` and `list` with `unnamed list` made the most sense. 

That's now been addressed in f91d69d.

rpkyle:
> I'll fix this and comment once done.

fixed in 0950605

rpkyle:
@alexcjohnson I've sorted out the previous issue, and have removed the extraneous code. I also added wrapping for the usage lines in R help (some used to extend extremely far to the right, on one line), as requested by a user.

dash/development/_r_components_generation.py
```diff
@@ -625,3 +641,74 @@ def generate_exports(
...
+        # React's PropTypes.oneOfType
+        union=lambda: "{}".format(
+            " | ".join(
+                "{}".format(get_r_type(subType))
+                for subType in type_object["value"]
+                if get_r_type(subType) != ""
+            )
+        ),
+        # React's PropTypes.arrayOf
+        arrayOf=lambda: "list".format(  # pylint: disable=too-many-format-args
```
alexcjohnson: I see that's in the python version too but it doesn't look workable. Perhaps it was meant to be
`"list" + (" of {}s".format(...) if... else "")`

dash/development/_r_components_generation.py
```diff
@@ -625,3 +641,74 @@ def generate_exports(
...
+
+    return dict(
+        array=lambda: "unnamed list",
+        bool=lambda: "logical",
+        number=lambda: "numeric",
+        string=lambda: "character",
+        object=lambda: "named list",
+        any=lambda: 'logical | numeric | character | named list | unnamed list',  # noqa:E501
+        element=lambda: "dash component",
+        node=lambda: "a list of or a singular dash " "component, string or number",  # noqa:E501
```
alexcjohnson: If this gets collapsed to 1 line (which I guess it can because black uses longer lines than pep8), smoosh it back into one string - ie just `...dash component...`, not `...dash " "component...`

alexcjohnson: and again, if we're going to use black we should just use black... tell flake8 to accept a line length of 88 and get rid of these `# noqa` directives.

rpkyle: Good point. I don't have a preference here. I can correct the reformatted lines until we discuss using `black` for Dash as well.

rpkyle: fixed in 2eb3a070adfd5b913e39d9981d951669b4c3814c

rpkyle:
> "dict" is a legit R type?

fixed in 984c004

dash/development/_r_components_generation.py
rpkyle: fixed in 083ff1b

rpkyle:
fixed in 73bc2be

rpkyle:
> 👍 would you mind fixing the Python one too while you're at it?

fixed in c9bc2fa

dash/development/_py_components_generation.py
```diff
@@ -488,11 +488,12 @@ def shape_or_exact():
                 if js_to_py_type(subType) != '')),
 
         # React's PropTypes.arrayOf
-        arrayOf=lambda: 'list'.format(  # pylint: disable=too-many-format-args
-            ' of {}s'.format(
-                js_to_py_type(type_object['value']))
-            if js_to_py_type(type_object['value']) != ''
-            else ''),
+        arrayOf=lambda: (
```
rpkyle: @alexcjohnson Hmm. Now we're seeing 5 unit test failures:

``
tests/unit/dash/development/test_base_component.py:944: in assert_docstring
    ])[i]
E   AssertionError: '- optionalArrayOf (list of numbers; optional)' != '- optionalArrayOf (list; optional)'
E   - - optionalArrayOf (list of numbers; optional)
E   ?                        -----------
E   + - optionalArrayOf (list; optional)
``

tests/unit/dash/development/test_base_component.py
```diff
@@ -540,15 +540,19 @@ def assert_no_trailing_spaces(self, s):
         for line in s.split('\n'):
             self.assertEqual(line, line.rstrip())
 
+    def match_lines(self, val, expected):
+        for val1, exp1 in zip(val.splitlines(), expected.splitlines()):
+            assert val1 == exp1
```
alexcjohnson: line-by-line matching, just to help me figure out where exactly the differences were

dash/development/_py_components_generation.py
alexcjohnson: Yep, you made a clean improvement and we had locked down the wrong behavior with our tests. Fixed in https://github.com/plotly/dash/pull/726/commits/4deb9000f3c056ba1022772fb5479b3d4f647048

dash/development/_r_components_generation.py
```diff
@@ -625,3 +641,104 @@ def generate_exports(
...
+    """Mapping from the PropTypes js type object to the R type"""
+
+    def shape_or_exact():
+        return 'lists containing elements {}.\n{}'.format(
+            ', '.join(
+                "'{}'".format(t) for t in list(type_object['value'].keys())
+            ),
+            'Those elements have the following types:\n{}'.format(
+                '\n'.join(
+                    create_prop_docstring(
```
alexcjohnson: oh hmm - `create_prop_docstring` means nested items are going to get python type info rather than R, right? It might be possible to pass an optional type conversion function into `create_prop_docstring`, if `get_r_type ` behaves compatibly with `js_to_py_type` (which it looks like it does...)

https://github.com/plotly/dash/blob/4deb9000f3c056ba1022772fb5479b3d4f647048/dash/development/_py_components_generation.py#L421

dash/development/_r_components_generation.py
```diff
@@ -625,3 +641,104 @@ def generate_exports(
...
+            ),
+
+        # React's PropTypes.shape
+        shape=shape_or_exact,
+        # React's PropTypes.exact
+        exact=shape_or_exact
+    )
+
+
+def get_r_type(type_object, is_flow_type=False, indent_num=0):
```
alexcjohnson: Please make an issue to come back to flow types. Don't do it in this PR, since we don't have any flow-typed packages of our own so we don't have an easy manual test. Come back to it after we have a test framework in place.

rpkyle: Issue https://github.com/plotly/dash/issues/763 will be closed when this is resolved.

dash/development/_r_components_generation.py
rpkyle: fixed in a68ace3342c7fdb2fa36e217bb6162670bb130f1

alexcjohnson: That'll work. I suspect we could do it 🌴 using the same function just passing in another arg to replace `js_to_py_type` - but we can come back to that after we have tests.

PR #744, opened 2019-05-23
Pytest fixtures

### LINKED ISSUE #728: merge pytest fixtures
the scope of the issue is to merge and restructure the pytest fixtures in https://github.com/plotly/pytest-dash into dash. It might include few integration tests rewrite  with the new fixtures and python state `assert`, but all tests rewrite is out of the scope of this issue. 

Note: all fixtures will be available for pytest as plugin thanks to the `pytest11` entry point by `pip install dash`.

byronz:

## Contributor Checklist

this fix #728 

   -  [x] fixture for app runner (flask server)  (thread and process)
  -   [x] fixture for selenium webdriver 
  -   [x] combo fixture binding thread + selenium for typical integration case
   -  [x] migrate a set of existing test cases using the fixtures

### optionals

    -  [ ] this github [#PR 520](https://github.com/plotly/dash-docs/pull/520) updates the dash docs
    -  [ ] update the pip install description about the pytest fixtures
    -  [ ] update the testing part in contributing guide


byronz:
@alexcjohnson  this is still WIP, but would like to get some early feedbacks.

.pylintrc
```diff
@@ -59,7 +59,9 @@ disable=fixme,
         invalid-name,
         too-many-lines,
         old-style-class,
-        superfluous-parens
+        superfluous-parens,
+        bad-continuation,
+        unexpected-keyword-arg
```
alexcjohnson: why do we need these extra disabled rules?

dash/exceptions.py
```diff
@@ -94,3 +94,27 @@ class SameInputOutputException(CallbackException):
 
 class MissingCallbackContextException(CallbackException):
     pass
+
+
+class DashTestingError(Exception):
+    """Base error for pytest-dash."""
```
alexcjohnson: perhaps these should go in `dash.testing.exceptions`?

dash/testing/browser.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def wait_for_text_to_equal(self, selector, text, timeout=None):
+        return self._wait_for(
+            text_to_equal,
+            (selector, text),
+            timeout,
+            "cannot wait until element contains expected text {}".format(text),
+        )
+
+    def wait_until_server_is_ready(self, timeout=10):
```
alexcjohnson: `wait_until_page_is_ready`? or even just `wait_for_page`?

.pylintrc
byronz: bad_continuation => is to workaround with `black`
unexpected-keyword-arg => is to handle python2 and 3 differences
e.g.  `self.proc.communicate(timeout=self.stop_timeout)` is python3 only and will have this in python2.
but we can also disable at code level.

dash/exceptions.py
byronz: I thought about that too, but since `.testing` is also part of the `dash`, I don't see a strong benefit of separating the exceptions. but I'm open to move if this makes the logic less confusing.

dash/testing/browser.py
```diff
@@ -0,0 +1,197 @@
...
+    def get_logs(self):
+        """get_logs works only with chrome webdriver"""
+        return (
+            [
+                entry
+                for entry in self.driver.get_log("browser")
+                if entry["timestamp"] > self._last_ts
+            ]
+            if self.driver.name == "chrome"
+            else []
```
alexcjohnson: empty could mask problems unexpectedly. Would it be better to throw an error if we're not in chrome?

tests/unit/test_app_runners.py
```diff
@@ -0,0 +1,34 @@
...
+    time.sleep(0.2)
+    r = requests.get(thread_server.url)
+    assert r.status_code == 200, "the threaded server is reachable"
+    assert 'id="react-entry-point"' in r.text, "the entrypoint is present"
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3,), reason="requires python3 for process testing"
+)
+def test_process_server_smoke(process_server):
```
byronz: I spent several hours to investigate why python2 has a problem here. even adding `__init__.py` in the target assets folder, the PYTHONPATH seems not to be effective, so under pytest, it's difficult to have the target app.py in the path and can be used by the subprocess command line as a valid python module.  I will open the issue as it is for now as most of the scenario will be with thread flavor.

dash/testing/plugin.py
```diff
@@ -0,0 +1,50 @@
...
+
+@pytest.fixture
+def process_server():
+    """Start a Dash server with subprocess.Popen and waitress-serve"""
+    with ProcessRunner() as starter:
+        yield starter
+
+
+@pytest.fixture
+def br(request):
```
alexcjohnson: that might be a little too terse, and could be confused with `<br>` - perhaps `browser`?

dash/testing/wait.py
```diff
@@ -0,0 +1,39 @@
...
+def until(
+    wait_cond,
+    timeout,
+    poll=0.1,
+    msg="expected condition not met within timeout",
+):  # noqa: C0330
+    end_time = time.time() + timeout
+    while wait_cond():
+        time.sleep(poll)
+        if time.time() > end_time:
```
alexcjohnson: swap the `time.sleep` and `if time > end_time` perhaps? So success right at the end still counts.

tests/integration/test_devtools.py
```diff
@@ -34,6 +34,12 @@
 TIMEOUT = 20
 
 
+def test_wdr001_simple_br_dash_docs(br):
```
byronz: demo case to be removed

dash/testing/wait.py
```diff
@@ -0,0 +1,39 @@
...
+    wait_cond, timeout, poll=0.1, msg="expected condition met within timeout"
+):  # noqa: C0330
+    end_time = time.time() + timeout
+    while not wait_cond():
+        time.sleep(poll)
+        if time.time() > end_time:
+            raise TestingTimeoutError(msg)
+
+
+class text_to_equal(object):  # pylint: disable=too-few-public-methods
```
alexcjohnson: nbd, but could be a higher-order function and the linter wouldn't complain :)
```py
def text_to_equal(selector, text)
    def _call(driver):
        ...

    return _call
```

byronz: this is actually a mimic from python EC conditions implementation. even though I feel that part is too-java, but just follow the pattern to make it more consistent.

dash/testing/plugin.py
byronz: I used to call that `wd` for webdriver, but found now `br` is nicer. I get your concern about confusion. 

in our test case context, it's less serious as the html component will be `html.Br`, and the main reason I would fight for being terse is we will use fixtures everywhere, and I'm kinda finger saver

dash/testing/browser.py
byronz: yep, I don't like the current api name, `wait_for_page` is cool for me

dash/testing/browser.py
byronz: the main concern with the throw-an-error approach is that user will be able to choose the browser on the fly, `pytest --webdriver Firefox tests\integration`, and that will raise a lot of errors.

tests/unit/test_app_runners.py
```diff
@@ -0,0 +1,34 @@
...
+    app = dash.Dash(__name__)
+
+    app.layout = html.Div(
+        [
+            html.Button("click me", id="clicker"),
+            html.Div(id="output", children="hello thread"),
+        ]
+    )
+    thread_server(app, debug=True, use_reloader=False, use_debugger=True)
+    time.sleep(0.2)
```
alexcjohnson: Testing the tests 🔁
Normally users won't need to use explicit sleeps like this, right? They'll just do `browser.wait_for_page` or something?

.pylintrc
alexcjohnson: OK - `bad-continuation` sounds fine then, but yes let's keep `unexpected-keyword-arg` local where it's needed, that one could hide some confusing real issues if used globally.

tests/unit/test_app_runners.py
byronz: yes, this is like a smoke test for server fixtures only, and most of our cases will have that helper API from browser

dash/exceptions.py
alexcjohnson: I don't feel too strongly about it, but I do think it's nice to keep testing-only code out of modules that are loaded in production.

tests/unit/test_app_runners.py
byronz: once we stabilize the api, we should further enhance this case with a more comprehensive app layout, so it can cover most of the common usage and alert the bugs at an earlier stage

dash/exceptions.py
byronz: OK, I will 🚚  them into testing.errors

dash/testing/browser.py
alexcjohnson: True. But we'll also presumably have occasional tests that *require* something to appear in the logs in order to succeed. Maybe returning `[]` is the best we can do for now, but perhaps we should at least print a warning?

dash/testing/plugin.py
alexcjohnson: right, due to pytest magic this becomes the var name people will need to use throughout their tests - eg `br.wait_for_element_by_css_selector(...)` - though TBH it seems a bit funny to worry about the length of the word `browser` when we have methods like `wait_for_element_by_css_selector` 😏

I could be overruled if other folks are on board with `br`, but personally I'd still prefer `browser`.

dash/testing/wait.py
byronz: with a poll freq at `0.1`,  I would argue what a poor performance that edge case could be :),  I will move it down anyway.

dash/testing/plugin.py
byronz: I kept the `wait_for_element_by_css_selector` for back compatibilities, we should use css selector by default, so that can actually be just `wait_for_element` and you remind me to add the default one.

dash/testing/browser.py
```diff
@@ -157,15 +161,17 @@ def _get_firefox():
...
+            return (
+                [
+                    entry
+                    for entry in self.driver.get_log("browser")
+                    if entry["timestamp"] > self._last_ts
+                ]
+            )
+        else:
+            warnings.warn("get_logs always return None with your webdriver")
+            return None
```
alexcjohnson: Ah, that's a nice compromise - so you can do eg:
`assert not self.get_logs()`
and you'll get the warning but not a test failure. But if you ask for something more explicit like
`assert self.get_logs() == []`
the test will actually fail (along with the warning that will show you *why* it failed)

Small point: perhaps explicitly say "with ~your webdriver~ webdrivers other than Chrome"

tests/integration/devtools/test_devtools_error_handling.py
```diff
@@ -0,0 +1,202 @@
+# -*- coding: UTF-8 -*-
+import dash_html_components as html
+import dash_core_components as dcc
+import dash
+from dash.dependencies import Input, Output
+from dash.exceptions import PreventUpdate
+
+
+def test_dveh001_python_errors(dash_duo):
```
byronz: @alexcjohnson  the tcid is added, it's more precise to communicate the test case with a unique id. and in case of test selection, mostly of time you can play with `pytest -k tcid`. 

the current convention is `mmffddd`  abbr for module name and file and three digits,  it also helps to navigate to the test case in IDE if you have the tcid

tests/integration/renderer/test_due_diligence.py
```diff
@@ -0,0 +1,114 @@
...
+
+def test_rddd001_initial_state(dash_duo):
+    app = dash.Dash(__name__)
+    my_class_attrs = {
+        "id": "p.c.4",
+        "className": "my-class",
+        "title": "tooltip",
+        "style": {"color": "red", "fontSize": 30},
+    }
+    # fmt:off
```
byronz: in general, I'm happy with the unarguable `black` default style, and here is an exception I explicitly turned it off

tests/integration/renderer/test_dependencies.py
```diff
@@ -0,0 +1,47 @@
...
+    dash_duo.percy_snapshot(name="dependencies")
+
+    dash_duo.find_element("#input").send_keys("a")
+    assert dash_duo.find_element("#output-1").text == "initial valuea"
+
+    assert output_1_call_count.value == 2 and output_2_call_count.value == 0
+
+    rqs = dash_duo.redux_state_rqs
+    assert len(rqs) == 1
+    assert rqs[0]["controllerId"] == "output-1.children" and not rqs[0]['rejected']
```
byronz: I don't like the abstraction we did before `assertion_request_queue`,  it encapsulates  the checkpoints elsewhere, and someone also introduces a bug by inserting another parameter, but forget to check all the API calls, so we end up having several API calls to check the length of the queue, but it actually passed to the first added parameter which is a boolean. 

perhaps the only acceptable exception to hide the assertion with an API is the check below about clear browser logs.

.circleci/config.yml
```diff
@@ -74,13 +74,13 @@ jobs:
...
-              git clone --depth 1 https://github.com/plotly/dash-core-components.git
-              git clone --depth 1 https://github.com/plotly/dash-html-components.git
+              git clone https://github.com/plotly/dash-core-components.git
+              git clone https://github.com/plotly/dash-html-components.git
               git clone --depth 1 https://github.com/plotly/dash-table.git
               git clone --depth 1 https://github.com/plotly/dash-renderer-test-components
               . venv/bin/activate
-              cd dash-core-components && npm install --ignore-scripts && npm run build && pip install -e . && cd ..
-              cd dash-html-components && npm install --ignore-scripts && npm run build && pip install -e . && cd ..
+              cd dash-core-components && git checkout 2932409 && npm install --ignore-scripts && npm run build && pip install -e . && cd ..
```
byronz: after these two commits, dev bundle was removed in dcc and html, which was the root cause of the 4 devtools props-check related failure. 

we need to address that with the formal release and this is a workaround for this PR to pass all tests

dash/testing/application_runners.py
```diff
@@ -0,0 +1,216 @@
...
+        return self
+
+    def __exit__(self, exc_type, exc_val, traceback):
+        if self.started and not self.keep_open:
+            try:
+                logger.info("killing the app runner")
+                self.stop()
+            except TestingTimeoutError:
+                raise ServerCloseError(
+                    "Cannot stop server within {} timeout".format(
```
alexcjohnson: `... {} second timeout"`

dash/testing/browser.py
```diff
@@ -0,0 +1,281 @@
...
+    def _is_windows():
+        return sys.platform == "win32"
+
+    def multiple_click(self, css_selector, clicks):
+        for _ in range(clicks):
+            self.find_element(css_selector).click()
+
+    def js_click(self, elem):
+        """click in native javascript way
+        note: this is NOT the recommended way to click"""
```
alexcjohnson: I don't see this used anywhere, is it? Is there a known reason you'd want to use this instead of `find_element().click()`?

dash/testing/wait.py
```diff
@@ -0,0 +1,91 @@
...
+logger = logging.getLogger(__name__)
+
+
+def until(
+    wait_cond,
+    timeout,
+    poll=0.1,
+    msg="expected condition not met within timeout",
+):  # noqa: C0330
+    res = None
```
alexcjohnson: minor, but do we want to init with a `wait_cond()` so if it's already satisfied we don't even need to wait one `poll` period?

dash/testing/composite.py
```diff
@@ -0,0 +1,17 @@
+from dash.testing.browser import Browser
+
+
+class DashComposite(Browser):
+
+    def __init__(self, server, browser, remote=None, wait_timeout=10):
+        super(DashComposite, self).__init__(browser, remote, wait_timeout)
+        self.server = server
+
+    def start_app_server(self, app, **kwargs):
```
alexcjohnson: shorten to `start_server`?

byronz: Yay! Will change that once back

byronz: Yay! Will change that once back

byronz: Yay! Will change that once back

dash/testing/browser.py
byronz: Yes this is a tricky one. Might not be useful at all for dash. I will remove it.

PR #748, opened 2019-05-24
Refactor <flavor> package generation

Marc-Andre-Rivet:
Very much a work in progress. Not attempting to rewrite all the generation, just to provide some level of abstraction between the generation and the file structure through parametrization + isolating package building from the nitty-gritty of the generation of the various parts. So a good portion of what is happening here is moving `with open(...) as f:` higher up the stack.

- support building R and Python artefacts to arbitrary folder (dist/<flavor>) folders  (`--dist` flag)
- untie R and Python build implementations
- use `root` folder otherwise if `--dist` is not present

- [ ] meta file for js artefacts (will untie R from Python)
- [ ] meta file for name/description/etc. -- can default to package.json -- untie ourselves from the pacakge.json entirely at some point

### DIFF
--- a/dash/development/_py_components_generation.py
+++ b/dash/development/_py_components_generation.py
@@ -110,7 +110,7 @@ def generate_class_string(typename, props, description, namespace):
     )
 
 
-def generate_class_file(typename, props, description, namespace):
+def generate_component_wrapper(typename, props, description, namespace):
     """
     Generate a python class file (.py) given a class string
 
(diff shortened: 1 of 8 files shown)


joseangel-sc:
In order for the test to pass you need to refactor the test file corresponding to all the renaming you did. 
`generate_class_file` is used in the file `tests/unit/dash/development/test_base_component.py`, check all of this and make sure your test pass before making the PR

PR #753, opened 2019-05-30
Props blacklist

alexcjohnson:
Fixes https://github.com/plotly/dash-core-components/issues/440

Pares down the attributes of dash component classes on the Python side by doing two things:
- Remove the `MutableMapping` inheritance - we weren't using any of the mixins from that class anyway, except in tests. This stops setting attributes: `clear`, `get`, `items`, `keys`, `pop`, `popitem`, `setdefault`, `update`, `values` - all of these are now allowable as prop names. We still set the magic methods of the mapping interface.
- Prefix a couple of other methods (`traverse` and `traverse_with_paths`) with underscores, so these are allowable as well.

Then what's left we blacklist: during `extract-meta`, if any prop has a leading underscore or matches one of the remaining disallowed names, we fail (not just print a warning) and refuse to generate the python file. The names we forbid are:
- `UNDEFINED`
- `REQUIRED`
- `to_plotly_json`
- `available_properties`
- `available_wildcard_properties`

These are all used within generated components, or (in the case of `to_plotly_json`) in other packages, so it would be difficult to allow these without forcing all packages to be rebuilt. Fortunately I don't see these as particularly valuable prop names (unlike, say, `values`, `items`, and some of the other `MutableMapping` methods)

Error messages look like:
```
ERROR: "_foo" matches forbidden prop name pattern: /^_.*$/

extract-meta failed.


Error generating metadata in dash_core_components (status=0)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
...
```

### DIFF
--- a/dash/development/component_generator.py
+++ b/dash/development/component_generator.py
@@ -19,12 +19,13 @@ from ._py_components_generation import generate_imports
 from ._py_components_generation import generate_classes_files
 
 
-forbidden_props = [
+reserved_words = [
     'UNDEFINED',
     'REQUIRED',
     'to_plotly_json',
     'available_properties',
-    'available_wildcard_properties'
+    'available_wildcard_properties',
+    '_.*'
 ]
 
 
@@ -56,14 +57,12 @@ def generate_components(
 
     extract_path = pkg_resources.resource_filename("dash", "extract-meta.js")
 
-    forbid_patterns = '|'.join(
-        '^{}$'.format(p) for p in forbidden_props + ['_.*']
-    )
+    reserved_patterns = '|'.join('^{}$'.format(p) for p in reserved_words)
 
     os.environ["NODE_PATH"] = "node_modules"
     cmd = shlex.split(
         "node {} {} {} {}".format(
-            extract_path, ignore, forbid_patterns, components_source
+            extract_path, ignore, reserved_patterns, components_source
         ),
         posix=not is_windows,
     )
--- a/dash/development/component_generator.py
+++ b/dash/development/component_generator.py
@@ -19,6 +19,15 @@ from ._py_components_generation import generate_imports
 from ._py_components_generation import generate_classes_files
 
 
+forbidden_props = [
+    'UNDEFINED',
+    'REQUIRED',
+    'to_plotly_json',
+    'available_properties',
+    'available_wildcard_properties'
+]
+
+
 class _CombinedFormatter(
         argparse.ArgumentDefaultsHelpFormatter,
         argparse.RawDescriptionHelpFormatter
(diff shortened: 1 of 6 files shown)


tests/unit/dash/development/test_base_component.py
```diff
@@ -244,17 +244,6 @@ def test_get_item_raises_key_if_id_doesnt_exist(self):
...
 
-    def test_equality(self):
-        # TODO - Why is this the case? How is == being performed?
-        # __eq__ only needs __getitem__, __iter__, and __len__
-        self.assertTrue(Component() == Component())
-        self.assertTrue(Component() is not Component())
-
-        c1 = Component(id='1')
-        c2 = Component(id='2', children=[Component()])
-        self.assertTrue(c1 == c2)
```
alexcjohnson: Kind of a nice side effect, IMO, that components are now no longer always equal to each other - though I don't know why this would come up...

tests/unit/dash/development/test_base_component.py
```diff
@@ -487,8 +489,7 @@ def test_iter(self):
 
 class TestGenerateClassFile(unittest.TestCase):
     def setUp(self):
-        json_path = os.path.join(
-            'tests', 'unit', 'dash', 'development', 'metadata_test.json')
+        json_path = os.path.join(_dir, 'metadata_test.json')
```
alexcjohnson: So that you can run the tests from any working directory, not just the repo root.

dash/extract-meta.js
```diff
@@ -20,13 +23,18 @@ const metadata = Object.create(null);
 componentPaths.forEach(componentPath =>
     collectMetadataRecursively(componentPath)
 );
-writeOut(metadata);
+if (failed) {
+    console.error('\nextract-meta failed.\n');
+}
+else {
+    writeOut(metadata);
+}
```
alexcjohnson: we don't write a response if the error was severe enough for us to set `failed` - this causes `generate_components` to fail as well.

tests/unit/dash/development/test_base_component.py
```diff
@@ -469,40 +464,37 @@ def test_iter(self):
...
+            self.assertFalse(hasattr(c, m), m)
+
+        keys = ['2', '3', '4', '5', '6', '7', '8']
 
-        # test __iter__()
         for k in keys:
             # test __contains__()
             self.assertTrue(k in c)
+            # test __getitem__()
+            self.assertEqual(c[k].id, k)
```
byronz: since we already switched to pytest now, I think it's better to use native `assert`  like 

assert c[k].id == k, "brief description on the criteria"

dash/development/component_generator.py
```diff
@@ -19,6 +19,15 @@
 from ._py_components_generation import generate_classes_files
 
 
+forbidden_props = [
```
byronz: reserved_props?

dash/development/component_generator.py
```diff
@@ -47,9 +56,15 @@ def generate_components(
 
     extract_path = pkg_resources.resource_filename("dash", "extract-meta.js")
 
+    forbid_patterns = '|'.join(
+        '^{}$'.format(p) for p in forbidden_props + ['_.*']
```
byronz: can we just add the last `_.*` in the blacklist above too?

dash/development/component_generator.py
alexcjohnson: But they're not props, they're words/patterns that you're not allowed to use as props. I suppose I could call them `reserved_words` if you don't like `forbidden`.

byronz: `reserved_words` is better, just feel it's more accurate with `reserved`. and forbidden might be harsh

PR #761, opened 2019-06-07
Config cleanup

### LINKED ISSUE #746: Clean up kwargs and config
Right now we have a lot of args to `dash.Dash`. Some of them may be unnecessary, some have confusing names. Then there are other settings that can only be done after constructing the app, by digging into sub-objects and changing the defaults that were set during construction. Ideally it seems to me we should make ALL configuration possible through the constructor, but perhaps using a smaller set of kwargs that accept objects. These kwargs and objects should map cleanly onto attributes that you can modify later if necessary (and possible).

Currently here's what we have, AFAICT. (Env vars are included as well, where available.)

**Constructor args:**
- misc:
  - `name` (`DASH_APP_NAME`)
  - `server`
  - `compress`
  - `suppress_callback_exceptions` (`DASH_SUPPRESS_CALLBACK_EXCEPTIONS` - can we remove the misspelled `supress_callback_exceptions` fallback? The fix dates to 2017)
  - `components_cache_max_age` (`DASH_COMPONENTS_CACHE_MAX_AGE` can we remove, as we have cache-busting urls?)
  - `plugins`
- url modifiers:
  - `assets_url_path`
  - `assets_external_path` (`DASH_ASSETS_EXTERNAL_PATH`)
  - `url_base_pathname` (`DASH_URL_BASE_PATHNAME`)
  - `routes_pathname_prefix` (`DASH_ROUTES_PATHNAME_PREFIX`)
  - `requests_pathname_prefix` (`DASH_REQUESTS_PATHNAME_PREFIX`)
- asset behavior:
  - `assets_folder`
  - `static_folder` (can we remove?)
  - `assets_ignore`
  - `include_assets_files` (`DASH_INCLUDE_ASSETS_FILES`)
- page content:
  - `meta_tags`
  - `index_string`
  - `external_scripts`
  - `external_stylesheets`
  - `show_undo_redo`
- unused: `**kwargs` - only used to look for unused `csrf_protect` kwarg that became obsolete in 2017 - `csrf_protect` will be removed but I'll use the same mechanism to help with kwargs I'm removing here.

**`run_server` args:** The idea here is that these may have different values when running locally than when deployed - seems like this division has been done well but a few potential tweaks mentioned below.

- dev tools (set through `enable_dev_tools`, or `run_server` with `dev_tools_` prefix)
  - `debug` (`DASH_DEBUG`, no prefix from `run_server`)
  - `ui` (`DASH_UI` - I know it fits the pattern but as an env var this name is confusing)
  - `props_check` (`DASH_PROPS_CHECK`)
  - `serve_dev_bundles` (`DASH_SERVE_DEV_BUNDLES`)
  - `hot_reload` (`DASH_HOT_RELOAD`)
  - `hot_reload_interval` (`DASH_HOT_RELOAD_INTERVAL`)
  - `hot_reload_watch_interval` (`DASH_HOT_RELOAD_WATCH_INTERVAL` - this is in seconds but `hot_reload_interval` is in msec, can we change one so they match?)
  - `hot_reload_max_retry` (`DASH_HOT_RELOAD_MAX_RETRY`)
  - `silence_routes_logging` (`DASH_SILENCE_ROUTES_LOGGING`)

- `run_server` options not for dev tools:
  - `port`
  - `**flask_run_options`

**Mutating `app` attributes after instantiation:**

- CSS / JS
  - `app.scripts.config.serve_locally` (add constructor arg to set `serve_locally` default for both?)
  - `app.css.config.serve_locally`
  - `app.scripts.config.infer_from_layout` (unused? remove?)
  - `app.css.config.infer_from_layout` (unused? remove?)

- `app.config`: all are available from the constructor. Suggestion: use only the constructor in docs, take out the item that cannot be modified post-init. If we want to keep `config` as a public part of the interface, standardize what is and isn't included - this seems like a pretty random subset.
  - `suppress_callback_exceptions` - config used in docs (in fact, constructor arg is never used)
  - `routes_pathname_prefix` - CANNOT BE MODIFIED, used in `__init__`
  - `requests_pathname_prefix`
  - `include_assets_files`
  - `assets_external_path`
  - `components_cache_max_age`

alexcjohnson:
closes #746 
## Some breaking changes to the `dash.Dash` API
Constructor / config breaking changes:
- Removed two obsolete constructor kwargs: `static_folder` (`assets_folder` is much smarter) and `components_cache_max_age` (all files have cache-busting URLs)
- Removed the very old misspelled `supress_callback_exceptions` fallback
- Removed the unused `resources.config.infer_from_layout`
- Revamped `app.config`: ALL constructor args are now stored in `config` (and none are stored in any other instance attributes anymore), with three exceptions:
  - `server` - `app.init_app(server)` is the right way to set this
  - `index_string` - it has a special setter method
  - `plugins` - You could add more plugins by calling `my_plugin.plug(app)` but not by modifying a config entry

### Dev tools breaking changes:
- Changed `hot_reload_interval` from msec to seconds, for consistency with `hot_reload_watch_interval`
- When called from `enable_dev_tools`, `debug=True` by default. It's still `False` by default from `run_server`.

### Non-breaking changes:
- The expanded `app.config` prevents you from adding new (ie misspelled or obsolete) items and makes some items read-only - the ones that will not take effect if changed after the constructor (and `init_app`, since that can happen in the constructor) have finished.
- The default `assets_url_path` is now just `'assets'`, not `'/assets'`. Since in actual usage we always strip the leading `'/'` and prepend the requests prefix, it's misleading as well as harder to explain with the slash there.
- New constructor kwarg `serve_locally` - default `True` as per #722 - sets both JS and CSS config
- Added a docstring for the `dash.Dash` constructor and improved docstrings for the `run_server` and `enable_dev_tools` methods

### Not implemented:

I looked into the `name` parameter and seems to me we should NOT make it mandatory, but would be interested to hear what others think about this. The only time this seems to matter is if the file that was run (`python run.py` in our common usage, or the `gunicorn` equivalent) is in a different directory from the one that defines `app` (let's call it `app.py`). All it seems to affect is where Flask looks for other files specified by relative paths, the key one being `assets_folder`.
- Our default, `'__main__'`, says these are relative to `run.py`.
- Our recommended usage, `name=__name__`, says they're relative to `app.py`
- You can specify something else with a package path (dotted, like `package.sub.path` as in `import package.sub.path`, and it will be resolved exactly as that import would be)
- AFAICT it does NOT matter what the working dir was, relative to `run.py`, when you ran Python

Most of the time this doesn't matter, as `run.py` and `app.py`, if they're even separate files, are usually in the same dir. `__name__` is nice in that you get the same assets folder no matter where you import `app.py` from, but on the other hand I suspect the more likely situation is that you set up your project and then at some point decide to move `app.py` into a subdirectory, without moving `assets` or `run.py`. In that case `__name__` would cause problems, and the current default `'__main__'` is probably correct. That to me is a sufficient reason to leave the default behavior as it is.

## Contributor Checklist


dash/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Unreleased
 ### Changed
+- [#761](https://github.com/plotly/dash/pull/761) Several breaking changes to the `dash.Dash` API:
```
byronz: a :boom: here?

dash/dash.py
```diff
@@ -86,106 +86,223 @@ class _NoUpdate(object):
 # pylint: disable=too-many-instance-attributes
 # pylint: disable=too-many-arguments, too-many-locals
 class Dash(object):
+    """
```
byronz: 😸  nice docstring.

 tangential question, do we have an agreed reference to follow?  I feel like it's not very helpful to use super-long-detailed-sphinx-style; chris is also building sth in the house.  

but from our users' perspective, data scientists might feel this more familiar https://sphinxcontrib-napoleon.readthedocs.io/en/latest/

dash/dash.py
```diff
@@ -86,106 +86,223 @@ class _NoUpdate(object):
...
+    Dash is a framework for building analytical web applications.
+    No JavaScript required.
+
+    If a parameter can be set by an environment variable, that is listed too.
```
byronz: that is listed with ``env:`` prefix?

dash/dash.py
```diff
@@ -86,106 +86,223 @@ class _NoUpdate(object):
...
+
+    :param server: Sets the Flask server for your app. There are three options:
+        ``True`` (default): Dash will create a new server
+        ``False``: The server will be added later via ``app.init_app(server)``
+            where ``server`` is a ``flask.Flask`` instance.
+        ``flask.Flask``: use this pre-existing Flask server.
+    :type server: boolean or flask.Flask
+
+    :param assets_folder: a path, relative to the current working directory,
+        for extra files to be used in the browser. Default ``'assets'`` By
```
byronz: the Default ``'asset'`` By default needs a better split.

dash/dash.py
```diff
@@ -86,106 +86,223 @@ class _NoUpdate(object):
...
+        such as images will be served if requested.
+    :type assets_folder: string
+
+    :param assets_url_path: The local urls for assets will be:
+        ``requests_pathname_prefix + assets_url_path + '/' + asset_path``
+        where ``asset_path`` is the path to a file inside ``assets_folder``.
+        Default ``'assets'``.
+    :type asset_url_path: string
+
+    :param assets_ignore: A regexp, as a string to pass to ``re.compile``, for
```
byronz: `regexp` is js flavor, pythonic name should be `regex`

dash/dash.py
```diff
@@ -86,106 +86,223 @@ class _NoUpdate(object):
...
+    :param show_undo_redo: Default ``False``, set to ``True`` to enable undo
+        and redo buttons for stepping through the history of the app state.
+    :type show_undo_redo: boolean
+
+    :param plugins: Extend Dash functionality by passing a list of objects
+        with a ``plug`` method, taking a single argument: this app, which will
+        be called after the Flask server is attached.
+    :type plugins: list of objects
+
+    :param **kwargs: Do not use. Gives error messages for obsolete arguments.
```
byronz: would it be more intuitive if we add an *example section* e.g. `man lsof` for less experienced users?

tests/integration/test_render.py
```diff
@@ -1164,7 +1164,7 @@ def test_removing_component_while_its_getting_updated(self):
             ),
             html.Div(id='body')
         ])
-        app.config.supress_callback_exceptions = True
+        app.config.suppress_callback_exceptions = True
```
byronz: 👍

byronz: predicted merge conflicts here

dash/dash.py
alexcjohnson: You mean rather than the errors on providing old kwargs? I was thinking of those just as helpers for experienced users of old dash versions. This documentation of `**kwargs` is just intended for people who see `**kwargs` in the function signature and wonder why it's there. Perhaps I can omit this documentation and just rename it to `**obsolete`, and assume that's sufficiently self-documenting?

dash/dash.py
alexcjohnson: I'd be up for changing to a cleaner style, I was just copying the style we already had for `run_server` and `enable_dev_tools`.

Feels to me like we should have a decent level of detail here, and then use this to generate the corresponding page in dash-docs. Let's keep this format until we decide how we want to do that, so we only need to alter it once. @chriddyp I'm not aware if you're working on something else to fill this gap.

dash/dash.py
byronz: Sorry for the confusion, I meant for all the docstring. Just happened to leave the comment at this line. Nothing wrong with kwargs.

byronz: Sorry for the confusion, I meant for all the docstring. Just happened to leave the comment at this line. Nothing wrong with kwargs.

alexcjohnson: Got it. I feel like once people get here they've seen examples already elsewhere in the docs, so this docstring is just about enumerating all the options.

Anyway I will change `**kwargs` to `**obsolete` and remove it from the docstring.

tests/integration/test_integration.py
```diff
@@ -870,6 +871,40 @@ def failure2(children):
             context.exception.args[0]
         )
 
+    def test_callback_dep_types(self):
```
byronz: @alexcjohnson  do you want to migrate this with new style?

dash/testing/browser.py
```diff
@@ -110,29 +110,32 @@ def wait_for_element(self, css_selector, timeout=None):
 
     # keep these two wait_for API for easy migration
     def wait_for_element_by_css_selector(self, selector, timeout=None):
+        _time = timeout or self._wait_timeout
         return self._wait_for(
             EC.presence_of_element_located,
             ((By.CSS_SELECTOR, selector),),
             timeout,
-            "timeout {} => waiting for selector {}".format(timeout, selector),
+            "timeout {}s => waiting for selector {}".format(_time, selector),
```
byronz: nice fix on the timeout
🐄  minor, I would use more pythonic way format `(timeout if timeout else self._wait_timeout, selector)`

alexcjohnson: 🐄 in https://github.com/plotly/dash/pull/761/commits/55274bea6ad084443a785ca6c1bef926e9de0060

tests/integration/test_integration.py
alexcjohnson: not too hard to migrate -> https://github.com/plotly/dash/pull/761/commits/40525079c6f6af873ca68faec8ee2743de3c1c2e

tests/integration/test_integration.py
```diff
@@ -17,325 +18,331 @@
...
+        callback2_count.value += 1
+        return value
+
+    dash_duo.start_server(app)
+
+    input_ = dash_duo.find_element('#input')
+    input_.send_keys('xyz')
+    dash_duo.wait_for_text_to_equal('#input', 'initial inputxyz')
+
+    # callback1 runs 4x (initial page load and 3x through send_keys)
```
byronz: I would put the comment as an assert msg, so you can see that directly in the assertion exception.   the comment does show up too with the code snippet.

tests/integration/test_integration.py
```diff
@@ -348,604 +355,568 @@ def test_index_customization(self):
...
-                      [Input('a', 'children')])
-        def multi(a):
-            # non-serializable inside a list
-            return [1, set([2])]
+    dash_duo.start_server(app)
 
-        with self.assertRaises(InvalidCallbackReturnValue):
-            multi('aaa')
+    btn_elements = [
+        dash_duo.find_element('#' + x) for x in btns
```
byronz: minor: prefer `btn` than `x` here

tests/integration/test_integration.py
```diff
@@ -348,604 +355,568 @@ def test_index_customization(self):
...
-        self.wait_for_text_to_equal('#a_out_short', 'aa')
-        # now ab_out *is* triggered - a_out_short got a new value
-        # even though that value is the same as the last value it got
-        self.wait_for_text_to_equal('#ab_out', 'aa bbbb')
-
-    def test_with_custom_renderer(self):
-        app = Dash(__name__)
+    btn = dash_duo.find_element('#btn')
+    for _ in range(10):
+        btn.click()
```
byronz: you can use this API `dash_duo.multiple_click()`. yep the API reference is coming soon 

``` def multiple_click(self, css_selector, clicks):
        for _ in range(clicks):
            self.find_element(css_selector).click()
```

tests/integration/test_integration.py
```diff
@@ -348,604 +355,568 @@ def test_index_customization(self):
...
+def test_inin019_callback_dep_types():
+    app = Dash(__name__)
+    app.layout = html.Div([
+        html.Div('child', id='in'),
+        html.Div('state', id='state'),
+        html.Div(id='out')
+    ])
+
+    with pytest.raises(IncorrectTypeException):
+        @app.callback([[Output('out', 'children')]],  # extra nesting
```
byronz: comment can be set as message='' inside raises

tests/integration/test_integration.py
```diff
@@ -17,325 +18,331 @@
...
+
+    dash_duo.start_server(app)
+
+    div = dash_duo.find_element('#data-element')
+
+    # React wraps text and numbers with e.g. <!-- react-text: 20 -->
+    # Remove those
+    comment_regex = r'<!--[^\[](.*?)-->'  # noqa: W605
+
+    # Somehow the html attributes are unordered.
```
byronz: this permutation was bad idea, I did similar clean up with beautiful soup parser for the initial state test case

alexcjohnson: yeah I noticed that - `BeautifulSoup` is a nice idea, I'll try it here.

PR #762, opened 2019-06-08
chore: unnecessary string

cctv1005s:
```js
            return <div className="_dash-error">Error loading layout</div>;
```

is better than 

```js
            return <div className="_dash-error">{'Error loading layout'}</div>;
```

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2
- [ ] I have run the tests locally and they passed. (refer to testing section in [contributing](../CONTRIBUTING.md))

### optionals


alexcjohnson:
Thanks @cctv1005s! There are 2 others like this farther down the same function, care to simplify them as well?

cctv1005s:
@alexcjohnson sure

PR #770, opened 2019-06-12
Replace dashR with dash in R package generation code

rpkyle:
This PR proposes minor changes to the R package generation code, which will use the R package name `dash` in lieu of `dashR` going forward.

### DIFF
--- a/dash/development/_r_components_generation.py
+++ b/dash/development/_r_components_generation.py
@@ -217,7 +217,8 @@ def generate_class_string(name, props, project_shortname, prefix):
 def generate_js_metadata(pkg_data, project_shortname):
     """
     Dynamically generate R function to supply JavaScript
-    and CSS dependency information required by dash package.
+    and CSS dependency information required by the dash
+    package for R.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


dash/development/_r_components_generation.py
```diff
@@ -221,7 +221,7 @@ def generate_class_string(name, props, project_shortname, prefix):
 def generate_js_metadata(pkg_data, project_shortname):
     """
     Dynamically generate R function to supply JavaScript
-    and CSS dependency information required by dashR package.
+    and CSS dependency information required by dash package.
```
alexcjohnson: hmm given that this is *in* the `dash` *Python* package this reads a bit confusingly. Perhaps "required by the dash R package"?

(this applies only to these 3 Python docstrings, the generated code string above and the `--r-prefix` help string are fine)

rpkyle: Right, that makes sense. I was hoping that using the word _package_ rather than _module_ would help clarify, but how about "required by the dash package for R" instead?

alexcjohnson: "required by the dash package for R" works 👍 

In the Python context, `module` is a single file and `package` (as in PyPI, the Python package index) is a collection of modules within one (possibly nested) folder, that together make a coherent set.

rpkyle: fixed in b578ce6f05021f84af80608afbd7fad2350c13f1

PR #773, opened 2019-06-15
Update setup

byronz:
This PR improves the requirement*.txt in dash repo, it unifies the requirements using the python version condition, and lists all dependencies in txt file instead of listing them in setup.py; this can give docker setup more flexibility. 

## Contributor Checklist

   -  [x] improve the requirements.txt in setup.py 
   -  [x] move the `dash.testing` dependencies into main install

### optionals


setup.py
```diff
@@ -2,65 +2,63 @@
...
-        'plotly',
-        'dash_renderer==0.24.0',
-        'dash-core-components==0.48.0',
-        'dash-html-components==0.16.0',
-        'dash-table==3.7.0'
-    ],
+    long_description=io.open("README.md", encoding="utf-8").read(),
+    long_description_content_type="text/markdown",
+    install_requires=read_req_file("install"),
+    extra_require={"intg": read_req_file("intg")},
```
byronz: each key here can be used by `pip install -e .[key]`

requires-intg.txt
```diff
@@ -0,0 +1,9 @@
+dash_flow_example==0.0.5
+dash-dangerously-set-inner-html
+six
+mock
+flake8
+pylint==1.9.4;python_version<"3.7"
+pylint==2.3.1;python_version=="3.7"
+astroid==2.2.5;python_version=="3.7"
```
byronz: this can avoid having two requirements per python version

requires-ci.txt
```diff
@@ -0,0 +1,9 @@
+dash_flow_example==0.0.5
+dash-dangerously-set-inner-html
+six
+mock
+flake8
+pylint==1.9.4;python_version<"3.7"
+pylint==2.3.1;python_version=="3.7"
+astroid==2.2.5;python_version=="3.7"
```
alexcjohnson: didn't know you could do that, cool! Would it be better to say `>="3.7"`? NBD, I guess we'll know pretty quickly if we move to 3.8 and things break...

byronz:
> Not sure what's going on with py2.7, but this is a nice simplification so if you can get it working I'm all for it! 💃

looks like a cache key issue, I will add python version as well

Issue #771, opened 2019-06-14
[Feature Request] Pick up app name from server instance


Batalex:
**Is your feature request related to a problem? Please describe.**
[It is known](https://community.plot.ly/t/dash-app-does-not-load-assets-and-app-index-string/12178/10?u=chriddyp) that the Dash app needs `name=__name__` to correctly load additional assets with gunicorn/waitress. However it is a shame that it cannot pickup the flask server's name if one is passed.
This leads to this:
```
server = Flask(__name__)
app = dash.Dash(server=server, name=__name__, external_stylesheets=external_stylesheets)
```

**Describe the solution you'd like**
Use flask server name if server in args but not name. See here https://github.com/plotly/dash/blob/master/dash/dash.py#L245

**Additional context**
I would be glad to submit a PR.

alexcjohnson:
Great idea @Batalex! I don't see any problems with doing that, so yeah, we would welcome a PR 🎉

alexcjohnson:
Closed by @Batalex in #774

PR #774, opened 2019-06-15
Pick up app name from server instance (#771)

Batalex:

## Contributor Checklist

   -  [X] use flask server name for dash app if name not passed in args. Fix https://github.com/plotly/dash/issues/771
- [X] I have run the tests locally and they passed. (refer to testing section in [contributing](../CONTRIBUTING.md))

### optionals

### Details

I changed the default name arg value from `__main__` to `None` in order to distinguish "not passing a name" from "setting explicitly the name to `__main__` ". I figured it would be the best course of action to avoid unwanted renaming cases.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -205,7 +205,7 @@ class Dash(object):
     """
     def __init__(
             self,
-            name='__main__',
+            name=None,
             server=True,
             assets_folder='assets',
             assets_url_path='assets',
@@ -239,10 +239,14 @@ class Dash(object):
 
         # We have 3 cases: server is either True (we create the server), False
         # (defer server creation) or a Flask app instance (we use their server)
-        if isinstance(server, bool):
-            self.server = Flask(name) if server else None
-        elif isinstance(server, Flask):
+        if isinstance(server, Flask):
             self.server = server
+            # GH 771
+            if name is None:
+                name = getattr(server, 'name', '__main__')
+        elif isinstance(server, bool):
+            name = name if name else '__main__'
+            self.server = Flask(name) if server else None
         else:
             raise ValueError('server must be a Flask app or a boolean')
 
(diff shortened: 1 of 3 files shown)


dash/dash.py
```diff
@@ -239,10 +239,14 @@ def __init__(
 
         # We have 3 cases: server is either True (we create the server), False
         # (defer server creation) or a Flask app instance (we use their server)
-        if isinstance(server, bool):
-            self.server = Flask(name) if server else None
-        elif isinstance(server, Flask):
+        if isinstance(server, Flask):
             self.server = server
+            # GH 771
```
alexcjohnson: I don't think we need this comment - that's what `git blame` is for.

alexcjohnson:
Looks great, thanks! Would you mind adding a test, perhaps in https://github.com/plotly/dash/blob/master/tests/unit/test_configs.py, that just instantiates an `app` in various ways and checks that `app.config.name` has the desired value?

dash/dash.py
Batalex: Sure thing I will remove it, each project has its own rules 😄

alexcjohnson:
> I changed the default name arg value from `__main__` to `None` in order to distinguish "not passing a name" from "setting explicitly the name to `__main__` ". I figured it would be the best course of action to avoid unwanted renaming cases.

💯 I don't know why a user would do this, but it should definitely behave the way you've done it!

Batalex:
@alexcjohnson I see that https://github.com/plotly/dash/blob/master/tests/unit/test_configs.py still use unitest, am I free to import pytest? `pytest.mark.parametrize` would be useful for this PR -> multiple cases to tests

alexcjohnson:
Yes, you are more than welcome to use `pytest`! We'll be converting this and the others the rest of the way soon, just haven't gotten to this file yet.

Batalex:
It should be good now. To keep a narrow scope for this PR I added the bare minimum for the test but I would be glad to continue this work in a new PR with a full convertion to pytest

PR #778, opened 2019-06-18
Unit tests migration to pytest

Batalex:
As a continuation of https://github.com/plotly/dash/pull/774, here is a PR migrating the remaining unitests to pytest:
- The behavior of each test should not have been changed, I just used pytest shortcuts for some of them. 
- I kept classes where I thought it was a nice scope delimiter

I do need some inputs however on:
- `tests.unit.development.test_base_component.TestGenerateClass#test_call_signature`. It has the following comment above: "# This one is kind of pointless now". Should I transform it to pytest anyway?
- `tests.unit.development.test_base_component.TestComponent#test_to_plotly_json_with_nested_children` does not seem to test anything, the assertion is enclosed in triple quotes. What should I do with it?

## Contributor Checklist

   -  [X] Migrate units tests to pytest
- [X] I have run the tests locally and they passed. (refer to testing section in [contributing](../CONTRIBUTING.md))

(edit by ACJ - removed unneeded steps. Thanks!)


tests/unit/test_configs.py
```diff
@@ -1,156 +1,168 @@
...
+    ],
+)
+def test_app_name_server(empty_environ, name, server, expected):
     app = Dash(name=name, server=server)
     assert app.config.name == expected
 
 
-if __name__ == '__main__':
-    unittest.main()
+if __name__ == "__main__":
```
byronz: we don't need this mechanism in `pytest` anymore, we can collect and run directly with `pytest tests/unit`

tests/unit/development/test_base_component.py
```diff
@@ -55,71 +57,67 @@ def nested_tree():
     return c, c1, c2, c3, c4, c5
 
 
-class TestComponent(unittest.TestCase):
+class TestComponent:
```
byronz: This is purely personal preferences, there is no right answer here. The purpose of keeping a class as a logical container is redundant in the sense that the test file name is doing the same thing, especially under the context of `pytest`.  another bonus point is that we can avoid `self`.  another practice is to use fixtures defined in `conftest.py` to share some common patterns in the test, which used to the reason for using `class` in `unittest`.

alexcjohnson: +1 rather than classes, let's just break into more files, and organize that way.

alexcjohnson:
> * `tests.unit.development.test_base_component.TestGenerateClass#test_call_signature`. It has the following comment above: "# This one is kind of pointless now". Should I transform it to pytest anyway?

I agree, that test is a bit silly. 🔪 

> * `tests.unit.development.test_base_component.TestComponent#test_to_plotly_json_with_nested_children` does not seem to test anything, the assertion is enclosed in triple quotes. What should I do with it?

LOL it's been that way for 2 years 
Screen Shot 2019-06-18 at 9 35 19 AM
🔪

tests/unit/development/test_base_component.py
Batalex: Sure, I would prefer as well

tests/unit/development/test_base_component.py
```diff
@@ -55,71 +57,67 @@ def nested_tree():
...
-        self.assertEqual(
-            len(c),
-            5 +  # 5 components
-            5 +  # c2 has 2 strings, 2 numbers, and a None
-            1    # c1 has 1 string
-        )
+        # 5 components
+        # c2 has 2 strings, 2 numbers, and a None
+        # c1 has 1 string
+        assert len(c) == 5 + 5 + 1
```
byronz: :cow: minor: a translation pattern we start doing now is to also translate the comment into an assertion message, by using clearly SHOULD or SHOULD NOT, so here we can do sth like

`assert len(c) == 5 + 5 + 1, "the length of the nested children should match the total of 5 components, 2 strings + 2 numbers + none in c2, and 1 string in c1"

another practice is to use `black` as an automatic formatting tool, it's inspired by the awesome `golang` official formatter.  we use `fmt: off` and `fmt: on` to disagree with the default style, but unlike in `yapf` world, people argument endlessly about the options and settings.

Batalex: +1 for black. As a matter of fact, I intentionnally turned it off because I though it would not play well with flake8 in CI, but used it anyway on `test_configs`

tests/unit/development/test_base_component.py
```diff
@@ -139,128 +137,126 @@ def test_set_item_with_nested_children_with_mixed_strings_and_without_lists(self
...
         c = nested_tree()[0]
         keys = reversed([k for k in c])
         for key in keys:
             c[key]
             del c[key]
-            with self.assertRaises(KeyError):
+            with pytest.raises(KeyError):
                 c[key]
 
     def test_traverse_with_nested_children_with_mixed_strings_and_without_lists(self):  # noqa: E501
```
byronz: another support point to remove the class and self, you can get rid of the `noqa: E501` too

tests/unit/development/test_base_component.py
```diff
@@ -487,8 +473,10 @@ def test_iter(self):
         assert len(keys) == len(keys2), 'iteration produces no extra keys'
 
 
-class TestGenerateClassFile(unittest.TestCase):
-    def setUp(self):
+class TestGenerateClassFile:
```
byronz: this might be a good sign we can shift them into another test file?

byronz:
Looks good! thanks @Batalex

tests/unit/development/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+    for line in s.split("\n"):
+        assert line == line.rstrip()
+
+
+def match_lines(val, expected):
+    for val1, exp1 in zip(val.splitlines(), expected.splitlines()):
+        assert val1 == exp1
+
+
+def assert_docstring(docstring):
```
byronz: @Batalex  this is a direct conversion from the original test right?  I don't like the check line by line, it looks long and error-prone with the long list of strings. 

Can we just check the hash of the content ?  like hashlib.md5(string).hexdigest(), the hardcode part will be a simple digest string.

tests/unit/development/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+                    "- in (string; optional)",
+                    "- id (string; optional)",
+                    "        ",
+                ]
+            )[i]
+        )
+
+
+def assert_flow_docstring(docstring):
+    for i, line in enumerate(docstring.split("\n")):
```
byronz: same here, and we should move the `assert`s back into the test case whenever possible.

Batalex: TBH I found it odd as well, but I did not investigate enough to see how and when it was decided to do so. Anyway, I'll change it !

tests/unit/development/__init__.py
Batalex: @byronz If you are okay with that, I replaced those functions with `difflib`. I think the hash approch is much harder to debug and equivalently error-prone.
Hardcoding the digest string would indeed be more concise, but we would not have any way of knowing why two hashes are different and this would require an extra step every time the docstring template is changed.
Even if we keep the hard coded docstring and compute the hash at runtime, the difficulty to figure out why the hashes are different remains

I also changed some other utils functions, all `assert` are back in the tests

byronz: @Batalex  I'm sorry for the late reply, I was on vacation last week. what's the difference of using the difflib vs pytest's own diff mechanism? did you find it more readable?

Batalex: No worries !
The main advantage of using difflib instead of what we had before is that we can now detect multiple errors in one test run. The previous version would fail a test on the first different line regardless of the total number of lines changed whereas here we get a git-like diff with all changes

tests/unit/development/test_generate_class_file.py
```diff
@@ -0,0 +1,83 @@
...
+        description=make_component_dir["description"],
+        namespace="TableComponents",
+    )
+    written_file_path = os.path.join("TableComponents", "Table.py")
+    with open(written_file_path, "r") as f:
+        return f.read()
+
+
+def test_class_string(expected_class_string, component_class_string):
+    assert not len(
```
byronz: is the `len()` here necessary?

Batalex: My bad, I'll remove it asap.

PR #780, opened 2019-06-18
add apis about tab and window storage

byronz:

## Contributor Checklist

   -  [x] add helper API for store component testing
- [ ] I have run the tests locally and they passed. (refer to testing section in [contributing](../CONTRIBUTING.md))

### optionals

### DIFF
--- a/.pylintrc37
+++ b/.pylintrc37
@@ -500,7 +500,7 @@ valid-metaclass-classmethod-first-arg=mcs
 max-args=5
 
 # Maximum number of attributes for a class (see R0902).
-max-attributes=7
+max-attributes=20
 
 # Maximum number of boolean expressions in an if statement.
 max-bool-expr=5
--- a/.pylintrc37
+++ b/.pylintrc37
@@ -515,7 +515,7 @@ max-locals=15
 max-parents=7
 
 # Maximum number of public methods for a class (see R0904).
-max-public-methods=20
+max-public-methods=40
 
 # Maximum number of return / yield for function / method body.
 max-returns=6
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -25,7 +25,7 @@ from dash.testing.errors import DashAppLoadingError, BrowserError
 logger = logging.getLogger(__name__)
 
 
-class Browser(DashPageMixin):
+class Browser(DashPageMixin):  # pylint: disable=too-many-instance-attributes
     def __init__(self, browser, remote=None, wait_timeout=10):
         self._browser = browser.lower()
         self._wait_timeout = wait_timeout
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -3,6 +3,7 @@ import os
 import sys
 import logging
 import warnings
+from time import sleep as wait
 import percy
 
 from selenium import webdriver
@@ -18,7 +19,7 @@ from selenium.common.exceptions import WebDriverException, TimeoutException
 
 from dash.testing.wait import text_to_equal, style_to_equal
 from dash.testing.dash_page import DashPageMixin
-from dash.testing.errors import DashAppLoadingError
+from dash.testing.errors import DashAppLoadingError, BrowserError
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


dash/testing/browser.py
```diff
@@ -18,13 +19,13 @@
...
 from dash.testing.dash_page import DashPageMixin
-from dash.testing.errors import DashAppLoadingError
+from dash.testing.errors import DashAppLoadingError, BrowserError
 
 
 logger = logging.getLogger(__name__)
 
 
-class Browser(DashPageMixin):
+class Browser(DashPageMixin):  # pylint: disable=too-many-instance-attributes
```
alexcjohnson: A limit of 7 instance attrs seems a bit draconian. The only other place we disable this is on the `Dash` class itself, which used to have more but since the config refactor it has 16. Could just bump the limit up in pylintrc to accommodate both of these.

dash/testing/browser.py
```diff
@@ -191,6 +194,19 @@ def wait_for_page(self, url=None, timeout=10):
                 )
             )
 
+    def toggle_window(self):
+        """switch between the current working window and the new opened one"""
+        if len(self.driver.window_handles) < 2:
+            raise BrowserError("there is no new opening window")
```
alexcjohnson:
```suggestion
            raise BrowserError("there is no second window")
```

alexcjohnson: perhaps actually move this check into `switch_window`, and check for `len(...) <= idx`?

dash/testing/browser.py
```diff
@@ -191,6 +194,19 @@ def wait_for_page(self, url=None, timeout=10):
...
+
+        idx = (self._window_idx + 1) % 2
+        self.switch_window(idx=idx)
+        self._window_idx += 1
+
+    def switch_window(self, idx=0):
+        self.driver.switch_to.window(self.driver.window_handles[idx])
+        wait(0.5)
```
alexcjohnson: `switch_to.window` doesn't incorporate sufficient wait on its own? Or are you waiting for something beyond just having the new window visible?

byronz: good question, the switch_to.window is a wrapper of `execute` lower level selenium command

``` python
        if self._driver.w3c:
            self._w3c_window(window_name)
            return
        data = {'name': window_name}
        self._driver.execute(Command.SWITCH_TO_WINDOW, data)

    def _w3c_window(self, window_name):
        def send_handle(h):
            self._driver.execute(Command.SWITCH_TO_WINDOW, {'handle': h})

        try:
            # Try using it as a handle first.
            send_handle(window_name)
        except NoSuchWindowException as e:
            # Check every window to try to find the given window name.```

I think it makes sense here to add a wait margin as window-switch is not that responsive, this can avoid having this extra wait everywhere in the test case. 

the  0.5s needs more on field testing, could be 0.1 in our case as dash is really fast, I remember an old project I had I put 2s before.

File: .pylintrc37 [hunk 4]
```diff
@@ -515,7 +515,7 @@ max-locals=15
 max-parents=7
 
 # Maximum number of public methods for a class (see R0904).
-max-public-methods=20
+max-public-methods=40
```
Marc-Andre-Rivet: Guessing this is browser.py below busting the limit? This rule might be meaningless for this file or context.

dash/testing/dash_page.py
```diff
@@ -35,3 +35,15 @@ def redux_state_rqs(self):
         return self.driver.execute_script(
             "return window.store.getState().requestQueue"
         )
+
+    @property
+    def local_storage(self):
+        return self.driver.execute_script(
+            "return JSON.parse(window.localStorage.getItem('local'));"
```
Marc-Andre-Rivet: not sure about that `local` arg, what are we doing here?

dash/testing/dash_page.py
```diff
@@ -35,3 +35,15 @@ def redux_state_rqs(self):
...
+        )
+
+    @property
+    def session_storage(self):
+        return self.driver.execute_script(
+            "return JSON.parse(window.sessionStorage.getItem('session'));"
```
Marc-Andre-Rivet: same here for `session`

dash/testing/dash_page.py
byronz: this is for `dcc.store` component, convention is set `local` as key for localStorage. that's why it's in dash_page.py as domain relevant helpers

dash/testing/dash_page.py
byronz: same answer for `local`

dash/testing/dash_page.py
byronz: good catch, I should either return localStorage or open an arg for that

PR #782, opened 2019-06-19
Docstring default values + minor corrections

shammamah:
## Contributor Checklist

   -  [x] Add default values to generated docstrings, if they are defined 
   -  [x] Ensure that component names that start with a vowel are preceded by "An" and not "A" 
   -  [x] Remove duplicate periods in top-level component descriptions, and add periods if they are missing
   -  [x] Change pluralization for "list of" prop when it's a list of dict (as opposed to an "s" being appended after all of the keys are defined for a dict) 

### optionals


dash/development/_py_components_generation.py
```diff
@@ -236,10 +236,12 @@ def create_docstring(component_name, props, description):
     props = reorder_props(props=props)
 
     return (
-        """A {name} component.\n{description}
+        """A{n} {name} component.\n{description}
 
 Keyword arguments:\n{args}"""
     ).format(
+        n='n' if component_name[0].lower() in ['a', 'e', 'i', 'o', 'u']
+        else '',
```
alexcjohnson: Classy touch 🥂

dash/development/_py_components_generation.py
```diff
@@ -249,6 +251,8 @@ def create_docstring(component_name, props, description):
                 else prop['flowType'],
                 required=prop['required'],
                 description=prop['description'],
+                default=prop['defaultValue']['value']
+                if 'defaultValue' in prop.keys() else '',
```
alexcjohnson: 🐄 `'defaultValue' in prop` is sufficient, `.keys()` is unneeded

dash/development/_py_components_generation.py
```diff
@@ -422,25 +426,36 @@ def create_prop_docstring(prop_name, type_object, required, description,
         type_object=type_object,
         is_flow_type=is_flow_type,
         indent_num=indent_num + 1)
-
     indent_spacing = '  ' * indent_num
+
+    is_required = 'optional'
+    if required:
+        is_required = 'required'
+    elif default and 'null' not in str(default) and '[]' not in str(default):
```
alexcjohnson: maybe also exclude `{}`?

Scares me a little to use `not in` here - like, what if we have a nested prop with a default like `{"mode": "local", "extras": []}` - can we just do `!=`?

dash/development/_py_components_generation.py
alexcjohnson: this gets repeated a few times - 🌴 ?

shammamah: I'm a bit conflicted about moving this out into a function or something. It would need to be at the topmost scope, and only gets used three times. Is there another way to do this without repeating myself?

dash/development/_py_components_generation.py
shammamah: That makes sense! Fixed in c70659e.

dash/development/_py_components_generation.py
shammamah: (The `.keys()` have been removed in 9271f5b and f5bbab4 (I found another instance of it being used, so I went ahead and fixed that too.)

shammamah: I tried doing this, just for brevity: bbccf47

alexcjohnson: I probably would have just made a top-level function but that's OK, we can leave it inline. I don't think bbccf47 is an improvement - shorter in one case, longer in the others, and less readable.

Though, the other thing you could do is move this logic into `create_prop_docstring` - ie pass the whole object in rather than the string value.

shammamah: The prop objects wouldn't have been trivial to extract in situations like this: https://github.com/plotly/dash/blob/cfa120ef4328a7895d7611020218b896d9ce29a3/dash/development/_py_components_generation.py#L245


So I compromised with 20cf393!

PR #783, opened 2019-06-19
Pre-release packaging improvements for Dash for R

### LINKED ISSUE #777: Generated R packages should support dynamic wildcards
In Dash for Python components, such as `html.Div`, wildcards need not be restricted to `data-*` and `aria-*`, though these are now standard in components which accept wildcards:

https://github.com/plotly/dash-html-components/blob/bf465eb83a80d6c761288e0ffce5a2f32de66f6c/dash_html_components/Div.py#L44-L50

In Dash for R, the set of wildcards is hardcoded:

https://github.com/plotly/dashR/blob/85915f8e6091f11c1da8e2e27780f7782bc5e55c/R/dash.R#L857

This line should be modified to accept dynamic/developer specified wildcards:

`grepl(c('^data-[a-zA-Z0-9]{1,}$|^aria-[a-zA-Z0-9]{1,}$'), x)`

Given that logic already exists to selectively generate wildcard props, the change required should be relatively minor:

https://github.com/plotly/dash/blob/176f2861a9c26d5e90d22501d87cba4fcdb513f7/dash/development/_r_components_generation.py#L183-L184

https://github.com/plotly/dash/blob/176f2861a9c26d5e90d22501d87cba4fcdb513f7/dash/development/_r_components_generation.py#L163-L168

rpkyle:
This PR proposes several modifications to the R package generator invoked by `dash-generate-components`:
- correct a :bug: in `assert_valid_wildcards` (use `names(args)` instead of `args`)
- correct a :bug: in usage blocks for generated R documentation (duplicated component names)
- support for _dynamic_ (i.e. developer-specified) wildcards besides `data-*` and `aria-*`
- support for documenting component-specific wildcards
- replace `assert_valid_wildcards` with `dash_assert_valid_wildcards ` in generated components
- :hocho: `filter_null` in generated components, inline the logic provided by this function
- remove the leading comma from the `Imports:` field, as only the `Depends:` field has a hardcoded entry (for the R binary itself)
- add the option to specify a package `Maintainer:` field
- add support for storing R package metadata in `dash-info.yaml`
- remove a hardcoded dependency on `dash` package (CLI arguments to `dash-generate-components` may be used instead)

Closes #777, related to #788.


dash/development/_r_components_generation.py
```diff
@@ -323,6 +320,9 @@ def write_help_file(name, props, description, prefix):
 
     has_wildcards = any("-*" in key for key in prop_keys)
```
alexcjohnson: not new here, but `"-*" in key` is different from `key.endswith("-*")` that we use elsewhere, in principle could lead to bugs.

dash/development/_r_components_generation.py
```diff
@@ -159,35 +166,25 @@ def generate_class_string(name, props, project_shortname, prefix):
...
-                                   for p in prop_keys if "*" in p)
-
-    if default_wildcards == "":
-        default_wildcards = "NULL"
-    else:
-        default_wildcards = "c({})".format(default_wildcards)
-
     # Filter props to remove those we don't want to expose
     for item in prop_keys[:]:
         if item.endswith("-*") or item in r_keywords or item == "setProps":
```
alexcjohnson: also not new in this PR, but `item in r_keywords` worries me to just be silently dropped - if nothing else it seems like this should log a warning. Do any of our own components use R keywords?

dash/development/_r_components_generation.py
```diff
@@ -821,3 +855,13 @@ def create_prop_docstring_r(prop_name, type_object, required, description,
                    ': {}'.format(description) if description != '' else ''
                ),
                is_required='required' if required else 'optional')
+
+
+def get_wildcards_r(prop_keys):
+    wildcards = ""
+    wildcards += ", ".join("'{}'".format(p)
+                           for p in prop_keys if "*" in p)
```
alexcjohnson: again, for consistency `p.endswith("-*")`

dash/development/_r_components_generation.py
```diff
@@ -425,6 +446,7 @@ def write_js_metadata(pkg_data, project_shortname):
     file_path = os.path.join("R", file_name)
     with open(file_path, "w") as f:
         f.write(function_string)
+        f.write(component_helpers)
```
alexcjohnson: `if has_wildcards`?

dash/development/_r_components_generation.py
```diff
@@ -560,6 +584,15 @@ def generate_rpkg(
...
 is on GitHub: plotly/dash-core-components."
+    if package_name in ["dashTable"]:
+        pkg_help_header = "Core Interactive Table Component for Dash"
+        pkg_help_desc = "Dash DataTable is an interactive table component\n\
+designed for viewing, editing, and exploring large datasets. DataTable is\n\
+rendered with standard, semantic HTML <table/> markup, which makes it\n\
+accessible, responsive, and easy to style. This component was written\n\
+from scratch in React.js specifically for the Dash community. Its API\n\
+was designed to be ergonomic and its behavior is completely customizable\n\
+through its properties."
```
alexcjohnson: I recognize that you're just continuing the pattern from above - but these multiline strings with trailing backslashes and no indentation are a nightmare. Let's put them up top in triple quotes like all the other long string constants.

Same goes for `component_helpers` - move it up top so it's appropriate for it to have no indentation.

dash/development/_r_components_generation.py
rpkyle: Good catch. Will fix.

rpkyle: fixed in c6a013a15eb90a5cc1cf939d4f711133800bd255

dash/development/_r_components_generation.py
rpkyle: fixed in fcd82a52a52584e50f6ceef4d43f9d39df10805f

dash/development/_r_components_generation.py
rpkyle: fixed in 1578e9ac76316be72f211e2a318ee764707243b4

dash/development/_r_components_generation.py
rpkyle: fixed in 8f50b8a17d064ac7da28de182235340e2ea3be43

rpkyle: I've made my initial attempt to support `dash-info.yaml` as we discussed, I am 💯certain that there will need to be changes, but felt it was better to start somewhere.

Have not relocated everything to the `.yaml` file, figured we would discuss.

dash/development/_r_components_generation.py
alexcjohnson: Somehow this change (only adding `dash_assert_valid_wildcards` to `internal.R` if needed by a component) was reverted - was that intended?

dash/development/_r_components_generation.py
```diff
@@ -129,22 +136,34 @@
...
+\\strong{{Maintainer}}: {maintainer}
 }}
 """
 
+component_helpers = """
+dash_assert_valid_wildcards <- function (attrib = list("data", "aria"), ...)
+{
+    args <- list(...)
+    validation_results <- lapply(names(args), function(x) {
+        grepl(paste0("^", attrib, "-[a-zA-Z0-9]{1,}$", collapse = "|"),
```
alexcjohnson: We want to accept `-` and `_` too (and maybe others, but let's start there) https://stackoverflow.com/questions/25897910/valid-characters-in-custom-data-attribute-name-in-html5

Also a little bit shorter regex (though longer code to create it 😄):

```suggestion
        grepl(paste0("^(", paste0(attrib, collapse="|"), ")-[a-zA-Z0-9_-]+$"),
```

dash/development/_r_components_generation.py
```diff
@@ -129,22 +136,34 @@
...
+            x)
+    })
+    if (FALSE %in% validation_results) {
+        stop(sprintf("The following wildcards are not currently valid in Dash: '%s'",
```
alexcjohnson:
```suggestion
        stop(sprintf("The following wildcards are not valid in this component: '%s'",
```

dash/development/_r_components_generation.py
rpkyle: Yes. It is preferable not to generate this function for all generated component libraries, but I am uncertain about how to check for wildcards without passing props to methods that ordinarily would not require them. 

Any advice you have on how to do this cleanly without refactoring the current code would be helpful.

alexcjohnson: I just looped through the metadata for wildcards again -> https://github.com/plotly/dash/pull/783/commits/15cbea84e4be40125b7ef3302de32ee6ec0fa92f

dash/development/_r_components_generation.py
alexcjohnson: Added a warning if we drop something from `r_keywords`, in https://github.com/plotly/dash/pull/783/commits/e35f040c112dd36642b9d1b235781a8d3451e130

We should keep an eye out for this when rebuilding all the components - if we never see this warning, perhaps we can lock the R keywords out by turning the warning into an error.

alexcjohnson:
I can't figure out what's going on with the python 3 tests here - some sort of error deep in Flask... what changed?

I notice BTW that our requirements call for `Flask>=0.12` and yet we're getting `Flask@1.0`.

FWIW I see these same errors running the tests locally, yet in the same environment I can run a normal dash app just fine.

@byronz can you take a look?

dash/development/component_generator.py
```diff
@@ -11,6 +11,7 @@
 import functools
 
 import pkg_resources
+import yaml
```
byronz: @rpkyle  I think the performance import pattern in official yaml doc is still valid.

```
try:
    from yaml import CLoader as Loader, CDumper as Dumper
except ImportError:
    from yaml import Loader, Dumper
```

requires-install.txt
```diff
@@ -7,12 +8,12 @@ dash-html-components==1.0.0
 dash-table==4.0.0
 
 # dash.testing
-pytest
+pytest<5.0.0
```
byronz: temporary ⬇️  pytest version until the fix in flask is released https://github.com/pallets/flask/pull/3278

@rpkyle  the unit tests all passed in circleci now

dash/development/_r_components_generation.py
```diff
@@ -331,62 +355,49 @@ def write_help_file(name, props, description, prefix):
...
+            props[p]["description"]
+        )
         for p in prop_keys
     )
 
-    if has_wildcards:
-        item_text += '\n\n\\item{...}{wildcards: `data-*` or `aria-*`}'
+    if any(key.endswith("-*") for key in prop_keys):
         default_argtext += ', ...'
+        item_text += wildcard_help_template.format(get_wildcards_r(prop_keys))
```
byronz: 🐄  minor: always try to use list for str operation `+=` might be expensive if it's in a big loop

dash/development/_r_components_generation.py
```diff
@@ -529,38 +550,16 @@ def generate_rpkg(
     # generate the internal (not exported to the user) functions which
     # supply the JavaScript dependencies to the dash package.
     # this avoids having to generate an RData file from within Python.
-    write_js_metadata(pkg_data=pkg_data, project_shortname=project_shortname)
+    write_js_metadata(pkg_data, project_shortname, has_wildcards)
 
-    with open("NAMESPACE", "w") as f:
+    with open("NAMESPACE", "w+") as f:
```
byronz: 🐮  minor it's personal preference, but I would avoid `f` with numbers, maybe for example fp_ns, fp_desr

Issue #787, opened 2019-06-20
[Feature Request] dynamically added callbacks


endremborza:
Hi!

this is somewhat of a mixture of a feature request and a bug report. here is my example:

```python
import dash
import dash_core_components as dcc
import dash_html_components as html
from dash.dependencies import Input, Output, State
import json
import os

external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']
app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
app.config.suppress_callback_exceptions=True

try:
    os.remove('cities.json')
except OSError:
    pass


def make_cb(city):
    @app.callback(Output('%s-pushed' % city, 'children'),
                  [Input('%s-but' % city, 'n_clicks')],
                  [State('%s-but' % city, 'children')])
    def fing(nc,cval):
        if nc:
            return cval
        return ''


def get_layout():
    
    try:
        cities = json.load(open('cities.json'))
    except FileNotFoundError:
        cities = []
    
    return html.Div([
                dcc.Input(id='input-1', type='text', value='Montréal'),
                html.Button(id='submit-button', n_clicks=0, children='Submit'),
                html.Div(id='pushed-city'),
                html.Div([html.Div([html.Button(c,id='%s-but' % c),
                                    html.Div(id='%s-pushed' % c)])
                          for c in cities],id='output-state')
            ])


app.layout = get_layout


@app.callback(Output('output-state', 'children'),
              [Input('submit-button', 'n_clicks')],
              [State('input-1', 'value')])
def update_output(n_clicks, input1):
    try:
        cities = set(json.load(open('cities.json')))
    except FileNotFoundError:
        cities = set()
    
    if input1 not in cities and input1 is not None:
        cities.add(input1)
        cities = list(cities)
        json.dump(cities,open('cities.json','w'))
        make_cb(input1)
    print(app.callback_map)
    return [html.Div([html.Button(c,id='%s-but' % c),
                                    html.Div(id='%s-pushed' % c)])
                          for c in cities]


if __name__ == '__main__':
    app.run_server(debug=True)
```

this is a simple app based on the documentation, that allows the user to add interactive elements, with some caching. There are two issues. The first one I don't understand, the second one I do.

1. If I click the submit button, to add a city, everything appears fine, the ```print(app.callback_map)``` line shows me that the callback has been registered, i can see everything in the browser, however the callback does not become active, until I refresh the browser.

[...]

PR #789, opened 2019-06-22
Callback map sharing

endremborza:
Hello! this is the idea referenced in #787 I thought I might as well show you what I'm on about. This gives me the chance to solve issue 2 there.

There is a lot here, but in the end, there is no significant change to currently running and tested code. On the other hand the new functionality can only be tested locally yet. ( circleci/python:#.#-stretch-node-browsers does not have a redis server set up, so i expect the ci tests to break. sorry for that, if this direction is approved, that obviously needs to be fixed)

I am merely suggesting, that this is a possible way towards the the feature of being able to dynamically add interactive elements in dash, which I find to be quite powerful. I'm working on a few projects involving dash, and all of them would greatly benefit from something like this, so if you have suggestions or some ideas how this can evolve into something that can be merged, I'm willing to put in the work.

## Contributor Checklist

   -  [x] ```validate_callback_output``` function needed to be moved from ```@staticmethod``` to top level function to allow dill pickling. it is not used anywhere else
   -  [x] callback_map assignment needed to be changed, this way it's elements can be immutable
   -  [x] a ```RedisDict``` class added and unit tested to optionally replace the dict type of callback_map
   -  [x] tests for the redis callback_map and multiple output callbacks (this was also a necessity, as multiprocessing.Value does not work with dill pickled functions)
- [x] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md)) - since the linked file does not work as dev-requirements.txt does not exist, I needed to improvise
  - [x] unit test for the redis dict
  - [x] simple callback tests for redis callback map

### DIFF
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -116,3 +116,30 @@ def create_callback_id(output):
     return '{}.{}'.format(
         output.component_id, output.component_property
     )
+
+
+class RedisDict:
+
+    def __init__(self, redis_kwargs):
+        import redis
+        import dill
+        self.serializer = dill
+        self.r = redis.Redis(**redis_kwargs)
+
+    def __iter__(self):
+
+        for k in self.r.keys():
+            yield k.decode('utf-8')
+
+    def __getitem__(self, k):
+
+        return self.serializer.loads(self.r.get(k))
+
+    def __setitem__(self, k, v):
+
+        self.r.set(k, self.serializer.dumps(v))
+
+    def items(self):
+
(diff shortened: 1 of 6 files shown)


alexcjohnson:
Thanks for advancing the discussion with this POC @endremborza - I'm going to close this as it's become stale, but feel free to follow up in #787 if there are further use cases not addressed or awkward using #1179

PR #790, opened 2019-06-24
Added send keys function to dash.testing

TakaakiFuruse:
**Problem**
Currently, dash.testing has [clear_input](https://github.com/plotly/dash/blob/149a8e853f7c94df55cd75c1735bec97717ae746/dash/testing/browser.py#L292) function which does
1. Click input box
2. Clear text

But, there's no function to input text to an input.
Adding such function would increase usability.

**Solution**

I have added a function to `dash/dash/testing/browser.py`

I know that we can do the same thing like...

```py
    input1 = dash_duo.find_element('#input')
    input1.send_keys('hello world')
```

And my way would be...

```py
    input1 = dash_duo.find_element('#input')
    dash_duo.send_key_input(input1, 'hello world')
```

I suppose, my way would make explanation and documenting easier.

# Tests
I haven't add tests as I could not figure out how. 
1. Deleting current way and replace them my new way.
2. Add my test code somewhere.

Which would be better?

## Contributor Checklist

   -  [x] Add code
   -  [ ] Add test
  - (I couldn't make tests run in locally as I got some errors, I would figure out later.)

### optionals

### DIFF
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -301,6 +301,14 @@ class Browser(DashPageMixin):
             .send_keys(Keys.DELETE)
         ).perform()
 
+    def send_key_input(self, elem, key_input):
+        """send key input to an input"""
+        (
+            ActionChains(self.driver)
+            .click(elem)
+            .send_keys(key_input)
+        ).perform()
+
     def get_logs(self):
         """return a list of `SEVERE` level logs after last reset time stamps
         (default to 0, resettable by `reset_log_timestamp`. Chrome only
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -301,13 +301,9 @@ class Browser(DashPageMixin):
             .send_keys(Keys.DELETE)
         ).perform()
 
-    def send_key_input(self, elem, key_input):
-        """send key input to an input"""
-        (
-            ActionChains(self.driver)
-            .click(elem)
-            .send_keys(key_input)
-        ).perform()
+    def send_keys(self, selector, keys):
+        """send keys to an input"""
+        self.find_element(selector).send_keys(keys)
 
     def get_logs(self):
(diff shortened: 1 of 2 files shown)


TakaakiFuruse:
I just realized that my solution would be confusing 

Rather than this..

```
    input1 = dash_duo.find_element('#input')
    dash_duo.send_key_input(input1, 'hello world')
```

This would be conprehensible..

```
    dash_duo.send_key_input('#input', 'hello world')
```

alexcjohnson:
Thanks for the contribution!
I like your latter suggestion. Can we shorten (and match method names) to:
```py
dash_duo.send_keys('#input', 'hello world')
```

Re: tests: just change one or two of the existing tests to use your method, that should be enough.

TakaakiFuruse:
Sure! Let mel recode my PR.

byronz:
@TakaakiFuruse  Thank you for taking the initiative and making a first PR about `dash.testing`.  I double checked the test changes and found it might not necessary to add this extra wrapper or shortcut in `browser.py`.  

As stated in the tutorial, the purpose of our browser API abstraction is to having some syntax sugar and making the test cases more resilient.  For example, the two `find_element` and `find_elements` set the CSS selector as default and avoid the `driver` property access everywhere.  the `clear_input` involves the `ActionChain` in selenium and is a simple migration from existing test cases.  

In the case of `send_keys`,  I feel it's clearer to directly use the *Element API* from selenium like Alex did in his first migration: get the elements and do the actions. *we have no intention to wrap the whole wheel from selenium and/or web element API*

It might be more helpful to abstract some patterns if we need to handle constantly the input with certain special `Key` in the tests. I'm pretty sure you can find a better use case in the near future.

TakaakiFuruse:
Thanks for your review!!

PR #801, opened 2019-06-28
fix for package generation with uninstalled new packages

alexcjohnson:
Small fix needed to get dash-component-boilerplate working for totally fresh new packages. Also simplified `import_module` syntax. This branch is linked to https://github.com/plotly/dash-component-boilerplate/pull/75 - ideally we'll make a new release before we merge that.

Yet another reason to move `js_dist` and `css_dist` into the YAML file - then we can stop importing the package we're in the middle of creating 😅

### DIFF
--- a/dash/development/_r_components_generation.py
+++ b/dash/development/_r_components_generation.py
@@ -228,10 +228,10 @@ def generate_js_metadata(pkg_data, project_shortname):
     -------
     function_string = complete R function code to provide component features
     """
-    importlib.import_module(project_shortname)
-
-    # import component library module into sys
-    mod = sys.modules[project_shortname]
+    # make sure the module we're building is available to Python,
+    # even if it hasn't been installed yet
+    sys.path.insert(0, os.getcwd())
+    mod = importlib.import_module(project_shortname)
 
     alldist = getattr(mod, "_js_dist", []) + getattr(mod, "_css_dist", [])
 


dash/development/_r_components_generation.py
```diff
@@ -228,10 +228,10 @@ def generate_js_metadata(pkg_data, project_shortname):
     -------
     function_string = complete R function code to provide component features
     """
-    importlib.import_module(project_shortname)
-
-    # import component library module into sys
-    mod = sys.modules[project_shortname]
+    # make sure the module we're building is available to Python,
+    # even if it hasn't been installed yet
+    sys.path.insert(0, os.getcwd())
```
rpkyle: Learned something new today; hadn't used this method previously.

rpkyle:
> Yet another reason to move `js_dist` and `css_dist` into the YAML file - then we can stop importing the package we're in the middle of creating 😅

(on the :metro:, so have a little "extra" time)

I was chatting with @Marc-Andre-Rivet about this, and the YAML metadata migration project got me thinking about what would be (from my perspective) an ideal package generator:
- building a new library for (current or future) backends should not require us to significantly refactor/rewrite package generating functions
- component developers should not need to write code to build packages, just need to modify a second, backend-specific YAML file and insert elements corresponding to features to generate, based on templates
- the system should be essentially modular and relatively easy for third-party developers to understand

Perhaps something like ...

```
backend: "R"
        requirements:
            DESCRIPTION: <path-to-description-file-template>
            NAMESPACE: <path-to-namespace-file-template>
            R: <path-to-component-file-template-directory>
```

Issue #804, opened 2019-07-03
Defer `pytest` import?


chriddyp:
Looks like `pytest` isn't the safest dependency, causing issues with scikit-learn: https://community.plot.ly/t/pytest-transient-dependency/25383

Could we move the `import pytest` into the testing module/class/function itself and not require it upon install? We could even have a separate install with setup.py's "extras" feature (https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) like `pip install dash[testing]` or something.

byronz:
@chriddyp  in 1.0, we already use the `extras` feature in `setup.py`,  so the `pip install dash[testing]` might be the way to go.

PR #808, opened 2019-07-04
remove strong testing dependencies.

byronz:
this solves the #804 

the separation could be easily handled by the extra_require. and the new `setup.py`. 

the only change might have an impact is for dash.testing users, they need to do `pip install dash[testing]` instead of `pip install dash`.

### DIFF
--- a/requires-testing.txt
+++ b/requires-testing.txt
@@ -1,0 +1,9 @@
+pytest<5.0.0
+pytest-sugar
+pytest-mock
+lxml
+selenium
+percy
+requests[security]
+beautifulsoup4
+waitress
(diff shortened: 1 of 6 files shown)


requires-testing.txt
```diff
@@ -0,0 +1,9 @@
+pytest<5.0.0
+pytest-sugar
+pytest-mock
+lxml
+selenium
+percy
+requests[security]
+beautifulsoup4
+waitress
```
Marc-Andre-Rivet: Does `testing` require

dash_flow_example==0.0.5
dash-dangerously-set-inner-html
virtualenv;python_version=="2.7"

if not in the `ci` environment?

byronz: they should go -ci.txt as only our integration tests need them

byronz: let me double check the `mock`

PR #812, opened 2019-07-08
:ok_hand: make headless store_true

byronz:


dash/testing/plugin.py
```diff
@@ -27,7 +27,7 @@ def pytest_addoption(parser):
 
     dash.addoption(
         "--headless",
-        action="store",
+        action="store_true",
         default=False,
```
alexcjohnson: not blocking, but I believe you get the default for free with `'store_true'`

PR #817, opened 2019-07-09
Maintenance

alexcjohnson:
2nd (and final) part of https://github.com/plotly/dash-renderer/pull/172 (the first was the ramda update #815)

A collection of random simplifications & generalizations...


dash-renderer/package.json
```diff
@@ -13,8 +13,8 @@
     "build:dev.watch": "webpack --mode development --watch",
     "build": "webpack --mode production",
     "start": "webpack-serve ./webpack.serve.config.js",
-    "format": "prettier --config .prettierrc --write src/**/*.js src/**/*.react.js",
-    "format:test": "prettier --config .prettierrc src/**/*.js src/**/*.react.js --list-different",
```
alexcjohnson: Not sure really what was going on here... but from https://prettier.io/docs/en/cli.html:

> Don't forget the quotes around the globs! The quotes make sure that Prettier expands the globs rather than your shell, for cross-platform usage. The glob syntax from the glob module is used.

Anyway with quotes we don't need separate `.js` and `.react.js` patterns, and we DO match files in the root of src.

alexcjohnson: Looks like this change needs to be propagated to all our repos in fact...

dash-renderer/src/APIController.react.js
```diff
@@ -45,7 +45,7 @@ class UnconnectedContainer extends Component {
         } = props;
 
         if (isEmpty(layoutRequest)) {
-            dispatch(getLayout());
+            dispatch(apiThunk('_dash-layout', 'GET', 'layoutRequest'));
```
byronz: 😸

byronz:
🧽💃 🧽

PR #818, opened 2019-07-09
:sparkles: add download support for both browser

byronz:
This PR adds file download automation in both Firefox and Chrome. It configures both browsers in a way that download file will be saved in a destinated folder (in current proposal, each time a tmpdir per test session), the system related prompt is also disabled.


dash/testing/browser.py
```diff
@@ -363,3 +371,7 @@ def server_url(self, value):
         """
         self._url = value
         self.wait_for_page()
+
+    @property
+    def download_path(self):
```
byronz: test developer can use this to sandbox the download file, as it will be different for the next test execution. In old way, normally it also comes with specific folder clean up as test setup.

dash/testing/browser.py
```diff
@@ -55,10 +58,18 @@ def __init__(
...
+        logger.debug("  download_path => %s", self._download_path)
+
     def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, traceback):
+        try:
+            shutil.rmtree(self._download_path)
+        except OSError:
+            logger.exception("deleting tmp download folder has an issue")
```
alexcjohnson: should we also raise an error if `self._download_path` already exists when we start? Would hate to have our framework delete something it wasn't supposed to!

alexcjohnson: OIC, in `dash_duo` and `dash_br`:
```py
download_path=tmpdir.mkdir('download').strpath
```

Feels weird to me - perhaps we should just have `Browser` make this dir automatically the first time the test asks for `dash_duo.download_path`, and then at the end only delete it if it was created.

byronz: hmm, probably I should just leave `pytest` to manage the lifecycle of the folder 
quote  ```tmpdir fixture provides a temporary directory unique to the test invocation, created in the base temporary directory.```

https://docs.pytest.org/en/latest/tmpdir.html#base-temporary-directory

byronz: > Feels weird to me - perhaps we should just have `Browser` make this dir automatically the first time the test asks for `dash_duo.download_path`, and then at the end only delete it if it was created.

it's agnostic at browser creation whether the fixture will be used for download.

alexcjohnson: ah that's right, the user wouldn't have any reason to access `download_path` until *after* the download has happened, which is of course too late to create it!

OK, so the way you have it now (after https://github.com/plotly/dash/pull/818/commits/998346656eab020019dc50da9d6e9d7dfe767bf8) the `tmpdir` fixture *always* makes this dir, and properly cleans it up afterward? As long as that's known to be reasonably robust across OSes, and the time overhead is minimal, seems good to me!

byronz: from the above link, `entries older than 3 temporary directories will be removed` so I think it's safer to rely on `pytest` than rebuilding the wheel.

PR #819, opened 2019-07-10
Migrate state paths

byronz:
migrate one of the longest test case in test_render.py


tests/integration/callbacks/test_layout_paths_with_callbacks.py
```diff
@@ -0,0 +1,238 @@
...
+    check_call_counts(("chapter2", "chapter3"), 0)
+
+    assert dash_duo.redux_state_paths == EXPECTED_PATHS["chapter1"]
+    check_chapter("chapter1")
+    dash_duo.percy_snapshot(name="chapter-1")
+
+    dash_duo.find_elements('input[type="radio"]')[1].click()  # switch chapters
+
+    # sleep just to make sure that no calls happen after our check
+    time.sleep(0.2)
```
alexcjohnson: couldn't we get rid of these sleeps too by changing the order? Move the `percy_snapshot` call at least after the `wait.until` calls? In fact usually it seems like we call percy as late as possible, which in this case would mean after the `check_chapter` call.

Issue #791, opened 2019-06-24
[BUG] error reporting callback definition errors


alexcjohnson:
From https://community.plot.ly/t/multiple-callback-error/24981

Probably came from removing the `MutableMapping` mixin https://github.com/plotly/dash/pull/753

This app has bad IDs in the second callback:
```py
import dash
import dash_html_components as html
import dash_core_components as dcc
from dash.dependencies import Input, Output

series = [str(i) for i in range(20)]

all_options = {
    'group_1': series[1:5],
    'group_2': series[6:10],
    'group_3': series[11:-1],
}

app = dash.Dash()

app.layout = html.Div(children=[
    html.P([
        html.Label('Choose a group:'),
        dcc.Dropdown(
            id='groups_dropdown',
            options=[{'label': k, 'value': k} for k in all_options.keys()],
            value='group_1'
        )
    ], style=dict(width='400px')),

    html.P([
        html.Label('Choose a series:'),
        dcc.Dropdown(id='series_dropdown')
    ], style=dict(width='400px')),
])


@app.callback(Output('series_dropdown', 'options'),
              [Input('groups_dropdown', 'value')])
def set_series_options(selected_group):
    return [{'label': i, 'value': i} for i in all_options[selected_group]]


@app.callback(Output('series-dropdown', 'value'),
              [Input('series-dropdown', 'options')])
def set_series_value(available_options):
    return available_options[0]['value']


if __name__ == '__main__':
    app.run_server(debug=True)
```

But the error is broken:
```
Traceback (most recent call last):
  File "f.py", line 40, in <module>
    [Input('series-dropdown', 'options')])
  File "/Users/alex/plotly/dash/dash/dash.py", line 1156, in callback
    self._validate_callback(output, inputs, state)
  File "/Users/alex/plotly/dash/dash/dash.py", line 865, in _validate_callback
    list(layout.keys()) + (
AttributeError: 'Div' object has no attribute 'keys'
```

shoegazerstella:
Hi,
I am having the same issue. These are the libraries versions if it can help:
```
dash
dash-html-components==0.13.4
dash-core-components==0.42.1
dash-daq
dash-table==3.1.11
```
Does this depend only on the new version of dash? Or maybe the other components are related?

alexcjohnson:
@shoegazerstella if the issue is what I think it is, it's specific to dash 1.0.0 via #753 - I expect we'll be able to fix this for the next release.

Buyannemekh:
[...]

PR #821, opened 2019-07-13
Fix callback error reporting

alexcjohnson:
#753 simplified the base `Component` class to prohibit fewer prop names - but the `keys` method was actually being used, leading to #791. Easy to replace it with a list comprehension (which calls `__iter__` under the hood).

### DIFF
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -69,7 +69,7 @@ class Browser(DashPageMixin):
             self.driver.quit()
             self.percy_runner.finalize_build()
         except WebDriverException:
-            logger.exception("webdriver quit was not successfully")
+            logger.exception("webdriver quit was not successful")
         except percy.errors.Error:
             logger.exception("percy runner failed to finalize properly")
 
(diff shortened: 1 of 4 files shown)


dash/testing/browser.py
```diff
@@ -247,16 +247,22 @@ def open_new_tab(self, url=None):
...
+                    else webdriver.Remote(
+                        command_executor=remote,
+                        desired_capabilities=getattr(
+                            DesiredCapabilities, self._browser.upper()
+                        ),
+                    )
+                )
+            except WebDriverException:
+                if not i:
+                    raise
```
alexcjohnson: @byronz I had 3 failures this morning - all from different tests, but with the same message:
```
WebDriverException: Message: unknown error:
Chrome failed to start: exited abnormally
(unknown error: DevToolsActivePort file doesn't exist)
```

It's a bit icky to add a blind retry loop like this, but it seemed to work.

tests/integration/test_integration.py
```diff
@@ -898,3 +896,48 @@ def test_inin022_no_callback_context():
...
+            dcc.Input(id='inner-input')
+        ], id='outer-div'),
+        dcc.Input(id='outer-input')
+    ], id='main')
+
+    ids = ['main', 'inner-div', 'inner-input', 'outer-div', 'outer-input']
+
+    with pytest.raises(NonExistentIdException) as err:
+        @app.callback(
+            Output('nuh-uh', 'children'),
```
byronz: might be slightly better with a multi-output edge case?

dash/testing/browser.py
byronz: I had an impression this crash comes always from one specific python version, like 3.7, or 2.7.   Might be good to figure out the real root cause, but assuming it's just turbulence from the circleci docker. it's more related to bad timing due to load performance of the selenium server.   should we try another approach like the `wait.until`? with a reasonable small wait time like 0.1s, I believe the next try should work (we can give a more tolerant 1s timeout).

byronz: @alexcjohnson  I created another PR to fix the deprecated message with `pytest.raises`, and I formatted the file with `black`. so if you can format this test_integration file in this PR, we could avoid a lot potential merge conflicts.

alexcjohnson: I saw this in both 2.7 and 3.7. Feel free to dig in more (preferably in another PR so we can get this bugfix merged), but it's not clear to me what we could wait for, seems like this is happening deep inside Selenium.

tests/integration/test_integration.py
alexcjohnson: added multi-output in https://github.com/plotly/dash/pull/821/commits/fbec9dcd866a847e6299d3130bbf8d8cdfce8619

dash/testing/browser.py
byronz: agree, we can leave it now in this PR.

PR #822, opened 2019-07-15
Fix deprecated messages

byronz:
follow the instructions in  https://docs.pytest.org/en/latest/deprecations.html#message-parameter-of-pytest-raises

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -16,87 +16,102 @@ from dash import Dash, callback_context, no_update
 
 from dash.dependencies import Input, Output, State
 from dash.exceptions import (
-    PreventUpdate, DuplicateCallbackOutput, CallbackException,
-    MissingCallbackContextException, InvalidCallbackReturnValue,
-    IncorrectTypeException
+    PreventUpdate,
+    DuplicateCallbackOutput,
+    CallbackException,
+    MissingCallbackContextException,
+    InvalidCallbackReturnValue,
+    IncorrectTypeException,
 )
 from dash.testing.wait import until
 
 
 def test_inin001_simple_callback(dash_duo):
     app = Dash(__name__)
-    app.layout = html.Div([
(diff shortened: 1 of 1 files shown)


dash/testing/browser.py
```diff
@@ -39,7 +44,7 @@ def __init__(
         self._download_path = download_path
         self._wait_timeout = wait_timeout
 
-        self._driver = self.get_webdriver(remote)
+        self._driver = until(lambda: self.get_webdriver(remote), timeout=1)
```
byronz: @alexcjohnson   I change the hardcode 3x trial with the `wait.until` mechanism, I can try later on remove this and get more info from the logs for debugging

alexcjohnson: Oh I see, `get_webdriver` swallows the error. I suppose that's OK, but the other way has the advantage that if the error persists all the way to the end it's still raised as a regular exception. perhaps build this in (optionally) to `until`, or make a new `until` variant `until_no_error`?

dash/testing/browser.py
```diff
@@ -247,22 +252,20 @@ def open_new_tab(self, url=None):
...
+                    desired_capabilities=getattr(
+                        DesiredCapabilities, self._browser.upper()
+                    ),
                 )
-            except WebDriverException:
-                if not i:
-                    raise
+            )
+        except WebDriverException:
+            logger.exception("<<<Webdriver not initialized correctly>>>")
```
byronz: @alexcjohnson  but the `logger.exception` will record all the exceptions, so it's not swallowed by this function. It depends on how we configure `pytest`, so all the logs could be accessible in the circleci.

alexcjohnson: true, and this has the advantage that you'll actually see *all* the exceptions - ie if it fails 5 times you'll see all 5 - and I guess the `until` will ensure that at the end of it all there at least *is* an exception. OK, let's do it.

PR #825, opened 2019-07-18
R: Exclude nested and private functions from generated NAMESPACE

### LINKED ISSUE #749: R package generator should allow exclusions for user-written anonymous and nested functions
The Dash R package generator code now allows users to bundle in their own functions, in addition to those generated within Python at R package assembly time (https://github.com/plotly/dash/pull/725). If these functions are saved in scripts with names other than `internal.R` or `utils.R`, they will automatically be exported individually via the package `NAMESPACE` file.

The currently proposed implementation does not account for anonymous functions within functions, e.g. such as may be passed via the argument  `FUN = function(` or others which should remain internal to the functions from which they are called.

Simple edits to remedy this would include
- not exporting `FUN` (see [here](https://github.com/plotly/dash-table/blob/ae03f11a6946bdd680259ea31ae8a3f1a8db4a19/R/df_to_list.R#L5))
- allowing a file with function names to ignore in the package root (i.e. `.ignorefns`)
- adding the `.ignorefns` file to `.Rbuildignore`

alexcjohnson:
Fixes #749 
Supersedes #824 

Two changes to the code generating NAMESPACE for R:
- Don't consider functions that are arguments to other function calls or nested inside other functions.
- Omit functions that start with `.` - which seems to be a relatively standard way to mark functions private in R. As @rpkyle points out, there may be other functions we want to exclude, such as S3/S4 methods; we can cross that bridge when we get there. For now I wanted to avoid a user-provided list of functions to exclude as in #824, as that seems likely to be forgotten.

Also fixes a unicode issue reading metadata.json, uncovered by building current master `dash-table` in py27. (cc @Marc-Andre-Rivet)

### DIFF
--- a/dash/development/_r_components_generation.py
+++ b/dash/development/_r_components_generation.py
@@ -639,22 +639,40 @@ def generate_exports(
     omitlist = ["utils.R", "internal.R"] + [
         "{}{}.R".format(prefix, component) for component in components
     ]
-    stripped_line = ""
     fnlist = []
 
     for script in os.listdir("R"):
         if script.endswith(".R") and script not in omitlist:
             rfilelist += [os.path.join("R", script)]
 
-    # in R, either = or <- may be used to create and assign objects
-    definitions = ["<-function", "=function"]
-
     for rfile in rfilelist:
         with open(rfile, "r") as script:
-            for line in script:
-                stripped_line = line.replace(" ", "").replace("\n", "")
-                if any(fndef in stripped_line for fndef in definitions):
-                    fnlist += set([re.split("<-|=", stripped_line)[0]])
+            # put the whole file on one line
+            s = script.read().replace("\n", " ").replace("\r", " ")
+
+            # empty out strings, in case of unmatched block terminators
+            s = re.sub(r"'([^'\\]|\\'|\\[^'])*'", "''", s)
+            s = re.sub(r'"([^"\\]|\\"|\\[^"])*"', '""', s)
+
+            # empty out block terminators () and {}
+            # so we don't catch nested functions, or functions as arguments
+            # repeat until it stops changing, in case of multiply nested blocks
+            prev_len = len(s) + 1
+            while len(s) < prev_len:
+                prev_len = len(s)
+                s = re.sub(r"\(([^()]|\(\))*\)", "()", s)
+                s = re.sub(r"\{([^{}]|\{\})*\}", "{}", s)
+
+            # now, in whatever is left, look for functions
+            matches = re.findall(
+                # in R, either = or <- may be used to create and assign objects
+                r"([^A-Za-z._]|^)([A-Za-z._]+)\s*(=|<-)\s*function", s
+            )
+            for match in matches:
+                fn = match[1]
+                # Allow users to mark functions as private by prefixing with .
+                if fn[0] != "." and fn not in fnlist:
+                    fnlist.append(fn)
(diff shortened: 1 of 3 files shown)


dash/development/_r_components_generation.py
```diff
@@ -639,22 +639,40 @@ def generate_exports(
...
-    # in R, either = or <- may be used to create and assign objects
-    definitions = ["<-function", "=function"]
-
     for rfile in rfilelist:
         with open(rfile, "r") as script:
-            for line in script:
-                stripped_line = line.replace(" ", "").replace("\n", "")
-                if any(fndef in stripped_line for fndef in definitions):
-                    fnlist += set([re.split("<-|=", stripped_line)[0]])
+            # put the whole file on one line
```
byronz: is this generic for loading R app file into a single line? 

 I can imagine a valid use case for testing using the same snippet.

we define the R app as a python triple quote string variable, and sanitize and convert the string to something we can run with `Rscript -e "{app}"`, @rpkyle  also points out other ways if we have the app defined in a R file.

alexcjohnson: I made no attempt to ensure this code would still execute, in fact in general it will not. For that you would need to figure out which line breaks delineate statements and which don't, then separate the statements with semicolons and turn the mid-statement line breaks into spaces. That's a nontrivial task though; much better IMO would be to save the triple-quoted string to a temp file and run it from there.

alexcjohnson:
> do we have some unit tests with some sample R chunks to cover these fancy regex ?

Good call. Now we do https://github.com/plotly/dash/pull/825/commits/4a858026e635c8f087837b4dca75d0d8188d40b0 ... and now they actually work right 😅 

Still look good @byronz? Is my fixture for the temp R directory reasonable?

tests/unit/development/test_r_component_gen.py
```diff
@@ -0,0 +1,85 @@
...
+
+import pytest
+
+from dash.development._r_components_generation import (
+    make_namespace_exports
+)
+
+
+@pytest.fixture
+def make_r_dir():
```
byronz: you can use tmpdir, it's built-in fixture from pytest

dash/development/_r_components_generation.py
```diff
@@ -622,6 +622,31 @@ def generate_exports(
         package_suggests,
         **kwargs
 ):
+    export_string = make_namespace_exports(components, prefix)
```
byronz: TDD always make the code more modular 😸

tests/unit/development/test_r_component_gen.py
alexcjohnson: Except that as the code stands we need the dir to be `./R`, and AFAICT `tmpdir` creates its own and you can't control it. I suppose I could make the R source path a param to `make_namespace_exports`... not sure that's worth it though.

byronz: that's not true, we have no control of the temp parent folder name, but  `tmpdir.mkdir("R")` :)

byronz: anyway, this is minor, and you have my 💃

alexcjohnson: Mmm, I suppose, then we `os.chdir` into the `tmpdir`, and back out when we're done. If you feel like cleaning this up after I merge go for it :)

PR #827, opened 2019-07-19
Dashr test support

### LINKED ISSUE #826: [Feature Request] - add integration support for dashR
dashR integration tests should use the main `dash.testing` framework, but there is no current support to make it easier to adapt. 

The ideal solution should have a harmonized test writing experience in both python and R, the tests for dashR should be the same except the way we host and define the dash app. 

the process app runner flavor needs ~1 for the server setup. might need to find a faster way (possible threading solution https://rpy2.readthedocs.io/en/version_2.8.x/)  in longer-term.

**Additional context**

in case we need to support another language testing (e.g. julia), make the tests structure more generic and flexible in terms of

- test selection
- app hosting
- circleci setup

Reference [issue](https://github.com/plotly/dashR/issues/82)

byronz:
This fixes #826 

the app as string sample
```python
app = '''
library(dash)
library(dashHtmlComponents)
app <- Dash$new()

app$layout(htmlDiv(list(htmlDiv(id='test',children='hello test'))))
app$run_server()
'''

def test_r001_r_with_string(dashr):
    dashr.start_server(app)
```

the filepath version
```python
def test_r002_r_with_path(dashr):
    dashr.start_server('/Users/byron/code/demo.R')
```


byronz:
@rpkyle  @Marc-Andre-Rivet   I need a review on this

dash/testing/application_runners.py
```diff
@@ -214,3 +223,55 @@ def stop(self):
...
+
+        # app is a R string chunk
+        if not (os.path.isfile(app) and os.path.exists(app)):
+            path = (
+                "/tmp/app_{}.R".format(uuid.uuid4().hex)
+                if not self.is_windows
+                else os.path.join(
+                    (os.getenv("TEMP"), "app_{}.R".format(uuid.uuid4().hex))
+                )
+            )
```
alexcjohnson: Seems like this will work, but it doesn't remove the file afterward, does it? Would there have been a way to do this with `tmpdir`? Not sure how to go about using one fixture from within the definition of another, but it feels like all the edge cases and cleanup have already been sorted out there.

PR #835, opened 2019-07-26
Provide support for inlining R examples in .Rd files via YAML

### LINKED ISSUE #831: Need a mechanism to inline Dash for R examples within generated package help
It is customary to provide sample code within R's native documentation which is executable at package build time. This is helpful to package users looking for a quick demo for functions in packages, but also to developers wanting to demonstrate packages work as expected.

Repositories such as CRAN require this to be present, although their current automated package checks pass when a vignette exists. 

To avoid delays for future submissions, and to streamline the path for Dash component developers, we should introduce functionality which allows for easy inlining of function examples.

The `dash-info.yaml` file is a logical avenue to implement this feature. If this file exists, and if it contains examples, we should insert them at the end of the corresponding `.Rd` file, e.g.

```    
r_examples:
    - name: df_to_list
      dontrun: TRUE
      code: |
           # first, create data frame
           df <- read.csv(url(
                  'https://raw.githubusercontent.com/plotly/datasets/master/solar.csv'
               ),
              check.names=FALSE,
              stringsAsFactors=FALSE
           )
           # then convert to list-of-lists format for use in dashTable
           # the following snippet below will print as JSON
           # see the help for dashDataTable to see an actual app example
           dashDataTable(
                  id = 'table',
                  columns = lapply(colnames(df), function(x) {
                         list(name = x, id = x)
                  }),
                  data = df_to_list(df)
           )
```

...should generate the following `.Rd` markup:

```
\examples{
\dontrun{
           # first, create data frame
           df <- read.csv(url(
                  'https://raw.githubusercontent.com/plotly/datasets/master/solar.csv'
               ),
              check.names=FALSE,
              stringsAsFactors=FALSE
           )
           # then convert to list-of-lists format for use in dashTable
           # the following snippet below will print as JSON
           # see the help for dashDataTable to see an actual app example
           dashDataTable(
                  id = 'table',
                  columns = lapply(colnames(df), function(x) {
                         list(name = x, id = x)
                  }),
                  data = df_to_list(df)
           )}
}
```

@alexcjohnson @Marc-Andre-Rivet

rpkyle:
This PR proposes to add functionality for storing R examples for help files in the `dash-info.yaml` file, as outlined in #831.

Closes #831.


rpkyle:
@byronz Not sure why it passed tests locally but failed on CircleCI, but made a minor changes and it looks 👌 now

dash/development/_r_components_generation.py
```diff
@@ -392,7 +392,7 @@ def write_help_file(name, props, description, prefix, rpkg_data):
             item_text=item_text,
             description=description.replace('\n', ' ')
         ))
-    if rpkg_data is not None:
+    if rpkg_data is not None and rpkg_data.get('r_examples') is not None:
```
byronz: 🐄  minor: `'r_examples' in rpkg_data`

dash/development/_r_components_generation.py
```diff
@@ -374,25 +384,40 @@ def write_help_file(name, props, description, prefix):
...
-
-
-def write_class_file(name, props, description, project_shortname, prefix=None):
+    if rpkg_data is not None and rpkg_data.get('r_examples') is not None:
+        ex = rpkg_data.get('r_examples')
+        the_ex = ([e for e in ex if e.get("name") == funcname] or [None])[0]
+        result = ""
+        if the_ex and "code" in the_ex.keys():
+            result += wrap("examples",
+                           wrap("dontrun" if the_ex["dontrun"] else "",
```
alexcjohnson: `if the_ex.get("dontrun")`
When would a user want or not want `dontrun`?

rpkyle: There's a slight delay when executing inline examples within R packages; CRAN requires that each example run and exit without errors in a second or two. Dash apps aren't typically terminated until the user interrupts them; this is problematic for CRAN (and devtools) checks.

In addition, CRAN checks will fail if a dependency does not exist in the repository when examples are executed. Setting `\dontrun` to `TRUE` allows the package author to include examples but specify that they not be executed during package tests.

dash/development/_r_components_generation.py
rpkyle: That does look cleaner! Will edit and test; thanks for the feedback.

alexcjohnson:
@rpkyle this looks good - just two small changes as requested by me (before my question about `dontrun` was a suggested syntax change) and by byron.

dash/development/_r_components_generation.py
rpkyle: fixed in a7883da

dash/development/_r_components_generation.py
rpkyle: My apologies for not catching the initial comment previously, change has been made!

rpkyle: fixed in cf0d476

alexcjohnson: no worries, the way I wrote it did look like I was quoting you rather than suggesting a change.

dash/development/component_generator.py
```diff
@@ -98,8 +98,15 @@ def generate_components(
             os.makedirs('man')
         if not os.path.exists('R'):
             os.makedirs('R')
+        if os.path.isfile("dash-info.yaml"):
+            with open("dash-info.yaml") as yamldata:
+                rpkg_data = yaml.safe_load(yamldata)
```
alexcjohnson: We need a fallback declaration of `rpkg_data` if `dash-info.yaml` doesn't exist - like we had [previously](https://github.com/plotly/dash/pull/835/files#diff-5704162f659bd0685c89d923fa806ed5L120):
```py
else:
    rpkg_data = None
```

(alternatively, initialize it to `None` up above the `if os.path.isfile("dash-info.yaml")` - just need to ensure it always exists by the time we ask for it below)

rpkyle: Good catch. Sometimes I wish there was a ⚾️ mitt emoji.

rpkyle: fixed in 139527d

PR #842, opened 2019-07-29
:camera_flash: using python3.7 for percy

### LINKED ISSUE #841: [Feature Request] CI Improvement
The CI job for dash is still about ~20 minutes, we would like to speed up the integration by enabling the headless mode (the race condition has a lot of hardcode sleep, so by adding headless in CI job, we can reduce about ~1 minute, but not by a factor of 50%), and redesign the workflow in a parallel run flavor.

byronz:
this fixes #841 

## Contributor Checklist

   -  [x] using python3.7 for percy and compare the performance 
   -  [x] use headless mode for integration and compare the performance
   -  [ ] integrate dashR test 
   -  [ ] integrate dash-table test 
   -  [ ] integrate dcc/html test 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 jobs:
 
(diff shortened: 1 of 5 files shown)


byronz:
https://docs.percy.io/docs/parallel-test-suites
https://circleci.com/blog/setting-up-continuous-integration-with-github/
https://circleci.com/docs/2.0/workflows/

PR #845, opened 2019-07-31
Ci speedup

### LINKED ISSUE #841: [Feature Request] CI Improvement

byronz:
this fixes #841 

## Contributor Checklist

   -  [x] using python3.7 for percy and compare the performance 
   -  [x] use headless mode for integration and compare the performance
   -  [x] improve CI performance by running integration in parallel and distribute build time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,92 +1,166 @@
 version: 2
-
 jobs:
+  "percy-finalize":
+    docker:
+      - image: percyio/agent
+    steps:
+      - run: percy finalize --all
 
-  "python-2.7": &test-template
+  "lint-unit-37": &lint-unit
     working_directory: ~/dash
     docker:
-      - image: circleci/python:2.7-stretch-node-browsers
+      - image: circleci/python:3.7-stretch-node-browsers
         environment:
-          PYLINTRC: .pylintrc
+          PYLINTRC: .pylintrc37
+          PYVERSION: python37
 
     steps:
       - checkout
-
-      - run:
-          name: ℹ️ CI Context
-          command: |
-            echo "TRIGGERER: ${CIRCLE_USERNAME}"
-            echo "BUILD_NUMBER: ${CIRCLE_BUILD_NUM}"
-            echo "BUILD_URL: ${CIRCLE_BUILD_URL}"
-            echo "BRANCH: ${CIRCLE_BRANCH}"
-            echo "RUNNING JOB: ${CIRCLE_JOB}"
-            echo "JOB PARALLELISM: ${CIRCLE_NODE_TOTAL}"
-            echo "CIRCLE_REPOSITORY_URL: ${CIRCLE_REPOSITORY_URL}"
-            echo $CIRCLE_JOB > circlejob.txt
-
+      - run: echo $PYVERSION > ver.txt
       - restore_cache:
-          key: v-{{ checksum "circlejob.txt" }}-{{ checksum "requires-ci.txt" }}-{{ checksum "requires-install.txt" }}-{{ checksum "requires-testing.txt" }}
+          key: dep-{{ checksum "ver.txt" }}-{{ checksum "requires-ci.txt" }}-{{ checksum "requires-install.txt" }}-{{ checksum "requires-testing.txt" }}
       - run:
           name: 🚧 pip dev requirements
           command: |
             sudo pip install --upgrade virtualenv
-            python -m venv venv || virtualenv venv
-            . venv/bin/activate
-            pip install -r requires-install.txt -r requires-ci.txt -r requires-testing.txt --quiet
+            python -m venv venv || virtualenv venv && . venv/bin/activate
+            pip install -e . -r requires-install.txt -r requires-ci.txt -r requires-testing.txt --quiet
       - save_cache:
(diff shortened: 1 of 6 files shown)


byronz:
useful references:

https://docs.percy.io/docs/parallel-test-suites
https://circleci.com/blog/setting-up-continuous-integration-with-github/
https://circleci.com/docs/2.0/workflows/

.circleci/config.yml
```diff
@@ -1,121 +1,218 @@
...
+      - attach_workspace:
+          at: ~/dash
       - run:
-          name: ⚙️ run integration test
+          name: ⚙️ Run Integration Tests
           command: |
             . venv/bin/activate
-            pytest --junitxml=test-reports/junit_intg.xml tests/integration/
+            cd packages && ls -la && pip install * && cd .. && pip list | grep dash
+            TESTFILES=$(circleci tests glob "tests/integration/**/test_*.py" | circleci tests split --split-by=timings)
```
byronz: circleci way to split and select tests based on previous execution duration infos

.circleci/config.yml
```diff
@@ -1,121 +1,218 @@
 version: 2
-
 jobs:
+  "percy-finalize":
+    docker:
+      - image: percyio/agent
+    steps:
+      - run: percy finalize --all
```
byronz: this works together with a `PERCY_PARALLEL_TOTAL: '-1'` in test job, percy command will compare the number of containers by fetching ENV variable value from circleci.

.circleci/config.yml
```diff
@@ -1,121 +1,218 @@
...
-            echo "BRANCH: ${CIRCLE_BRANCH}"
-            echo "RUNNING JOB: ${CIRCLE_JOB}"
-            echo "JOB PARALLELISM: ${CIRCLE_NODE_TOTAL}"
-            echo "CIRCLE_REPOSITORY_URL: ${CIRCLE_REPOSITORY_URL}"
-            echo $CIRCLE_JOB > circlejob.txt
-
+      - run: echo $PYVERSION > ver.txt
       - restore_cache:
-          key: v-{{ checksum "circlejob.txt" }}-{{ checksum "requires-ci.txt" }}-{{ checksum "requires-install.txt" }}-{{ checksum "requires-testing.txt" }}
+          key: dep-{{ checksum "ver.txt" }}-{{ checksum "requires-ci.txt" }}-{{ checksum "requires-install.txt" }}-{{ checksum "requires-testing.txt" }}
```
byronz: the cache key formula might need further tuning later

byronz:
the whole config.yml can be improved using some new features in version 2.1, but out of scope of this PR.

byronz:
we might need to change the documents as the current change have an impact on local circlecli runs.

tests/integration/test_race_conditions.py
```diff
@@ -16,7 +16,7 @@ def setUp(self):
         pass
 
 
-DELAY_TIME = 1
+DELAY_TIME = 0.2
```
alexcjohnson: heh that in itself should be a huge speedup. Should be just fine - if for some reason that's ever not a long enough delay to actually have the responses arrive at the front end in the desired order, it would be a (presumably rare) missed test, or testing a *different* order, but one that should still be valid. So I'm happy to make this change, but some day we should refactor this test to something a little smarter that actually triggers the responses based on evidence that the previous response has been received and acted upon, rather than just time.

tests/integration/test_render.py
```diff
@@ -957,7 +957,7 @@ def update_output(value):
         self.wait_for_text_to_equal('#output-post', 'request_post changed this text!')
         self.wait_for_text_to_equal('#output-post-payload', '{"output":"output-1.children","changedPropIds":["input.value"],"inputs":[{"id":"input","property":"value","value":"fire request hooks"}]}')
         self.wait_for_text_to_equal('#output-post-response', '{"props":{"children":"fire request hooks"}}')
-        self.percy_snapshot(name='request-hooks')
+        self.percy_snapshot(name='request-hooks render')
```
alexcjohnson: oh interesting - was one of these snapshots blocking the other one previously? How did you discover this? Do we need some additional mechanism (like a list of previously-used names inside `percy_snapshot` - though I guess in this case that wouldn't help as one's the new framework and the other is still using the old system) to ensure we don't double up a name again?

alexcjohnson: And I guess the next question: do these two tests actually both have distinct value, or should we delete one of them?

byronz: I got a 400 error when the test job start to have parallelism > 1

alexcjohnson: oh weird - so it allows multiple snapshots with the same name from the same process but not from different processes? very strange, though useful I guess as any test we do in our own code could only catch duplicates in the same process.

byronz: there was another issue with the previous approach, as we spin up a browser per test vs per session, the percy_finalize was called after each test case. 

anyway, there are some extra integration under the hood from percy client, it adds some extra protection for all major CI tools.

alexcjohnson: Ah, got it - right, it's probably the extra finalizing that buried the duplicate, and now percy will be taking care of ensuring uniqueness.

PR #848, opened 2019-07-31
Delete contents of inst/deps when rebuilding R packages

### LINKED ISSUE #847: Generator should delete contents of inst/deps for R packages prior to rebuild
Currently, assets for component libraries which are shared between Python and R implementations of Dash are copied from within `project_shortname` to `inst/deps` without removing any artifacts of previous build cycles.

To avoid confusion in case older versions of assets remain from earlier build cycles, we should remove any files in `inst/deps` before assets are copied. The relevant code is here:

https://github.com/plotly/dash/blob/57dae3f0a64d82b2e87d67cfdcc2448950ffd324/dash/development/_r_components_generation.py#L442-L455

Documents in other `inst/*` subdirectories should continue to be left as-is, since these resources are not necessarily 1:1 with the Python package (e.g. vignettes, sample data in R format, etc.)

@alexcjohnson @Marc-Andre-Rivet @shammamah

rpkyle:
This PR proposes to modify the package generation code such that the `inst/deps` folder within generated R packages is emptied prior to copying over assets from the `project_shortname` directory:

https://github.com/plotly/dash/blob/4a689e01dcdbc0a2ad09871078c537d754ddbff6/dash/development/_r_components_generation.py#L447-L449

Closes #847.

### DIFF
--- a/dash/development/_r_components_generation.py
+++ b/dash/development/_r_components_generation.py
@@ -442,11 +442,10 @@ def write_js_metadata(pkg_data, project_shortname, has_wildcards):
     # now copy over all JS dependencies from the (Python) components dir
     # the inst/lib directory for the package won't exist on first call
     # create this directory if it is missing
-    if not os.path.exists("inst/deps"):
-        os.makedirs("inst/deps")
-    else:
+    if os.path.exists("inst/deps"):
         shutil.rmtree("inst/deps")
-        os.makedirs("inst/deps")
+
+    os.makedirs("inst/deps")
 
     for javascript in glob.glob("{}/*.js".format(project_shortname)):
         shutil.copy(javascript, "inst/deps/")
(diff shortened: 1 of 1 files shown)


dash/development/_r_components_generation.py
```diff
@@ -444,6 +444,9 @@ def write_js_metadata(pkg_data, project_shortname, has_wildcards):
     # create this directory if it is missing
     if not os.path.exists("inst/deps"):
         os.makedirs("inst/deps")
+    else:
+        shutil.rmtree("inst/deps")
+        os.makedirs("inst/deps")
```
alexcjohnson: 🌴 
```py
if os.path.exists("inst/deps"):
    shutil.rmtree("inst/deps")

os.makedirs("inst/deps")
```

rpkyle: 🐠

byronz: in case you need a more robust tool, try and catch

rpkyle: fixed in c8dd6e9

PR #849, opened 2019-07-31
Ci speedup

byronz:
this PR allows we control the percy snapshot finalize process by pytest flag, so we can do `pytest --percyfinalize` in case we need parallelism in the integration jobs.

### DIFF
--- a/dash/testing/plugin.py
+++ b/dash/testing/plugin.py
@@ -30,7 +30,15 @@ def pytest_addoption(parser):
     )
 
     dash.addoption(
-        "--headless", action="store_true", help="Run tests in headless mode"
+        "--headless",
+        action="store_true",
+        help="set this flag to run in headless mode",
+    )
+
+    dash.addoption(
+        "--percyfinalize",
+        action="store_false",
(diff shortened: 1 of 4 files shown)


dash/testing/plugin.py
```diff
@@ -30,7 +30,15 @@ def pytest_addoption(parser):
...
-        "--headless", action="store_true", help="Run tests in headless mode"
+        "--headless",
+        action="store_true",
+        help="set this flag to run in headless mode",
+    )
+
+    dash.addoption(
+        "--percyfinalize",
+        action="store_false",
+        help="set this flag to control percy finalize at CI level",
```
alexcjohnson: would this make more sense to be called `--percynofinalize`? `"store_false"` just seems inherently confusing...

byronz: `--nopercyfinalize`?

alexcjohnson: > `--nopercyfinalize`

even better!

byronz:
dancer without actual approve can't get merged lol

PR #854, opened 2019-08-06
Improve ci

byronz:
improvement based on 1.1 release:

- make the pip install robust for build jobs
- adding artifacts based on branch and tags,

### DIFF
--- a/requires-testing.txt
+++ b/requires-testing.txt
@@ -1,4 +1,4 @@
-pytest<5.0.0
+pytest
 pytest-sugar
 pytest-mock
 lxml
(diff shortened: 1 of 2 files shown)


requires-testing.txt
```diff
@@ -1,4 +1,4 @@
-pytest<5.0.0
+pytest
```
alexcjohnson: Oh, I didn't notice when <5 came in with https://github.com/plotly/dash/pull/808 - what's the story here? Why did you add it then, why did the restriction get removed now?

byronz: There was an issue with pytest at the moment, I was looking for a candidate to trigger a cache key not found, a natural pick for the purpose.

PR #860, opened 2019-08-10
dev_tools_prune_errors

### LINKED ISSUE #843: [Feature Request] Simplified tracebacks in devtools
Currently devtools provides extremely long tracebacks when there's an error in a callback. They look something like this:
```
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/app.py", line 2463, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/app.py", line 2449, in wsgi_app
    response = self.handle_exception(e)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/app.py", line 1866, in handle_exception
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/app.py", line 2446, in wsgi_app
    response = self.full_dispatch_request()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/app.py", line 1951, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/app.py", line 1820, in handle_user_exception
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/flask/app.py", line 1935, in dispatch_request
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/dash/dash.py", line 1287, in dispatch
    response.set_data(self.callback_map[output]['callback'](*args))
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/dash/dash.py", line 1171, in add_context
    output_value = func(*args, **kwargs)
  File "/Users/alex/plotly/dash-dev/agricultural_exports/agricultural_exports.py", line 58, in update_graph
    raise ValueError('boo')
ValueError: boo
```

Normally an app developer is only interested in the part in their own code (the last entry above, but could be more if their code was more involved). Even dash core devs wouldn't normally care about anything above the two extra dash entries, though @bpostlethwaite claims to have had to chase something through the flask stack in the past.

I'd propose an option like `dev_tools_short_tracebacks` with a default `True`, so normally all that shows up is the app code, but if necessary you can set it `False` and see it in all the current gory detail.

alexcjohnson:
Closes #843 

Adds a new `run_server` / `enable_dev_tools` arg `dev_tools_prune_errors` that defaults to `True`. In that default case, tracebacks only include user code (and anything below it in the stack), excluding entries from Dash and Flask. Set it to `False` to revert to the current behavior of showing the complete stack.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -19,7 +19,7 @@ from textwrap import dedent
 import flask
 from flask import Flask, Response
 from flask_compress import Compress
-from werkzeug.debug import get_current_traceback
+from werkzeug.debug.tbtools import get_current_traceback
 
 import plotly
 import dash_renderer
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -19,6 +19,7 @@ from textwrap import dedent
 import flask
 from flask import Flask, Response
 from flask_compress import Compress
+from werkzeug.debug import get_current_traceback
 
 import plotly
 import dash_renderer
@@ -1168,7 +1169,8 @@ class Dash(object):
         def wrap_func(func):
             @wraps(func)
             def add_context(*args, **kwargs):
-                output_value = func(*args, **kwargs)
+                # don't touch the comment on the next line - used by debugger
+                output_value = func(*args, **kwargs)  # %% callback invoked %%
                 if multi:
                     if not isinstance(output_value, (list, tuple)):
                         raise exceptions.InvalidCallbackReturnValue(
@@ -1390,7 +1392,8 @@ class Dash(object):
                 'props_check',
                 'serve_dev_bundles',
                 'hot_reload',
-                'silence_routes_logging'
+                'silence_routes_logging',
+                'prune_errors'
         ):
             dev_tools[attr] = get_combined_config(
                 attr, kwargs.get(attr, None), default=debug
@@ -1419,7 +1422,8 @@ class Dash(object):
             dev_tools_hot_reload_interval=None,
             dev_tools_hot_reload_watch_interval=None,
             dev_tools_hot_reload_max_retry=None,
-            dev_tools_silence_routes_logging=None):
+            dev_tools_silence_routes_logging=None,
+            dev_tools_prune_errors=None):
         """
         Activate the dev tools, called by `run_server`. If your application is
         served by wsgi and you want to activate the dev tools, you can call
@@ -1483,6 +1487,11 @@ class Dash(object):
             env: ``DASH_SILENCE_ROUTES_LOGGING``
         :type dev_tools_silence_routes_logging: bool
 
+        :param dev_tools_prune_errors: Reduce tracebacks to just user code,
+            stripping out Flask and Dash pieces. `True` by default, set to
+            `False` to see the complete traceback.
+        :type dev_tools_prune_errors: bool
+
         :return: debug
         """
         if debug is None:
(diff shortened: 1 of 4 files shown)


tests/integration/devtools/test_props_check.py
```diff
@@ -15,22 +15,22 @@
         "fail": True,
         "name": 'missing required "value" inside options',
         "component": dcc.Checklist,
-        "props": {"options": [{"label": "hello"}], "values": ["test"]},
+        "props": {"options": [{"label": "hello"}], "value": ["test"]},
```
alexcjohnson: This is what infuriates me about Percy... OK, here it's just a test problem, not a real bug, but the images were completely different from what they should have been, and got auto-approved, I assume, because the change came from DCC and so likely happened first in master.

dash/dash.py
```diff
@@ -1168,7 +1169,8 @@ def callback(self, output, inputs=[], state=[]):
         def wrap_func(func):
             @wraps(func)
             def add_context(*args, **kwargs):
-                output_value = func(*args, **kwargs)
+                # don't touch the comment on the next line - used by debugger
+                output_value = func(*args, **kwargs)  # %% callback invoked %%
```
byronz: is the `%% callback invoked %%` designed for testing or there is some black magic hook underneath?

dash/dash.py
```diff
@@ -1527,6 +1537,21 @@ def enable_dev_tools(
...
+            def _wrap_errors(_):
+                # find the callback invocation, if the error is from a callback
+                # and skip the traceback up to that point
+                # if the error didn't come from inside a callback, we won't
+                # skip anything.
+                tb = get_current_traceback()
+                skip = 0
+                for i, line in enumerate(tb.plaintext.splitlines()):
+                    if "%% callback invoked %%" in line:
+                        skip = int((i + 1) / 2)
```
byronz: why the skip is about half of code to the callback stub? just curious about the algorithm

dash/dash.py
```diff
@@ -19,6 +19,7 @@
 import flask
 from flask import Flask, Response
 from flask_compress import Compress
+from werkzeug.debug import get_current_traceback
```
byronz: minor: 
```suggestion
from werkzeug.debug.tbtools import get_current_traceback
```

dash/dash.py
alexcjohnson: The plaintext traceback has 2 lines for each stack frame: first line lists the file, line #, and function; second line is the actual line of code. +1 gives the first line of the next pair - the first frame we want to report - then /2 gives the number of frames above it.

dash/dash.py
alexcjohnson: Right, that's where it's actually defined... but is that better? I thought it'd be safest to import from as high up in the package as possible, in case the package internals get restructured later. That said it looks like even internally to werkzeug they generally use the import you suggest, and I guess `werkzeug.debug.__init__` isn't really importing things to export a clean namespace, it just wants to use `get_current_traceback` itself. OK, I'll change it.

PR #863, opened 2019-08-12
Fix ci

byronz:


.circleci/config.yml
```diff
@@ -192,10 +192,10 @@ jobs:
       - attach_workspace:
           at: ~/dash
       - run:
-          name: ⚙️ Run Integration Tests
+          name: 🧪 Run Integration Tests
           command: |
-            . venv/bin/activate
-            cd packages && ls -la && pip install * && cd .. && pip list | grep dash
+            . venv/bin/activate && cd packages && ls -la
+            find . -name "*.gz" | xargs pip install --force-reinstall && pip list | grep dash && cd ..
```
byronz: lessons learned in dcc, better to install this way.

alexcjohnson: Why? Is it just that you want the `--force-reinstall` flag and that doesn't work with `pip install *`?

byronz: I found the pip install * with tar.gz were not working as expected in DCC CI. The built dcc.tar.gz contains a fix for one test case, but it seems the DCC was not installed with built version,but the one from dash dependencies. I can tell as the test failed and a more awkward step including uninstall and install can make it pass.

PR #881, opened 2019-08-22
Assets #597

byronz:
this is a follow-up of https://github.com/plotly/dash-core-components/issues/597 

mainly make the build tool more DRY, so the `build` has a clean abstraction pipeline defined in the base class.


dash/development/build_process.py
```diff
@@ -0,0 +1,210 @@
...
+            ),
+        )
+
+
+class DCC(BuildProcess):
+    def __init__(self, main=None):
+        main = (
+            os.path.abspath(os.path.expanduser(main)) if main else os.getcwd()
+        )
+        logger.info("start the build process at %s", main)
```
byronz: unlike renderer, dcc folder cannot be computed from location of `__file__`,  user can either run it at root, or specify the path like this `dcc build --main=~/project/dash-core-components`

dash/development/build_process.py
```diff
@@ -0,0 +1,210 @@
...
+        if os.path.basename(main) != "dash-core-components":
```
byronz: this extra protection might make people changed the folder name unpleasant,  like someone git clone and checkout to `dcc`

dash/development/build_process.py
alexcjohnson: I'm uneasy with putting dcc-specific code into the dash repo - can't this whole class live in the dcc repo? perhaps in the `scripts` dir? Then perhaps we wouldn't need this hack?

byronz: I know the current code splits between two repos, but what about the part of the generation of Python & R components?  if we want to have a clean logical structure, we should probably 1: move all the relevant code into DCC 2. merge dcc into dash?

alexcjohnson: Component generation is a generic process that applies to all component packages - and all of the package-specific pieces of that are controlled within the package it applies to. Initially that was done just by command-line args in the package scripts, now additional configuration is going into the `dash-info.yaml` file. There's nothing package-specific in `dash.development`. So if at all possible I'd like to continue that structure. That way other component packages can choose to use these facilities as well, including 3rd-party packages.

byronz: sounds reasonable

OK,  I can move this class to dcc, also fix the issues @shammamah  reported in https://github.com/plotly/dash-core-components/issues/590

the dash is a dependency for dcc to run the generate-component and also import the base process class

alexcjohnson: > the dash is a dependency for dcc to run the generate-component and also import the base process class

Right. It's currently an *unlisted* dependency, because we don't support dcc as an entry point and don't want a circular version conflict, you just get it when you install dash. Anyway the circularity is only there for devs, not for users. Is there a way to specify this kind of thing cleanly in `setup.py`?

byronz: yes, the circular dependency was already present implicitly like this generated code in component

```from dash.development.base_component import Component, _explicitize_args```

but in dash and dash-renderer, actually, there is no such code to do direct import from DCC
we do that only with app writing. 

I feel DCC has more strong dependency on dash than the other direction.  Dash requires dcc mainly for app writing convenient.

byronz: the only clear solution I see is merging dcc into dash :)

PR #894, opened 2019-08-29
Remote grid

### LINKED ISSUE #829: [BUG] dash_duo does not set remote thus fails when get_webdriver is called
**Describe your context**
```
dash                                           1.0.2
dash-core-components                           1.0.0
dash-html-components                           1.0.0
dash-renderer                                  1.0.0
dash-table                                     4.0.2

```
- OS: Windows 10
- Browser Chrome 
- Version 75.0.3770.142

**Describe the bug**

Trying to use new dash.testing. Running the example code copied from the documentation (https://dash.plot.ly/testing) in a headless mode (with a Chrome webdriver running). 

An error gets thrown in 
```
dash\testing\browser.py:47: in __init__
    self._driver = until(lambda: self.get_webdriver(remote), timeout=1)
```

It appears that the variable remote is never set in the call stack of dash_duo and thus the call to get_webdriver always fails.

**Expected behavior**

It should work. I don't know what else to say beyond the basic example doesn't appear to work. If there is some setup requirement that is missing, then a clarification on the documentation needs to be added, otherwise at some point in the setup of `dash_duo` the value of remote needs to be set such that the `get_webdriver` call doesn't fail and it correctly fetches the webdriver when initialising the browser.

byronz:
This closes #829 

added a new argument `--remote-url` for the usage of selenium grid in the cloud, to test with a local host grid, users just need to add `--remote` 

- [x] needs a follow-up PR in dash-docs https://github.com/plotly/dash-docs/pull/639


byronz:
I will probably add few unit tests

PR #899, opened 2019-08-30
Async support and general performance work

Marc-Andre-Rivet:
Realted to component-specific implementations:
https://github.com/plotly/dash-core-components/pull/616
https://github.com/plotly/dash-table/pull/554

Component loading state is exposed on the component class / function.

The renderer evals the loading state of all component types on layout changes (possibility of new component types).

`notifyObserver` uses this state to decide whether to go forward with callbacks or wait.

Because of callback chains and multi-input & state parameters, it's not possible to only depend on the state of the current component. This may result in more overhead vs. checking out only the state of the (1) components in the dependency graph, (2) components impacted by the callback's dependency chain, but it feels like a good first implementation.

---

Dash now supports a new attribute on `_js_dist` assets: `async`.
`async` can't be used with `dynamic` and is instead used to determine at runtime if an asset should be `eager` or `lazy`. Takes three values: True, `eager`, `lazy`.

This is necessary to cover cases like the dcc.Graph plotly.js dependency -- we need `window.Plotly` to be set automatically if eagerly loaded, for the value to already be set after DOM render, for legacy purposes.

---

- [x] Special webpack plugin to redirect Dash dynamic imports correctly, byt taking into account the Dash path structure [webpack-dash-dynamic-import](https://github.com/plotly/dash/tree/exp-dynamic/%40plotly/webpack-dash-dynamic-import)
- [x] Common code for all async components to use to make the renderer aware of their 'aysnc' nature [dash-component-plugins](https://github.com/plotly/dash/tree/exp-dynamic/%40plotly/dash-component-plugins) -- this was made more general with thoughts of adding other plugins here (e.g. the private layout could be abstracted here for components that need it)

- [x] Test Dash resolution of `async` with different values for `eager_loading` and `force_eager_loading` flags
- [x] Validate `lazy` requests are not loaded right away (e.g. DOM comparison)
- [x] Test `setAppIsReady` with shim -- make sure callbacks are blocked
---

Additional performance work that could be done here or spin off new PRs:
- [ ] Investigate the possibility of dynamically deciding to be lazy/eager/prefetch/preload (determining the mechanism by which dash would provide the info is out of scope - pure FE sub-poc)
- [ ] Chunk the renderer into two parts: (1) initiate layout request to server, (2) get the react/redux code -- one can be made as small as possible and put in the `<head>`, the other can stay were it is atm
- [ ] Load the renderer with script tags in `<head>` instead of `<body>` -- wait for DOM from there
- [ ] More aggressive chunking to parallelize js processing / download
- [ ] Shared `core-js` in renderer, externalized for all other components (renderer to pull `all` instead of `usage`) -- need to coordinate targets
- [ ] in prod, use attribute `?v=x.y.z` with the version instead of the timestamp -- keep using the timestamp for debug
- [ ] In dcc, make additional components lazy
- [ ] apply a different priority to css and other resources (not sure there's something to be done here yet)

In order of priority / potential gains, stopping cache busting in production code is probably the biggest (and easiest) gain we can get.


dash-renderer/src/APIController.react.js
```diff
@@ -54,6 +55,7 @@ class UnconnectedContainer extends Component {
                     dispatch
                 );
                 dispatch(setLayout(finalLayout));
+                dispatch(setAppIsReady());
```
Marc-Andre-Rivet: The layout changes on `SET_LAYOUT`, `UNDO`, `REDO`, `REVERT`. When doing one, do the other.

dash-renderer/src/actions/index.js
```diff
@@ -216,10 +223,15 @@ function reduceInputIds(nodeIds, InputGraph) {
 }
 
 export function notifyObservers(payload) {
-    return function(dispatch, getState) {
+    return async function(dispatch, getState) {
```
Marc-Andre-Rivet: notifyObservers may now have to wait defensively before sending callbacks

dash-renderer/src/actions/index.js
```diff
@@ -216,10 +223,15 @@ function reduceInputIds(nodeIds, InputGraph) {
...
         const {id, props, excludedOutputs} = payload;
 
-        const {graphs, requestQueue} = getState();
+        const {graphs, isAppReady, requestQueue} = getState();
+
+        if (isAppReady !== true) {
+            await isAppReady;
+        }
```
Marc-Andre-Rivet: If a known component is not ready, wait for it

dash-renderer/src/actions/setAppReadyState.js
```diff
@@ -0,0 +1,80 @@
...
+        await ready;
+        /**
+         * All known async is ready.
+         *
+         * Callbacks were blocked while waiting, we can safely
+         * assume that no update to layout happened to invalidate.
+         */
+        dispatch(setAction(true));
+    }
+};
```
Marc-Andre-Rivet: On layout updates, determine the loading state of all known component types. Set `isAppReady` to a promise if it's not ready and to `true` if it is -- the action will wait until all component types are ready to update the state a 2nd time. Further layout updates will trigger a re-evaluation. Given that with the current implementation of Dash, once we reach `true` we know there can't be additional changes to that state, I plan on updating the logic here + explanation and forgo the check once `true`.

dash/dash.py
```diff
@@ -213,6 +213,7 @@ def __init__(
         assets_url_path="assets",
         assets_ignore="",
         assets_external_path=None,
+        eager_loading=False,
```
Marc-Andre-Rivet: By default Dash wants to load resources lazily

dash/dash.py
```diff
@@ -228,6 +229,11 @@ def __init__(
         plugins=None,
         **obsolete
     ):
+        # Apply _force_eager_loading overrides from modules
+        for module_name in ComponentRegistry.registry:
+            module = sys.modules[module_name]
+            eager = getattr(module, '_force_eager_loading', False)
+            eager_loading = eager_loading or eager
```
Marc-Andre-Rivet: jerry-rigged from Scripts & Css, not sure it warrants a wrapper

dash/resources.py
```diff
@@ -14,48 +14,73 @@ def __init__(self, resource_name):
...
         filtered_resources = []
         for s in all_resources:
             filtered_resource = {}
             if 'dynamic' in s:
                 filtered_resource['dynamic'] = s['dynamic']
+            if 'async' in s:
+                if 'dynamic' in s:
+                    raise exceptions.ResourceException(
+                        '{} can\'t have both \'dynamic\' and '
+                        '\'async\'.'.format(json.dumps(filtered_resource))
```
Marc-Andre-Rivet: async and dynamic are incompatible

dash-renderer/src/actions/index.js
```diff
@@ -128,6 +131,8 @@ export function redo() {
             })
         );
 
+        dispatch(setAppIsReady());
```
alexcjohnson: Do undo and redo really need this? They're both returning to a state you've already been in at some point, so doesn't that mean you've already waited for that state to be ready?

dash-renderer/src/APIController.react.js
alexcjohnson: What about on `children` callbacks?

In terms of what kind of layout crawling is required and when it seems to me it matches `applyPersistence` - which also has to crawl the new layout - or layout chunk - in order to find potential matches. We call that immediately above this `setAppIsReady`, but also down in `updateOutput->doUpdateProps`:

https://github.com/plotly/dash/blob/d972c02e8b7842ca9ead4e3965143cac40b5b0b6/dash-renderer/src/actions/index.js#L551-L553

The key for performance there is we're able to crawl only the new piece of layout, and it seems like we should be able to do the same here. If there's any chance the previous components haven't completed loading (perhaps there are 2 simultaneously-invoked children callbacks?) it feels like it should be possible to chain the new promise onto the previous one.

dash-renderer/src/actions/index.js
```diff
@@ -176,7 +183,7 @@ function reduceInputIds(nodeIds, InputGraph) {
     /*
      * Create input-output(s) pairs,
      * sort by number of outputs,
-     * and remove redudant inputs (inputs that update the same output)
```
alexcjohnson: re-dud-ant
*adjective*
- Equally bodacious and/or gnarly as the previous dude.

dash-renderer/src/actions/setAppReadyState.js
```diff
@@ -0,0 +1,80 @@
+import {contains, filter, type} from 'ramda';
+import {createAction} from 'redux-actions';
+
+import Registry from './../registry';
+import {getAction} from './constants';
+
+export const SIMPLE_COMPONENT_TYPES = ['String', 'Number', 'Null', 'Boolean'];
+export const isSimpleComponent = component =>
+    contains(type(component), SIMPLE_COMPONENT_TYPES);
```
alexcjohnson: ~`contains`~ -> `includes` https://github.com/ramda/ramda/pull/2452

Do the `export`s here imply you're planning to share this with `TreeContainer`?

dash-renderer/src/actions/setAppReadyState.js
```diff
@@ -0,0 +1,80 @@
...
+
+const isAppReady = layout => {
+    const queue = [layout];
+
+    const res = {};
+
+    /* Would be much simpler if the Registry was aware of what it contained... */
```
alexcjohnson: Would it? You still only want to wait for components in the layout...

dash/resources.py
alexcjohnson: 🐄 moot point if we `black` this, but if you find yourself using `\'`, switch the outer quotes to `"`

Also the error msg would probably be easier to read with the json dump at the end so you don't have to hunt for the error.

dash/resources.py
```diff
@@ -14,48 +14,73 @@ def __init__(self, resource_name):
...
+                if s['async'] is True:
+                    filtered_resource[
+                        'dynamic'
+                    ] = not self.config.eager_loading
+                else:
+                    filtered_resource['dynamic'] = bool(
+                        (s['async'] == 'eager' and
+                            not self.config.eager_loading)
+                        or s['async'] == 'lazy'
+                    )
```
alexcjohnson: The `# pylint: disable=too-many-branches` might be telling us a subroutine would be useful 😅 
Also your `bool` is redudant.

alexcjohnson:
Re:
- Find a place to share the `isLazyComponent...` promise wrapper that both dcc and the table use (for now, thinking of a new `window.xyz` exposed by the renderer)
- Shared `core-js` in renderer, externalized for all other components (renderer to pull `all` instead of `usage`) -- need to coordinate targets

The renderer is in charge of everything; rather than dumping stuff on `window`, can we inject these deps the first time `Registry` encounters any new component suite? Something like:

```js
if (!nsSeen[ns]) {
    const setCommon = window[ns]._setCommon;
    if (setCommon) {
        setCommon({isLazyComponent, coreJs, ramda, ...});
    }
    nsSeen[ns] = 1;
}
```
Hmm, not sure about `core-js`, we're using that as polyfills I guess. But that might be a clean way to handle other common code.

dash-renderer/src/APIController.react.js
Marc-Andre-Rivet: https://github.com/plotly/dash/pull/899/commits/b4e105a59bd6e8239377d1c7737f7f7ae73f4e1b

dash-renderer/src/actions/setAppReadyState.js
Marc-Andre-Rivet: https://github.com/plotly/dash/pull/899/commits/f32edbd8b220e32d895f9a611e24abc5642523f2

dash-renderer/src/actions/setAppReadyState.js
Marc-Andre-Rivet: You're right, it wouldn't.

dash-renderer/src/actions/index.js
Marc-Andre-Rivet: No. Being too defensive here.

dash/resources.py
Marc-Andre-Rivet: https://github.com/plotly/dash/pull/899/commits/91771d1d94ce69a6033f4a7445dbebd6faa3275f

dash-renderer/src/index.js
```diff
@@ -1,6 +1,3 @@
-/* eslint-env browser */
-
-'use strict';
```
Marc-Andre-Rivet: Code packaged by webpack is always strict, this is useless.

Marc-Andre-Rivet:
https://github.com/plotly/dash/pull/899#issuecomment-536863417

The lazy decorator is now being handled by a small package of its own [dash-component-plugins](https://github.com/plotly/dash/tree/exp-dynamic/%40plotly/dash-component-plugins) that exists inside the dash monorepo.

Ramda could certainly be loaded and shared by the renderer in the future to save a maybe a few tens of kilobyes. core-js normally gets bundled up and scoped, I'm confidant we can do something, but would have to explore.

The standalone plugins package does cause some amount of duplicated code but as it's transpiled with the other component repos, the overhead is negligeable (~20 additional lines / package)

@plotly/dash-component-plugins/src/dynamicImport.js
```diff
@@ -0,0 +1,31 @@
+import { lazy } from 'react';
+
+export const decorate = (target, promise) => {
```
alexcjohnson: do you want to just call this `asyncDecorator` if that's how it'll be used?

@plotly/dash-component-plugins/src/index.js
```diff
@@ -0,0 +1,3 @@
+import { decorate, isReady } from './dynamicImport';
+
+export { decorate, isReady };
```
alexcjohnson: or just: `export { decorate, isReady } from './dynamicImport';`

I guess the reason to have separate files (rather than putting everything in `index`) is just preemptive structure for future expansion?

Marc-Andre-Rivet: yes

@plotly/dash-component-plugins/src/dynamicImport.js
Marc-Andre-Rivet: sure

Marc-Andre-Rivet: https://github.com/plotly/dash/pull/899/commits/af3500637e1403dc2719654118c835e72212c4b4

dash-renderer/src/actions/index.js
```diff
@@ -560,6 +568,8 @@ function updateOutput(
             })
         );
 
+        dispatch(setAppIsReady());
```
alexcjohnson: In principle we should be able to do a lot less work than this - only walk the new props, and chain the resulting promise (if any) onto any existing isReady promise.

tests/integration/test_scripts.py
```diff
@@ -0,0 +1,102 @@
...
+    assert (findAsyncPlotlyJs(scripts) is None) is True
+
+    dash_duo.find_element("#btn").click()
+
+    # Give time for the async dependency to be requested (if any)
+    time.sleep(2)
+
+    scripts = dash_duo.driver.find_elements(By.CSS_SELECTOR, "script")
+    assert (findSyncPlotlyJs(scripts) is None) is True
+    assert (findAsyncPlotlyJs(scripts) is None) is False
```
Marc-Andre-Rivet: I'm of two minds about where these tests should live. On the one hand this is about checking behavior for a specific component / package, so it should live there. BUT it specifically cares about the behavior of Dash when a sync/async resource is (not)requested and how it gets loaded in the end.

I'm leaning towards the 2nd concern winning over the first and leaving this here because it will guarantee that as dash evolves, this loading behavior either stays the same or is maintained alongside server changes (e.g. we add prefetch, etc.)

alexcjohnson: Sounds right, let's leave it here.

minor:
```py
assert findSyncPlotlyJs(scripts) is None
assert findAsyncPlotlyJs(scripts) is not None
```

Marc-Andre-Rivet:
devtools were using `dcc.Graph` for prop type errors - started using the DataTable instead as it (a) is not async, (b) has the same type of complex / deeply nested props as the graph

https://github.com/plotly/dash/pull/899/commits/13e274ed258b3cf3c35fd99ac079b14947d7e423

Marc-Andre-Rivet:
added a shimmed test for `notifyObservers`- `isAppReady` https://github.com/plotly/dash/pull/899/commits/7c3edaadea38d82ad2bc264f42ea010f7a540b7a

dash-renderer/src/actions/index.js
```diff
@@ -917,6 +915,8 @@ function updateOutput(
                                 );
                             });
                         }
+
+                        dispatch(setAppIsReady());
```
Marc-Andre-Rivet: This is possibly the smallest change that brings the most performance benefits: only evaluate the ready state if (1) the layout was set or (2) this code here - when a `children` prop is updated

Marc-Andre-Rivet: @alexcjohnson This may not be as granular as what you had in mind but it's still a pretty significant improvement over the previous usage. I'm having a look at the surrounding code and familiarizing myself with it. For now I don't feel comfortable making a more involved change here. This can either be seen as an intermediary step (the tests are all still passing) or final.

Marc-Andre-Rivet: Tested this out against the docs `/all` page - 31 rounds of render occur with `1.4.1` over about 20 seconds with 2138 components being iterated on each prop update. Total execution time of `setAppIsReady` for these 31 executions is under 50ms. I don't see a point optimizing this further at this point even if it is possible to do so.

PR #900, opened 2019-09-03
:wrench: remove map files from release tar

### LINKED ISSUE #810: [Feature Request] Don't include development assest in Python package

The `dash-renderer` sdist on PyPI is quite big large (2.3M compressed, 10M uncompressed) compared to the actual required components. This is due to at least the following files that shouldn't be required:

* `dash_renderer.dev.js` (3.9M)
* `dash_renderer.dev.js.map` (4.5M)
* `dash_renderer.min.js.map` (1.3M)

Don't distribute files for development purposes in release artifacts. This would reduce the size of the package to 1/10 (200K compressed, 720K uncompressed).

As an alternative to not shipping the debug maps at all, you could also distribute them as separate `-dbg` packages. This would enable a user to get debug information but not having these files in all installations. This requires a lot more effort though.

byronz:
This closes #810  

exclude the *.map file for release tar.gz file

the compressed file size is 1.1M

### DIFF
--- a/dash-renderer/MANIFEST.in
+++ b/dash-renderer/MANIFEST.in
@@ -1,4 +1,4 @@
 include package.json
 include digest.json
 include dash_renderer/*.js
-include dash_renderer/*.map
+exclude dash_renderer/*.map


alexcjohnson:
The other piece of #810 was minifying `dash_renderer.dev.js` - which should be OK if our sourcemap is working correctly, right?

byronz:
> The other piece of #810 was minifying `dash_renderer.dev.js` - which should be OK if our sourcemap is working correctly, right?

does minifying mean using the same production bundle for dev.js? or there are other technics I can apply to just reduce the size?

alexcjohnson:
As discussed, the prod bundle still needs the `GlobalErrorContainerPassthrough` transform, but otherwise the two builds should be identical.

byronz:
@alexcjohnson  @Marc-Andre-Rivet 
the firefox will complain about the absence of the source map if the `Enable Source Maps` is enabled (looks like it's the default setting).

Screen Shot 2019-09-04 at 11 08 06 AM
Screen Shot 2019-09-04 at 11 08 15 AM

alexcjohnson:
Can we just fix the file serving code to give a 404 instead of a 500 in this case? Would that make FF happy?

PR #901, opened 2019-09-03
Jchang10 mycustom

jchang10:
Some fairly minor updates. 

## Contributor Checklist

- [X ] I have broken down my PR scope into the following TODO tasks
   -  [X ] task 1. Make mypy happy by keeping property getter and setter consecutive to each other.
   -  [X] task 2. Set cached_property to None when property is set.
- [X ] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))
  Tests should not be required.
### optionals


byronz:
@jchang10  Thank you for this PR, I'm curious to understand more about your usage with `mypy`, are you using `mypy` for the static check of your own App, or it's purely for dash code itself?

jchang10:
hi @byronz. Yeah, you are right. My MYPY comment could be removed. It is a personal preference of mine but does not make sense for all other projects. I will remove it.

Thus, the only really relevant line I wanted to see updated is invalidating cached_property.

I will see how I can change this PR request.

byronz:
hey @jchang10 

to be clear, there is nothing wrong for adopting `mypy` changes, but we would like to know more about your context to understand better the potential benefits for dash itself. e.g. what's the pros and cons you see by applying the static check for your application code? thanks

jchang10:
IMHO. I like using mypy, among other linters, for more or less the standard reasons. My main reasons are:

1. mypy is a personal project for Guido, so it is likely to continue improving over time.
2. Helps with a sort of "code commenting", so later I have more "documentation" of what variable and paramter types are expected to be later.
3. I use VSCode, which integrates very easily with various linters.  Instead of having to compile code like with C/C++, the linters constantly run in the background. CodeLens, highlights potential problem areas automatically for you.

Anyway, those are some of my own reasons. Thanks for all the work on Dash, too! ;)

PR #903, opened 2019-09-04
Generalized persistence

alexcjohnson:
Building generalized prop persistence into the renderer. Engineered somewhat similarly to `uirevision` in plotly.js, but geared more toward saving state across page reloads, user changes are saved and, when the entire component or larger is refreshed with the same original value, these changes are reapplied. 

Components can use this by defining props `persistence`, `persisted_props`, and `persistence_type` - and if necessary the class property `persistenceTransforms`. See the comment at the top of `persistence.js` for details.

Linked PR coming in dash-table as the primary use case.


dash-renderer/src/persistence.js
```diff
@@ -0,0 +1,346 @@
...
+        element,
+        persistence,
+        persisted_props,
+        persistence_type,
+    } = getProps(layout);
+    if (!persistence) {
+        return;
+    }
+
+    forEachObjIndexed((newPropVal, propName) => {
```
Marc-Andre-Rivet: The persisted props will be few vs. the actual props of the components. Maybe we could loop on the `persisted_props` instead and `const newPropVal = newProps[propName]` instead? These actions are user driven, so not in a tight loop, either way, I do not have a strong opinion.

dash-renderer/src/persistence.js
```diff
@@ -0,0 +1,346 @@
...
+        // note: _storage.getItem returns null on missing keys
+        // and JSON.parse(null) returns null as well
+        const gotVal = this._storage.getItem(storePrefix + key);
+        return gotVal === UNDEFINED ? void 0 : JSON.parse(gotVal);
+    }
+
+    setItem(key, value) {
+        const setVal = value === void 0 ? UNDEFINED : JSON.stringify(value);
+        this._storage.setItem(storePrefix + key, setVal);
+    }
```
Marc-Andre-Rivet: I think `setItem` calls might need to be more defensive.

1. Some browsers and some setups will not define `window.localStorage` and/or `window.sessionStorage` (I do not think this is a concern in 2019 anymore... -- everything >= IE8 supports this)
2. Some browsers will expose local/session storage with storage size of 0 byte - user configurations on certain browsers could also cause an issue here
3. `setItem` willthrow an error if the storage space is exceeded - this will disrupt the app's render flow

e.g. Each prop in the loop at https://github.com/plotly/dash/pull/903/files#diff-1f9fdd7834ce2c4e31ca7ce6aa890a1fR203 could be guarded in a `try ... catch` and/or LocalStorage be defaulted to MemoryStorage

More generally, should failure to store be conveyed to the user in debug mode and how?

dash-renderer/src/persistence.js
alexcjohnson: Most of the time `newProps` only has one key, regardless of how many are in the component. Table is a bit of an outlier in that the 11 `derived_*` props all arrive at once, but it also seems to call `setProps` twice in many cases - once with just input prop, and again with all the `derived_*` - so I think even there it's roughly equal cost either way: as I have it, looping over 1 then 11 items; flipping it we'd loop over 6+6 items.

That said, `if(newProps[propName])` is probably much faster than `if(persisted_props.indexOf(propName) !== -1)` and most components other than table will probably only have one or two persisted props. So yeah, I'll make the change. 🎉

dash-renderer/src/persistence.js
```diff
@@ -0,0 +1,346 @@
...
+export function recordUiEdit(layout, newProps) {
+    const {
+        id,
+        props,
+        element,
+        persistence,
+        persisted_props,
+        persistence_type,
+    } = getProps(layout);
+    if (!persistence) {
```
Marc-Andre-Rivet: Might want to quick-exit on an undefined `id`.

Marc-Andre-Rivet: Or maybe that's handled by https://github.com/plotly/dash/pull/903/files#diff-1f9fdd7834ce2c4e31ca7ce6aa890a1fR167 already, but it's not clear.

dash-renderer/src/persistence.js
```diff
@@ -0,0 +1,346 @@
...
+ *   case is to maintain settings across reloads.
+ *
+ * In addition, if any props require special behavior - particularly for storing
+ * a piece nested inside a larger prop, a component can define a class property
+ * (not a React prop) `persistenceTransforms`. This should be an object:
+ * {
+ *   [propName]: {
+ *     extract: propValue => valueToStore,
+ *     apply: (storedValue, propValue) => newPropValue
+ *   }
```
Marc-Andre-Rivet: Should we only support one extract from a complex prop or should we support an array of extracts? If so, should we consider a way to expose/select these extracts in the `persistence_props`

Marc-Andre-Rivet:
One optimization, one API question, one storage question. Otherwise this looks pretty good!

dash-renderer/src/persistence.js
```diff
@@ -0,0 +1,346 @@
...
+                    )
+                ) {
+                    layoutOut = set(
+                        lensPath(path.concat('props', propName)),
+                        transform.apply(
+                            storage.getItem(newValKey),
+                            props[propName]
+                        ),
+                        layoutOut
+                    );
```
Marc-Andre-Rivet: Concern here: If a component gets re-renderered and a prop has persisted changes, will the same value / object instance be returned as long as (baseValue, persisted) are the same?

e.g. If the table's column names are persisted and this returns a new array of columns on each render, the table's caching will bust on each render and degrade performance significantly.

dash-renderer/src/persistence.js
alexcjohnson: Right, that was the point of the TODO comment below https://github.com/plotly/dash/pull/903/files#diff-1f9fdd7834ce2c4e31ca7ce6aa890a1fR195

Even if we don't have a current use case, the intent would certainly be clearer if we listed the sub-prop specifically. And then we don't lock ourselves out for later.

So what should this look like? Try to match the structure fully (`"columns[i].name"` or `"columns[].name"`) or just a simple namespace of the outer prop (`"columns.name"`)? I think I'd go for the simpler one; I don't think we want to try to automate extract/apply, there are too many cases to consider. What if we want to extract names keyed by column ID, for example?

(Also note to self: if we do this we also need to only store a change if the extracted piece changed vs. the previous value)

Marc-Andre-Rivet: >> namespace of the outer prop (`"columns.name"`)

Yes. That's what I had in mind. 

>> I don't think we want to try to automate extract/apply

I don't think so either. Might be possible for simple cases, but that could be added later if we have a clear target.

dash-renderer/src/persistence.js
alexcjohnson: Right, good call. (1) I think you're right, not an issue. But (2) and (3) are worth protecting against. Also I notice `dcc.Store` needs this treatment -> https://github.com/plotly/dash-core-components/issues/635

(2) seems like all we can do is fall back on memory.
(3) we perhaps have some more options: I can imagine in fact app developers filling up their storage with data from old apps - perhaps we can clear out data associated with IDs that don't exist on the page? Certainly not ideal, as those IDs may pop up later. I suppose we could do better by including a timestamp with the data, and delete the oldest ones first. That's a lot of overhead though for a hopefully rare edge case.

dash-renderer/src/persistence.js
alexcjohnson: I don't *think* this is an issue - `applyPersistence` is not being called as part of the render loop, only on new data coming with the initial layout or from a callback. And that data (aside from clientside callbacks) all went through JSON anyway.

dash-renderer/src/persistence.js
Marc-Andre-Rivet: > (2) seems like all we can do is fall back on memory.

Yes. At least in memory will preserve the behavior while in a session.

> (3) we perhaps have some more options: I can imagine in fact app developers filling up their storage with data from old apps `[...]`

Interesting. All callbacks are defined though.. maybe we could use (dependencyGraph + persisted_props) to identify ids that are not significant for this app? Seems weird..

dash-renderer/src/persistence.js
Marc-Andre-Rivet: Ok. Only re-evaluating If the value changed.

Marc-Andre-Rivet:
Random thought: IndexedDB does not suffer the same limitations as LocalStorage/SessionStorage:
- it's faster (no need to parse / stringify - supports blobs and files natively) and non-blocking (sync / async APIs)
- it allows larger data sets (up to ~50MB)
- it's also widely supported (>= IE 10; at least for the major parts of the API; https://caniuse.com/#feat=indexeddb)
- ...but is also more complex to implement. 

https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API
https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB

This could be added as an additional `persistence_type` later.

dash-renderer/src/persistence.js
Marc-Andre-Rivet: Call me paranoid, tested it out with the table. As you said, only called on prop update.

dash-renderer/src/persistence.js
alexcjohnson: https://github.com/plotly/dash/pull/903/commits/b737a3fc9104134b3dfb57c04ef1ad731b1c10bc

dash-renderer/src/persistence.js
alexcjohnson: Defensive behavior added in https://github.com/plotly/dash/pull/903/commits/83b8cd3cb0910dd79e00f0cc2d1fdbfa25d7768e

- Don't try to access stores until requested; That way the app is ready, so we can `dispatch` any errors to devtools.
- Try to add some reasonably large data to the store - I chose 65K chars (130kB) as a meaningful quantity (more than most persistence needs, unless you store table `data` or something) but still far less than normal webstore limits
- If it fails on initialization, try to clear out the persistence-related keys and try again
- Then give up and use memory if it still fails
- If it fails later on (not during init), don't try to recover or fall back - just report the error to both the console and devtools. We could explore ways to be smarter about this but it's more involved - see TODO https://github.com/plotly/dash/pull/903/commits/83b8cd3cb0910dd79e00f0cc2d1fdbfa25d7768e#diff-1f9fdd7834ce2c4e31ca7ce6aa890a1fR134

dash-renderer/src/persistence.js
alexcjohnson: namespaced in https://github.com/plotly/dash/pull/903/commits/78df979b65d60d21e2027a3474931f0281995174

dash-renderer/src/persistence.js
```diff
@@ -0,0 +1,497 @@
...
+// That should be plenty for common persistence use cases,
+// without getting anywhere near typical browser limits
+const pow = 16;
+function longString() {
+    let s = 'Spam';
+    for (let i = 2; i < pow; i++) {
+        s += s;
+    }
+    return s;
+}
```
Marc-Andre-Rivet: 😀

dash-renderer/src/persistence.js
```diff
@@ -0,0 +1,497 @@
...
+        const gotVal = this._storage.getItem(storePrefix + key);
+        return gotVal === UNDEFINED ? void 0 : JSON.parse(gotVal);
+    }
+
+    /*
+     * In addition to the regular key->value to set, setItem takes
+     * dispatch as a parameter, so it can report OOM to devtools
+     */
+    setItem(key, value, dispatch) {
+        const setVal = value === void 0 ? UNDEFINED : JSON.stringify(value);
```
Marc-Andre-Rivet: First time I've seen someone use `void 0` in a while in JS :)

As ES5 explicitly states that `undefined` is read-only, I don't think we need to protect against reassignment.

IE9 and Edge10 have full ES5 support so I don't think this is a concern.

Was there an additional concern here?

dash-renderer/src/persistence.js
```diff
@@ -0,0 +1,497 @@
...
+            this._storage.setItem(storePrefix + key, setVal);
+        } catch (e) {
+            if (dispatch) {
+                dispatch(err(e));
+            } else {
+                throw e;
+            }
+            // TODO: Should we clear storage here? Or fall back to memory?
+            // Probably not, unless we want to handle this at a higher level
+            // so we can keep all 3 items in sync
```
Marc-Andre-Rivet: I think this is fine for our first implementation of persistence.
1. Make sure there's some space available in storage, clear if not, fallback to memory if still not.
2. Write until there's no space left.

Not sure what the internal error looks like but maybe we want to make it cleaner / user focused, like in the initialization step.

Marc-Andre-Rivet:
@alexcjohnson Looks good, waiting for the tests / changelog / etc. and 📤

dash-renderer/src/persistence.js
alexcjohnson: yeah, just satisfying the linter but you're right, it would be better to just disable that check in the linter.

dash-renderer/src/persistence.js
alexcjohnson: > Not sure what the internal error looks like but maybe we want to make it cleaner / user focused, like in the initialization step.

good point - gave it a nicer error message -> https://github.com/plotly/dash/pull/903/commits/3669df63e7b9806fdafa6e22111a7b05d76762c0

dash-renderer/src/persistence.js
alexcjohnson: 🔪 no-undefined https://github.com/plotly/dash/pull/903/commits/15ca0e5ae6747af2fb8e5c898a89c48f4ddef0a3

.circleci/config.yml
```diff
@@ -156,7 +156,7 @@ jobs:
           name: ️️🏗️ build misc
           command: |
               . venv/bin/activate && pip install --upgrade -e . --quiet && mkdir packages
-              git clone --depth 1 https://github.com/plotly/dash-table.git
+              git clone --depth 1 -b persistence https://github.com/plotly/dash-table.git
```
alexcjohnson: The test failures look as though I didn't get the right branch of table installed. @byronz did I do something wrong here? I want to use https://github.com/plotly/dash-table/pull/566

tests/integration/renderer/test_persistence.py
```diff
@@ -0,0 +1,353 @@
...
+    def report_props(columns, hidden_columns):
+        return 'names: [{}]; hidden: [{}]'.format(
+            ', '.join([col['name'] for col in columns]),
+            ', '.join(hidden_columns or [])
+        )
+
+    return app
+
+
+def rename_and_hide(dash_duo, rename=0, new_name='x', hide=1):
```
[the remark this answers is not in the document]
alexcjohnson: > do you think it makes sense to create a small set of APIs to handle common dash operations in the future?

Sure, but only after we find that we're reusing similar functionality in multiple places. So far this one seems pretty specialized to me.

> I would probably make the new_name more "newer" than a simple 'x'

Good call -> https://github.com/plotly/dash/pull/903/commits/5fa4d559b54f7c7e10209188f4f3673bed76d90e

> I also like to add sometimes randomness to avoid mistakes in tests

As long as the test outcome is completely deterministic (to the best of our abilities) I suppose, but my preference is single-use unusual but not really random values.

alexcjohnson:
multi-key persistence ended up being pretty easy, and actually simplifies our storage usage -> https://github.com/plotly/dash/pull/903/commits/09b93934e23c45b4ac41e3cb4ee5b1edac28dc9f

bg-bi:
I noticed an error in the documentation, please see #1012. Thanks.

Issue #913, opened 2019-09-08
[BUG] dash-generate-components ignore not working


stevej2608:
Hi

The ignore argument on `dash-generate-components` is not working correctly, eg files in the
**\_\_test\_\_** sub-folder are not ignored by:

```
dash-generate-components --ignore=.*?__tests__
```
The ignore pattern is failing due to the regex being quoted when passed into `extract-meta.js`

The one-line fix is in: `dash/extract-meta.js:8`

```
const ignorePattern = new RegExp(process.argv[2]);
```
Should be:

```
const ignorePattern = new RegExp(process.argv[2].split('"').join(''));
```

Cheers, Steve

alexcjohnson:
Closed by @stevej2608 in #915

PR #915, opened 2019-09-09
dash-generate-components crashes on windows

stevej2608:
dash-generate-components crashes on windows. This fix puts quotes around some of the offending command line arguments. This stops dash-generate-components crashing. The quotes placed around the `ignore` attribute need to be removed on entry to `extract-meta.js`

These problems have been reported here:

[Issue with executing dash-component-boilerplate on Windows](https://github.com/plotly/dash-component-boilerplate/issues/74)

and here:

[dash-generate-components ignore not working](https://github.com/plotly/dash/issues/913)

### DIFF
--- a/dash/development/component_generator.py
+++ b/dash/development/component_generator.py
@@ -60,7 +60,7 @@ def generate_components(
     os.environ["NODE_PATH"] = "node_modules"
 
     cmd = shlex.split(
-        "node {} \"{}\" \"{}\" {}".format(
+        'node {} "{}" "{}" {}'.format(
             extract_path, ignore, reserved_patterns, components_source
         ),
         posix=not is_windows,
--- a/dash/development/component_generator.py
+++ b/dash/development/component_generator.py
@@ -59,10 +59,8 @@ def generate_components(
 
     os.environ["NODE_PATH"] = "node_modules"
 
-    fmt = "node {} \"{}\" \"{}\" {}" if is_windows else "node {} {} {} {}"
-
     cmd = shlex.split(
-        fmt.format(
+        "node {} \"{}\" \"{}\" {}".format(
             extract_path, ignore, reserved_patterns, components_source
         ),
         posix=not is_windows,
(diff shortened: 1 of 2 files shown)


dash/development/component_generator.py
```diff
@@ -58,8 +58,11 @@ def generate_components(
     reserved_patterns = "|".join("^{}$".format(p) for p in reserved_words)
 
     os.environ["NODE_PATH"] = "node_modules"
+
+    fmt = "node {} \"{}\" \"{}\" {}" if is_windows else "node {} {} {} {}"
```
alexcjohnson: Thanks @stevej2608 !
Is it necessary that this change is *only* made on Windows? Can't we just *always* quote these arguments? Obviously we (Plotly devs) don't build on Windows very often, so I worry it'll just break again if Windows has special code.

stevej2608:
I agree that allways quoting is the way to go,  I'm not in a position to
test on anything other than Windows so
I went with the least risky solution. If you could give the "always quoted
solution" a go on unix then I'm sure
it will also run OK on Windows.

Cheers.

alexcjohnson:
Confirmed that we can use these quotes in all contexts - in fact they get stripped out immediately by `shlex.split` if we're not on Windows:

```py
>>> shlex.split('a "b" c')
['a', 'b', 'c']
>>> shlex.split('a "b" c', posix=False)
['a', '"b"', 'c']
```

So @stevej2608 if you can update to always add these quotes, I will be happy to merge.

dash/development/component_generator.py
```diff
@@ -58,8 +58,9 @@ def generate_components(
     reserved_patterns = "|".join("^{}$".format(p) for p in reserved_words)
 
     os.environ["NODE_PATH"] = "node_modules"
+
     cmd = shlex.split(
-        "node {} {} {} {}".format(
+        "node {} \"{}\" \"{}\" {}".format(
```
byronz: minor:  a more pythonic way is to use a single quote to avoid the escape 
e.g.` 'node {} "{}"'.format()`

PR #918, opened 2019-09-10
add APIs for dash docs testings

byronz:

## Contributor Checklist

   -  [x] add timeout argument in process runner
   -  [ ] add wait_for_element_by_id

### optionals

### DIFF
--- a/dash/CHANGELOG.md
+++ b/dash/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Added
+
+- [#918](https://github.com/plotly/dash/pull/918) Adds `wait_for_element_by_id` and `visit_and_snapshot` APIs, adds optional `start_timeout` argument for process runner
+
 ### Fixed
 
 - [#915](https://github.com/plotly/dash/issues/915) Fixes `dash-generate-components` on Windows
--- a/dash/CHANGELOG.md
+++ b/dash/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Added
 
-- [#918](https://github.com/plotly/dash/pull/918) Adds `wait_for_element_by_id` and `visit_and_snapshot` APIs, adds optional `start_timeout` argument for process runner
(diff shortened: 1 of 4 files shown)


dash/testing/browser.py
```diff
@@ -173,6 +173,20 @@ def wait_for_element_by_css_selector(self, selector, timeout=None):
             ),
         )
 
+    def wait_for_element_by_id(self, element_id, timeout=None):
```
alexcjohnson: Is it really worth adding this, vs `wait_for_element_by_css_selector('#mi-id')`?

byronz: Yes the id we have like `#wait-/testing` is an invalid CSS selector

alexcjohnson: hmm OK - seems like poor form to use a character in an ID that can't be used in CSS, but I guess sometimes we can't control that.

dash/testing/application_runners.py
```diff
@@ -179,23 +179,34 @@ def __init__(self, keep_open=False, stop_timeout=3):
         self.proc = None
 
     # pylint: disable=arguments-differ
-    def start(self, app_module, application_name="app", port=8050):
+    def start(
```
byronz: @alexcjohnson  played a little bit on the dash-docs side

this is somehow a little break change, but I guess it's not used by anyone in the community.  the raw_command option would give us more flexibility in the future.  like in dash-docs case, our usage should just use python run.py instead a gunicorn or waitress which targets for production.

dash/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## Unreleased
 
+### Added
+
+- [#918](https://github.com/plotly/dash/pull/918) Adds `wait_for_element_by_id` and `visit_and_snapshot` APIs in browser, adds `raw_command` option (it aslo has higher priority than
```
alexcjohnson:
```suggestion
- [#918](https://github.com/plotly/dash/pull/918) Adds `wait_for_element_by_id` and `visit_and_snapshot` APIs in browser, adds `raw_command` option (it also has higher priority than
```

PR #923, opened 2019-09-14
Percy runner

byronz:


dash/testing/browser.py
```diff
@@ -456,6 +456,9 @@ def visit_and_snapshot(self, resource_path, hook_id):
             self.driver.get(self.server_url + resource_path)
             self.wait_for_element_by_id(hook_id)
             self.percy_snapshot(resource_path)
+            assert not self.driver.find_elements_by_css_selector(
+                "div.dash-debug-alert"
+            ), "devtools should not raise an error alert"
```
alexcjohnson: 🏆

alexcjohnson: though, might need an option to disable this in case someone *wants* to snapshot the errors

byronz: Screen Shot 2019-09-13 at 11 54 45 PM
works like a charm

dash/testing/browser.py
```diff
@@ -447,11 +451,15 @@ def reset_log_timestamp(self):
             if entries:
                 self._last_ts = entries[-1]["timestamp"]
 
-    def visit_and_snapshot(self, resource_path, hook_id):
+    def visit_and_snapshot(self, resource_path, hook_id, assert_check=True):
         try:
-            self.driver.get(self.server_url + resource_path)
+            self.driver.get("{}/{}".format(self.server_url, resource_path))
```
alexcjohnson: do we need to worry about someone giving us a `server_url` that ends in a `/`, or a `resource_path` that starts with `/`? Maybe add `.rstrip('/')` and `.lstrip('/')` respectively, just in case?

byronz: agree to add `lstrip` but we should probably not touch `rstrip`, as if someone is testing a self-hosted site, we don't really have a conclude on the last '/'

alexcjohnson: I meant `rstrip` on `self.server_url`, since we’re adding `/` immediately after it anyway... but yes, `lstrip` on `resource_path` like you did.

byronz: 🙈 was not clear after one day work in the garden.

PR #928, opened 2019-09-17
try more

byronz:


.circleci/config.yml
```diff
@@ -201,7 +201,7 @@ jobs:
           name: 🧪 Run Integration Tests
           command: |
             . venv/bin/activate && cd packages && ls -la
-            find . -name "*.gz" | xargs pip install --force-reinstall && pip list | grep dash && cd ..
+            find . -name "*.gz" | xargs pip install --ignore-installed && pip list | grep dash && cd ..
```
byronz: the real problem is this option.   `--force-reinstall` claims to uninstall the old package, but looks like -I  or --ignore-installed should be the preferred way to go

alexcjohnson:
@byronz @Marc-Andre-Rivet this seems a bit unpleasant:

> Upgrading pip with --ignore-installed did not uninstall previous package version first, and due to new file structure, new files did not overwrite the old ones. As a consequence, old files are now orphaned and not picked up by any package; even pip uninstall pip will not remove the orphaned files. One would need to clean them up manually.

https://stackoverflow.com/questions/51913361/difference-between-pip-install-options-ignore-installed-and-force-reinstall

If `--force-reinstall` isn't doing what it's supposed to be doing, should we be explicitly uninstalling everything before reinstalling?

byronz:
@alexcjohnson  I was referring to the same resource attempting the `--force-reinstall` option, but I do find the problem we had during the release is dash-renderer was not **reinstalled** at all, even though we built a perfect dash-renderer tar file with 1.1.0, in the test job, the end result is still 1.0.1.

the concern for --ignore-installed doesn't really have a real impact for the users, and we do know what we are doing in a sense that  we just use it for testing in a sandbox (will never use it again) 

the really unpleasant part about this is the chicken-egg problem we have when building the dash-renderer, the ultimate solution would be replacing those Dash entry points with a node-js tool

alexcjohnson:
> will never use it again

that's the part I'm concerned about... we apparently DO use it again, as the starting cache for the next build. If we ever have a commit that deletes a file, and we depend on it being deleted, will it actually disappear from the cache? Probably not something we'll hit often, if ever, but I wanted to voice the concern.

> the ultimate solution would be replacing those Dash entry points with a node-js tool

🏆

PR #929, opened 2019-09-17
Reword sentence in auto-generated docstrings.

mkcor:
This is a tiny PR inspired from my reviewing https://github.com/plotly/dash-bio/pull/411. Why the distance (expressed by "those")? I believe the meaning we want is that of "these."

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals

### DIFF
--- a/dash/development/_py_components_generation.py
+++ b/dash/development/_py_components_generation.py
@@ -449,8 +449,8 @@ def create_prop_docstring(prop_name, type_object, required, description,
 
     if '\n' in py_type_name:
         return '{indent_spacing}- {name} (dict; {is_required}): ' \
-            '{description}{period}' \
-            '{name} has the following type: {type}'.format(
+            '{description}{period}\n' \
+            '{indent_spacing}{name} is a {type}'.format(
                 indent_spacing=indent_spacing,
                 name=prop_name,
                 type=py_type_name,
@@ -472,11 +472,7 @@ def map_js_to_py_types_prop_types(type_object):
     """Mapping from the PropTypes js type object to the Python type"""
 
     def shape_or_exact():
-        return 'dict containing keys {}.\n{}'.format(
-            ', '.join(
-                "'{}'".format(t) for t in list(type_object['value'].keys())
-            ),
-            'Those keys have the following types:\n{}'.format(
+        return 'dict with keys:\n{}'.format(
                 '\n'.join(
                     create_prop_docstring(
                         prop_name=prop_name,
@@ -487,7 +483,6 @@ def map_js_to_py_types_prop_types(type_object):
                         indent_num=1
                     ) for prop_name, prop in
                     list(type_object['value'].items())))
-            )
 
     return dict(
         array=lambda: 'list',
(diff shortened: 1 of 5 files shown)


alexcjohnson:
hmm, to me it actually sounds a little better with "Those" - but I'd actually prefer us to rewrite it just using (potentially nested) indentation, rather than duplicating the keys (so those/these will disappear entirely). Also we need to rewrap lines so the indentation will be preserved in auto-wrapping contexts. To take a simple example, here's the current `active_cell` from the table:
```
- active_cell (dict; optional): The row and column indices and IDs of the currently active cell. active_cell has the following type: dict containing keys 'row', 'column', 'row_id', 'column_id'.
Those keys have the following types:
  - row (number; optional)
  - column (number; optional)
  - row_id (string | number; optional)
  - column_id (string; optional)
```

This should be cleaned up to look something like:
```
- active_cell (dict; optional): The row and column indices and IDs of the
  currently active cell.
  active_cell is a dict with keys:
  - row (number; optional)
  - column (number; optional)
  - row_id (string | number; optional)
  - column_id (string; optional)
```

mkcor:
It should be "These" because we're referring to those keys we *just* mentioned... ;)
Anyhow, I like the deeper change you're suggesting.

I also agree on the formatting improvement. Speaking of which, I've been wondering whether "active_cell is a dict with keys:" should be "`active_cell` is a dict with keys:" (the way other docstrings are in the SciPy ecosystem). Maybe this has been discussed and decided before.

Leaving that ;) detail aside, I'll rework the PR.

mkcor:
@shammamah I've taken care of Dash Bio components in terms of running `black` over there: https://github.com/plotly/dash-bio/pull/420

mkcor:
I have rebased (getting rid of the original "Those" --> "These" change) and rewritten the new docstrings we want in the tests (TDD, so tests are expected to fail at 97dd9e71f7983b152117c89c8317c0834b06514e).

There are four instances of "Those keys have the following types:" left (two in `test_metadata_conversions.py` and two in `test_flow_metadata_conversions.py`) because I don't see where the previous sentence "... dict containing keys ..." begins. If I don't figure it out, I'll leave it to another PR.

mkcor:
Ok, I'm afraid this is getting unreasonable... This is the kind of diff I'm battling with when testing docstrings:

ouch_diff

One culprit I believe is:

https://github.com/plotly/dash/blob/75cbcd8ee133904cc4531391cee8cced0188cc5b/dash/development/_py_components_generation.py#L487

This value should not be constant but instead inherit, to account for the nested cases.

I'll take care of the general reformatting issue (including `black` in the Python component generation here in Dash, replacing https://github.com/plotly/dash-bio/pull/420) in a new PR, which will be small and beautiful.

dash/development/_py_components_generation.py
```diff
@@ -449,8 +449,8 @@ def create_prop_docstring(prop_name, type_object, required, description,
 
     if '\n' in py_type_name:
         return '{indent_spacing}- {name} (dict; {is_required}): ' \
-            '{description}{period}' \
-            '{name} has the following type: {type}'.format(
+            '{description}{period}\n' \
+            '{indent_spacing}{name} is a {type}'.format(
```
mkcor: This should be:
```suggestion
            '{indent_spacing}  {name} is a {type}'.format(
```

alexcjohnson:
@mkcor I'm afraid we never got around to finishing this and it's gone stale... hopefully if/when we finish #1429 we can take these ideas into account.

PR #930, opened 2019-09-18
elaborate about append_(scripts|css) in serve_locally warning

alexcjohnson:
re: https://github.com/plotly/dash-docs/issues/651 cc @michaelbabyn - look reasonable?


alexcjohnson:
Eh, I'd consider this just noise if mentioned in the changelog. Behavior is unchanged, just better documented.

PR #935, opened 2019-09-20
add --no-cache-dir

byronz:


byronz:
@Marc-Andre-Rivet  this should at least solve the problem that sometimes the built tar files were not installed properly in the integration job.  I would like to see if this help your issues on the lazy loading part after the merge.

byronz:
> This seems like a clear win for our own packages. Possibly not necessary (and a performance hit) for external deps in the `requires` files, but if you think there's any way it would fix any incorrect test builds, that's worth a lot.

the `--no-cache-dir` is a very common practice Dockerfiles commands. Actually, I don't think it will slow the installation speed in CI context. but since we are using some `venv` caching, it's better to make it cleaner.

PR #937, opened 2019-09-23
add two APIs

byronz:

### DIFF
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -434,18 +434,19 @@ class Browser(DashPageMixin):
         ).perform()
 
     def zoom_in_graph_by_ratio(
-        self, elem, start_point_ratio=0.5, zoom_ratio=0.2, compare=True
+        self, elem, start_fraction=0.5, zoom_box_fraction=0.2, compare=True
     ):
-        """zoom out a graph by its dimension ratio
-        default start at middle with a rectangle of 20% of the dimension
+        """zoom out a graph with a zoom box fraction of component dimension
+        default start at middle with a rectangle of 1/5 of the dimension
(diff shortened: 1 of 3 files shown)


dash/testing/browser.py
```diff
@@ -429,6 +433,32 @@ def clear_input(self, elem):
             .send_keys(Keys.DELETE)
         ).perform()
 
+    def zoom_in_graph_by_ratio(
+        self, elem, start_point_ratio=0.5, zoom_ratio=0.2, compare=True
```
alexcjohnson: I'd call `start_point_ratio` -> `start_fraction`. But then it's not quite clear to me what `zoom_ratio` means - is it relative to the top left corner of the element like the starting point is (in which case I'd call it `end_fraction`)? Or is it relative to the starting point (in which case I'd call it `zoom_box_fraction`)? Regardless a short description of each arg, (including `compare`) would be good.

For simple "did it do something" tests on single-subplot graphs, this is fine. If anyone wants to do something more specific or precise though, we could also accept a 2-tuple for each one so you can separately specify x and y of each point.

byronz: the zoom ratio controls the rectangle size when you click and drag to zoom out, the start point decides the left-top point of the rectangle

alexcjohnson: OK good, so let's call them `start_fraction` and `zoom_box_fraction`. At least to me, `ratio` doesn't make it so clear what the denominator is, but `fraction` implies "fraction of the size of the element (that's the first arg)"

byronz:
hmm, the approval disappear after update

alexcjohnson:
:dancer:

PR #939, opened 2019-09-25
Components suite logs

### LINKED ISSUE #938: [Feature Request]
Thanks so much for your interest in Dash!

Before posting an issue here, please check the Dash [community forum](https://community.plot.ly/c/dash) to see if the topic has already been discussed. The community forum is also great for implementation questions. When in doubt, please feel free to just post the issue here :)

after an investigation in recent DCC-CI setup, we identified a long-existing trouble maker, at run time, that the dash app loads `dash_core_components` folder in check-out source code, rather than the expected installed path in `venv`.

there is no logging trace at dash backend side to tell which package we are serving and from where. 

add logging in python backend

byronz:

## Contributor Checklist

- [x ] I have broken down my PR scope into the following TODO tasks
   -  [x] fix

### optionals

this closes #938

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -695,10 +695,8 @@ class Dash(object):
         )[path_in_package_dist.split(".")[-1]]
 
         loader = pkgutil.get_loader(package_name)
-        msg = getattr(
-            self.logger, "info" if self.config.serve_locally else "debug"
-        )
-        msg(
+
+        self.logger.debug(
             "serving -- package: %s resource: %s => location: %s",
             package_name,
             path_in_package_dist,
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -695,7 +695,7 @@ class Dash(object):
         )[path_in_package_dist.split(".")[-1]]
 
         package = sys.modules[package_name]
-        self.logger.warning(
+        self.logger.debug(
             "serving -- package: %s[%s] resource: %s => location: %s",
             package_name,
             package.__version__,
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -678,7 +678,9 @@ class Dash(object):
         })[path_in_package_dist.split('.')[-1]]
 
         loader = pkgutil.get_loader(package_name)
-        self.logger.info(
+        msg = getattr(
+            self.logger, 'info' if self.config.serve_locally else 'debug')
+        msg(
             "serving -- package: %s resource: %s => location: %s",
             package_name,
             path_in_package_dist,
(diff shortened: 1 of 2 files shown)


dash/dash.py
```diff
@@ -677,6 +677,14 @@ def serve_component_suites(self, package_name, path_in_package_dist):
             'map': 'application/json'
         })[path_in_package_dist.split('.')[-1]]
 
+        loader = pkgutil.get_loader(package_name)
+        self.logger.info(
+            "serving package %s resource %s => %s",
+            package_name,
+            path_in_package_dist,
+            loader.path if hasattr(loader, 'path') else loader,
```
byronz: python 2 has different `loader` protocol, so we will see the path only in python3

byronz:
the logs will appear in CI logs only if the test case failed

dash/dash.py
alexcjohnson: [elsewhere](https://github.com/plotly/dash/blob/a42b712e1a88d17400726b5b28ef20a36e826d21/dash/development/base_component.py#L22) we just reference `sys.modules[package_name]` - can we just get the package there and grab its `__path__`?

dash/dash.py
```diff
@@ -678,7 +678,9 @@ def serve_component_suites(self, package_name, path_in_package_dist):
         })[path_in_package_dist.split('.')[-1]]
 
         loader = pkgutil.get_loader(package_name)
-        self.logger.info(
+        msg = getattr(
+            self.logger, 'info' if self.config.serve_locally else 'debug')
```
alexcjohnson: do we actually get here if we're not serving locally?

byronz: Interesting question. the flag that I really want is the `debug` argument but in this function `self.config` is easier to access. 

the logic behind using serve_locally is from here https://github.com/plotly/dash/blob/a42b712e1a88d17400726b5b28ef20a36e826d21/dash/dash.py#L1557-L1560

it's False by default and set to True when debug is True

byronz: I tested with debug=True/False,  the logging level get changed as expected

dash/dash.py
byronz: tried in python 2.7, I think it's basically the same, as the protocol is at the lower level, so you won't even get `__path__` from the loader object

alexcjohnson: Ah I see, you're right, `module.__path__` and `module.__file__` *exist* but they aren't really paths.

PR #940, opened 2019-09-26
Fix spelling

jbampton:

### DIFF
--- a/dash/development/build_process.py
+++ b/dash/development/build_process.py
@@ -147,7 +147,7 @@ class BuildProcess(object):
         os.chdir(self.main)
         run_command_with_process("npm run {}".format(_script))
 
-        logger.info("generate the `__init__.py` from template and verisons")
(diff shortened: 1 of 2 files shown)


dash/development/build_process.py
```diff
@@ -147,7 +147,7 @@ def bundles(self, build=None):
         os.chdir(self.main)
         run_command_with_process("npm run {}".format(_script))
 
-        logger.info("generate the `__init__.py` from template and verisons")
```
alexcjohnson: haha I'd noticed that one before. First time I read it as venison 🦌

jbampton:
Thanks 👍

byronz:
@jbampton  the failures on CI is a known issue with the percy runner. It just needs some timeout to kill the sessions, which unfortunately we don't have any control. We won't get the 400s after few hours.

PR #944, opened 2019-09-27
Persistence fix

alexcjohnson:
Edge case discovered while writing persistence docs: failure when enabling/disabling persistence on an existing component.

As discussed with @byronz I also generalized some of our test methods to accept either an element or a CSS selector.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,6 @@ jobs:
           name: ️️🏗️ build core
           command: |
               . venv/bin/activate && pip install --no-cache-dir --upgrade -e . --progress-bar off && mkdir packages
-              python setup.py sdist && mv dist/* packages/
               cd dash-renderer && renderer build && python setup.py sdist && mv dist/* ../packages/ && cd ..
               git clone --depth 1 https://github.com/plotly/dash-core-components.git
               cd dash-core-components && npm install --ignore-scripts && npm run build && python setup.py sdist && mv dist/* ../packages/  && cd ..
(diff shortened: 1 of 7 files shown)


dash-renderer/src/persistence.js
```diff
@@ -280,38 +280,41 @@ const getValsKey = (id, persistedProp, persistence) =>
...
     const element = Registry.resolve(layout);
-    const persisted_props =
-        props.persisted_props || element.defaultProps.persisted_props;
-    const persistence_type =
-        props.persistence_type || element.defaultProps.persistence_type;
-    if (!persisted_props || !persistence_type) {
-        return {props};
-    }
-    return {id, props, element, persistence, persisted_props, persistence_type};
+    const getVal = prop => props[prop] || (element.defaultProps || {})[prop];
```
Marc-Andre-Rivet: `null` and `0` could be valid non-default values

Marc-Andre-Rivet: Ah ok, just used for persisted_props / persistence_type

.circleci/config.yml
```diff
@@ -200,10 +199,17 @@ jobs:
...
           at: ~/dash
       - run:
-          name: 🧪 Run Integration Tests
+          name: ️️🏗️  Install packages
           command: |
             . venv/bin/activate && cd packages && ls -la
-            find . -name "*.gz" | xargs pip install --no-cache-dir --ignore-installed && pip list | grep dash && cd ..
+            find . -name "*.gz" | xargs pip install --no-cache-dir --ignore-installed && cd ..
+            sed -i '/dash/d' requires-install.txt
+            pip install --no-cache-dir --ignore-installed .
```
alexcjohnson: FYI @byronz - the issue I was having in this PR is related to the original one that led to the `sed` call during `build-core`: sometimes (seemingly at random) the `dash-renderer` we prepared in `packages` was being overwritten by a published one from PyPI - so I used essentially the same solution, but extended to `dash-renderer` by using the pattern `/dash/d` rather than `/dash-/d`.

Also I skipped creating a tarball for dash itself because (a) we already have everything we need here, dash has no build artifacts, and (b) that made it easier to modify its requirements.

tests/integration/renderer/test_persistence.py
```diff
@@ -354,3 +364,102 @@ def reset_names(n):
...
+    @app.callback(Output('out', 'children'), [Input('persisted2', 'value')])
+    def set_out(val):
+        return val
+
+    dash_duo.start_server(app)
+
+    dash_duo.wait_for_text_to_equal('#out', 'a')
+
+    dash_duo.find_element('#persistence-val').send_keys('s')
+    time.sleep(0.2)
```
Marc-Andre-Rivet: I see you've removed all the other `sleep`. Is this one special / still required?

alexcjohnson: yes - I'm expecting the same text both before and after the `send_keys`, so there's really nothing to wait for; I inserted an explicit wait just to ensure the keypress had fully registered.

PR #947, opened 2019-10-01
Combine dash and dash-renderer changelogs into one at the repo root

alexcjohnson:
The two separate changelogs in the `dash/` and `dash-renderer/` folders were hard to discover, and unnecessarily disjoint. Also we have various links to the changelog in the root of the repo in community posts; these are currently broken. (PR coming in dash-docs though to update changelog links there, which *do* point to the current nested files)

In addition to combining the two files, I tried to standardize the style in a few ways:
- Subheadings should be ordered by rough level of interest to upgrading users:
  - `Added`
  - `Removed`
  - `Changed`
  - `Fixed`
- Start each entry with a link to the PR making the change.
- Write in the active present tense ("Fix a bug...") rather than past ("Fixed") or passive ("Fixes"). But **clarity is more important** than this rule.
- Release numbers get heading level 2 `##`, subheadings get level 3 `###`


alexcjohnson:
Matching dash-docs PR https://github.com/plotly/dash-docs/pull/657

PR #948, opened 2019-10-01
cwd for running dashr tests - auto and override

alexcjohnson:
@rpkyle with any luck this will let you remove that `os.chdir` call in `test_clientside.py`.

Sets the cwd for R test apps during the `subprocess.Popen` call - so it won't affect any other tests. If the app was provided as a string, it tries to find the directory of the calling Python file and use that as cwd. If it was provided as a path to a file it just uses the directory of that file. Either way you can override this behavior with `dashr.start_server(app, cwd='/some/path/')`

### optionals

### DIFF
--- a/dash/testing/application_runners.py
+++ b/dash/testing/application_runners.py
@@ -257,6 +257,7 @@ class RRunner(ProcessRunner):
             # app is already a file in a dir - use that as cwd
             if not cwd:
                 cwd = os.path.dirname(app)
+                logger.info("RRunner inferred cwd from app path: %s", cwd)
         else:
             path = (
                 "/tmp/app_{}.R".format(uuid.uuid4().hex)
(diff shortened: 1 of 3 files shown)


dash/testing/application_runners.py
```diff
@@ -248,28 +249,39 @@ def __init__(self, keep_open=False, stop_timeout=3):
...
 
             with open(path, "w") as fp:
                 fp.write(app)
 
             app = path
 
+            # try to find the path to the calling script to use as cwd
+            if not cwd:
+                for entry in inspect.stack():
+                    if "/dash/testing/" not in entry[1].replace("\\", "/"):
```
byronz: a little bit scared about this hack here

dash/testing/application_runners.py
```diff
@@ -248,28 +249,39 @@ def __init__(self, keep_open=False, stop_timeout=3):
...
+                        cwd = os.path.dirname(os.path.realpath(entry[1]))
```
byronz: or at least we should log the smarter decision and let the user know the result.

dash/testing/application_runners.py
alexcjohnson: Yes, it's ugly. Do you see an alternative? I suppose if we're too worried about it we could document the official usage pattern as "start your server with `dashr.run_server(app, cwd=os.path.dirname(__file__))`" but it would be nice if we can avoid that.

byronz: can we just use the `os.path.dirname(__file__)` by default (in case `cwd=None`) ?

dash/testing/application_runners.py
alexcjohnson: added log messages -> https://github.com/plotly/dash/pull/948/commits/8e287c38cace9f9bead0b26c315df61f600b19e7

dash/testing/application_runners.py
alexcjohnson: but we don't want `__file__` from here, we want it from the user's test module

PR #950, opened 2019-10-02
Attempt to standardize Python Docstrings to PEP 257.

jbampton:
Remove whitespace.  Full stop at end of first sentence. Use triple double quotes.  Capital letter to start Docstring. Add missing docstring params.

https://www.python.org/dev/peps/pep-0257/

Found this great Docstring autoformater called `docformatter`.

https://pypi.org/project/docformatter/ 

`docformatter --in-place --recursive tests` is an example run 👍 

Minor fix to remove unneeded parens.

Another minor fix to remove the first and last blank lines from HTML in view source mode.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -87,8 +87,7 @@ no_update = _NoUpdate()
 # pylint: disable=too-many-instance-attributes
 # pylint: disable=too-many-arguments, too-many-locals
 class Dash(object):
-    """
-    Dash is a framework for building analytical web applications.
+    """Dash is a framework for building analytical web applications.
     No JavaScript required.
 
     If a parameter can be set by an environment variable, that is listed as:
@@ -347,9 +346,7 @@ class Dash(object):
             self.init_app()
 
     def init_app(self, app=None):
-        """
-        Initialize the parts of Dash that require a flask app
-        """
+        """Initialize the parts of Dash that require a flask app."""
         config = self.config
 
         if app is not None:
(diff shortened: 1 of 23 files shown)


dash/dash.py
```diff
@@ -87,9 +87,8 @@ class _NoUpdate(object):
 # pylint: disable=too-many-instance-attributes
 # pylint: disable=too-many-arguments, too-many-locals
 class Dash(object):
-    """
-    Dash is a framework for building analytical web applications.
-    No JavaScript required.
+    """Dash is a framework for building analytical web applications. No
+    JavaScript required.
```
alexcjohnson: "No JavaScript required" loses a bit of its punch with the "No" off on a separate line 😏

tests/unit/development/metadata_test.py
```diff
@@ -4,46 +4,46 @@
 
 
 class Table(Component):
-    """A Table component.
-This is a description of the component.
-It's multiple lines long.
```
alexcjohnson: Don't touch this file, as it's intended to match the output of our component generators. Unless you want to fix the component generator to match - but I'd prefer you not, in fact, as we have some other work in the wings to update that. Anyway that's a bigger project...

alexcjohnson:
@jbampton this looks good. There's the position of "No" https://github.com/plotly/dash/pull/950#discussion_r330695315, then I don't know if you can see it in circleci but for some reason this PR triggered a pylint error (spurious, in my opinion, it's complaining about duplicate code - between the signature of a function and its invocation 🙄 ). If you change the line in both `.pylintrc` and `.pylintrc37` from:
```
min-similarity-lines=4
```
to:
```
min-similarity-lines=10
```
then the lint tests should pass again. And if you merge the tip of `dev` the [spurious failure](https://github.com/plotly/dash/pull/952) in `test-37` should also go away.

Issue #956, opened 2019-10-04
[Feature Request] Inline clientside callbacks


alexcjohnson:
Right now our clientside callbacks need to reference a separate file - typically loaded in via assets - where the functions are defined in a special namespace:

```js
window.dash_clientside = Object.assign({}, window.dash_clientside, {
    large_params_function: function(largeValue1, largeValue2) {
        return someTransform(largeValue1, largeValue2);
    }
});
```

Then in the app you reference these functions by namespace and name:

```py
app.clientside_callback(
    ClientsideFunction('clientside', 'large_params_function'),
    Output('out-component', 'value'),
    [Input('in-component1', 'value'), Input('in-component2', 'value')]
)
```

From a user standpoint, at least for simple functions it would be convenient to be able to provide the function just as a string with the callback definition. We would then package these up with some automatic namespace and naming, either as a separate endpoint or just a special case of `_dash-component-suites`, and reference it in the generated index alongside all the other scripts.

```py
app.clientside_callback(
    """
    function(largeValue1, largeValue2) {
        return someTransform(largeValue1, largeValue2);
    }
    """,
    Output('out-component', 'value'),
    [Input('in-component1', 'value'), Input('in-component2', 'value')]
)
```

jjaraalm:
Taking this one step further, it would be great if we could use a transpiler like [Transcrypt](https://github.com/QQuick/Transcrypt) to write the callback in python then convert it to js, Then the API for clientside callbacks and server callbacks could be integrated like

```python
@app.callback(
    Output('out', 'value'),
    [Input('in', 'value')],
    clientside=True)
def callback(input):
    return input in range(5)
```

The simple and dumb way to do this would be to use `inspect` to dump the source to a temporary file, transpile it, then load and deploy the js. We would probably want to do some preprocessing on the source to make sure the callback is "simple." 

The entire global context probably shouldn't be included in the transpiling, but we could allow users to pass in a context so that calls like

```python
@app.callback(
    Output('out', 'value'),
    [Input('in', 'value')],
    clientside=True,
    context={'i': 4})
def callback(input):
    return input in range(i)
```
also work. In this case `i` is always a fixed constant, but this would be helpful if/when programatically generating callbacks.

Marc-Andre-Rivet:
@alexcjohnson Closing this as fixed. Let's open a new issue if there's a need for a follow up.

PR #967, opened 2019-10-16
Inline clientside callbacks

jjaraalm:
Addresses at least part of #956 and adds support for writing JS clientside callbacks in python source.

Adds explicit JS source injection through a new keyword argument to `Dash.clientside_callback`:

```python
clientside_callback(
    ClientsideFunction('namespace_to_inject_to', 'function_to_inject'),
    Output('output', 'value'),
    [Input('input', 'value')],
    source="""
    function(value) {
        console.log(value);
        return parseInt(value, 10) + 1;
    }
    """
)
```

Also adds source injection via the `Dash.callback` decorator. This probably should be explained or documented better so that users don't think they can write arbitrary python code here. Eventually I would like to add support for transpiling python -> JS, but this is just a stepping stone.

```python
@callback(
    Output('output', 'value'),
    [Input('input', 'value')],
    clientside=True
)
def native_js_callback(value): 
    # Interior source is JS that will be added to the 
    # dash_clientide._inline_callbacks namespace.
    # Single-line comments are properly escaped.
    console.log(value);
    return parseInt(value, 10) + 1;
```

## Contributor Checklist

   -  [X] Add inline clientside callbacks to `Dash.clientside_callback`
   -  [X] Add decorated clientside callbacks to `Dash.callback`

### optionals


alexcjohnson:
@jjaraalm you're on a roll 🏎

I very much like your implementation with `inline_scripts`, that's simpler than what I was envisioning 🎉 - that said if it's going to be in `config` - and thereby intentionally exposed to users - we would need to (a) include it in the `Dash` constructor args, and (b) document it fully. For now, unless there are clear use cases for users to add other things to it, let's keep it private and out of `config`, just set it as something like `self._inline_scripts`.

I'd like to see if we can get rid of the `source` kwarg, and just replace the `ClientsideFunction` first arg with the function source string, as proposed in #956. This would require choosing an implicit namespace and function name. Is there a reason to want explicit names? I suppose in principle that could allow multiple callbacks to use the same function, but that feels to me like an unnecessary optimization that will lead to more confusing code, especially since such sharing would only generally be useful for very short functions. I'd propose a namespace like `_dashprivate`, and the name could be just the (first) output id and prop, joined with maybe a double underscore, ie `output__value`?

I'm not a fan of writing JS code directly in a Python function body, and I don't see how it can be a stepping stone without later becoming a breaking change. If it's alright with you, let's leave this flavor out unless and until we're ready to do the transpiled version.

jjaraalm:
Thanks for the feedback. I wasn't thinking so much about the implications of putting it in `config`, I just liked the symmetry with `external_scripts`. I could see why some people might use it in the constructor (to avoid writing simple JS assets), but I agree it probably shouldn't be there, I'll fix that. 

I agree with your assessment of the utility of namespace/function name for inline scripts. Really I only see that being used for intricate, overly complex, and probably not well-documented callback ... which I'm guilty of sometimes haha.

I think JS in python feels dirty, but that said I dislike code-in-strings even more. It's more clear that it's a foreign language, but also there's no easy way to get syntax highlighting or linting in editors (that I know of). I've always found code-in-strings to be a pain to write and debug. I understand your points though, and we can leave it out.

jjaraalm:
I decided to create a new namespace prefixed by `_dashprivate_` for each output id. This is to prevent the highly unlikely case where someone creates a custom component with horrible prop names then decides to use even worse id names which result in a conflict.

dash/dash.py
```diff
@@ -1161,13 +1168,14 @@ def clientside_callback(
         (JavaScript) function instead of a Python function.
 
         Unlike `@app.calllback`, `clientside_callback` is not a decorator:
-        it takes a
+        it takes either a
         `dash.dependencies.ClientsideFunction(namespace, function_name)`
         argument that describes which JavaScript function to call
         (Dash will look for the JavaScript function at
-        `window[namespace][function_name]`).
+        `window.dash_clientside[namespace][function_name]`), or it may take
```
alexcjohnson: good catch! 🔬

dash/dash.py
```diff
@@ -1178,28 +1186,75 @@ def clientside_callback(
...
+
+            namespace = '_dashprivate_{}'.format(out0.component_id)
+            function_name = '{}'.format(out0.component_property)
+
+            self._inline_scripts.append(
+                """
+                if (!window.dash_clientside) {{
+                    window.dash_clientside = {{}};
+                }}
+                window.dash_clientside["{0}"] = Object.assign(
```
alexcjohnson: Oh hmm, not sure this is safe since we still support IE11. Looks like right now it works as our current bundle pipeline includes a polyfill for `Object.assign`, but especially with all the performance work @Marc-Andre-Rivet is doing I'm not sure I trust us to always do it that way, could potentially be converted to transpiling and this would break.

Also while we're on the topic of pathological ids and props (good call putting the id in the namespace!), I think `"` is technically allowed in either one, so should be escaped.

Depending on how golf-like you want to be, could be something like
```
var clientside = window.dash_clientside = window.dash_clientside || {{}};
var ns = clientside["{0}"] = clientside["{0}"] || {{}};
ns["{1}"] = {2};
```

jjaraalm: works for me!

alexcjohnson:
Alright, Looks good to me! Just needs a changelog (and updating so it doesn't conflict with your other PR 😏) and we'll be ready to go.

Marc-Andre-Rivet:
@alexcjohnson @jjaraalm Looking into the [DashR parity implementation](https://github.com/plotly/dashR/issues/135) and came to the conclusion that we wanted to prepend the namespace with `_dashprivate_` -- glad to see you've already thought of everything! This looks great. As previously stated, just needs a changelog and an update 🎉

jjaraalm:
Sorry guys, was traveling. This should take care of it I think.

ieipi:
I've updated to dash=1.7 and want to use the new inline callbacks feature, but got the following error:

TypeError: clientside_callback() got an unexpected keyword argument 'source'
(base)

alexcjohnson:
@ieipi the final implementation is not as described in the lead comment for the PR. See https://community.plot.ly/t/dash-v1-7-0-released/31961 for an example.

PR #969, opened 2019-10-16
Clean up react devtools warnings on our devtools

alexcjohnson:
and remove partially-implemented resolveError, which was making everything
much more complicated, especially when I tried to clean up ComponentErrorBoundary


dash-renderer/src/components/error/ComponentErrorBoundary.react.js
```diff
@@ -13,9 +12,14 @@ class UnconnectedComponentErrorBoundary extends Component {
             myID: props.componentId,
             myUID: uniqid(),
             oldChildren: null,
+            hasError: false,
         };
     }
 
+    static getDerivedStateFromError(_) {
+        return {hasError: true};
+    }
```
Marc-Andre-Rivet: Not finding code calling this function

alexcjohnson: https://reactjs.org/docs/react-component.html#static-getderivedstatefromerror

PR #973, opened 2019-10-22
Improve resource caching

Marc-Andre-Rivet:
Currently Dash uses caching on the browser side combined with query string cache busting (e.g. `?v=141&m=29384720349`).

This is generally considered an anti-pattern as browsers have sometimes problems dealing with the query string correctly and cause the cache to always be bust. This is our case. When checking caching on `dash.plot.ly` it turns out that no JS from `_dash-component-suites` is ever cached (tested on latest Chrome). Apparently we do not set `cache-control max-age`.

To keep with the existing behavior but also providing a second caching mechanism, adding suport for `ETag` and `If-None-Match` provides the BE with the ability to tell the browser that the cached version of the resource is indeed the latest version. When the tag received matches the resource's tag, a 304 is returned instead of a 200+data, significantly reducing network traffic and load time. This won't impact debug/dev as the resource updating will update the etag.

Fingerprinting for async artifacts has also been added as part of this PR. ETag is now a fallback for resources that are not fingerprinted, whatever the reason.

-----

(before) against `dash.plot.ly` 1st and 2nd run, all resources return `200` and are fetched, no caching

(after) against local dash-docs 1st run (200) and 2nd run (304, or loaded from cache with fingerprinting) -- the calls are still being made to the server but the server confirms the cached resource is not stale.

On local dash-docs this is enough to reduce the time to DOMContentLoaded for `http://127.0.0.1:8000/datatable` down from 600ms to 300ms without throttling, from ~11 seconds to ~2.5 seconds with throttling the network speed to `Fast 3G` and from ~35 to ~6 seconds on `Slow 3G`.

** Having trouble uploading animated gifs with the browser's behavior at the moment - will try again later

### DIFF
--- a/@plotly/webpack-dash-dynamic-import/src/index.js
+++ b/@plotly/webpack-dash-dynamic-import/src/index.js
@@ -1,4 +1,16 @@
-const resolveImportSource = `\
+const fs = require('fs');
+
+function getFingerprint() {
+    const package = fs.readFileSync('./package.json');
+    const packageJson = JSON.parse(package);
+
+    const timestamp = Math.round(Date.now() / 1000);
+    const version = packageJson.version.replace(/[.]/g, '_');
+
+    return `"v${version}m${timestamp}"`;
+}
+
+const resolveImportSource = () => `\
 Object.defineProperty(__webpack_require__, 'p', {
     get: (function () {
         let script = document.currentScript;
@@ -15,7 +27,16 @@ Object.defineProperty(__webpack_require__, 'p', {
             return url;
         };
     })()
-});`
+});
+
+const __jsonpScriptSrc__ = jsonpScriptSrc;
+jsonpScriptSrc = function(chunkId) {
+    const srcFragments = __jsonpScriptSrc__(chunkId).split('.');
+    srcFragments.splice(-1, 0, ${getFingerprint()});
+
+    return srcFragments.join('.');
+}
+`
 
 class WebpackDashDynamicImport {
     apply(compiler) {
@@ -23,7 +44,7 @@ class WebpackDashDynamicImport {
             compilation.mainTemplate.hooks.requireExtensions.tap('WebpackDashDynamicImport > RequireExtensions', (source, chunk, hash) => {
                 return [
                     source,
-                    resolveImportSource
+                    resolveImportSource()
                 ]
             });
         });
--- a/@plotly/webpack-dash-dynamic-import/src/index.js
+++ b/@plotly/webpack-dash-dynamic-import/src/index.js
@@ -32,7 +32,7 @@ Object.defineProperty(__webpack_require__, 'p', {
 const __jsonpScriptSrc__ = jsonpScriptSrc;
 jsonpScriptSrc = function(chunkId) {
     const srcFragments = __jsonpScriptSrc__(chunkId).split('.');
-    srcFragments.splice(-1, 0, ${getFingerprint()});
+    srcFragments.splice(1, 0, ${getFingerprint()});
 
     return srcFragments.join('.');
 }
--- a/@plotly/webpack-dash-dynamic-import/src/index.js
+++ b/@plotly/webpack-dash-dynamic-import/src/index.js
@@ -1,0 +1,33 @@
+const resolveImportSource = `\
+Object.defineProperty(__webpack_require__, 'p', {
+    get: (function () {
+        let script = document.currentScript;
+        if (!script) {
+            /* Shim for IE11 and below */
+            /* Do not take into account async scripts and inline scripts */
+            const scripts = Array.from(document.getElementsByTagName('script')).filter(function(s) { return !s.async && !s.text && !s.textContent; });
+            script = scripts.slice(-1)[0];
+        }
+
+        var url = script.src.split('/').slice(0, -1).join('/') + '/';
+
+        return function() {
+            return url;
+        };
+    })()
+});`
+
+class WebpackDashDynamicImport {
+    apply(compiler) {
+        compiler.hooks.compilation.tap('WebpackDashDynamicImport', compilation => {
+            compilation.mainTemplate.hooks.requireExtensions.tap('WebpackDashDynamicImport > RequireExtensions', (source, chunk, hash) => {
+                return [
+                    source,
+                    resolveImportSource
+                ]
+            });
+        });
+    }
+}
+
+module.exports = WebpackDashDynamicImport;
--- a/@plotly/webpack-dash-dynamic-import/src/index.js
+++ b/@plotly/webpack-dash-dynamic-import/src/index.js
@@ -31,10 +31,13 @@ Object.defineProperty(__webpack_require__, 'p', {
 
 const __jsonpScriptSrc__ = jsonpScriptSrc;
 jsonpScriptSrc = function(chunkId) {
-    const srcFragments = __jsonpScriptSrc__(chunkId).split('.');
-    srcFragments.splice(1, 0, ${getFingerprint()});
+    const srcFragments = __jsonpScriptSrc__(chunkId).split('/');
+    const fileFragments = srcFragments.slice(-1)[0].split('.');
 
-    return srcFragments.join('.');
+    fileFragments.splice(1, 0, ${getFingerprint()});
+    srcFragments.splice(-1, 1, fileFragments.join('.'))
+
+    return srcFragments.join('/');
 }
 `
 
--- a/@plotly/webpack-dash-dynamic-import/src/index.js
+++ b/@plotly/webpack-dash-dynamic-import/src/index.js
@@ -11,15 +11,25 @@ function getFingerprint() {
 }
 
 const resolveImportSource = () => `\
+const getCurrentScript = function() {
+    let script = document.currentScript;
+    if (!script) {
+        /* Shim for IE11 and below */
+        /* Do not take into account async scripts and inline scripts */
+        const scripts = Array.from(document.getElementsByTagName('script')).filter(function(s) { return !s.async && !s.text && !s.textContent; });
+        script = scripts.slice(-1)[0];
+    }
+
+    return script;
+};
+
+const isLocalScript = function(script) {
+    return /\/_dash-components-suite\//.test(script.src);
+};
+
 Object.defineProperty(__webpack_require__, 'p', {
     get: (function () {
-        let script = document.currentScript;
-        if (!script) {
-            /* Shim for IE11 and below */
-            /* Do not take into account async scripts and inline scripts */
-            const scripts = Array.from(document.getElementsByTagName('script')).filter(function(s) { return !s.async && !s.text && !s.textContent; });
-            script = scripts.slice(-1)[0];
-        }
+        let script = getCurrentScript();
 
         var url = script.src.split('/').slice(0, -1).join('/') + '/';
 
@@ -31,14 +41,23 @@ Object.defineProperty(__webpack_require__, 'p', {
 
 const __jsonpScriptSrc__ = jsonpScriptSrc;
 jsonpScriptSrc = function(chunkId) {
-    const srcFragments = __jsonpScriptSrc__(chunkId).split('/');
+    let script = getCurrentScript();
+    let isLocal = isLocalScript(script);
+
+    let src = __jsonpScriptSrc__(chunkId);
+
+    if(!isLocal) {
+        return src;
+    }
+
+    const srcFragments = src.split('/');
     const fileFragments = srcFragments.slice(-1)[0].split('.');
 
     fileFragments.splice(1, 0, ${getFingerprint()});
     srcFragments.splice(-1, 1, fileFragments.join('.'))
 
     return srcFragments.join('/');
-}
+};
 `
 
 class WebpackDashDynamicImport {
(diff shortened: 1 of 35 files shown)


dash/dash.py
```diff
@@ -711,10 +711,19 @@ def serve_component_suites(self, package_name, path_in_package_dist):
             package.__path__,
         )
 
-        return flask.Response(
+        response = flask.Response(
             pkgutil.get_data(package_name, path_in_package_dist),
             mimetype=mimetype,
         )
+        response.add_etag()
+        (tag,) = response.get_etag()[:1]
```
Marc-Andre-Rivet: not sure what the preferred syntax is for unpacking only part of a tuple

alexcjohnson:
```suggestion
        tag = response.get_etag()[0]
```

chriddyp:
Nice!

alexcjohnson:
Wow, good catch! Did this change recently? I could swear I saw this working not too long ago...

I wonder though, wouldn't it be pretty easy to include the same mod time info we do now, just as part of the path rather than as a query string, and simply strip it back out when serving the file?

ie right now we take a path like:
`dash_core_components/dash_core_components.min.js`
and turn it into a url like:
`/_dash-component-suites/dash_core_components/dash_core_components.min.js?v=1.3.1&m=1571759511`

The query string gets automatically dropped by the [route](https://github.com/plotly/dash/blob/c4db2efd24c5a1e0f75a0ca262ba5aeeb0ff3c7f/dash/dash.py#L404-L411)
`"{}_dash-component-suites/<string:package_name>/<path:path_in_package_dist>"`

But without changing the file names/paths on disk at all we could just construct urls like:
`/_dash-component-suites/dash_core_components/dash_core_components.min.v1_3_1m1571759511.js`
and have the [route handler](https://github.com/plotly/dash/blob/c4db2efd24c5a1e0f75a0ca262ba5aeeb0ff3c7f/dash/dash.py#L678) strip it out before looking for the file
```py
path_parts = path_in_package_dist.split(".")
actual_path = ".".join(path_parts[:-2] + [path_parts[-1]])
```

dash/dash.py
byronz: we should avoid the unpacking here as it's a little bit tricky to handle the syntax in python 2 and 3. with pure python3 it's easier.

byronz:
```suggestion
tag, _ = response.get_etag()
```

Marc-Andre-Rivet:
Based on feedback / discussion with @alexcjohnson, expending the scope to cover:

1. moving from `query string` to `fingerprinting` for both standard and async js resources
2. updating `@plotly/webpack-dash-dynamic-import` to fingerprint the async resources in accordance with the DashPy expectations. This will require the DashR backend to implement these changes too @rpkyle 
3. DashPy checks for fingerprinting information through pattern matching - if there is a fingerprint, use `cache-control max-age` with a value of 1 year for caching purposes, otherwise, use ETag/If-None-Match as a general catch-all mechanism

Libraries using the old version of `@plotly/webpack-dash-dynamic-import` will need to update their usage to benefit from the `cache-control`, otherwise `ETag` will be used.

@plotly/webpack-dash-dynamic-import/src/index.js
```diff
@@ -15,15 +27,24 @@ Object.defineProperty(__webpack_require__, 'p', {
...
-});`
+});
+
+const __jsonpScriptSrc__ = jsonpScriptSrc;
+jsonpScriptSrc = function(chunkId) {
+    const srcFragments = __jsonpScriptSrc__(chunkId).split('.');
+    srcFragments.splice(-1, 0, ${getFingerprint()});
+
+    return srcFragments.join('.');
+}
```
Marc-Andre-Rivet: Override webpack's async chunks path resolution; add fingerprint information.

dash/dash.py
```diff
@@ -541,12 +541,15 @@ def _relative_url_path(relative_package_path="", namespace=""):
...
                 self.config.requests_pathname_prefix,
                 namespace,
-                relative_package_path,
-                importlib.import_module(namespace).__version__,
+                '.'.join(relative_package_path.split('.')[:-1]),
+                importlib.import_module(namespace).__version__.replace(
+                    '.', '_'
+                ),
                 modified,
+                '.'.join(relative_package_path.split('.')[-1:]),
```
Marc-Andre-Rivet: new pattern!

dash/dash.py
```diff
@@ -676,6 +679,18 @@ def _generate_meta_html(self):
...
+        res = re.match(
+            r'^(.*)[.]v\d+_\d+_\d+(-\w+[.]?\d+)?m\d{10}([.]js)$',
+            path_in_package_dist,
+        )
+
+        fingerprint = res is not None
+        # Resolve real resource name from fingerprinted resource path
+        path_in_package_dist = (
+            res[1] + res[3] if res is not None else path_in_package_dist
+        )
```
Marc-Andre-Rivet: If the resource was requested with the correct pattern (from dash filling the html page or from async resources override) - use cache-control, otherwise use Etag

dash/dash.py
```diff
@@ -711,11 +728,28 @@ def serve_component_suites(self, package_name, path_in_package_dist):
             package.__path__,
         )
 
-        return flask.Response(
+        response = flask.Response(
             pkgutil.get_data(package_name, path_in_package_dist),
             mimetype=mimetype,
         )
 
+        if fingerprint:
```
byronz: 🐮  I found that the variable assignment before is confusing, and it might be better to put the `res is not None` directly here.

Marc-Andre-Rivet: sure

dash/dash.py
Marc-Andre-Rivet: https://github.com/plotly/dash/pull/973/commits/5e61384996035ff47542134cf5e2a8ad978c1a5c

dash/dash.py
```diff
@@ -676,6 +679,19 @@ def _generate_meta_html(self):
 
     # Serve the JS bundles for each package
     def serve_component_suites(self, package_name, path_in_package_dist):
+        # Check if the resource has a fingerprint
+        res = re.match(
+            r'^(.*)[.]v\d+_\d+_\d+(-\w+[.]?\d+)?m\d{10}([.]js)$',
```
alexcjohnson: Isn't this going to be used for `.css` and `.js.map` too? Anyway those are included in the `mimetype` dict below.

Also wondering if we really need to be that specific with the pattern, and if it really needs to have a `.` in the `(-\w+[.]?\d+)` group? What about:
```
r'^(.*)[.]v[\w_-]+m\d+([.][\w.]+)$'
```

dash/dash.py
alexcjohnson: oh hmm, `.js.map` again...

dash/dash.py
Marc-Andre-Rivet: Good point about the various extensions.

`.js.map` is not really an issue here as the request resource comes from the `//# sourceMappingURL=bundle.js.map` at the end of the bundle file and will never get fingerprinted in the front-end or be provided by the sever in the html.

As for the regex being stronger vs. weaker, as I was coming at it from the perspective that we have full control over the format that will be provided, so might as well make a strong check here too. That said, there's no obvious advantage to making a stronger check and maybe we end up losing some flexibility if edge-cases appear. Will update with a more general pattern.

alexcjohnson: OK cool, I'll buy that about the maps - if you tell me you've seen it working 😅 

Yeah my thinking about the regex was just it's safer to match anything we're sure *won't* be in the regular filename. Like, what if there's some system where mod time isn't 10 digits for whatever reason, or someone has a really nonstandard version identifier.

The `mimetype` dict as written right now is explicit - whether or not intentionally so - that the final extension can only be `js`, `css`, or `map` - anything else will throw a `KeyError`. Currently anything else you need would go through assets (which would be interesting to investigate for caching, though generally much less important than component js), not component suites.

Marc-Andre-Rivet: > if you tell me you've seen it working 😅

Yup. They work :) The "routing/fingerprint" done by Dash only affects the resources put in the HTML, in that sense it's more trickery than routing 😸 

For the regex, generalized the pattern to:
`^(?P<filename>.*)[.]v[\w-]+m[0-9a-fA-F]+(?P<extension>(?:(?:(?<![.])[.])?[\w@])+)$` and its inverse `^(?P<filename>[\w@-]+)(?P<extension>.*)$` which behaves as followed for various valid / invalid strings:

image

image

These strings are now tested in unit tests within Dash.

> The mimetype dict as written right now is explicit

My understanding is that this was intentional. Learned about it when I added the `map` extension. There's apparently a library that offers mimetype mapping but I was told at the time that it was sketchy. Needs to be revisited. If only for the woff/ddk issue.

@plotly/webpack-dash-dynamic-import/src/index.js
```diff
@@ -15,15 +27,24 @@ Object.defineProperty(__webpack_require__, 'p', {
             return url;
         };
     })()
-});`
+});
+
+const __jsonpScriptSrc__ = jsonpScriptSrc;
+jsonpScriptSrc = function(chunkId) {
+    const srcFragments = __jsonpScriptSrc__(chunkId).split('.');
+    srcFragments.splice(1, 0, ${getFingerprint()});
```
Marc-Andre-Rivet: Changed so that only the first fragment is considered the filename.

dash/dash.py
```diff
@@ -541,12 +542,14 @@ def _relative_url_path(relative_package_path="", namespace=""):
...
                 self.config.requests_pathname_prefix,
                 namespace,
-                relative_package_path,
-                importlib.import_module(namespace).__version__,
-                modified,
+                build_fingerprint(
+                    relative_package_path,
+                    importlib.import_module(namespace).__version__,
+                    modified,
+                ),
```
Marc-Andre-Rivet: Isolated the fingerprinting logic with its inverse in a easily testable file / function

dash/dash.py
Marc-Andre-Rivet: Updated to use a specialized fingerprint building function that should handle these cases consistently in regards to the fingerprint check. https://github.com/plotly/dash/pull/973/files#diff-2b63b39f9669123b2341a57a37913ad8R10

dash/dash.py
```diff
@@ -676,6 +679,10 @@ def _generate_meta_html(self):
 
     # Serve the JS bundles for each package
     def serve_component_suites(self, package_name, path_in_package_dist):
+        (path_in_package_dist, has_fingerprint) = check_fingerprint(
```
alexcjohnson: non-🚫 - you don't need the parens
```suggestion
        path_in_package_dist, has_fingerprint = check_fingerprint(
```

Marc-Andre-Rivet:
@alexcjohnson Good catch on https://github.com/plotly/dash-table/pull/632#discussion_r339557176 -- testing it out.

Marc-Andre-Rivet:
@alexcjohnson Non-async resources will work fine because Dash Py distinguishes the `relative_package_path` usage (https://github.com/plotly/dash/blob/dev/dash/dash.py#L544) from the `external_url` usage (https://github.com/plotly/dash/blob/dev/dash/dash.py#L564), causing only the relative path to get a fingerprint.

Async resources do not distinguish between being external or internal and will cause problems as currently implemented. My recommendation at this point is to revert the webpack fingerprinting for async resources for now and add it back / take external resources into consideration when adding more aggressive async into the various packages. This will make the existing async files (dcc graph, dcc plotly, table xlsx) default back to ETag, which is not ideal but still provides some caching improvements over a no-caching impl.

Marc-Andre-Rivet:
@alexcjohnson My recommendation would be to drop the `1.1 plugin` update in favor of publishing the cache improvements as-is. Async resources will still benefit from `Etag` and having serve_locally cache *most* of the resource files is still a very good improvement vs. the current state of things.

Marc-Andre-Rivet:
I think the Dash Py code can be updated relatively easily to decorate the script tag with an attribute telling the piugin if it is local/external and allowing toggling fingerprinting that way.

Marc-Andre-Rivet:
> I think the Dash Py code can be updated relatively easily to decorate the script tag with an attribute telling the plugin if it is local/external and allowing toggling fingerprinting that way.

Turns out there's a few scenarios for async resources. Amongst others, chaining `import(...)` has impact over the "currently executing script" and the webpack implementation for `script` creation can't be augmented/decorated. Instead of some complex BE logic and weird webpack overrides, pushing this into the webpack plugin: if the script `src` contains `_dash-components-suite` it's local and will get fingerprinted, otherwise it's not.

@plotly/webpack-dash-dynamic-import/src/index.js
```diff
@@ -1,29 +1,72 @@
...
+
+const __jsonpScriptSrc__ = jsonpScriptSrc;
+jsonpScriptSrc = function(chunkId) {
+    let script = getCurrentScript();
+    let isLocal = isLocalScript(script);
+
+    let src = __jsonpScriptSrc__(chunkId);
+
+    if(!isLocal) {
+        return src;
```
Marc-Andre-Rivet: Don't fingerprint if using an external resource!

PR #974, opened 2019-10-23
add extra wait for callbacks in queue

byronz:
after the discussion with @alexcjohnson  about the incident of dash-daq PR on dash docs, this PR adds one extra wait in `visit_and_snapshot` API to make sure all the callbacks in the queue get fired. tested with the local snapshots and also dash-docs tests, the results are positive. 

https://percy.io/plotly/dash-docs/builds/2904660
https://percy.io/plotly/dash/builds/2904645

### DIFF
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -509,7 +509,7 @@ class Browser(DashPageMixin):
             self.wait_for_element_by_id(hook_id)
             until(
                 lambda: self.redux_state_rqs
-                and all((_["status"] == 200 for _ in self.redux_state_rqs)),
+                and all((_["responseTime"] for _ in self.redux_state_rqs)),
                 timeout=10,
             )
 
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -504,7 +504,11 @@ class Browser(DashPageMixin):
             if path != resource_path:
                 logger.warning("we stripped the left '/' in resource_path")
             self.driver.get("{}/{}".format(self.server_url.rstrip("/"), path))
+
+            # wait for the hook_id to present and all callbacks get fired
             self.wait_for_element_by_id(hook_id)
+            until(lambda: self.redux_state_rqs, timeout=10)
+
             self.percy_snapshot(path)
             if assert_check:
                 assert not self.driver.find_elements_by_css_selector(
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -445,7 +445,7 @@ class Browser(DashPageMixin):
         elem_or_selector,
         start_fraction=0.5,
         zoom_box_fraction=0.2,
-        compare=True
+        compare=True,
     ):
         """Zoom out a graph with a zoom box fraction of component dimension
         default start at middle with a rectangle of 1/5 of the dimension use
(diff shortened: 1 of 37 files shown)


dash/testing/browser.py
```diff
@@ -504,7 +504,15 @@ def visit_and_snapshot(self, resource_path, hook_id, assert_check=True):
             if path != resource_path:
                 logger.warning("we stripped the left '/' in resource_path")
             self.driver.get("{}/{}".format(self.server_url.rstrip("/"), path))
+
+            # wait for the hook_id to present and all callbacks get fired
             self.wait_for_element_by_id(hook_id)
+            until(
+                lambda: self.redux_state_rqs
+                and all((_["responseTime"] for _ in self.redux_state_rqs)),
```
alexcjohnson: I don't think `responseTime` is even *in* the object while it's waiting for a response? so perhaps
```suggestion
                and all((_.get("responseTime") for _ in self.redux_state_rqs)),
```

alexcjohnson:
FYI the PR that sparked this push for more robust percy snapshots is https://github.com/plotly/dash-docs/pull/676

This strategy looks great - I didn't know about `self.redux_state_rqs`, that's really nice to have!

Just the one comment above, then I think this will be good to go.

BTW I had to approve one spurious percy diff here that seems to be frequently changing, `chapter-1-again` (I've also seen `chapter-1` fail similarly). I think it's somehow autosizing the graph differently in different runs? If that's it, we can 🔒 it by changing this line:

https://github.com/plotly/dash/blob/c4db2efd24c5a1e0f75a0ca262ba5aeeb0ff3c7f/tests/integration/callbacks/test_layout_paths_with_callbacks.py#L125

to have fixed height & width like:
```py
"layout": {"title": value, "width": 500, "height": 400}
```

byronz:
> This strategy looks great - I didn't know about `self.redux_state_rqs`, that's really nice to have!

it's in the dash_page.py which is not documented, we should add a section for that part later

alexcjohnson:
Oh actually, one more concern: what if some existing cases are snapshotting pages that don't even have callbacks? I guess we could look at `store.getState().dependenciesRequest.content.length` - if that's 0, don't wait for anything to show up in `requestQueue`?

byronz:
> Oh actually, one more concern: what if some existing cases are snapshotting pages that don't even have callbacks? I guess we could look at `store.getState().dependenciesRequest.content.length` - if that's 0, don't wait for anything to show up in `requestQueue`?

In the context of dash docs (where we use mostly this API) there is always at least one request like the home page.  

In the context of an app, you are right that no callback, no `requestqueue`, but I'm also wondering the purpose of using this API in this case. isn't a simple percy_snapshot more straightforward?

alexcjohnson:
> isn't a simple percy_snapshot more straightforward?

Oh... right, I forgot about the `visit_and_snapshot` vs `percy_snapshot` distinction - in fact if it were robust (against not even being a dash page, for example what if you get a wekzeug debug page) I could see this being part of `percy_snapshot` - perhaps with the ability to turn it off in case you actually want to run ahead of a response for some reason.

byronz:
> > isn't a simple percy_snapshot more straightforward?
> 
> Oh... right, I forgot about the `visit_and_snapshot` vs `percy_snapshot` distinction - in fact if it were robust (against not even being a dash page, for example what if you get a wekzeug debug page) I could see this being part of `percy_snapshot` - perhaps with the ability to turn it off in case you actually want to run ahead of a response for some reason.


we dont' have window.store outside the scope of dash apps. :)

PR #976, opened 2019-10-25
Quart

sjtrny:
This PR migrates Dash from Flask to Quart. This change should not break existing Dash apps.

This is an effort to move the ship a few degrees closer to supporting some form of web socket update or callback. However I would expect support for this to be added in another PR.

TODO:
- [] Replace minor dependencies on flask such as 'flask.helpers.get_root_path'
- [] Swap references from flask to quart in code and docs
- [] Update docs to discuss async callbacks including pitfalls such as calling sync code


alexcjohnson:
@sjtrny thanks for the contribution. Looks like this went stale before getting to a reviewable state, so I'm going to close it. Apologies for leaving this with no response for so long.

We'd be amenable to discussing this kind of change, but it should probably go through an issue first so we can hash out an architecture that will satisfy everyone. Given that there are a lot of folks who plug Dash into existing Flask apps, we can't completely swap it out, but we may well be able to build in pluggable support for alternative servers.

PR #983, opened 2019-10-29
Dbg r reload

byronz:
this PR solves the problem when dashR app is defined in a string chunk, and the app needs to work with assets or clientside js file.


dash/testing/application_runners.py
```diff
@@ -309,13 +327,13 @@ def start(self, app, start_timeout=2, cwd=None):
...
+            wait.until(lambda: self.accessible(self.url), timeout=start_timeout)
 
         except (OSError, ValueError):
             logger.exception("process server has encountered an error")
             self.started = False
             return
 
         self.started = True
+
+        return app
```
byronz: @rpkyle  in case you need to get the tmp path, `start()` returns it

dash/testing/application_runners.py
```diff
@@ -261,13 +267,18 @@ def start(self, app, start_timeout=2, cwd=None):
...
         else:
-            path = (
-                "/tmp/app_{}.R".format(uuid.uuid4().hex)
-                if not self.is_windows
-                else os.path.join(
-                    (os.getenv("TEMP"), "app_{}.R".format(uuid.uuid4().hex))
-                )
+            self._tmp_app_path = os.path.join(
+                "/tmp" if not self.is_windows else os.getenv("TEMP"),
+                uuid.uuid4().hex,
```
byronz: as @rpkyle  mentioned, the hot reloading on R side will actively monitor all the folder under the main app file, in this case, the old approach using /tmp/app_{}.R is not safe

dash/testing/application_runners.py
```diff
@@ -297,16 +308,40 @@ def start(self, app, start_timeout=2, cwd=None):
                     "dashr.run_server(app, cwd=os.path.dirname(__file__))"
                 )
 
+            # try copying all valid sub folders (i.e. assets) in cwd to tmp
+            # note that the R assets folder name can be any valid folder name
+            assets = [
+                os.path.join(cwd, _)
+                for _ in os.listdir(cwd)
+                if not _.startswith("__")
+                and os.path.isdir(os.path.join(cwd, _))
```
alexcjohnson: 🐄 why `_`? I prefer to use that only when the var isn't going to be used at all.

byronz: I also like to use that in list comprehension especially when it gets more complicated. 
- shorter
- I think it's easier to read once you used to the convention, and can be more focus on the operation and logic side.

alexcjohnson: Fair enough. I'd tend to use a 1-character name for that but I see your point.

byronz: > Fair enough. I'd tend to use a 1-character name for that but I see your point.

1-char name is anti-pylint

alexcjohnson: `_` isn't a 1-char name? 🤔

byronz: btw, it's a valid 1-char name, certified.

PR #985, opened 2019-10-29
Ddk eager flag fix

### LINKED ISSUE #984: [BUG]  _force_eager_loading depends on import/instantiation order
<details>
 <summary>click for obligatory `pip list | grep dash`</summary>

| package  | version |
| ------------- | ------------- |
| dash               |                1.5.0      |
| dash-auth        |                  1.2.0    |   
| dash-bio           |                0.1.1      | 
| dash-bio-utils    |                 0.0.2     |  
| dash-canvas       |                 0.0.11  |    
| dash-core-components    |           1.4.0     |  
| dash-cytoscape           |          0.0.5       |
| dash-dangerously-set-inner-html |   0.0.2 |      
| dash-daq              |             0.1.0      | 
| dash-design-kit         |           1.3.0       /home/plotly/Workspace/dash-design-kit         |
| dash-design-kit-docs     |          1.3.0       |
| dash-enterprise-auth       |        0.0.2       |
| dash-extra-components     |         0.0.1    |   
| dash-html-components       |        1.0.1      | 
| dash-renderer              |        1.2.0   |
| dash-table                |         4.5.0       |

</details>

In `dash v1.5.0`, instantiating or importing an "app" (i.e. `app = dash.Dash(__name__)`) before importing a package with `_force_eager_loading = True` defined in its `__init__.py` file will still result in lazy loading. Expected behavior is that eager loading will be forced, as implied by the flag name. This can result in missing dependencies, e.g. `plotly.js`: 

image

Confirmed as reproducible by @Marc-Andre-Rivet

Marc-Andre-Rivet:
Fixes #984 - delays (reevaluates) Css and Scripts at the moment when the server is ready instead of doing so during initialization.

This should allow us to handle both (1) app.run_server and (2) gunicorn / multi-page apps correctly.


dash/dash.py
```diff
@@ -1434,14 +1429,22 @@ def _validate_layout(self):
...
+        # Apply _force_eager_loading overrides from modules
+        eager_loading = self.config.eager_loading
+        for module_name in ComponentRegistry.registry:
+            module = sys.modules[module_name]
+            eager = getattr(module, '_force_eager_loading', False)
+            eager_loading = eager_loading or eager
+
+        # static files from the packages
+        self.css = Css(self.css.config.serve_locally)
+        self.scripts = Scripts(self.scripts.config.serve_locally, eager_loading)
```
alexcjohnson: Do we need to recreate these here? Wouldn't it be sufficient to just say
```py
self.scripts.config.eager_loading = eager_loading
```
or even
```py
for module_name in ComponentRegistry.registry:
    module = sys.modules[module_name]
    if getattr(module, '_force_eager_loading', False):
        self.scripts.config.eager_loading = True
        break
```

dash/dash.py
```diff
@@ -1434,14 +1429,22 @@ def _validate_layout(self):
...
+
         if self.config.include_assets_files:
             self._walk_assets_directory()
 
         self._validate_layout()
 
-        self._generate_scripts_html()
-        self._generate_css_dist_html()
```
alexcjohnson: Seems like we need these for their side-effects, or perhaps recreating `self.scripts` and `self.css` was the problem? Regardless, I don't see any harm bringing this back.

Marc-Andre-Rivet: Agreed. Undoing this.

dash/dash.py
Marc-Andre-Rivet: Ah! I think you're right. I think I misunderstood how/when this got evaluated. But the filter only gets called later.

Marc-Andre-Rivet:
@alexcjohnson Closing this in favor of https://github.com/plotly/dash/pull/986 as the baseline was `master` and caused a bunch of artifact files to be included..

PR #987, opened 2019-10-30
Fix caching with resources in subdirs

alexcjohnson:
Fixes the issue described in https://github.com/plotly/dash/commit/340d960b93bb9cc0ec91664d05992143cf94d265#commitcomment-35719028

### DIFF
--- a/tests/unit/test_fingerprint.py
+++ b/tests/unit/test_fingerprint.py
@@ -1,51 +1,66 @@
-
 from dash.fingerprint import build_fingerprint, check_fingerprint
 
 version = 1
 hash_value = 1
 
 valid_resources = [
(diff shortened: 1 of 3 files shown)


tests/unit/test_fingerprint.py
```diff
@@ -1,56 +1,122 @@
...
+        # The inverse operation returns that the fingerprint was valid
+        # and the original path
         assert has_fingerprint
-        assert original_path == resource.get('path')
+        assert original_path == resource.get("path")
 
     for resource in valid_fingerprints:
         (_, has_fingerprint) = check_fingerprint(resource)
-        assert has_fingerprint
+        assert has_fingerprint, resource
```
alexcjohnson: To report which resource it is when one fails, rather than just `assert False`

PR #996, opened 2019-11-04
Fix display of stack traces within RStudio's viewer pane

### LINKED ISSUE #994: Stack traces do not render in the dev tools UI when using RStudio's viewer pane
RStudio's IDE does not allow embedding of `iframe` elements which include content from "external" URLs, as noted in https://github.com/rstudio/rstudioapi/issues/2.

The way that dash-renderer currently handles stack traces places them into an `iframe` so that Werkzeug's CSS doesn't affect the display of the rest of the page:

https://github.com/plotly/dash/blob/b3b9de9923d82ff1281889b3695d638e95bb1590/dash-renderer/src/components/error/FrontEnd/FrontEndError.react.js#L111-L143

This has the unfortunate consequence of showing no stack trace within RStudio, though it appears properly in standalone browsers:

image

The `plot_ly` function gets around this problem by embedding source from `localhost`, which is allowed:

```
<iframe class="gwt-Frame ace_editor_theme" src="http://localhost:29325/session/viewhtml98b753d4ffc6/index.html?viewer_pane=1" style="width: 100%; height: 100%; position: absolute; left: 0px; top: 0px; right: 0px; bottom: 0px; background-color: rgb(255, 255, 255);"></iframe>
```

We can preserve the existing behaviour in the renderer for Python apps by implementing a condition which checks whether the content of the stack trace begins with `<!DOCTYPE HTML>`, then enclose in an `iframe` if `true`. 

Dash for R would then be modified to send back a stack trace that is raw text, which the renderer will insert directly within `<pre>` tags (since the above condition would be `false`), and this should enable support for stack trace display in the RStudio viewer as well as standalone browsers.

@alexcjohnson

rpkyle:
This PR proposes to modify `dash-renderer` to check whether the stack trace being sent up by the backend includes the string `<!DOCTYPE HTML`. If it does, the content is embedded within an `iframe` to avoid CSS leaking caused by the Werkzeug HTML which Dash for Python provides. 

If it does not, the content is included in a simple nested `div` instead, with customized CSS to properly line-wrap the preformatted content, and to automatically use a scrollbar if the stack trace is sufficiently long.

After merging, the stack trace within the dev tools UI for Dash for R apps will look like this:

Viewer Pane Screenshot

To be merged concurrently with plotly/dashR#137. Fixes #994.


dash-renderer/src/components/error/FrontEnd/FrontEndError.react.js
```diff
@@ -108,7 +108,9 @@ function UnconnectedErrorContent({error, base}) {
                 </div>
             )}
             {/* Backend Error */}
-            {typeof error.html !== 'string' ? null : (
+            {typeof error.html !== 'string' ? null : error.html.indexOf(
+                  '<!DOCTYPE HTML'
+              ) !== -1 ? (
```
alexcjohnson:
```suggestion
              ) === 0 ? (
```
Just in case this were to show up somewhere in the error message, only look at the beginning of the string.

dash-renderer/src/components/error/FrontEnd/FrontEndError.react.js
```diff
@@ -140,6 +142,19 @@ function UnconnectedErrorContent({error, base}) {
...
                     </div>
                 </div>
+            ) : (
+                <div className="dash-be-error__str">
+                    <div
+                        className="dash-backend-error"
+                        style={{
+                            width: 'calc(600px - 67px)',
+                            height: '75vh',
+                            border: 'none',
```
alexcjohnson: Do we actually need any of these style attrs?

rpkyle: I kept them because we used them previously, but it might be ok to 🔪 from both R and Python. The important bits are in the `FrontEndError.css` file anyhow.

alexcjohnson: The `<iframe>` needs explicit sizing because the containing page doesn't know what's in it, and border because perhaps iframes have a border by default. But the unframed string version should be closer to the front-end version, which gets most of its styling from `.dash-fe-error__info`.

rpkyle: fixed in bf089447f31618973412356e690d011e911789c9

PR #999, opened 2019-11-04
Issue 995 - Correctly support `version-prerelease+metadata` in fingerprint

### LINKED ISSUE #995: [BUG] + in version string breaks fingerprint system
**Describe your context**
```
dash                                  1.5.1                           
dash-core-components                  1.4.0                           
dash-daq                              0.2.2                           
dash-html-components                  1.0.1                           
dash-renderer                         1.2.0                           
dash-table                            4.5.0  
```

**Describe the bug**

When going from `dash==1.4` to `dash==1.5`, we experienced a breaking change in the custom Dash components we use.

It took some hours to debug, but the reason was found to be related to the new "fingerprint" system in Dash. In our project, we use the [setuptools_scm](https://github.com/pypa/setuptools_scm) package (by the Python Packaging Authority) in order to have a versioning system that automatically is linked to the git repo tags. This makes continuous deployment to e.g. Pypi easy and robust wrt. keeping versions consistent.

I.e. instead of
```python
__version__ = package['version']
```
in the component package, we use something like
```
__version__ = get_distribution(__name__).version
```
This worked until `dash==1.5`, then it broke on non-release-versions due to automatic tags of the type
`1.0.0.dev5+af4304c.d20191103`, where the tag includes a `+`. See [the default tag formats](https://github.com/pypa/setuptools_scm#default-versioning-scheme).

Changing the line above to
```
__version__ = get_distribution(__name__).version.replace("+", ".")
```
is one workaround that gets the third party components to also work on `dash==1.5`

**Expected behavior**

`setuptools_scm` provided versions to work also in `dash>=1.5`.

**Suggested solution**

Change [this line](https://github.com/plotly/dash/blob/40b5357f262ac207f94ac980e6cb928d94df65b7/dash/fingerprint.py#L12) in Dash's `build_fingerprint` to also replace `+` with `_`?

Marc-Andre-Rivet:
Fixes #995 

Manually tested against

1. Custom table build with version
```
  "version": "4.5.1-alpha1.2.3+metadata.1.2.3-abc",
```

2. The modified version of the `@plotly/webpack-dash-dynamic-import` plugin

3. App, with both `eager_loading` true / false to trigger async resources from both DashPy and the plugin
```
import json
import pandas as pd
from datetime import datetime

import dash

import dash_table

app = dash.Dash(__name__, eager_loading=False)

app.layout = html.Div([
    dcc.DatePickerRange(
        id="date-picker-range",
        start_date_id="startDate",
        end_date_id="endDate",
        start_date=datetime(1997, 5, 3),
        end_date_placeholder_text="Select a date!",
    ),
    dash_table.DataTable(
        columns=[],
        data=[]
    )
])

if __name__ == '__main__':
    app.run_server(debug=False)
```

### DIFF
--- a/dash/fingerprint.py
+++ b/dash/fingerprint.py
@@ -1,7 +1,7 @@
 import re
 
 cache_regex = re.compile(r"^v[\w-]+m[0-9a-fA-F]+$")
-
+version_clean = re.compile(r"[^\w-]")
 
 def build_fingerprint(path, version, hash_value):
     path_parts = path.split("/")
@@ -9,7 +9,7 @@ def build_fingerprint(path, version, hash_value):
 
     return "{}.v{}m{}.{}".format(
         "/".join(path_parts[:-1] + [filename]),
-        str(version).replace(".", "_").replace("+", "__"),
+        re.sub(version_clean, "_", str(version)),
         hash_value,
         extension,
     )
--- a/dash/fingerprint.py
+++ b/dash/fingerprint.py
@@ -9,7 +9,7 @@ def build_fingerprint(path, version, hash_value):
 
     return "{}.v{}m{}.{}".format(
         "/".join(path_parts[:-1] + [filename]),
-        str(version).replace(".", "_"),
+        str(version).replace(".", "_").replace("+", "__"),
         hash_value,
         extension,
     )
--- a/@plotly/webpack-dash-dynamic-import/src/index.js
+++ b/@plotly/webpack-dash-dynamic-import/src/index.js
@@ -5,7 +5,7 @@ function getFingerprint() {
     const packageJson = JSON.parse(package);
 
     const timestamp = Math.round(Date.now() / 1000);
-    const version = packageJson.version.replace(/[.]/g, '_');
+    const version = packageJson.version.replace(/[.]/g, '_').replace(/[+]/g, '__');
 
     return `"v${version}m${timestamp}"`;
 }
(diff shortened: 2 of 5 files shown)


dash/fingerprint.py
```diff
@@ -9,7 +9,7 @@ def build_fingerprint(path, version, hash_value):
 
     return "{}.v{}m{}.{}".format(
         "/".join(path_parts[:-1] + [filename]),
-        str(version).replace(".", "_"),
+        str(version).replace(".", "_").replace("+", "__"),
```
alexcjohnson: This works for `+`, but how confident are we that `+`, `-`, and `.` are the only characters outside of `\w` (`[A-Za-z0-9_]`) that we're ever going to see? Is there anything to stop people from making a package with
`__version__ = "the $%# *best* apple π"`?
and then, to be ultra-paranoid, their next version is:
`__version__ = "the $@# *best* apple π"`
Or perhaps more likely, some foreign character used as a serial number gets incremented to the next such foreign character.

Feels to me the super-safe way to handle this would be to replace each `[^\w-]` character with an escape code, something like `"_{:x}".format(ord(char))`. If we want to keep `.` as just `_` that seems pretty safe, even if there could be a pathological case that breaks it - in fact starting with any unescaped character then replacing it with itself escaped in the next version (including, today, replacing `_` with `.`) would yield a matching string... but I'd be willing to ignore that possibility.

Marc-Andre-Rivet: `[0-9a-zA-Z]` and `-` are the only allowed characters in the fragments as per semver.
https://semver.org/

Versions need to be non-negative numbers.
Pre-release begins with `-` with n-fragments separated by `.`
Metadata beings with `+` with n-fragments seperated by `.`

Checking out what https://www.python.org/dev/peps/pep-0440/ expects.. turns out they also support `version epoch` which would add `!` into the mix (https://www.python.org/dev/peps/pep-0440/#id30)

Marc-Andre-Rivet: Alright.. let's go full pathological.

PR #1000, opened 2019-11-04
update ci

byronz:
the change in the CI config should fix the problem where a test case failed, percy cannot be finalized properly, i.e. a simple rerun will introduce a 409 error.


.circleci/config.yml
```diff
@@ -189,6 +189,7 @@ jobs:
       - image: circleci/python:3.7-stretch-node-browsers
         environment:
           PERCY_PARALLEL_TOTAL: -1
+          PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: True
```
alexcjohnson: What's this about?

.circleci/config.yml
```diff
@@ -218,6 +219,10 @@ jobs:
           path: test-reports
       - store_artifacts:
           path: /tmp/dash_artifacts
+      - run:
+          name: 🦔 percy finalize
+          command: npx percy finalize --all
+          when: on_fail
```
alexcjohnson: interesting... how does this work with parallelism? I'm just wondering what will happen if it fails early in one container, runs `percy finalize` from there, then in another container more snapshots are recorded. If it just silently fails to add those to the run that would be fine, but if it makes more errors in the other container that could get very confusing.

.circleci/config.yml
byronz: the npx needs implicitly the installation of puppeteer, which will then download the big guy `chromium`, this option is to skip the download.

.circleci/config.yml
byronz: good point about the parallelism. this step will be executed only after the full run of `pytest` in each container, no matter if it failed at an earlier stage. the whole duration is close from one to another, but you are right, it still has some risk to confuse other containers. but I think this way it's better than not having the finalize at all.

Issue #992, opened 2019-11-01
[BUG] requestQueue get stuck when browser is experiencing intensive navigation


byronz:
Thank you so much for helping improve the quality of Dash!

We do our best to catch bugs during the release process, but we rely on your help to find the ones that slip through.

**Describe your context**
Please provide us your environment so we can easily reproduce the issue.

-  replace the result of `pip list | grep dash` below
```
dash                            1.5.1     
dash-auth                       1.2.0
dash-bio                        0.1.4
dash-bio-utils                  0.0.3rc3
dash-canvas                     0.0.11
dash-core-components            1.4.0
dash-cytoscape                  0.0.5
dash-dangerously-set-inner-html 0.0.2
dash-daq                        0.2.1
dash-flow-example               0.0.5
dash-html-components            1.0.1
dash-renderer                   1.2.0
dash-table                      4.5.0
plotly                          3.3.0
```
-  if frontend related, tell us your Browser, Version and OS

    - OS: [Mac OS]
    - Browser [Chrome]
    - Version [77]

**Describe the bug**

This is kind of reproducible with the current dash-docs PR for 1.5.1 upgrade, but we don't have fixed steps to reproduce it each time.

To give a little background about issue. We introduced a new feature/parameter is `visit_and_snapshot` API, what it does is to detect if window.store exists and makes sure all 
the callbacks get fired and have a `responseTime` in the queue (refer to https://github.com/plotly/dash/blob/b3b9de9923d82ff1281889b3695d638e95bb1590/dash/testing/dash_page.py#L43-L48) 

this mechanism works fine with the older dash version prior to the 1.5.x release. so the behavior is new. what I observed is that the tests in  test_component.py are doing intensively url navigation in the browser, it basically goes through each sub section of dash docs in a very fast way, at some point, the request will get stuck with status = 'loading', a refresh of the page might solve the issue, but not always the case. the stuck point is always consistent the same on CircleCI and doing it locally. but not the same in the two environments : i.e. circleci get stuck with /datatables/callbacks while local run might be /datatables/editable

this looks like a leak or something related to the recent async work 

sample of a loading state  https://circleci.com/api/v1.1/project/github/plotly/dash-docs/8326/output/106/0?file=true
[...]

PR #1001, opened 2019-11-05
:mag: tune snapshots API

byronz:

## Contributor Checklist

### optionals


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+- [#1001](https://github.com/plotly/dash/pull/1001) Fix and imporve the `clear_input()` API in `dash.testing`, so it's more robust handling react `input`
```
alexcjohnson:
```suggestion
- [#1001](https://github.com/plotly/dash/pull/1001) Fix and improve the `clear_input()` API in `dash.testing`, so it's more robust handling react `input`
```

dash/testing/browser.py
```diff
@@ -108,8 +112,26 @@ def percy_snapshot(self, name="", wait_for_callbacks=False):
...
         logger.info("taking snapshot name => %s", snapshot_name)
-        if wait_for_callbacks:
-            until(self._wait_for_callbacks, timeout=10)
+        try:
+            if wait_for_callbacks:
+                until(
+                    self._wait_for_callbacks,
+                    timeout=40,
+                    poll=0.5,
+                    sleep_first=True,
```
alexcjohnson: I'm not so sure about adding `sleep_first`... it means you always get `poll` amount of sleep time, whereas you could get the same effect without having to set an otherwise-unnecessarily large `poll` if you just put `time.sleep(0.5)` before the `until` call, right? That feels clearer to read as well.

dash/testing/dash_page.py
```diff
@@ -42,9 +42,17 @@ def window_store(self):
 
     def _wait_for_callbacks(self):
         if self.window_store:
-            return self.redux_state_rqs and all(
```
alexcjohnson: We don't need this first `self.redux_state_rqs and` anymore?

dash/testing/dash_page.py
```diff
@@ -42,9 +42,17 @@ def window_store(self):
...
-                (_.get("responseTime") for _ in self.redux_state_rqs)
+            # note that there is still a small chance of Negative True
+            # where we get two earlier requests in the queue, this returns
+            # True but there are still more requests to come
+            return all(
+                (
+                    _.get("responseTime")
+                    for _ in self.redux_state_rqs
+                    if _.get("controllerId")
```
alexcjohnson: ha, I think I see what's going on here, thanks for bringing this up. Interestingly this was a question from the beginning when multi-outputs was implemented:

https://github.com/plotly/dash/blob/046c438928c17e0f7faf5343f5f4e6d98d0c9efd/dash-renderer/src/actions/index.js#L933-L934

Yes, there seem to be some implications... probably due to the fact that `dispatch` is not synchronous. Not clear to me why we need this dummy request in the first place, but I'll figure that out.

alexcjohnson: For followup in #1008

dash/testing/dash_page.py
byronz: good catch, I was wrong about the `all`, an empty list will return True 🐛

dash/testing/dash_page.py
byronz: another issue in the same basket is #992

dash/testing/browser.py
byronz: ah, I forgot to add the flag in the `percy_snapshot` API. my initial intention is to make it more flexible, so for some `visit_and_snapshot` calls, we should let it do a `sleep` before the first `check` to have some safe margins with `wait_for_callbacks`. I will revise it

alexcjohnson: OK - if you do want to keep `sleep_first` as part of `until`, perhaps better would be to make it a number instead of a boolean, so it would be decoupled from `poll`.

byronz: yes,  revisit the code, inclines to add hard sleep than sleep switch, and I will just use `wait_for_callbacks` to control that.

dash/testing/dash_page.py
```diff
@@ -42,9 +42,17 @@ def window_store(self):
 
     def _wait_for_callbacks(self):
         if self.window_store:
+            # note that there is still a small chance of FP(Fasle Positive)
```
alexcjohnson:
```suggestion
            # note that there is still a small chance of FP (False Positive)
```

Issue #1003, opened 2019-11-06
[BUG] Dash async causes syntax error in IE11


shammamah:
Thank you so much for helping improve the quality of Dash!

We do our best to catch bugs during the release process, but we rely on your help to find the ones that slip through.

**Note - found while working on https://github.com/plotly/dash-core-components/issues/627** 

**Describe your context**
Please provide us your environment so we can easily reproduce the issue.

-  replace the result of `pip list | grep dash` below
```
dash                            1.6.0    
dash-core-components            1.5.0    
dash-dangerously-set-inner-html 0.0.2    
dash-flow-example               0.0.5    
dash-html-components            1.0.1    
dash-renderer                   1.2.0    
dash-table                      4.5.0    
```
-  if frontend related, tell us your Browser, Version and OS

    - OS: Windows
    - Browser: Internet Explorer
    - Version: 11

**Describe the bug**

Using the `dash-core-components` library installed with the latest version of Dash that includes async (`1.6.0`) in an app causes the app to be stuck on "Loading..." and the following errors to show up in the console: 

![screen_shot_2019-11-06_at_1 29 43_pm](https://user-images.githubusercontent.com/15622051/68332696-3e1e5300-00a5-11ea-9b7b-6b95fbdd4177.png)

This is due to an arrow function: 

![Screen Shot 2019-11-06 at 2 51 05 PM](https://user-images.githubusercontent.com/15622051/68332474-e67fe780-00a4-11ea-8692-9362591e89e1.png)

**Expected behavior**

The components should render. 

**Screenshots**

See above.

Marc-Andre-Rivet:
Caused by `/node_modules/`, `^[@]plotly\/.*/` should be included as the non-transpiled source is used directly for the plugins.

This affects dash-renderer, html, dcc, table, bio, daq in production.
This affects the in-review changes:
https://github.com/plotly/dash-cytoscape/pull/75
https://github.com/plotly/dash-canvas/pull/29

Marc-Andre-Rivet:
Also, there are other issues unrelated with async that are breaking IE11. Details in #1006

Marc-Andre-Rivet:
Closed by #1006

PR #1006, opened 2019-11-07
IE11 compatibility

Marc-Andre-Rivet:
Related to #1003 and actually solves the issue with additional PRs (will be linked)

Also fixes https://github.com/plotly/dash/issues/1011.

|repo|pr|
|-|-|
|dash-table|https://github.com/plotly/dash-table/pull/637|
|dash-core-components|https://github.com/plotly/dash-core-components/pull/696|
|dash-html-components|https://github.com/plotly/dash-html-components/pull/143|
|dash-bio|https://github.com/plotly/dash-bio/pull/437|
|dash-canvas|https://github.com/plotly/dash-canvas/pull/29|
|dash-cytoscape|https://github.com/plotly/dash-cytoscape/pull/75|
|dash-daq|https://github.com/plotly/dash-daq/pull/73|

The latest async changes and following tests uncovered that Dash is broken for IE11 / Edge < 15. This comes from many things:

1. `@plotly/dash-component-plugins` was not transpiled when used in projects
2. `@plolty/webpack-dash-dynamic-import` was not ES5 (can't be transpiled, because added directly to the webpack artifact)
3. Multiple projects were not using `core-js` or `@babel/polyfill`. Even with correctly transpiled code they would not have worked
4. Multiple projects depend on npm dependencies that either (1) have source code as the `main` in package.json -- combined with standard webpack settings excluding `node_modules` this caused this code not to be transpiled, or (2) transpiled npm resources not transpiled in ES5, or (3) npm resources with `dist` but with `main` still pointing to the source

Handling source code from 3rd parties as part of your build process is complicated.. maybe your configuration is not compatible, or maybe the code uses another scheme (e.g. `import/export` vs. `required` vs. `module.exports`). The `module.exports` case turned out to be problematic for certain 3rd parties because of `core-js`+`useBuiltIns: usage` side-effects, namely,the code evals everything to commonjs, wrapping `module.exports` in a particular way that is almost but not quite normal `module.exports` usage in Node.

This PR and the attached PRs solves all of the above in 3 ways.

1. The artifacts are now ES5-validated with `es-check`, making it impossible to get a successful build with hanging ES6+ syntax
2. The polyfills have been moved into the renderer -- they are now shared like `react`, `react-dom`, and `prop-types` -- this removes the responsibility from the component developer's hands.. making for less chances of mistakes.. a component developer can still continue using their own polyfills if some additional specifics are needed
3. npm resources that are not ES5 compatible are now transpiled either from source code or from the `dist` (or whatever) artifact -- this is time consuming as the webpack `exclude` needs to be tweaked manually (this is mostly not an issue) -- or a `alias` is provided to use a `dist` instead of the source if the project was not setup correctly..

---

Not covered here is whether the code *actually* runs in IE11 / Edge < 15, but it establishes a good baseline with rapid (build-time) feedback. Further steps would involve setting up a test environment on (say) the docs with IE11 or the oldest version of Edge available.

---

Also, this version of Dash will become the minimal version required for all components that are dropping `@babel/polyfill` or `core-js` from their build -- effectively, will make this version of Dash the minimal version for all core repos, bio, canvas, cytoscape, daq.

Interesting side-effect, externalizing polyfills will reduce the JS size for core (renderer, html, dcc, table) by about 70kB and the externalized polyfill can be cached / will probably be stable through multiple releases or 7% of the current minimal JS load.

---

Other PRs will need to update `@plotly/webpack-dash-dynamic-import` to `1.1.3` when approved / published and some will require a version bump.


@plotly/webpack-dash-dynamic-import/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@plotly/webpack-dash-dynamic-import",
-  "version": "1.1.2",
+  "version": "1.1.3",
```
Marc-Andre-Rivet: Need a new version. The other PRs are using a rc release `1.1.3-rc1`

@plotly/webpack-dash-dynamic-import/src/index.js
```diff
@@ -39,19 +43,19 @@ Object.defineProperty(__webpack_require__, 'p', {
...
+    var src = __jsonpScriptSrc__(chunkId);
 
     if(!isLocal) {
         return src;
     }
 
-    const srcFragments = src.split('/');
-    const fileFragments = srcFragments.slice(-1)[0].split('.');
+    var srcFragments = src.split('/');
+    var fileFragments = srcFragments.slice(-1)[0].split('.');
```
Marc-Andre-Rivet: Use only ES5 syntax. This will be validated in the end by `es-check`

dash-renderer/babel.config.js
```diff
@@ -1,8 +1,8 @@
 module.exports = {
-    presets: [['@babel/preset-env', {
-        useBuiltIns: 'usage',
-        corejs: 3
-    }], '@babel/preset-react'],
+    presets: [
+        '@babel/preset-env',
```
Marc-Andre-Rivet: The polyfills are externalized. We only use `preset-env` to know what the transpilation target is.

dash-renderer/init.template
```diff
@@ -7,23 +7,27 @@ _js_dist_dependencies = [
...
         },
         "relative_package_path": {
             "prod": [
+                "polyfill@$polyfill.min.js",
                 "react@$react.min.js",
                 "react-dom@$reactdom.min.js",
                 "prop-types@$proptypes.min.js",
             ],
             "dev": [
+                "polyfill@$polyfill.min.js",
```
Marc-Andre-Rivet: Externalized polyfills, to be used by all components

dash-renderer/package.json
```diff
@@ -10,6 +10,7 @@
     "build:dev": "webpack --build local",
     "build:local": "renderer build local",
     "build": "renderer build",
+    "postbuild": "es-check es5 dash_renderer/*.js",
```
Marc-Andre-Rivet: postbuild check, automagically runs after `build`

dash-renderer/webpack.config.js
```diff
@@ -10,7 +10,7 @@ const defaults = {
         rules: [
             {
                 test: /\.js$/,
-                exclude: /node_modules/,
+                exclude: /node_modules\/(?!@plotly\/dash-component-plugins\/|uniqid\/|check-prop-types\/)/,
```
Marc-Andre-Rivet: Exclude all that can be excluded. In this case it means that `@plotly/dash-component-plugins`, `uniqid` and `check-props`types` need to be transpiled to the correct target b/c we're using the source or a dist with a less stringent target.

dash/development/build_process.py
```diff
@@ -126,8 +126,8 @@ def bundles(self, build=None):
             "version": self.version,
             "package": self.name.replace(" ", "_").replace("-", "_"),
         }
-        for name, subfolder, filename, target in self.deps_info:
-            version = self.deps[name]["version"]
+        for scope, name, subfolder, filename, target in self.deps_info:
```
Marc-Andre-Rivet: Handle scoped packages with `@SCOPE/NAME` format

dash/development/build_process.py
```diff
@@ -166,12 +166,13 @@ def __init__(self):
                 "dash-renderer",
             ),
             (
-                ("react", "umd", "react.production.min.js", None),
-                ("react", "umd", "react.development.js", None),
-                ("react-dom", "umd", "react-dom.production.min.js", None),
-                ("react-dom", "umd", "react-dom.development.js", None),
-                ("prop-types", None, "prop-types.min.js", None),
-                ("prop-types", None, "prop-types.js", None),
+                ("@babel", "polyfill", "dist", "polyfill.min.js", None),
```
Marc-Andre-Rivet: Add polyfill, add `None` scope for all other deps

dash-renderer/jest.init.js
```diff
@@ -0,0 +1 @@
+import '@babel/polyfill';
```
Marc-Andre-Rivet: jest now needs the polyfills..

alexcjohnson: seems fine, but I'm surprised jest is running in an environment that needs polyfills.

dash-renderer/webpack.config.js
```diff
@@ -10,7 +10,7 @@ const defaults = {
         rules: [
             {
                 test: /\.js$/,
-                exclude: /node_modules\/(?!@plotly\/dash-component-plugins\/|uniqid\/|check-prop-types\/)/,
+                exclude: /node_modules\/(?!uniqid\/|check-prop-types\/)/,
```
alexcjohnson: 😍

dash-renderer/jest.init.js
Marc-Andre-Rivet: jest doesn't, but jest needs to polyfills to handle our code!

@plotly/webpack-dash-dynamic-import/src/index.js
```diff
@@ -39,25 +43,27 @@ Object.defineProperty(__webpack_require__, 'p', {
     })()
 });
 
-const __jsonpScriptSrc__ = jsonpScriptSrc;
-jsonpScriptSrc = function(chunkId) {
-    let script = getCurrentScript();
-    let isLocal = isLocalScript(script);
+if (typeof jsonpScriptSrc !== 'undefined') {
```
Marc-Andre-Rivet: `jsonpScriptSrc` is not define is there are no async resources, the `typeof ... === 'undefined'` check is the only way in JS to test for the existence/non-existence of a variable in a safe way.

dash-renderer/src/TreeContainer.js
```diff
@@ -94,6 +94,60 @@ CheckedComponent.propTypes = {
     id: PropTypes.string,
 };
 class TreeContainer extends Component {
+    constructor(props) {
+        super(props);
+
+        this.setProps = this.setProps.bind(this);
+    }
+
+    setProps(newProps) {
```
Marc-Andre-Rivet: We were creating new `setProps` on each TreeContainer render. Everything in here is dependent on the injected props, it can be reused if the instance is reused. Should reduce the number of re-renders, if nothing else.

chriddyp: :tada:

.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ jobs:
             flake8 dash setup.py && flake8 --ignore=E123,E126,E501,E722,E731,F401,F841,W503,W504 --exclude=metadata_test.py tests
             pylint dash setup.py --rcfile=$PYLINTRC
             pylint tests/unit tests/integration/devtools tests/integration/renderer tests/integration/dash_assets -d all -e C0410,C0411,C0412,C0413,W0109
-            cd dash-renderer && npm install --ignore-scripts
+            cd dash-renderer && npm ci
```
alexcjohnson: Does `npm ci` just always ignore scripts? That seems implied, but I can't find it documented anywhere...

@plotly/webpack-dash-dynamic-import/src/index.js
```diff
@@ -10,26 +10,30 @@ function getFingerprint() {
     return `"v${version}m${timestamp}"`;
 }
 
+/**
+ * This code is injected as-is in the webpack artifact and must be ES5 compatible to work in all scenarios.
+ * This will not get transpiled.
```
alexcjohnson: Just to note, because I was curious: it doesn't get transpiled, but it *does* get minified in the pipeline you've got set up - as we would hope 🎉

.circleci/config.yml
Marc-Andre-Rivet: `--ignore-scripts` prevents `prepublish` and `prepare` from being run, these commands are not used anymore in any of the core packages

PR #1015, opened 2019-11-15
:mute: remove the warning as it might confuse non dash usage

### LINKED ISSUE #946: pytest plugin should not be registered by default
Hello 

I suggest that the pytest plugin should not be registered by default on dash installation.

A clean install of pytest without any plugins automatically activates `dash-1.3.1` plugin (which increases the time of test suite by unnecessary import of the module and also annoys with the warning).

On a clean virtual environment:
```
$ python3 -m venv venv
$ source venv/bin/activate
$ pip install dash
$ pip install pytest
$ mkdir tests
$ cd tests/
```
an empty run of pytest results in
```
$ pytest
/Users/user/tmp/venv/lib/python3.7/site-packages/dash/testing/plugin.py:17: UserWarning:

run `pip install dash[testing]` if you need dash.testing

===================================================================== test session starts ======================================================================
platform darwin -- Python 3.7.1, pytest-5.2.0, py-1.8.0, pluggy-0.13.0
rootdir: /Users/user/tmp/tests
plugins: dash-1.3.1
collected 0 items                                                                                                                                              

==================================================================== no tests ran in 0.00s =====================================================================
```

Though the `pytest-dash` plugin was never installed:
```
$ pip uninstall pytest-dash
WARNING: Skipping pytest-dash as it is not installed.
```

byronz:
this PR will fix one part of #946 and added missing `python_requires` in `setup.py`

### DIFF
--- a/dash/testing/plugin.py
+++ b/dash/testing/plugin.py
@@ -1,11 +1,8 @@
 # pylint: disable=missing-docstring,redefined-outer-name
-import warnings
+import pytest
 from .consts import SELENIUM_GRID_DEFAULT
 
-
 try:
-    import pytest
-
     from dash.testing.application_runners import (
         ThreadedRunner,
         ProcessRunner,
@@ -14,7 +11,7 @@ try:
     from dash.testing.browser import Browser
     from dash.testing.composite import DashComposite, DashRComposite
 except ImportError:
-    warnings.warn("run `pip install dash[testing]` if you need dash.testing")
+    pass
 
 WEBDRIVERS = {"Chrome", "Firefox"}
 
@@ -51,8 +48,8 @@ def pytest_addoption(parser):
     dash.addoption(
         "--percy-assets",
         action="store",
-        default='tests/assets',
-        help="configure how Percy will discover your app's assets"
+        default="tests/assets",
+        help="configure how Percy will discover your app's assets",
     )
 
     dash.addoption(
--- a/dash/testing/plugin.py
+++ b/dash/testing/plugin.py
@@ -13,15 +13,13 @@ try:
 except ImportError:
     pass
 
-WEBDRIVERS = {"Chrome", "Firefox"}
-
 
 def pytest_addoption(parser):
     dash = parser.getgroup("Dash", "Dash Integration Tests")
 
     dash.addoption(
         "--webdriver",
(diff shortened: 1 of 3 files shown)


dash/testing/plugin.py
```diff
@@ -14,7 +11,7 @@
     from dash.testing.browser import Browser
     from dash.testing.composite import DashComposite, DashRComposite
 except ImportError:
-    warnings.warn("run `pip install dash[testing]` if you need dash.testing")
```
alexcjohnson: Wouldn't we also need to put all the below definitions into the `try` block, so that without the necessary imports it won't try to define our fixtures?

That said, this still doesn't seem like it completely satisfies the request of #946 - this takes care of the "annoys with the warning" part, but not the "increases the time of test suite by unnecessary import of the module" part. As I said in the issue, it's not clear to me how one can resolve that part, but the ideal seems to me determining whether `setup.py` was called with `[testing]` or not - and only include `"pytest11": ["dash = dash.testing.plugin"]` if yes.

setup.py
```diff
@@ -1,6 +1,20 @@
...
+        "dash-generate-components = "
+        "dash.development.component_generator:cli",
+        "renderer = dash.development.build_process:renderer",
+    ]
+}
+
+# this is not a complete guess, but picking two typical dash dependencies in
+# require-testing.txt
+if pkgutil.find_loader("waitress") and pkgutil.find_loader("percy"):
+    ENTRY_POINTS["pytest11"] = ["dash = dash.testing.plugin"]
```
alexcjohnson: Interesting... but will these exist already if we're installing `dash[testing]` for the first time? Is there no way to directly determine whether the `testing` extra was requested?

byronz: I tried with a clear env, it we install with `dash[testing]` it adds the entry_point as expected. in the case of only `pip install dash`, no pytest-dash is registered. 

I don't think there is an easy way to directly determine if the `testing` extras is used:
- if we install with editable mode, in `dash.egg-info/PKG-INFO`, we can get information like this. but this will depends on how the package is installed. the simple `pip install dash[testing]` does not contain such information. not sure what happens with `conda install`
```
Provides-Extra: dev
Provides-Extra: testing
```

Issue #1005, opened 2019-11-07
[Feature Request] Better python dependency management


byronz:
**Is your feature request related to a problem? Please describe.**
this is mainly happening in the dash app repositories like dash-docs or dash gallery monorepo projects. Given a specific dash release, we don't really lock down the dependencies to a specific version.  Taken the default `requirements` in `require-install.txt` as example
```
Flask>=1.0.2
flask-compress
plotly
dash_renderer==1.2.0
dash-core-components==1.5.0
dash-html-components==1.0.1
dash-table==4.5.0
future
```
we have a loose requirement on Flask, and we use the same txt file in `dash-doc` circleci setup as a cache key.  we have renovate bot setup to some extent, but obviously it's not doing the job to keep dependencies updated in a more precise way.  Our staging environment for dash-docs is deployed using heroku, this is where the problem starts to reveal that it also refers to the same txt file, but without cache, each time it will install a more recent version of flask until some regression bug happens in a new version, and we also have the same bug not detected in an earlier stage (CI), but in heroku.

@cldougl  also have similar problem on on-prem side with the demo apps. 

**Describe the solution you'd like**
we need to at least lock/freeze the version during the dash release process ( @Marc-Andre-Rivet )  in a way that these versions can also be accessible and used by other apps to prevent any dependency conflict or bug happening.

the simplest way can update the requirements files with `pip freeze` during the release as part of the content of the release snapshot (both source code and release artifacts)

**Describe alternatives you've considered**

we can also share the system dependency management tool like poetry, it has a complete command toolchain to do the job from adding the dependency to publishing dash

byronz:
@shammamah feel free to report your incident in dash-bio if you think it's relevant so we can have more incident reports from the dash eco-system.

shammamah:
The issue I was having with https://github.com/plotly/dash-bio/pull/434 is that the pip environment was loading from a cache that was made a couple weeks ago (October 23), so the version of `dash` was never updated. This led to a `ContextualVersionConflict`. 

The 3.6 CI runs went smoothly, but the 3.7 run has an error that is to do with conflicting versions of `dash` (the cached version had `1.4.1`, but the new version should have `1.6.0` as defined in the `requirements.txt` and the `install_requires` for the package). 

[...]

PR #1023, opened 2019-11-22
:alembic: trial on pip-tools

byronz:
1. use pip-compile to pin the deps
2. use makefile to simplify the CI

This PR is trying to solve the #1005 with pip-tools, I found that `poetry` is still not mature enough to be ready for the whole workflow management. the `poetry install` is hard to control and sometimes it misses the dependencies of a dependency, like `flake8` => `entrypoints`. though it's still a solid tool to consider in the future for `dash`, because it can put all the constraints for different python versions in a single place. 

The pip-tools has different approach, it contains `pip-compile` and `pip-sync`. I use the first here to do one simple job, it will upgrade and freeze the version from a simple dependency list file. we can use this tool to lock the dependency in a consistent way.   

## Contributor Checklist

   -  [x] use pip-compile to lock
   -  [x] use makefile to simplify the config.yml

### optionals


setup.py
```diff
@@ -27,7 +27,12 @@ def read_req_file(req_type):
     long_description_content_type="text/markdown",
     install_requires=read_req_file("install"),
     extras_require={
-        "dev": read_req_file("dev"),
+        "dev": read_req_file("dev")
+        + [
```
byronz: this is a compromise of the solution, the `pip-compile` has no flexibility of locking down a specific version based on python version.

Issue #1024, opened 2019-11-22
[BUG] Unsupported format string passed to Button.__format__


astariul:
**Context**

```
dash                 1.6.0
dash-core-components 1.5.0
dash-html-components 1.0.1
dash-renderer        1.2.0
dash-table           4.5.0
dash-daq                  0.2.1
dash-bootstrap-components 0.7.2
```

**Bug description**

When I run my application, I receive following error :

>Traceback (most recent call last):
  File "app.py", line 452, in <module>
    @app.callback(
  File "C:\Users\user\AppData\Local\Programs\Python\Python38\lib\site-packages\dash\dash.py", line 1264, in callback
    self._validate_callback(output, inputs, state)
  File "C:\Users\user\AppData\Local\Programs\Python\Python38\lib\site-packages\dash\dash.py", line 921, in _validate_callback
    if arg_id not in layout and arg_id != layout_id:
  File "C:\Users\user\AppData\Local\Programs\Python\Python38\lib\site-packages\dash\development\base_component.py", line 257, in __iter__
    for t in self._traverse():
  File "C:\Users\user\AppData\Local\Programs\Python\Python38\lib\site-packages\dash\development\base_component.py", line 217, in _traverse
    for t in self._traverse_with_paths():
  File "C:\Users\user\AppData\Local\Programs\Python\Python38\lib\site-packages\dash\development\base_component.py", line 252, in _traverse_with_paths
    for p, t in i._traverse_with_paths():
  File "C:\Users\user\AppData\Local\Programs\Python\Python38\lib\site-packages\dash\development\base_component.py", line 244, in _traverse_with_paths
    "(id={:s})".format(i.id)
**TypeError: unsupported format string passed to Button.__format__**

The error appear on a line where I define this callback :

```
@app.callback(
    Output('reportPid', 'data'),
    [Input('report1Button', 'n_clicks')],
    [State('pid1', 'data')]
)
def report1(n, data):
    return data
```

---

The error seems to be about naming of one of my button.

The weird thing is, if I remove this callback, no error.  
I must be doing something wrong in this callback but I cannot see what, and the error is not helpful...

alexcjohnson:
I suspect somewhere in your layout you have a set of `children` that should be wrapped in an array and the second one is a `Button` - ie
`html.Div(Something(...), Button(...), ...)`
should be
`html.Div([Something(...), Button(...), ...])`
(isn't necessarily a `Div` but you get the point)

[...]

PR #1026, opened 2019-11-22
Children array error

alexcjohnson:
Trying to give a more helpful error message than users encounter today, like #1024, when forgetting to wrap `children` in an array inside components. Obviously if we ever do Components as Props that would need to be taken into account here - but I suspect that would be fine, each component would need to register internally which props can accept components.

### optionals


dash/development/base_component.py
```diff
@@ -94,6 +94,14 @@ def __init__(self, **kwargs):
                         ", ".join(sorted(self._prop_names))
                     )
                 )
+
+            if k != "children" and isinstance(v, Component):
+                raise TypeError(
+                    "Component detected as a prop other than `children`\n" +
+                    "Did you forget to wrap multiple `children` in an array?" +
```
Marc-Andre-Rivet: Either a `\n` or whitespace after the end of the sentence

alexcjohnson: good call -> https://github.com/plotly/dash/pull/1026/commits/4d96e0dbf3c96ec19683c4e68c9e350302cb6da8

Marc-Andre-Rivet:
Diff in https://percy.io/plotly/dash/builds/3264807/view/203019307/1280?mode=diff&browser=firefox&snapshot=203019307 looks related to async / the behavior is fine but without `eager_loading=True` the graph displays locally with a noticeable delay, after the `1` appears on the devtools icon -- the screenshot occurs before it renders.

ohaibbq:
Hey y'all,

We're running into this exception while trying to use multiple props as `PropType.node`. [React's docs suggest](https://reactjs.org/docs/composition-vs-inheritance.html#containment) that if you have multiple pieces of content to dynamically add to a component, you should use your own props scheme rather than `props.children`. i.e. `props.leftChildren` and `props.rightChildren`.

Was this diff supposed to intentionally disallow this case? If not, I think we should add a `PropType` check to ensure that the prop type for `k` is not a `PropType.node` before raising this error.

PR #1027, opened 2019-11-22
Issue 1010 - Non-rendered async components using `asyncDecorator` lock renderer callbacks

### LINKED ISSUE #1010: [BUG] Dash 1.6 Datatable disables all callbacks if not in initial layout
**Describe your context**
```
dash                 1.6.0  
dash-core-components 1.5.0  
dash-html-components 1.0.1  
dash-renderer        1.2.0  
dash-table           4.5.0  
```
Same issue as #998 . 
If DataTable is used in the app but not included in the initial layout callbacks are not working.

Small example using dcc.Tabs below.
Here the first rendered tab has a dcc.Graph which is populated in a callback.
The second tab has a empty DashTable.
With dash > 1.5.1 it is necessary to first navigate to the tab with the DashTable before the callback on the first tab is working.

```python
# -*- coding: utf-8 -*-
import dash
import dash_table

app = dash.Dash()

app.layout = html.Div(
    children=[
        dcc.Tabs(
            children=[
                dcc.Tab(
                    children=[
                        html.Button(id="btn", children="Update graph"),
                        dcc.Graph(id="example-graph"),
                    ]
                ),
                dcc.Tab(children=dash_table.DataTable(id="example-table")),
            ]
        ),
    ]
)

@app.callback(Output("example-graph", "figure"), [Input("btn", "n_clicks")])
def update_graph(clicks):
    return {
        "data": [{"x": [1, 2, 3], "y": [4, 1, 2], "type": "bar"},],
    }

if __name__ == "__main__":
    app.run_server(debug=True)
```
**Describe the bug**

If a dash_table.DataTable is included in the app, but not in the initial layout, app callbacks are not working. After the table has been added to the layout the callbacks works correctly.

**Expected behavior**
The callbacks should work.

Marc-Andre-Rivet:
Fixes #1010 

It is possible for a component in the layout not to be displayed (e.g. tabs) and as such, async resources associated with it will not be loaded. If the component uses the `asyncDecorator` this means that the component could block the renderer's callbacks indefinitely.

This change reduces the scope of the renderer's callback lock by only using the components required as input/state to determine the required lock.

A new test has been added and demonstrated to fail prior to adding the fix (https://github.com/plotly/dash/pull/1027/commits/f5e577e82fe936609b3ffd9c74775d5a7effce13)


dash-renderer/src/actions/setAppReadyState.js
```diff
@@ -35,43 +41,47 @@ const isAppReady = layout => {
...
+                        promises.push(ready);
+                    }
+                });
+            });
+        }
 
-    return promises.length ? Promise.all(promises) : true;
+        return promises.length ? Promise.all(promises) : true;
+    };
 };
```
Marc-Andre-Rivet: Change `isAppReady` to return either the global state of the app or the state for the target ids given.

tests/integration/callbacks/test_basic_callback.py
```diff
@@ -140,3 +142,35 @@ def update_input(value):
...
+            )
+        ]
+    )
+
+    @app.callback(Output("output", "children"), [Input("btn", "n_clicks")])
+    def update_graph(n_clicks):
+        if n_clicks is None:
+            raise PreventUpdate
+
+        return "Bye"
```
Marc-Andre-Rivet: Before the changes in this PR, this callback is never triggered because of the blocking dash-table (uses the asyncDecorator) in the other tab. Now, since the callback only involves `btn` and `output`, the table is non-blocking.

dash-renderer/src/actions/setAppReadyState.js
```diff
@@ -1,87 +1,43 @@
...
-        const id = elementLayout.props && elementLayout.props.id;
-        const children = elementLayout.props && elementLayout.props.children;
-        const namespace = elementLayout.namespace;
-        const type = elementLayout.type;
-
-        components[namespace] = components[namespace] || {};
-        components[namespace][type] = type;
+const isAppReady = (layout, paths) => targets => {
+    if (!layout || !paths || !Array.isArray(targets)) {
+        return true;
```
alexcjohnson: Is there a use case for this? Feels like an error rather than a silent success.

dash-renderer/src/actions/setAppReadyState.js
```diff
@@ -1,87 +1,43 @@
...
-        }
-
-        const id = elementLayout.props && elementLayout.props.id;
-        const children = elementLayout.props && elementLayout.props.children;
-        const namespace = elementLayout.namespace;
-        const type = elementLayout.type;
-
-        components[namespace] = components[namespace] || {};
-        components[namespace][type] = type;
+const isAppReady = (layout, paths) => targets => {
```
alexcjohnson: As long as we get the order of operations right (ie always updating `isAppReady` after any update to `layout` or `paths`) this should be fine, but it would be more robust and simpler - without any cost anymore - to not put this in the state at all.

alexcjohnson: ie to just make it a function `(layout, paths, targets) => {...}`

Marc-Andre-Rivet: exactly the problem making the tests fail, the store evals `IS_APP_READY` before `COMPUTE_PATH` and the first calls ends up not blocking

dash-renderer/src/actions/setAppReadyState.js
Marc-Andre-Rivet: Relates to the comment below. Refactoring to take `isAppReady` outside of the state, timing issues will be moot.

tests/integration/test_render.py
```diff
@@ -972,11 +972,11 @@ def render_content(tab):
         )[0].click()
 
         graph_1_expected_clickdata = {
-            "points": [{"curveNumber": 0, "pointNumber": 1, "pointIndex": 1, "x": 2, "y": 10}]
+            "points": [{"curveNumber": 0, "pointNumber": 1, "pointIndex": 1, "x": 2, "y": 10, "label": 2, "value": 10}]
         }
 
         graph_2_expected_clickdata = {
-            "points": [{"curveNumber": 0, "pointNumber": 1, "pointIndex": 1, "x": 3, "y": 10}]
+            "points": [{"curveNumber": 0, "pointNumber": 1, "pointIndex": 1, "x": 3, "y": 10, "label": 3, "value": 10}]
```
Marc-Andre-Rivet: @alexcjohnson Getting the same error you have in #1026 -- there seems to be additional label/value props

Marc-Andre-Rivet: Added https://github.com/plotly/dash/issues/1031 for follow up

PR #1035, opened 2019-12-02
Simplify build

alexcjohnson:
A few changes to get our webpack config a little more manageable...

- `raw-loader` was only used for the werkzeug css - I just wrapped that in js to export a string explicitly.
- `uniqid` wasn't even being used, AFAICT. We were generating them and ignoring them. Probably was intended at some point for being able to clear errors. But that's something we've removed the vestiges of, at least for the time being. This is one more.
- `check-prop-types` hasn't been maintained recently, and anyway if given the opportunity (by bringing it into our own src) we would make some slightly different choices anyway - the one I made already was returning *all* prop errors, not just the first one.

### DIFF
--- a/dash-renderer/src/checkPropTypes.js
+++ b/dash-renderer/src/checkPropTypes.js
@@ -1,14 +1,13 @@
 /*
- * Copied out of prop-types and modified - similar to check-prop-types, but
- * simplified and tweaked to our needs, also so we don't need a special
- * transpiling inclusion for this in node_modules - check-prop-types hasn't
- * been modified in years and yet includes
+ * Copied out of prop-types and modified - inspired by check-prop-types, but
+ * simplified and tweaked to our needs: we don't need the NODE_ENV check,
+ * we report all errors, not just the first one, and we don't need the throwing
+ * variant `assertPropTypes`.
  */
 import ReactPropTypesSecret from 'prop-types/lib/ReactPropTypesSecret';
 
 /**
  * Assert that the values match with the type specs.
- * Error messages are memorized and will only be shown once.
  *
  * @param {object} typeSpecs Map of name to a ReactPropType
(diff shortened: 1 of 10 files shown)


dash-renderer/src/checkPropTypes.js
```diff
@@ -0,0 +1,89 @@
...
+                var stack = (getStack && getStack()) || '';
+
+                errors.push(
+                    'Failed ' + location + ' type: ' + error.message + stack
+                );
+            }
+        }
+    }
+    return errors.join('\n\n');
+}
```
Marc-Andre-Rivet: Can you explain the additional changes that have been made to make it `simplified and tweaked to our needs`. Arguably, the source file hasn't been changed in a long time either and it might be no worse to have our own copy than it is to depend on someone else's copy.. 

Since the original file is under a BSD license, we should make sure that we either retain the original disclaimer as done in check-prop-type (https://github.com/ratehub/check-prop-types/blob/master/index.js#L7) or make sure the disclaimer for React is included explicitly in our bundled code -- which it wouldn't right now as we externalize React dependencies.

Marc-Andre-Rivet:
One licensing comment for check-prop-types, otherwise 👍

dash-renderer/src/checkPropTypes.js
alexcjohnson: Current `prop-types` [license is MIT](https://github.com/facebook/prop-types/blob/master/LICENSE) (it was changed from BSD shortly after check-prop-types was published), as is [check-prop-types](https://github.com/ratehub/check-prop-types/blob/master/LICENSE) - that presumably means this concern is moot, yes?

Marc-Andre-Rivet: Moot as moot can be

PR #1048, opened 2019-12-13
Provide string interpolation support for async/dynamic dep attributes in R

### LINKED ISSUE #1002: Need async metadata support when generating R versions of component libraries
Currently, Dash for Python leverages `__init__.py` to identify whether a package has resources which should be loaded asynchronously, e.g.

https://github.com/plotly/dash-core-components/blob/0e1449813c620adfd337568cee8cc55292d76c02/dash_core_components_base/__init__.py#L42-L48

In Dash for R, the `internal.R` script provides much of the same information, but currently lacks metadata related to async resources:

https://github.com/plotly/dash-core-components/blob/master/R/internal.R

The package generator should be modified to insert the requisite details about async resources when appropriate.

@Marc-Andre-Rivet @alexcjohnson

rpkyle:
This PR proposes to add support for generating `async` or `dynamic` attributes within the JavaScript metadata Dash for R uses to manage dependencies.

Additionally, improvements to the package generator have been introduced:
- R package descriptions and titles are now populated using `dash-info.yaml` first, and then from information in `package.json` when it is unavailable or the relevant keys are missing
- R package dependency versions (`.js`, `.map`, `.css`) are now properly generated to match versioning within the library of origin
- if a `vignettes` directory is present, the generator will add a `VignetteBuilder` entry to `DESCRIPTION` and append `knitr, rmarkdown` to the `Suggests` field, but it will not build the vignettes itself (that requires running `R CMD build` and is out of scope at this time)

Closes #1002. Fixes plotly/dash-core#69, resolves plotly/dash-core#71.


dash/development/_r_components_generation.py
```diff
@@ -296,23 +301,30 @@ def generate_js_metadata(pkg_data, project_shortname):
                     project_shortname=project_shortname,
                     script_name=script_name,
                     css_name=css_name,
+                    async_or_dynamic=async_or_dynamic,
                 )
             ]
             function_frame_body = ",\n".join(function_frame)
     elif len(alldist) == 1:
-        rpp = alldist[0]["relative_package_path"]
+        dep = alldist[0]
```
Marc-Andre-Rivet: Not part of this change, but I'm having trouble understanding why we process an array of length one differently from an array of length N

dash/development/_r_components_generation.py
```diff
@@ -322,6 +334,24 @@ def generate_js_metadata(pkg_data, project_shortname):
     return function_string
 
 
+# determine whether dependency uses async or dynamic flag
+# then return the properly formatted string if so, i.e.
+# " async = TRUE,". a dependency can have async or
+# dynamic elements, neither of these, but never both.
+def check_async_type(dep):
```
Marc-Andre-Rivet: Would suggest renaming to `get_**` instead as this doesn't check but rather returns some specific value to use as-is

dash/development/_r_components_generation.py
```diff
@@ -509,7 +539,17 @@ def generate_rpkg(
 
     package_name = snake_case_to_camel_case(project_shortname)
     lib_name = pkg_data.get("name")
-    package_description = pkg_data.get("description", "")
+
+    if rpkg_data is not None:
+        if rpkg_data.get("pkg_help_title"):
+            package_title = rpkg_data.get("pkg_help_title",
+                                          pkg_data.get("description",
+                                                       ""))
```
Marc-Andre-Rivet: Is the extra return on purpose?

Generating for the table I now get extra empty lines:
```
Package: dashTable
Title: Core Interactive Table Component for Dash

Version: 4.5.1
Description: Dash DataTable is an interactive table component designed for designed for viewing, editing, and exploring large datasets. DataTable is rendered with standard, semantic HTML <table/> markup, which makes it accessible, responsive, and easy to style. This component was written from scratch in React.js specifically for the Dash community. Its API was designed to be ergonomic and its behaviour is completely customizable through its properties.

Depends: R (>= 3.0.2)
Imports: 
Suggests: 
License: MIT + file LICENSE
URL: https://github.com/plotly/dash-table
BugReports: https://github.com/plotly/dash-table/issues
Encoding: UTF-8
LazyData: true
Author: Chris Parmer [aut]
Maintainer: Ryan Patrick Kyle <ryan@plot.ly>

```

rpkyle: The funky formatting of the lines from 544:547 is to satisfy the linting overlords.

I'm not seeing the extra linefeeds you have above, but that's probably because I edited my `dash-info.yaml` files such that the `>` operator has been replaced with the "chomping" operator `>-`. I've since back-committed them to `dev` via new PRs, so hopefully you'll generate the same `DESCRIPTION` I have committed to this and the other two core component repositories.

package-lock.json
```diff
@@ -0,0 +1,14115 @@
+{
```
Marc-Andre-Rivet: This looks like 🔪  - there's no corresponding `package.json` file.

dash/development/_r_components_generation.py
```diff
@@ -369,6 +399,18 @@ def write_help_file(name, props, description, prefix, rpkg_data):
...
+        print(
+            "Error, aborting R package generation: "
+            "percent signs cannot be used in docstrings "
+            "without escaping with a backslash, e.g; \\%.\n"
+            "Please omit or escape any percent signs in "
+            "your JavaScript source before rebuilding the "
+            "R package.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
alexcjohnson: As discussed on slack: javascript docstrings should NOT contain escapes for % characters, that would look weird in JS (and in Python). So the right way to handle this is to automatically insert the escapes when translating from JS to R.

rpkyle: fixed in 52a3394ad4f07b9bd92d1eec9e4ff4ca29b4ef63

package-lock.json
rpkyle: fixed in 1d5a1121d5bd2137aa28d26d35d6aa62f0d37d4b

dash/development/_r_components_generation.py
rpkyle: fixed in 3ff7607bec58bc128d4367669d31f121cc49bbb4

dash/development/_r_components_generation.py
rpkyle: The code doesn't really make much sense as written, but that's because I had trouble getting it to work with `enumerate` when I wrote it a year ago. I had only been programming in Python for a week or two, so this code is pretty much a giant pile of 💩. It needs some love and refactoring, no disagreement there.

Marc-Andre-Rivet:
The tests failures we are seeing here are caused by a regression from the dcc.Graph responsiveness -- not related to this PR (https://circleci.com/gh/plotly/dash/13778#tests/containers/1)

rpkyle:
I'm going to add one last feature here -- to scrub examples from docstrings in packages such as DDK when generating the R packages, will relocate them to `dash-info.yaml` (at least for R, can leave them as-is for Python users).

rpkyle:
> I'm going to add one last feature here -- to scrub examples from docstrings in packages such as DDK when generating the R packages, will relocate them to `dash-info.yaml` (at least for R, can leave them as-is for Python users).

fixed in b471d35

PR #1078, opened 2020-01-10
Issue 481 - Support arbitrary file extensions in component suites

### LINKED ISSUE #481: Support arbitrary file extensions
Dash components are packaged with webpack and it's natural to end up having standalone files in webpack bundled projects. For example, static images, fonts or icons would typically not be bundled inside the main bundle.js when they exceed a certain size.

This issue relates to
#96
#480 

Using a combination of ```mimetypes``` as suggested here (https://github.com/plotly/dash/pull/478#discussion_r239234446), explicit overrides (e.g. .map files) and defaulting to no content type (or application/octet-stream) (https://github.com/plotly/dash/pull/478#discussion_r239458219) would allow Dash to expose arbitrary files in a predictable fashion, while providing default behavior for unknowns (some clients will be able to figure out the type when defaulted, some will not)

For known types (mimetypes, map) we could add this additional header
X-Content-Type-Options=nosniff
to prevent the browser from trying to figure out the content type by itself (the browser can sometimes use this to override the default -- content-type)

Additional possible improvement: https://github.com/plotly/dash/issues/480#issuecomment-445039469

Marc-Andre-Rivet:
Closes #481 

Add arbitrary file support for component suite.

This is being done with the idea of packaging fonts with the components that require them without the need for adding assets or loading everything upfront as base64 inside css files. Images would be handled similarly, also with the `file-loader`.

Once added, component repos using webpack can use the following loader configuration:

```
{
    test: /\.(woff|woff2|eot|svg|ttf)$/,
    use: [{
        loader: 'file-loader',
        options: {
            limit: 0, // no resource gets inlined in the js bundle
            esModule: false, // this option is required for file-loader >= 5.0.0
            name: '[name].[ext]'
        }
    }]
}
```

And include source resources in css files like so:
```
@font-face {
    font-family: 'My-Font';
    src: url(./My-Font.ttf) format('truetype');
}
```

`url(./My-Font.ttf)` will be resolved to the correct path by the same plugin used for async (https://github.com/plotly/dash/tree/dev/%40plotly/webpack-dash-dynamic-import)

And include the resource in the component's `__init__.py` (and MANIFEST):
```
{
    'relative_package_path': 'My-Font.ttf',
    'external_url': (
       'https://unpkg.com/my-dash-components@{}'
        '/my_dash_component/My-Font.ttf'
    ).format(__version__),
    'namespace': 'my_dash_component',
    'dynamic': True
},
```

Adding `dynamic:True` will ensure that the file is only loaded when requested by the css (e.g. async scenarios).

Do note that as implemented, these resources will not be fingerprinted and will default to using `eTag` for caching purposes.


dash/dash.py
```diff
@@ -703,13 +707,10 @@ def serve_component_suites(self, package_name, path_in_package_dist):
...
 
-        mimetype = (
-            {
-                "js": "application/javascript",
-                "css": "text/css",
-                "map": "application/json",
-            }
-        )[path_in_package_dist.split(".")[-1]]
+        resource_path = os.path.dirname(sys.modules[package_name].__file__)
+        full_path = os.path.join(resource_path, path_in_package_dist)
```
Marc-Andre-Rivet: Will update this implementation to get the mapped mimetype from the `mimetypes` mapping instead of requiring the fullpath.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [1.8.0] - 2020-01-14
 ### Added
+- [#1078](https://github.com/plotly/dash/pull/1078) Add support for arbitrary file extensions in component packages
```
rpkyle:
```suggestion
- [#1078](https://github.com/plotly/dash/pull/1078) Permit usage of arbitrary file extensions for assets within component libraries
```

wbrgss:
This PR is closer than I thought; I can load custom fonts. The CSS and JS files are being loaded too, but the CSS is not being applied as per https://github.com/plotly/dash/pull/1078#issuecomment-576370388. That is, the rules are not being applied to the elements they select, even though the CSS files *are* there.

Marc-Andre-Rivet:
@wbrgss Fixed some issues - can you give it another go?

dash/development/_r_components_generation.py
```diff
@@ -505,7 +505,7 @@ def write_js_metadata(pkg_data, project_shortname, has_wildcards):
         for filename in filenames:
             extension = os.path.splitext(filename)[1]
 
-            if extension not in [".css", ".js", ".map"]:
+            if extension in [".py", ".pyc", ".json"]:
```
Marc-Andre-Rivet, line 508: Pick up everything but the Python artifacts

tests/integration/test_generation.py
```diff
@@ -18,3 +23,39 @@ def test_gene001_simple_callback(dash_duo):
...
+
+    assert (
+        dash_duo.driver.execute_script("return document.fonts.check('1em godfather')")
+        is False
+    )
+
+    dash_duo.wait_for_element("#btn").click()
+    assert dash_duo.wait_for_element("#standard").text == "Standard"
+
+    WebDriverWait(dash_duo.driver, 10).until(
```
alexcjohnson, line 54: Is this better than `dash.testing.wait.until`? It's barely any more complicated so fine to keep it, just curious if there's a difference.

Regardless, very nice test!

Marc-Andre-Rivet: Wasn't aware I could do that. I thought it was either this of `dash_duo._wait_for`, will keep in mind for the future.

PR #1080, opened 2020-01-11
feat: disable cookie access under restricted sandboxes

josegonzalez:
When dash is embedded into an iframe with a sandbox attribute that only has allow-scripts, cookie access is disabled and dash fails to load. As such, we need to restrict our cookie usage by disabling functionality.

This patch removes the disabled functionality in a graceful manner, allowing dash to load in very restricted iframes.

## Contributor Checklist

- [x] ~I have broken down my PR scope into the following TODO tasks~ not needed

### optionals

### DIFF
--- a/dash-renderer/src/AccessDenied.react.js
+++ b/dash-renderer/src/AccessDenied.react.js
@@ -1,9 +1,12 @@
 /* global window:true, document:true */
 import React from 'react';
-import {mergeRight} from 'ramda';
+import {mergeRight, once} from 'ramda';
 import PropTypes from 'prop-types';
 import * as styles from './styles/styles.js';
 import * as constants from './constants/constants.js';
+
+/* eslint-disable-next-line no-console */
+const logWarningOnce = once(console.warn);
 
 function AccessDenied(props) {
     const {config} = props;
@@ -34,8 +37,7 @@ function AccessDenied(props) {
                             `${constants.OAUTH_COOKIE_NAME}=; ` +
                             'expires=Thu, 01 Jan 1970 00:00:01 GMT;';
                     } catch (e) {
-                        /* eslint-disable-next-line no-console */
-                        console.warn(e);
+                        logWarningOnce(e);
                     }
                     window.location.reload(true);
                 }}
--- a/dash-renderer/src/AccessDenied.react.js
+++ b/dash-renderer/src/AccessDenied.react.js
@@ -28,9 +28,15 @@ function AccessDenied(props) {
             <a
                 style={styles.base.a}
                 onClick={() => {
-                    document.cookie =
-                        `${constants.OAUTH_COOKIE_NAME}=; ` +
-                        'expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+                    /* eslint no-empty: ["error", { "allowEmptyCatch": true }] */
+                    try {
+                        document.cookie =
+                            `${constants.OAUTH_COOKIE_NAME}=; ` +
+                            'expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+                    } catch (e) {
+                        /* eslint-disable-next-line no-console */
+                        console.warn(e);
+                    }
                     window.location.reload(true);
                 }}
             >
--- a/dash-renderer/src/actions/index.js
+++ b/dash-renderer/src/actions/index.js
@@ -55,9 +55,15 @@ export function hydrateInitialOutputs() {
 }
 
 export function getCSRFHeader() {
-    return {
-        'X-CSRFToken': cookie.parse(document.cookie)._csrf_token,
-    };
+    try {
+        return {
+            'X-CSRFToken': cookie.parse(document.cookie)._csrf_token,
+        };
+    } catch (e) {
+        /* eslint-disable-next-line no-console */
+        console.warn(e);
+        return {};
+    }
 }
 
 function triggerDefaultState(dispatch, getState) {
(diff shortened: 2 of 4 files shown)


josegonzalez:
Not sure how to fix the `Empty block statement` error...

alexcjohnson:
> Not sure how to fix the `Empty block statement` error...

May be worth `console.log`-ing the error anyway, could be useful to sandbox users for debugging. That would fill the block :)

josegonzalez:
heh:

```
Unexpected console statement  no-console
```

josegonzalez:
@alexcjohnson I've ignored the prettier issue in code for now since there isn't a great way to handle it otherwise.

Do you have any thoughts on how I should write a test for the code? The change itself was manually tested and confirmed working within a large-ish dash app, so I'm fairly confident it works, but I'm not sure how to write tests for the change...

alexcjohnson:
> how I should write a test for the code?

The most robust thing I can think of is to make a container html file (in an `assets/` folder next to the test file so dash will serve it) with the actual dash app in a sandboxed iframe - then go interact with the app in the iframe - I've never done that but it looks manageable http://allselenium.info/handling-iframes-using-selenium-webdriver/

josegonzalez:
@alexcjohnson once tests pass (I rebased everything together, they previously passed) this should be good for a re-review/merge :)

dash-renderer/src/AccessDenied.react.js
```diff
@@ -28,9 +28,12 @@ function AccessDenied(props) {
...
                 onClick={() => {
-                    document.cookie =
-                        `${constants.OAUTH_COOKIE_NAME}=; ` +
-                        'expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+                    /* eslint no-empty: ["error", { "allowEmptyCatch": true }] */
+                    try {
+                        document.cookie =
+                            `${constants.OAUTH_COOKIE_NAME}=; ` +
+                            'expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+                    } catch (e) {}
```
Marc-Andre-Rivet: Between an eslint exception allowing error swallowing and an eslint exception for a `console.warn` entry, I'd rather have the warning. Dash will be operating with limited capabilities and the information should be exposed. Maybe later we can approve it with https://github.com/plotly/dash/issues/1088

dash-renderer/src/actions/index.js
```diff
@@ -55,9 +55,13 @@ export function hydrateInitialOutputs() {
...
 export function getCSRFHeader() {
-    return {
-        'X-CSRFToken': cookie.parse(document.cookie)._csrf_token,
-    };
+    try {
+        return {
+            'X-CSRFToken': cookie.parse(document.cookie)._csrf_token,
+        };
+    } catch (e) {
+        return {};
```
Marc-Andre-Rivet: Same as above.

Marc-Andre-Rivet:
@josegonzalez I suppose that test failure was predictable.. we now log a warning.. and the test picks it up! You'll want to update `assert not dash_duo.get_logs()` accordingly

tests/integration/renderer/test_iframe.py
```diff
@@ -0,0 +1,51 @@
...
+    <iframe src="{0}" sandbox="allow-scripts">
+    </iframe>
+    </html>
+    """
+
+    html_content = iframe.format(dash_duo.server_url)
+
+    dash_duo.driver.get("data:text/html;charset=utf-8," + html_content)
+
+    assert not dash_duo.get_logs()
```
Marc-Andre-Rivet: @josegonzalez I suppose that test failure was predictable.. we now log a warning.. and the test picks it up! You'll want to update `assert not dash_duo.get_logs()` accordingly

release
```diff
@@ -0,0 +1,20 @@
...
+cd dash-renderer
+rm -rf dist node_modules dash_renderer*
+# build renderer bundles, this will build all bundles from source code
+# the only true source of npm version is defined in package.json
+npm run build  # or `renderer build`
+# install dash-renderer for development
+pip install -e .
+# you should see both dash and dash-renderer are pointed to local source repos
+pip list | grep dash
+python setup.py sdist
```
Marc-Andre-Rivet: Can you explain the goal here?

tests/integration/renderer/test_iframe.py
josegonzalez: Yeah, the other failures i was talking about previously were eslint errors, but i got them covered.

release
josegonzalez: My fault. I wrote this script so I could run the release process in docker so I could build the package locally.

tests/integration/renderer/test_iframe.py
```diff
@@ -0,0 +1,51 @@
...
+    <iframe src="{0}" sandbox="allow-scripts">
+    </iframe>
+    </html>
+    """
+
+    html_content = iframe.format(dash_duo.server_url)
+
+    dash_duo.driver.get("data:text/html;charset=utf-8," + html_content)
+
+    assert len(dash_duo.get_logs()) == 2
```
Marc-Andre-Rivet: @josegonzalez The `==2` part is brittle but ok for now I think -- if you could open an issue to improve this part of the test later, I think I'm fine as-is. What I would like so see here is some validation that the Dash app actually loaded and works as expected. Maybe select the `#btn`, click it, and make sure at least one of the divs got updated by the callback.

PR #1083, opened 2020-01-12
☑️ Fix Pylint for the tests folder.

jbampton:
Add both a .pylintrc and a .pylintrc37 file to the tests folder.

Enable the classic superfluous-parens on Python 3 to test the difference between Python 2 and 3.

Lint some Python.

Pylint has now run twice.

pylint-ran-twice


alexcjohnson:
Thanks @jbampton - superseded by #1151

Issue #1084, opened 2020-01-13
[BUG] Multi-output leads to callback results being ignored


slishak:
**Describe your context**
```
dash                    1.7.0
dash-core-components    1.6.0
dash-html-components    1.0.2
dash-renderer           1.2.2
dash-table              4.5.1
```
**Describe the bug**

When using multi-outputs in the following code snippet (set `MULTI = True`), the labels are not reliably updated when clicking the button to update the slider values.

To reproduce:
 - Start the app
 - Click the button to set both slider values to 5
 - Change both slider values manually to a different number. Both labels update correctly
 - Click the button again. Only the second label is updated, not the first, despite both `update_slider1_label` and `update_slider2_label` being called

```python
import dash

import dash_html_components as html
import dash_core_components as dcc
from dash.dependencies import Output, Input
from dash.exceptions import PreventUpdate

APP = dash.Dash(__name__)
APP.layout = html.Div([
    html.Div([
        html.Button('Button', id='button'),
    ]),
    html.Div([
        html.Label(id='label1'),
        dcc.Slider(id='slider1', min=0, max=10, value=0),

    ]),
    html.Div([
        html.Label(id='label2'),
        dcc.Slider(id='slider2', min=0, max=10, value=0),
    ])
])


MULTI = True

if MULTI:
    @APP.callback(
        [Output('slider1', 'value'),
         Output('slider2', 'value')],
        [Input('button', 'n_clicks_timestamp')]
    )
    def update_slider_vals(n_clicks_timestamp):
        if not n_clicks_timestamp:
            raise PreventUpdate
        return 5, 5
else:
    @APP.callback(
        Output('slider1', 'value'),
        [Input('button', 'n_clicks_timestamp')]
    )
    def update_slider1_val(n_clicks_timestamp):
        if not n_clicks_timestamp:
            raise PreventUpdate
        return 5

    @APP.callback(
        Output('slider2', 'value'),
        [Input('button', 'n_clicks_timestamp')]
    )
    def update_slider2_val(n_clicks_timestamp):
        if not n_clicks_timestamp:
            raise PreventUpdate
        return 5


@APP.callback(
    Output('label1', 'children'),
    [Input('slider1', 'value')]
)
def update_slider1_label(val):
    ret = f'Slider1 value {val}'
    print(ret)
    return ret


@APP.callback(
    Output('label2', 'children'),
    [Input('slider2', 'value')]
)
def update_slider2_label(val):
    ret = f'Slider2 value {val}'
[...]

Issue #1071, opened 2020-01-08
[BUG] UI components not updating from dash_table connected callback


tsiq-bertram:
**Describe your context**
```
dash                 1.7.0
dash-core-components 1.6.0
dash-html-components 1.0.2
dash-renderer        1.2.2
dash-table           4.5.1
```

- macOS Catalina Version 10.15.2
- Chrome Version 79.0.3945.88

**Describe the bug**

The issue is not apparent in dash 1.6.1, but appeared in 1.7.0.
For certain callback graphs, UI components with callbacks connected to a dash_table will not update their state.

Example callback graph and code:
<img width="480" alt="dash_callback_graph" src="https://user-images.githubusercontent.com/41931276/71971577-cb476a80-3202-11ea-9af3-2a42daa7884e.png">

```
import dash
import dash_html_components as html
import dash_table
from dash.dependencies import Input, Output
import random


app = dash.Dash(__name__)
app.layout = html.Div(
    [
        html.Div(id="level-1_1"),
        html.Div(id="level-1_2"),
        html.Div(id="level-2_1"),
        html.H1(id="level-2_2"),
        html.Button("Update", id="button"),
        dash_table.DataTable(id="table"),
    ],
)


@app.callback(
    # Changing the order of outputs here fixes the issue
    [Output("level-1_2", "children"), Output("level-1_1", "children")],
    [Input("button", "n_clicks")],
)
def level1(n_clicks):
    if n_clicks is None:
        return "initializing", "initializing"
    return "level-1_2", "level-1_1"


@app.callback(
    Output("level-2_1", "children"), [Input("level-1_1", "children")],
)
def level2_1(data):
    if data is None:
        return "initializing"
    return "level-2_1"


@app.callback(
    Output("level-2_2", "children"),
    [Input("level-1_2", "children"), Input("table", "selected_cells")],
)
def level2_2(data, selected_cells):
    if data is None:
        return "initializing"
    return "hello " + str(random.randint(0,100))

if __name__ == "__main__":
    app.run_server(debug=True)
```

**Expected behavior**

The UI displays the H1 element with "hello" and a random number. When clicking the "Update" button, the H1 element will be updated.

**Actual behavior and investigation**

The H1 UI element nondeterministically displays either "hello XX" or "initializing" (see screenshot below) on startup. Clicking the update button does not change the H1 element. This is despite the browser receiving the correct callback response from the backend on `_dash-update-component`, e.g.
```
[...]

Issue #1053, opened 2019-12-19
[BUG] Callback with canceled update intermittently prevents later callback


alexcjohnson:
If I make a callback chain where a final output depends on an input via multiple paths, a `PreventUpdate` on one of those paths can intermittently prevent the final callback from ever firing.
```py
import dash
import dash_core_components as dcc
import dash_html_components as html
from dash.dependencies import Input, Output
from dash.exceptions import PreventUpdate

app = dash.Dash(__name__)
app.layout = html.Div([
    dcc.Input(id='a', value="x"),
    html.Div('b', id='b'),
    html.Div('c', id='c'),
    html.Div(id='out')
])

@app.callback(
    Output("out", "children"),
    [Input("a", "value"), Input("b", "children"), Input("c", "children")]
)
def set_out(a, b, c):
    print("out")
    return "{}/{}/{}".format(a, b, c)

@app.callback(Output("b", "children"), [Input("a", "value")])
def set_b(a):
    print("b")
    raise PreventUpdate

@app.callback(Output("c", "children"), [Input("a", "value")])
def set_c(a):
    print("c")
    return a

if __name__ == "__main__":
    app.run_server(debug=True)
```

Both on initial load and later as you type in the input box, sometimes you'll see the last line correctly concatenating the three lines before it, sometimes it lags behind. I don't notice a pattern, this happens anywhere from about a quarter to three quarters of the time.

I'll try to sort this out along with wildcards #475

Issue #635, opened 2019-03-08
Callbacks called multiple times unnecessarily


slishak:
Firstly - thanks so much for the new callback context and multi-output features added recently. They have significantly improved the development experience!

In the example app below, I'm using both of the new features. When a button is clicked, both sliders are updated from the same callback, and then the graph is updated based on the sliders. It all works fine, but on every button click the plot callback is actually triggered twice (once for each slider) even though both sliders are updated "simultaneously". 

Is there a way to prevent this? In my real app, the plot callback is potentially quite heavy (represented by the `time.sleep`) and there are more than 2 sliders linked to it, so I'd prefer to only call it once when all of the sliders are updated by one multi-output callback.

```python
from time import sleep

import dash
from plotly import graph_objs as go

import dash_html_components as html
import dash_core_components as dcc
from dash.dependencies import Output, Input, State
from dash.exceptions import PreventUpdate

app = dash.Dash(__name__)
app.scripts.config.serve_locally = True

app.layout = html.Div([
    html.Button('Button 1', id='b1'),
    html.Button('Button 2', id='b2'),

    dcc.Slider(id='slider1', min=-5, max=5),
    dcc.Slider(id='slider2', min=-5, max=5),

    dcc.Graph(id='graph')

])


@app.callback(
    [Output('slider1', 'value'),
     Output('slider2', 'value')],
    [Input('b1', 'n_clicks'),
     Input('b2', 'n_clicks')]
)
def update_sliders(button1, button2):

    if not dash.callback_context.triggered:
        raise PreventUpdate

    triggered = dash.callback_context.triggered[0]['prop_id'].split('.')[0]

    if triggered == 'b1':
        return -1, -1
    else:
        return 1, 1


@app.callback(
    Output('graph', 'figure'),
    [Input('slider1', 'value'),
     Input('slider2', 'value')]
)
def update_graph(s1, s2):

    sleep(1)

    print('Called x={}, y={}'.format(s1, s2))

    scatter = go.Scatter(x=[s1], y=[s2], mode='markers')
    layout = go.Layout(xaxis={'range': [-5, 5]}, yaxis={'range': [-5, 5]})
    fig = go.Figure(data=[scatter], layout=layout)

    return fig


if __name__ == '__main__':
    app.run_server(debug=True, port=6060, threaded=False)
```

```
dash==0.39.0
dash-core-components==0.44.0
dash-html-components==0.14.0
dash-renderer==0.20.0
```

PR #1103, opened 2020-01-30
475 Pattern Matching Callbacks (codenamed wildcards)

### LINKED ISSUE #475: Dynamic Layouts (Wildcard Props)
It would be really powerful if we could have the layout of an app, not just individual component properties, respond to state changes. Some use cases that come to mind from previous (non-dash) apps I've made:
- Creating an arbitrary number of graphs, based on a multi-select dropdown
- Chained filters to a database query, where all previous filters determine the options available to the next filter
- You might want a new component (or set of components) for each item in some query results. Perhaps each items gets its own graph, or each gets a full report page.

There would be one new piece of user API: wildcards in property callbacks. But first some notes about callbacks affecting the layout (via `children`)

Thoughts @plotly/dash ?

## Layout callbacks

(edit: we can already do this by having a callback output the `children` of some component - So these are just regular `app.callback`, I've updated the pseudocode to reflect this but there are some notes I wanted to add)

```py
app.layout = html.Div([
    dcc.Dropdown(id='dd', multi=True, ...),
    html.Div(id='graphs')
])

@app.callback(Output('graphs', 'children'), [Input('dd', 'value')])
def graphs(selected):
    return html.Div([graph(val) for val in selected])

def graph(val):
    return dcc.Graph(id='graph-{}'.format(val), figure={...})
```

That example uses `val` to construct the id. Importantly, `Graph` elements [set the `key` prop to match `id`](https://github.com/plotly/dash-core-components/blob/6558624125505547ee6702d928852ef428c9349b/src/components/Graph.react.js#L240) independent of their original location, which means an existing `Graph` will be reused independent of its position among siblings, which both improves performance and would preserve any GUI interactions on the graph for the original item even if a new item is added before it.

Two related thoughts:
- For dynamic layout support, should we extend `key=id` to all core components? It doesn't look like most of them do this yet, though one or two do. `id` is tied to the meaning of the component via callbacks so I can't see any downside to keeping this correspondence everywhere, anyone disagree?
- `dash_html_components` allows users to separately specify `id` and `key` for all elements. That's probably fine, but the dynamic layout docs should highlight importance `key`. For example if you make a larger `html.Div` of a graph and some controls surrounding it, you'd like that whole item to be reused.

There's at least one very common use case that on its surface requires a circular dependency, which would be nice to avoid: "remove this item", like an ❎ button to remove a pane. For the case of creating panes from a dropdown or other list, we could just say "don't do that" (since this would be two separate controls for the same thing, and you can always deselect the item in the list where you selected it). But in other cases you just have an "add new item" button (so you'd use `n_clicks` for that button, and `State` for the existing list, to add new items). I suppose we could have the ❎ button just permanently hide the item, and you need to check the item visibility in the rest of your callbacks? Any better solution to this case? I suppose it's *possible* that we could just plumb it up to have the ❎ delete its own parent... maybe that won't get flagged as a circular dep, though in principle it is one.

## Wildcards in regular (property) callbacks

When we have these repeating items, a single callback will need to manage a whole class of output, and some callbacks will want to use an entire class or a subset of that class as an input. I propose a wildcard syntax based on `str.format` that can handle integers and strings, and returns the matched values in `kwargs`.
- Single items can be used in `Output`, this creates one set of dependencies for each matching item. If an identifier has already been used in `Output`, that same identifier can be used in `Input` or `State` too.
  - Single integer: `Output('graph-{n:d}', 'figure')` -> `kwargs['n'] = int`
  - Single string: `Output('info-{item:s}-summary', 'children')` -> `kwargs['item'] = str`
    (should we make `:s` the default, so you only need to specify it if you want an integer match?)
- Multi-item matches can be used only in `Input` and `State`, and the corresponding `args` (and `kwargs`) will be lists. I suppose if you have two multi-item matches in a single id you would get a nested list...
  - All integers: `Input('text-input-{*m:d}', 'value')` -> `kwargs['m'] = list(int)`
  - All previous integers: `Input('dropdown-{*m<n:d}', 'value)` -> `kwargs['m'] = list(int)`
    (requires `n` to be used in the `Output`)
  - All strings: `Input('options-{*v:s}', 'value')` -> `kwargs['v'] = list(str)`

Some examples:

Interdependencies within one dynamically-created pane:
```py
@app.callback(
    Output('graph-{name:s}', 'figure'),
    [Input('log-scale-{name:s}', 'value')]
)
def make_graph(is_log_scale, name):
    return {'data': get_data(name), 'layout': {'yaxis': {'type': 'log' if is_log_scale else 'linear'}}
```

Use a list of filters to construct a query:
```py
@app.callback(
    Output('results', 'children'),
    [Input('filter-field-{*n:d}', 'value'), Input('filter-values-{*n:d}', 'value')]
)
def get_count(fields, values, n):
    # we don't use n but our function needs to accept it - it's a list of integers
    # or could just take **kwargs
    filters = ['{} in ("{}")'.format(field, '","'.join(vals)) for field, vals in zip(fields, values)]
    res = run_query('select count(*) from users where ' + ' and '.join(filters))[0][0]
    return 'Found {} matching users'.format(res)
```

Use all previous filters to find values available to this filter:
```py
@app.callback(
    Output('filter-values-{n:d}', 'options'),
    [Input('filter-field-{n:d}', 'value'),
     Input('filter-field-{*m<n:d}', 'value'),
     Input('filter-values-{*m<n:d}', 'value')]
)
def get_values(field, prev_fields, prev_vals, n, m):
    filters = ['{} in ("{}")'.format(f, '","'.join(v)) for f, v in zip(prev_fields, prev_vals)]
    res = run_query('select distinct(' + field + ') from users where ' + ' and '.join(filters))
    return [{
        'label': capitalize(v[0]),
        'value': v[0]
    } for v in res]
```

### LINKED ISSUE #832: [BUG] callbacks called in wrong order / too often
Hi plotly/dash people,

Great project, nicely designed.

**Versions**
```
dash                          1.0.1
dash-core-components          1.0.0
dash-html-components          1.0.0
dash-renderer                 1.0.0
dash-table                    4.0.1
```

**Describe the bug**

I created this demo of dash + vaex to interactively explore large datasets (150 million rows in this case):
https://github.com/maartenbreddels/dash-taxi-example

Talking to @alexcjohnson at glue-con, he suggested I use the Store, to avoid doing calculations when for instance I change between linear to log.

This seems to expose a bug, that I tried to isolate, but that didn't make it easier to digest.

I opened a PR here for that: https://github.com/maartenbreddels/dash-taxi-example/pull/2, which exposes the bug that I will try to describe.

This dash app shows a heatmap on the left containing the number of taxi trips. On the right, it shows the taxi trips per hour for the zoomed-in region and the total dataset:
image

Optionally, one can filter the month (not relevant now). When zooming in the heatmap, it should trigger a computation to update the heatmap and barchart.

**Expected behavior**

When I zoom in:
 * `update_limits` callback should be called, which outputs to the `limits` store.
 * `update_data` should now be called since it depends on `limits`, and it outputs to the `data-heatmap` store.
 * `update_figure_2d` should be called after `update_data`, since it depends on the `limits` **and** `data-heatmap` store.
 * `update_figure_bar` should be called as well (not relevant for the discussion).

**Observed behaviour**

Instead, what I see is that `update_figure_2d` is called before **and** after `updata_data` , as shown in the terminal output:
```
update_limits {'xaxis.range[0]': -73.80614443872746, 'xaxis.range[1]': -73.77961602289301, 'yaxis.range[0]': 40.6426960947285, 'yaxis.range[1]': 40.67774895854463}
update_figure_2d
updating data: limits [[-73.80614443872746, -73.77961602289301], [40.6426960947285, 40.67774895854463]]
update_figure_2d
update_figure_bars
```

This leads to a very bad user experience, as shown in this screen capture:
screencapture-bug

What happens in the screencapture is:
 * I zoom in (and this is visually reflected in the browser, by the low res heatmap) and it also triggers the `update_limits` callback.
 * This is followed by `update_figure_2d` (the one that shouldn't be triggered, only after update_data), which will update the figure, but with the new limits, and the old data. This is seen as the original heatmap, but shown at the zoomed in region.
 * Now the `update_data`, followed by the second `update_figure_2d` is triggered, and the correct heatmap is shown.

This to me seems like a bug.

My workaround is to put all dependencies of `update_figure_2d` in the Store (as a list or dict), but this seems to go against the design of dash.

Regards,

Maarten Breddels

### LINKED ISSUE #1146: Component Div inside tab cannot render when upgrading from dash 1.6.1 to 1.7.0
App related: https://github.com/plotly/dash-sample-apps/tree/master/apps/dash-manufacture-spc-dashboard
Callback related: https://github.com/plotly/dash-sample-apps/blob/854b8df105e24b04f6b3f9124c016f1e01b23691/apps/dash-manufacture-spc-dashboard/app.py#L943-L993

```
dash                 1.7.0  
dash-core-components 1.6.0  
dash-daq             0.4.0  
dash-html-components 1.0.2  
dash-renderer        1.2.2  
dash-table           4.5.1  
```

    - OS: [e.g. iOS]: MacOS Mojave
    - Browser [e.g. chrome, safari]: MacOS Chrome Version 80.0.3987.132 (Official Build) (64-bit), Firefox 73.0.1

**Describe the bug**

Div(as callback Output) within tab1 cannot be updated by dropdown in the same tab after dropdown option change - > tab switching back. 

**Expected behavior**

I noticed this problem when updating from **dash1.6.1** to **dash1.7.0**. As shown on the gif below,  a callback is fired from dropdown selection within first tab to update the control sections on the right. This section doesn't render for me for dash1.7.0 when I : 1) Click on tab1 2) change dropdown value to a different one 3) Navigate to tab2 then Navigate back to tab1. 

I've tried to simplify the [output](https://github.com/plotly/dash-sample-apps/blob/854b8df105e24b04f6b3f9124c016f1e01b23691/apps/dash-manufacture-spc-dashboard/app.py#L957-L988) to an individual `html.Div("update time" + str(time.time())),` and still behaves the same. My callback is fired as usual upon tab navigation/dropdown-toggling. Looking into dev-tool has indicated that the dash_update_component has been sent back to front-end, so I wonder it could simply be a render-related issue.

app

Screen_Shot_2020-03-06_at_10_37_11_AM

alexcjohnson:
Pattern Matching Callbacks! (Codenamed, Wildcards!)

➡️ See preliminary documentation here: https://github.com/plotly/dash-docs/pull/831

Closes #475 - using the API discussed there (I'll post some examples soon, but for now here's the description):
- String IDs still work the same way in both layout and callbacks
- Now you can also make dict IDs. Keys should be strings, values can be strings, numbers, booleans, like:
  - `id={"type": "cat", "age": 4, "purrs": True}`
  - IDs must still be unique
- Callbacks can use dict IDs with wildcard values for any of the keys. The supported wildcards are `MATCH`, `ALL`, and `ALLSMALLER` - and can be imported from `dash.dependencies`. There are various rules about how these can combine, driven by the need to resolve these to a unique callback invocation for each set of outputs and identify uniquely the inputs required for this invocation...

To do before this is ready:
- [x] Move most callback validation logic to the front end (so it doesn't need to be replicated in R)
- [x] Do something with the callback graph when there are wildcards
- [x] Test that this PR closes callback failure bugs: #1053, #1071, #1084, #1105
- [x] See if it also closes redundant callback bugs #635, #832

## Contributor Checklist

Closes #832 (among other callback bugs - the others all have commits that close them with matching tests, but #832 is a bit heavy for a test of its own)
Closes #1146


christianwengert:
Dear @alexcjohnson 
This looks promising!
If you deem it useful, I might start some tests for my usecases next week.

alexcjohnson:
@christianwengert Absolutely! Thank you for your patience with this. I will try to post some usage examples here later today.

christianwengert:
Yes, an usage example would be great for getting started

alexcjohnson:
## To-Do list example

The `edit_list` callback creates pairs of elements: a checkbox for marking items done and a div displaying the item text. Both of these have dict IDs, with matching `item` values, but the checkbox is distinguished by the `"action": "done"` entry. The callbacks then refer to these items with `"item": ALL` or `"item": ANY` in their input/output/state wildcard component IDs.
```py
import dash
from dash.dependencies import Input, Output, State, ANY, ALL

app = dash.Dash(__name__)

app.layout = html.Div([
    html.Div('Dash To-Do list'),
    dcc.Input(id="new-item"),
    html.Button("Add", id="add"),
    html.Button("Clear Done", id="clear-done"),
    html.Div(id="list-container"),
    html.Hr(),
    html.Div(id="totals")
])

style_todo = {"display": "inline", "margin": "10px"}
style_done = {"textDecoration": "line-through", "color": "#888"}
style_done.update(style_todo)


@app.callback(
    [
        Output("list-container", "children"),
        Output("new-item", "value")
    ],
    [
        Input("add", "n_clicks"),
        Input("new-item", "n_submit"),
        Input("clear-done", "n_clicks")
    ],
    [
        State("new-item", "value"),
        State({"item": ALL}, "children"),
        State({"item": ALL, "action": "done"}, "value")
    ]
)
def edit_list(add, add2, clear, new_item, items, items_done):
    triggered = [t["prop_id"] for t in dash.callback_context.triggered]
    adding = len([1 for i in triggered if i in ("add.n_clicks", "new-item.n_submit")])
    clearing = len([1 for i in triggered if i == "clear-done.n_clicks"])
    new_spec = [
        (text, done) for text, done in zip(items, items_done)
        if not (clearing and done)
    ]
    if adding:
        new_spec.append((new_item, []))
    new_list = [
        html.Div([
            dcc.Checklist(
                id={"item": i, "action": "done"},
                options=[{"label": "", "value": "done"}],
                value=done,
                style={"display": "inline"}
            ),
            html.Div(text, id={"item": i}, style=style_done if done else style_todo)
        ], style={"clear": "both"})
        for i, (text, done) in enumerate(new_spec)
    ]
    return [new_list, "" if adding else new_item]


@app.callback(
    Output({"item": ANY}, "style"),
    [Input({"item": ANY, "action": "done"}, "value")]
)
def mark_done(done):
    return style_done if done else style_todo


@app.callback(
    Output("totals", "children"),
    [Input({"item": ALL, "action": "done"}, "value")]
)
def show_totals(done):
    count_all = len(done)
    count_done = len([d for d in done if d])
    result = "{} of {} items completed".format(count_done, count_all)
    if count_all:
        result += " - {}%".format(int(100 * count_done / count_all))
    return result


if __name__ == "__main__":
    app.run_server(debug=True)
```

alexcjohnson:
I noticed a problem with the To-Do app: if you delete all the items, the totals callback will not fire to say "0 of 0 items completed" as it should. This is because we currently only fire callbacks based on inputs that exist when they're fired, but for `ALL` (and `ALLSMALLER`) an input being removed is itself a change to the input. In fact this could happen when removing just some of the items, but in this particular case the callback still fires because we replace all the other list items (having not yet implemented #968 array operation callbacks).

Working on a fix for this...

alexcjohnson:
Extended To-Do app using `ALLSMALLER` to link up the list items in sequence:
Screen Shot 2020-02-01 at 10 48 37 PM

`ALLSMALLER` can only be used in `Input` and `State` items, and must be used on a key that has `ANY` in the `Output` item(s).

```py
import dash
from dash.dependencies import Input, Output, State, ANY, ALL, ALLSMALLER

app = dash.Dash(__name__)

app.layout = html.Div([
    html.Div('Dash To-Do list'),
    dcc.Input(id="new-item"),
    html.Button("Add", id="add"),
    html.Button("Clear Done", id="clear-done"),
    html.Div(id="list-container"),
    html.Hr(),
    html.Div(id="totals")
])

style_todo = {"display": "inline", "margin": "10px"}
style_done = {"textDecoration": "line-through", "color": "#888"}
style_done.update(style_todo)


@app.callback(
    [
        Output("list-container", "children"),
        Output("new-item", "value")
    ],
    [
        Input("add", "n_clicks"),
        Input("new-item", "n_submit"),
        Input("clear-done", "n_clicks")
    ],
    [
        State("new-item", "value"),
        State({"item": ALL}, "children"),
        State({"item": ALL, "action": "done"}, "value")
    ]
)
def edit_list(add, add2, clear, new_item, items, items_done):
    triggered = [t["prop_id"] for t in dash.callback_context.triggered]
    adding = len([1 for i in triggered if i in ("add.n_clicks", "new-item.n_submit")])
    clearing = len([1 for i in triggered if i == "clear-done.n_clicks"])
    new_spec = [
        (text, done) for text, done in zip(items, items_done)
        if not (clearing and done)
    ]
    if adding:
        new_spec.append((new_item, []))
    new_list = [
        html.Div([
            dcc.Checklist(
                id={"item": i, "action": "done"},
                options=[{"label": "", "value": "done"}],
                value=done,
                style={"display": "inline"}
            ),
            html.Div(text, id={"item": i}, style=style_done if done else style_todo),
            html.Div(id={"item": i, "preceding": True}, style=style_todo)
        ], style={"clear": "both"})
        for i, (text, done) in enumerate(new_spec)
    ]
    return [new_list, "" if adding else new_item]


@app.callback(
    Output({"item": ANY}, "style"),
    [Input({"item": ANY, "action": "done"}, "value")]
)
def mark_done(done):
    return style_done if done else style_todo


@app.callback(
    Output({"item": ANY, "preceding": True}, "children"),
    [
        Input({"item": ALLSMALLER, "action": "done"}, "value"),
        Input({"item": ANY, "action": "done"}, "value")
    ]
)
def show_preceding(done_before, this_done):
    if this_done:
        return ""
    all_before = len(done_before)
    done_before = len([1 for d in done_before if d])
    out = "{} of {} preceding items are done".format(done_before, all_before)
    if all_before == done_before:
        out += " DO THIS NEXT!"
    return out


@app.callback(
    Output("totals", "children"),
    [Input({"item": ALL, "action": "done"}, "value")]
)
def show_totals(done):
    count_all = len(done)
    count_done = len([d for d in done if d])
    result = "{} of {} items completed".format(count_done, count_all)
    if count_all:
        result += " - {}%".format(int(100 * count_done / count_all))
    return result


if __name__ == "__main__":
    app.run_server(debug=True)
```

alexcjohnson:
At a meeting where I demoed this feature, the consensus was `ALL` and `ANY` are easy to confuse with each other (and in a separate discussion @christianwengert seems to agree) - The proposal I think we agreed on is to change `ANY` to `MATCH` - which makes it clear that the values will match across `Input` and `Output` items for a given callback invocation.

We didn't discuss `ALLSMALLER`, as I didn't have a demo ready for it yet. Is it clear from the extended To-Do demo above what this does and how to use it?

Give me a 👍 reaction (not a comment please) if the names `ALL`, `MATCH`, and `ALLSMALLER` look good to you. If not and you have another suggestion please comment.

nicolaskruchten:
`ALLSMALLER` seems oddly specific to me... why not also `ALLBIGGER` ?

alexcjohnson:
`ALLSMALLER` is kind of shorthand for "everything earlier in the dependency chain." In a lot of cases - including the demo - it's not strictly necessary, just makes the logic a bit simpler. But in some cases you would get a circular dependency graph if you didn't have it.

I suppose people could come up with cases where "earlier" items have a larger value in their ids, or where you want some attributes to flow up the list and some to flow down. I wouldn't be opposed to adding `ALLBIGGER`, and this would be easy to do; but `ALLSMALLER` already seemed like an edge case so I stopped there.

Incidentally, I explored for a while adding a `PREVIOUS` wildcard, that instead of all preceding items would just give you the last one. I dropped it as this turns out to be far more complicated than `ALLSMALLER` both for the API (how do you distinguish "there is no previous item" from "the previous item has null value"?) and implementation (because the input and output items can come from different lists, it's possible to have one input be previous to any number of outputs).

alexcjohnson:
`callback_context.triggered` changes a fair amount with this PR: before, it was almost always a one-element array (and never an empty array), so you could safely write code like
```
changed_id, changed_prop = callback_context.triggered[0]["prop_id"].split(".")
```
without error. But there are plenty of cases where a single trigger doesn't make sense.
- On initialization, either on page load or when a `children` section is replaced, many inputs are just present on the page and we need to calculate the initial value of the output. I don't think it makes sense to treat these as triggers. Certainly in as far as triggers are mostly used to detect user input, we shouldn't pretend there was some!
  - **In this PR initial callbacks have no triggers** (ie `callback_context.triggered==[]`).
- Callbacks further along a chain may have multiple triggers at once. Say a user action triggers two callbacks, and both of their outputs feed into a third callback. Before, we would just take the last one to arrive, which could be random.
  - **In this PR all of the props that changed are listed as triggers.**

If we were creating `callback_context.triggered` from scratch today, I would just leave this PR as is, because the previous behavior seems clearly a bug. And I don't see any problem with the multi-trigger case. But a lot of apps depend on there always being an element 0 of the array (and that element having a `"prop_id"` item of a certain form) so I'm thinking perhaps of making a dummy entry `triggered = [{"prop_id": ".", "value": None}]` but then maybe also subclass the array so `bool(triggered)==False` in this case, so that users aware of the new behavior can just use it like `if triggered: ...`. Does that seem like a reasonable solution?

christianwengert:
> I'm thinking perhaps of making a dummy entry `triggered = [{"prop_id": ".", "value": None}]` but then maybe also subclass the array so `bool(triggered)==False` in this case, so that users aware of the new behavior can just use it like `if triggered: ...`. Does that seem like a reasonable solution?

I guess, even a dummy value would be breaking, because - at least in my case - I am not awaiting such a dummy value so I will need to adapt my code nevertheless or experience even weirder bugs. 

Thus: I would advocate to clean this up

Marc-Andre-Rivet:
@alexcjohnson Using the example from https://github.com/plotly/dash/pull/1103#issuecomment-581095515 and changing:

`html.Div(text, id={"item": i}, style=style_done if done else style_todo),` for
`html.Div(text, id={"item": i, "action": "text"}, style=style_done if done else style_todo),`

and the related callbacks, I now get this error in Dash Devtools:
image

Marc-Andre-Rivet:
I'm good with https://github.com/plotly/dash/pull/1103#issuecomment-581096486 `MATCH`. I personally prefer `ONE` but don't have a really good case for it.

Marc-Andre-Rivet:
@alexcjohnson Using https://github.com/plotly/dash/pull/1103#issuecomment-581095515 as a baseline and changing:

`html.Div(text, id={"item": i}, style=style_done if done else style_todo),` for

```
html.Div(text, id={"item": i, "action": "text", "sub": "text"}, style=style_done if done else style_todo),
html.Div(text, id={"item": i, "action": "text2", "sub": "text"}, style=style_done if done else style_todo),
```

and updating the callback to:

```
@app.callback(
    Output({"item": ANY, "action": ANY, "sub": "text"}, "style"),
    [Input({"item": ANY, "action": "done"}, "value")]
)
def mark_done(done):
    return style_done if done else style_todo
```

I get no error either in Dash DevTools or in DashPy, but don't see the styles update. The callbacks do occur:

image
image

Using this callback instead:
```
@app.callback(
    Output({"item": ANY, "action": ALL, "sub": "text"}, "style"),
    [Input({"item": ANY, "action": "done"}, "value")]
)
def mark_done(done):
    return [style_done if done else style_todo, style_done if done else style_todo]
```

works without errors:
image

dash-renderer/src/actions/constants.js
```diff
@@ -1,18 +1,18 @@
...
+    SET_APP_LIFECYCLE: 1,
+    SET_CONFIG: 1,
+    ON_ERROR: 1,
+    SET_HOOKS: 1,
 };
 
 export const getAction = action => {
     if (actionList[action]) {
-        return actionList[action];
+        return action;
```
Marc-Andre-Rivet: Minor point, `actionList.hasOwnProperty(action)` is supported since IE 5.5+, does the same and is actually slightly faster on Chrome (100ms for 1e6 iterations vs. 130ms) and both implementations have identical performance on FF.

dash-renderer/src/components/core/Toolbar.react.js
```diff
@@ -33,7 +33,7 @@ function UnconnectedToolbar(props) {
                 },
                 styles.parentSpanStyle
             )}
-            onClick={() => dispatch(undo())}
+            onClick={() => dispatch(undo)}
```
Marc-Andre-Rivet: I know we do this in lots of places but re-rendering with a new function can in theory have an impact on performance and trigger GC more often. Since this is a functional component, could be wrapped in the `useCallback` hook which is a memoization hook offered by React since `16.6`.

https://reactjs.org/docs/hooks-reference.html#usecallback

dash-renderer/src/actions/constants.js
alexcjohnson: Do we still find `getAction` valuable? (cc @chriddyp) Its whole point, I guess, is to make debugging easier by throwing an error if an unrecognized action is dispatched, rather than just silently doing nothing. But we haven't been following that pattern particularly, and our project doesn't lend itself to new actions being added very often anyway, which seems like it's where this would come in handy. I'm inclined to just take `getAction` out entirely.

chriddyp: I always found the debugging helpful, I've wasted many hours without it. But I agree that not many actions are being added lately.

alexcjohnson: @Marc-Andre-Rivet points out that, of course, this is exactly the kind of issue TypeScript was made for - keep all the overhead and debugging in the compile step 😁 
Anyway for now I'll leave `getAction` in place but I'm not going to invest in applying the pattern consistently.

dash/_validate.py
```diff
@@ -0,0 +1,571 @@
...
+        # properties of the elements in the callback.
+        raise exceptions.LayoutIsNotDefined(
+            """
+            Attempting to assign a callback to the application but
+            the `layout` property has not been assigned.
+            Assign the `layout` property before assigning callbacks.
+            Alternatively, suppress this warning by setting
+            `suppress_callback_exceptions=True`
+            """
+        )
```
Marc-Andre-Rivet: Not in scope but I've run across the issue a few times where I've defined my callbacks before assigning layout. I wonder if it would be possible to relax that requirement. I think this is because we validate the callback ids on the spot, but if we were to accumulate and validate them on `run_server` this wouldn't be necessary.

dash/_validate.py
```diff
@@ -0,0 +1,571 @@
...
+
+    outputs = output if is_multi else [output]
+
+    for args, cls in [(outputs, Output), (inputs, Input), (state, State)]:
+        validate_callback_args(args, cls, layout, validate_ids)
+
+    prevent_duplicate_outputs(app, outputs)
+    prevent_input_output_overlap(inputs, outputs)
+    prevent_inconsistent_wildcards(outputs, inputs, state)
```
Marc-Andre-Rivet: I've been wondering if it would make sense to move deeper graph validation to the renderer. If we later decide to allow multiple non-overlapping callbacks to update the same output for example, we would then only update the renderer, not every backend. The renderer could show the error in the DevTools and send back the error through a logging service to the backend to use. Might also be necessary or useful if/when we would come to support adding callbacks on the fly.

dash/dash.py
```diff
@@ -23,23 +21,29 @@
...
-from ._utils import generate_hash as _generate_hash
-from ._utils import patch_collections_abc as _patch_collections_abc
-from . import _watch
-from ._utils import get_asset_path as _get_asset_path
-from ._utils import create_callback_id as _create_callback_id
-from ._utils import get_relative_path as _get_relative_path
-from ._utils import strip_relative_path as _strip_relative_path
-from ._configs import get_combined_config, pathname_configs
+from .development.base_component import ComponentRegistry
+from .exceptions import PreventUpdate, InvalidResourceError
```
Marc-Andre-Rivet: Nice clean up / refactoring of this file. Thanks.

josegonzalez:
One annoying thing so far is that the `prop_id` that is sent in from `dash.callback_context.triggered` includes the `component_value`. So you end up getting a `prop_id` like:

```
{"action": "click", "id": 123}.n_clicks
```

for an input of:

```
[Input({"id": ALL, "action": "click"}, "n_clicks")]
```

The annoying part about this is that - at least in my system - `item` contains an external identifier that the callback needs in order make requests, but I have to fetch this out by manually parsing the prop_id:

```
value = dash.callback_context.triggered[0]["prop_id"]
value = value.replace(".n_clicks", "")
value = json.loads(value)
```

Without wildcards, you still have to do some parsing:

```
# breaks if the prop_id has periods elsewhere for something like... an email address :)
value = dash.callback_context.triggered[0]['prop_id'].split('.')[0]
```

So now there are two ways of parsing the `prop_id` to show folks :(

Given that we cannot change the value of `prop_id` without breaking BC, would it be possible to add a new value to the `triggered` that contains the value of the `component_id` as is? Same for `component_value`.

dash-renderer/src/actions/isAppReady.js
```diff
@@ -22,7 +31,14 @@ export default (layout, paths, targets) => {
...
         if (ready && typeof ready.then === 'function') {
-            promises.push(ready);
+            promises.push(
+                Promise.race([
+                    ready,
+                    rendered.then(
+                        () => document.getElementById(stringifyId(id)) && ready
+                    ),
+                ])
+            );
```
alexcjohnson: For each component, we wait until either its component type is ready OR we've finished the current render loop and it's not in the DOM (if it _is_ in the DOM we chain on `ready` to ensure this condition is moot)

Note that if there are two components of this type, one rendered and the other not, the `Promise.all` below ensures we still wait for the `ready`. That said I should probably add a test for this case...

dash-renderer/src/APIController.react.js
```diff
@@ -19,18 +20,29 @@ class UnconnectedContainer extends Component {
...
         this.initialization = this.initialization.bind(this);
+        this.emitReady = this.emitReady.bind(this);
         this.state = {
             errorLoading: false,
         };
+
+        // Event emitter to communicate when the DOM is ready
+        this.events = new EventEmitter();
+        // Flag to determine if we've really updated the dash components
+        this.renderedTree = false;
```
Marc-Andre-Rivet: This is probably not an issue here but the APIController instance, like any component instance, could be reused for another APIController and the ctor will not be run a 2nd time. As such it would be safer to assign `this.renderedTree = false` in `componentWillMount` instead.

Marc-Andre-Rivet: As discussed with @alexcjohnson the above is not an issue. Only components with `key` now get reused.

dash-renderer/package.json
```diff
@@ -25,6 +25,7 @@
     "cookie": "^0.4.0",
     "dependency-graph": "^0.9.0",
     "prop-types": "15.7.2",
+    "fast-isnumeric": "^1.1.3",
```
Marc-Andre-Rivet: Given the numerous changes in `package-lock.json`, I assume that at some point `npm install` was run from a clean slate. To have a fully consistent dev/prod environment with minimal changes, would be preferable to revert the lock file, `npm ci` + `npm install --save-dev npm-run-all` + `npm install --save fast-isnumeric` and recommit the newly updated lock file with limited changes. Alternatively, if we are fully confident that these version changes do not impact Dash's behavior, they can be accepted as the new baseline.

dash-renderer/src/APIController.react.js
```diff
@@ -23,15 +28,21 @@ import {STATUS} from './constants/constants';
...
+        events.current = new EventEmitter();
+    }
+    const renderedTree = useRef(false);
+
     useEffect(() => {
         const {
             appLifecycle,
             dependenciesRequest,
             dispatch,
+            error,
```
Marc-Andre-Rivet: This component might be easier to read if we make this effect a standalone function
```
function storeEffect(props) { ... }

useEffect(storeEffect.call(this, props))
```

dash-renderer/src/TreeContainer.js
```diff
@@ -102,49 +109,48 @@ class TreeContainer extends Component {
...
-                itempath: _dashprivate_path,
-            })
+        const oldProps = this.getLayoutProps();
+        const {id} = oldProps;
+        const changedProps = pickBy(
+            (val, key) => !equals(val, oldProps[key]),
+            newProps
         );
+        const changedKeys = keys(changedProps);
+        if (changedKeys.length) {
```
Marc-Andre-Rivet: Since none of this is time critical, maybe we can keep with the Ramda abstraction by using `isEmpty` instead of `array.length`

CHANGELOG.md
```diff
@@ -4,12 +4,21 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+- [#1103](https://github.com/plotly/dash/pull/1103) Wildcard IDs and callbacks. Component IDs can be dictionaries, and callbacks can reference patterns of components, using three different wildcards: `All`, `MATCH`, and `ALLSMALLER`, available from `dash.dependencies`. This lets you create components on demand, and have callbacks respond to any and all of them. To help with this, `dash.callback_context` gets three new entries: `outputs_list`, `inputs_list`, and `states_list`, which contain all the ids, properties, and except for the outputs, the property values from all matched components.
```
Marc-Andre-Rivet:
```suggestion
- [#1103](https://github.com/plotly/dash/pull/1103) Wildcard IDs and callbacks. Component IDs can be dictionaries, and callbacks can reference patterns of components, using three different wildcards: `ALL`, `MATCH`, and `ALLSMALLER`, available from `dash.dependencies`. This lets you create components on demand, and have callbacks respond to any and all of them. To help with this, `dash.callback_context` gets three new entries: `outputs_list`, `inputs_list`, and `states_list`, which contain all the ids, properties, and except for the outputs, the property values from all matched components.
```

CHANGELOG.md
```diff
@@ -4,12 +4,21 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+
 - [#1134](https://github.com/plotly/dash/pull/1134) Allow `dash.run_server()` host and port parameters to be set with environment variables HOST & PORT, respectively
 
 ### Changed
+- [#1103](https://github.com/plotly/dash/pull/1103) Multiple changes to the callback pipeline:
+  - `dash.callback_context.triggered` now does NOT reflect any initial values, and DOES reflect EVERY value which has been changed either by activity in the app or as a result of a previous callback. That means that the initial call of a callback with no prerequisite callbacks will list nothing as triggering. For backward compatibility, we continue to provide a length-1 list for `triggered`, but its `id` and `property` are blank strings, and `bool(triggered)` is `False`.
+  - A callback which returns the same property value as was previously present will not trigger the component to re-render, nor trigger other callbacks using that property as an input.
```
Marc-Andre-Rivet: Is it important to distinguish between value-props and object/array-props here? Returning an object or array with the same props/values will trigger re-render.

CHANGELOG.md
```diff
@@ -4,12 +4,21 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+  - Callback validation is now mostly done in the browser, rather than in Python. A few things - mostly type validation, like ensuring IDs are strings or dicts and properties are strings - are still done in Python, but most others, like ensuring outputs are unique, inputs and outputs don't overlap, and (if desired) that IDs are present in the layout, are done in the browser. This means you can define callbacks BEFORE the layout and still validate IDs to the layout; and while developing an app, most errors in callback definitions will not halt the app.
```
Marc-Andre-Rivet: 🎉

dash-renderer/src/components/error/CallbackGraph/CallbackGraphContainer.react.js
```diff
@@ -55,19 +64,29 @@ class CallbackGraphContainer extends Component {
...
+
+    return <div className="dash-callback-dag--container" ref={el} />;
+};
 
 CallbackGraphContainer.propTypes = {
-    dependenciesRequest: PropTypes.object,
+    graphs: PropTypes.object,
 };
 
 export {CallbackGraphContainer};
```
Marc-Andre-Rivet: Out of scope: Could be a `export default` instead

dash-renderer/src/store.js
```diff
@@ -19,16 +19,18 @@ const initializeStore = reset => {
...
+        store = createStore(reducer, applyMiddleware(thunk));
+    } else {
+        // only attach logger to middleware in non-production mode
+        const reduxDTEC = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__;
+        if (reduxDTEC) {
+            store = createStore(reducer, reduxDTEC(applyMiddleware(thunk)));
+        } else {
+            store = createStore(reducer, applyMiddleware(thunk));
+        }
+    }
```
Marc-Andre-Rivet: We've talked about this in DM, can you explain what this is about again?

dash-renderer/src/reducers/reducer.js
```diff
@@ -49,19 +49,21 @@ function mainReducer() {
 function getInputHistoryState(itempath, props, state) {
     const {graphs, layout, paths} = state;
     const {InputGraph} = graphs;
-    const keyObj = filter(equals(itempath), paths);
+    // TODO: wildcards?
```
Marc-Andre-Rivet:
```suggestion
```

dash-renderer/src/reducers/reducer.js
```diff
@@ -49,19 +49,21 @@ function mainReducer() {
...
+    const keyObj = filter(equals(itempath), paths.strs);
     let historyEntry;
     if (!isEmpty(keyObj)) {
         const id = keys(keyObj)[0];
         historyEntry = {id, props: {}};
         keys(props).forEach(propKey => {
             const inputKey = `${id}.${propKey}`;
             if (
+                // TODO: wildcards?
```
Marc-Andre-Rivet:
```suggestion
```

chriddyp:
Found one potential issue, unless I'm thinking about it wrong.

The TODO app linked above works fine for me.

However when I render that app in a callback, something gets fired and the initial state looks like this:
image

Instead of this:
image

Repro example:
```python
import dash
from dash.dependencies import Input, Output, State, MATCH, ALL

app = dash.Dash(__name__, suppress_callback_exceptions=True)

app.layout = html.Div([
    html.Div(id='content'),
    dcc.Location(id='url')
])

@app.callback(Output('content', 'children'), [Input('url', 'pathname')])
def display_content(_):
    return html.Div([
        html.Div('Dash To-Do list'),
        dcc.Input(id="new-item"),
        html.Button("Add", id="add"),
        html.Button("Clear Done", id="clear-done"),
        html.Div(id="list-container"),
        html.Hr(),
        html.Div(id="totals")
    ])


style_todo = {"display": "inline", "margin": "10px"}
style_done = {"textDecoration": "line-through", "color": "#888"}
style_done.update(style_todo)


@app.callback(
    [
        Output("list-container", "children"),
        Output("new-item", "value")
    ],
    [
        Input("add", "n_clicks"),
        Input("new-item", "n_submit"),
        Input("clear-done", "n_clicks")
    ],
    [
        State("new-item", "value"),
        State({"item": ALL}, "children"),
        State({"item": ALL, "action": "done"}, "value")
    ]
)
def edit_list(add, add2, clear, new_item, items, items_done):
    triggered = [t["prop_id"] for t in dash.callback_context.triggered]
    adding = len([1 for i in triggered if i in ("add.n_clicks", "new-item.n_submit")])
    clearing = len([1 for i in triggered if i == "clear-done.n_clicks"])
    new_spec = [
        (text, done) for text, done in zip(items, items_done)
        if not (clearing and done)
    ]
    if adding:
        new_spec.append((new_item, []))
    new_list = [
        html.Div([
            dcc.Checklist(
                id={"item": i, "action": "done"},
                options=[{"label": "", "value": "done"}],
                value=done,
                style={"display": "inline"}
            ),
            html.Div(text, id={"item": i}, style=style_done if done else style_todo)
        ], style={"clear": "both"})
        for i, (text, done) in enumerate(new_spec)
    ]
    return [new_list, "" if adding else new_item]


@app.callback(
    Output({"item": MATCH}, "style"),
    [Input({"item": MATCH, "action": "done"}, "value")]
)
def mark_done(done):
    return style_done if done else style_todo


@app.callback(
    Output("totals", "children"),
    [Input({"item": ALL, "action": "done"}, "value")]
)
def show_totals(done):
    count_all = len(done)
    count_done = len([d for d in done if d])
    result = "{} of {} items completed".format(count_done, count_all)
    if count_all:
        result += " - {}%".format(int(100 * count_done / count_all))
    return result


if __name__ == "__main__":
    app.run_server(debug=True)
```

chriddyp:
Another potential issue.

 It seems to run fine, but an exception is thrown:
```
can't assign to property "initialCall" on false: not an object
```

Screen Shot 2020-03-30 at 5 26 49 PM

```
import dash
from dash.dependencies import Input, Output, State, MATCH, ALL

app = dash.Dash(__name__, suppress_callback_exceptions=True)

app.layout = html.Div([
    html.Button("Add Filter", id="dynamic-add-filter", n_clicks=0),
    html.Div(id='dynamic-dropdown-container', children=[]),
    html.Div(id='dynamic-dropdown-container-output')
])

@app.callback(
    Output('dynamic-dropdown-container', 'children'),
    [Input('dynamic-add-filter', 'n_clicks')],
    [State('dynamic-dropdown-container', 'children')])
def display_dropdowns(n_clicks, children):
    new_element = html.Div([
        dcc.Dropdown(
            id={
                'type': 'dynamic-dropdown',
                'index': n_clicks
            },
            options=[{'label': i, 'value': i} for i in ['NYC', 'MTL', 'LA', 'TOKYO']]
        ),
        html.Div(
            id={
                'type': 'dynamic-output',
                'index': n_clicks
            }
        )
    ])
    children.append(new_element)
    return children


@app.callback(
    Output({'type': 'dynamic-output', 'index': MATCH}, 'children'),
    [Input({'type': 'dynamic-dropdown', 'index': MATCH}, 'value')],
    [State({'type': 'dynamic-dropdown', 'index': MATCH}, 'id')],
)
def display_output(value, id):
    return html.Div('Dropdown {} = {}'.format(id['index'], value))


if __name__ == '__main__':
    app.run_server(debug=True)
```

***

Also, side note: the actual JavaScript exception for this error isn't in the console so I'm not able to dig into the unminified code. Seems like devtools is catching the error but not re-raising it or printing it. I think that it used to send it to the console (hence the "open your browser's console" message), but I'm not sure.

dash-renderer/src/store.js
alexcjohnson: For connecting up the dev bundle to the redux devtools chrome extension, if it happens to be installed. I'm not quite sure when the previous pattern stopped working, the [extension docs](http://extension.remotedev.io/#usage) make it look like it should have been prior to redux 3.1, but maybe the old version kept working until the 4.x bump in #1145 - at any rate prior to my change this would throw an error in dev mode. See also #1166

Marc-Andre-Rivet:
Validation for multiple callbacks updating the same value don't seem to be working. The following does not trigger an error (two callbacks on item1.children):

```
import dash
from dash.dependencies import Input, Output, State, MATCH, ALL, ALLSMALLER

app = dash.Dash(__name__)


@app.callback(
    [Output('item1', 'children')],
    [Input('item3', 'children')]
)
def callback(children):
    if children is None:
        return PreventUpdate

    return children,

@app.callback(
    [Output('item1', 'children')],
    [Input('item2', 'children')]
)
def callback(children):
    if children is None:
        return PreventUpdate

    return children,

app.layout = html.Div([
    html.Div([
    ], id='item1'),
    html.Div([
        'Bye'
    ], id='item2'),
    html.Div([
        'Hello'
    ], id='item3')
])


if __name__ == "__main__":
    app.run_server(debug=True)
```

In Dash 1.9.1, this would result in:
```
dash.exceptions.DuplicateCallbackOutput:
Multi output ..item1.children.. contains an `Output` object
that was already assigned.
Duplicates:
{'item1.children'}
```

Marc-Andre-Rivet:
One-step and multi-step cycles don't seem to be detected, in either `1.9.1` or this version. The following code updates `item1.children` from `item2.children` and vice versa. Commented out callbacks do a multi-step version of that with 1 -> 2 -> 3 -> 1

```
import dash

app = dash.Dash(__name__)



app.layout = html.Div([
    html.Div([
    ], id='item1'),
    html.Div([
        'Bye'
    ], id='item2'),
    html.Div([
        'Hello'
    ], id='item3')
])


@app.callback(
    [Output('item1', 'children')],
    [Input('item2', 'children')]
)
def callback(children):
    if children is None:
        return PreventUpdate

    return children,

@app.callback(
    [Output('item2', 'children')],
    [Input('item1', 'children')]
)
def callback(children):
    if children is None:
        return PreventUpdate

    return children,

# @app.callback(
#     [Output('item2', 'children')],
#     [Input('item1', 'children')]
# )
# def callback(children):
#     if children is None:
#         return PreventUpdate

#     return children,

# @app.callback(
#     [Output('item3', 'children')],
#     [Input('item2', 'children')]
# )
# def callback(children):
#     if children is None:
#         return PreventUpdate

#     return children,

# @app.callback(
#     [Output('item1', 'children')],
#     [Input('item3', 'children')]
# )
# def callback(children):
#     if children is None:
#         return PreventUpdate

#     return children,    

if __name__ == "__main__":
    app.run_server(debug=True)
```

Marc-Andre-Rivet:
Using a component ID with either `.` or `{` crashes the app instead of displaying the error in the DevTools, is this as intended?

Marc-Andre-Rivet:
Duplicated component ids are still validated in the BE, is this as intended?

Marc-Andre-Rivet:
@alexcjohnson Sorry about that close/reopen.. hand slipped to the worse possible button!

chriddyp:
Another potential issue. 

When running the docs locally, we have some callback error related to a `dcc.Store` component. This issue existed before wildcards. However, now the tab says "Updating...", so it seems like there might be some new logic with failed callbacks that aren't clearing the `requestQueue` (or w/e that part of the store is called now)
image

You should be able to reproduce just by running https://github.com/plotly/dash-docs/pull/831 locally (that branch has the builds as tarballs referenced in `requirements.txt`)

alexcjohnson:
> Using a component ID with either `.` or `{` crashes the app instead of displaying the error in the DevTools, is this as intended? Follow up issue for FE implementation?

Yes, for now that needs to stay on the back end, otherwise we can't properly interpret the callback definitions on the front end - users could create string IDs that look like dict IDs. If we were to change the API for `dependenciesRequest` we could relax this.

Alternatively, back-end errors that happen prior to starting the server could be put into a queue instead of throwing, and served up to the front end along with one of the initialization requests. Out of scope for this PR though.

> Duplicated component ids are still validated in the BE, is this as intended? Follow up issue for FE implementation?

You mean duplicate IDs in the layout? I only moved callback validation to the FE, not layout validation. But you're right, it could be nice to move that as well.

#1173 for followup.

dash-renderer/package.json
alexcjohnson: good call - not sure how that happened... -> https://github.com/plotly/dash/pull/1103/commits/3f46fe658dd48d17ece406d88015798291b5c6cb

dash-renderer/src/APIController.react.js
alexcjohnson: sure -> https://github.com/plotly/dash/pull/1103/commits/883a6ef80ad42f0ba0ec640a95deb66189bf031f

dash-renderer/src/TreeContainer.js
alexcjohnson: yep, that's a little simpler -> https://github.com/plotly/dash/pull/1103/commits/14b7e3ca90b0e33e63ef9ebc0178503644848aab

CHANGELOG.md
alexcjohnson: Actually turns out I didn't implement this for callback returns, only for front-end changes via `setProps`. And I used `R.equals` so no need to distinguish, it'll deep compare. I suppose that might be a problem for very large props (table data for example)... perhaps `===` would be safer? Might be good to do what I originally said I had done 😏 esp if we switch to `===`.

Anyway I fixed the changelog to match the current behavior in this branch in https://github.com/plotly/dash/pull/1103/commits/1e75e9b3fb17e37f261d0b7541a49f658c7fabc8

dash-renderer/src/actions/dependencies.js
```diff
@@ -1247,11 +1247,13 @@ export function getCallbacksInLayout(graphs, paths, layoutChunk, opts) {
         if (outIdCallbacks) {
             for (const property in outIdCallbacks) {
                 const cb = getCallbackByOutput(graphs, paths, id, property);
-                // callbacks found in the layout by output should always run,
-                // ie this is the initial call of this callback even if it's
-                // not the page initialization but just a new layout chunk
-                cb.initialCall = true;
-                addCallback(cb);
+                if (cb) {
```
alexcjohnson, line 1250: Fix for the "can't set initialCall of false" that @chriddyp saw in https://github.com/plotly/dash/pull/1103#issuecomment-606326673

Turns out I had been mainly testing using different keys in the ID dict for different classes of objects - which, given the way I implemented this, is slightly more performant, but that shouldn't be a big deal and keeping the keys consistent is probably more intuitive for users - but we should make clear at least in our examples that this isn't necessary.

tests/integration/test_integration.py
```diff
@@ -745,3 +746,28 @@ def test_inin_024_port_env_success(dash_duo):
...
+    def out(n):
+        return text
+
+    return app
+
+
+def test_inin025_url_base_pathname(dash_br, dash_thread_server):
+    server = flask.Flask(__name__)
+    app = nested_app(server, "/app1/", "The first")
+    nested_app(server, "/app2/", "The second")
```
alexcjohnson, line 765: Shockingly we had no tests for multiple apps in one server... thanks for pointing this out @christianwengert!

`dash_duo` currently needs an app running at `/` or it will time out; a clean fix for this wasn't obvious to me, so I just used `dash_br` and `dash_thread_server` (effectively the two pieces that make the "duo" of `dash_duo`) separately.

dash-renderer/src/reducers/error.js
```diff
@@ -8,6 +8,11 @@ const initialError = {
 export default function error(state = initialError, action) {
     switch (action.type) {
         case 'ON_ERROR': {
+            // log errors to the console for stack tracing and so they're
+            // available even with debugging off
+            /* eslint-disable-next-line no-console */
+            console.error(action.payload.error);
```
alexcjohnson, line 14: re: @chriddyp 's note:
> Also, side note: the actual JavaScript exception for this error isn't in the console so I'm not able to dig into the unminified code. Seems like devtools is catching the error but not re-raising it or printing it. I think that it used to send it to the console (hence the "open your browser's console" message), but I'm not sure.

Not clear to me when that logging disappeared but yes it's certainly useful. Any objection to logging it from the reducer here? A bit odd to have a reducer make side-effects, but this seems like the DRYest way to handle it, and it'll work whether debug mode is on or off.

dash-renderer/src/actions/isAppReady.js
Marc-Andre-Rivet: Thinking about this some more, prior to the DOM check approach, I thought about being able to trigger the component’s download from the renderer, extending the existing component plugin, eg:

https://github.com/plotly/dash/blob/dev/@plotly/dash-component-plugins/src/dynamicImport.js#L23

```
    const get = lazy(() => {
        return Promise.resolve(promise()).then(res => {
            setTimeout(async () => {
                await resolve(true);
                state.isReady = true;
            }, 0);

            return res;
        });
    });

    Object.defineProperty(target, ‘_dashprivate_forceLoadComponent’, {
        get: () => get
    });
```

If the renderer comes across a component using the new version of the plugin, it will be able to trigger the component download by itself. I think the usefulness is limited as the component would still need to be rendered in the DOM in order for its calculated props (table’s `derived_*`) or side-effects (graph’s `extendData`) to take effect.

I’m thinking that instead it might be possible to have preprocessing code that lives outside the components themselves doing `props -> props*` transform, with the renderer being responsible for asking for this transform to be run. This would allow these component props to behave consistently whether they are displayed or not. In the grand scheme of things, this is an edge case, but inconsistent behaviors are always annoying to devs :grimacing: 

I wonder if it’s possible to isolate Plotly.js `fiture + extendData -> figure` logic so that it’s not necessary to load only a small fraction of Plotly.js to do that operation.

In all cases, this is totally out of scope for wildcards. Just want to see what other reviewer's thoughts are about this behavior before going ahead and creating an issue for it.

alexcjohnson:
> The TODO app linked above works fine for me.
> However when I render that app in a callback, something gets fired and the initial state looks like this:

Great catch @chriddyp - this is inconsistent, but I'm trying to figure out what makes most sense in order to make it consistent. Here's what's happening right now:
- The TODO with its skeleton in the initial layout: initially we only look through this layout for callback *outputs*, because we're visiting the entire layout so that will capture everything. And since the associated inputs are just part of the initial layout at that point, they don't count as "changed."
- But when the skeleton is itself the result of a callback, we need to also look for callback *inputs* in the new layout fragment, because they could be connected to outputs elsewhere in the layout. And when we find inputs in this new fragment, we DO consider them "changed."
- Then in the current logic we merge the two ways we found these callbacks - by output and by input - which currently gives the union of the two `changedPropIds` sets.

That's where we are today. The distinction I'm thinking we need to make in order to have these two flavors of the TODO app match is if ALL the outputs of the callback are in the new layout fragment too, then none of the inputs in that fragment count as changed; but if even one output is outside that fragment, then the inputs do count as changed. Or in fact, ideally in this case we would check the previous state of those inputs and see if they really did change - but still, if all the outputs are inside the fragment we wouldn't even do this check: even if prior to the new fragment these inputs were present and had different values, they still wouldn't count as changed. That would be necessary in this case if for example you had a button "reset the whole TODO list" that triggered the `content` callback again - then the various `n_clicks` props might change - they'd *decrease* - but we would not want them to appear changed to the `edit_list` callback.

One rationalization of this distinction is that if the callback has outputs outside the new fragment, then the structure of the fragment must be constant (or the callbacks wouldn't make sense) so providing a new fragment is really just editing it in a perhaps more efficient or convenient way than specifying all the changed props individually - but either way the important thing is that the props changed. Whereas if the callback is entirely internal to the new fragment then the fragment is effectively a new initial state of this sub-app.

dash-renderer/src/actions/isAppReady.js
Marc-Andre-Rivet: In any case, the callbacks with components that are not part of the layout seem to behave better now.

alexcjohnson: I like the idea of pushing component loading up to the renderer, though would need to ponder a little bit if there are performance implications.

Re: props preprocessing - yay for consistency 😁 but I suspect there will still occasionally be props that do depend on real component rendering. An example is the [live figure](https://github.com/plotly/dash-core-components/issues/584) proposal for `dcc.Graph` - you can't know what that's going to be until the graph actually renders and writes things like axis ranges (when these are auto-determined) back into the figure.

alexcjohnson:
> Validation for multiple callbacks updating the same value don't seem to be working.

@Marc-Andre-Rivet that's hilarious, hooray for code reviews! I tested all the complicated cases but the simple case - two callbacks with exactly the same output(s) - I didn't catch, because `app.callback_map` is a dict keyed on a stringified version of the callback outputs so the second one simply overwrites the first. Fixed in https://github.com/plotly/dash/pull/1103/commits/c25bb38a2d7a59e3b78cd34f0e01a59ecb88f464

Marc-Andre-Rivet:
Did sanity test runs against table and DCC tests. Table is fine, DCC tests fail for `pytest tests/integration/store/`

alexcjohnson:
> One-step and multi-step cycles don't seem to be detected, in either 1.9.1 or this version.

I see errors in both versions of your app

Screen Shot 2020-04-02 at 8 53 19 PM
Screen Shot 2020-04-02 at 8 53 59 PM

I don't see a test of it though - I'll add one.

alexcjohnson:
> DCC tests fail for `pytest tests/integration/store/`. Hinting at the same store issues @chriddyp described above.

Turns out this was a generic derived props problem (surprised the table tests didn't surface this...) -> fixed in https://github.com/plotly/dash/pull/1103/commits/6f772a9339d8f2b9e143f338fe51d581db68b094

chriddyp:
@alexcjohnson  re your comment about changed props and ambiguity. I tried to write out all of the cases with partially appearing / rendering inputs _and_ outputs and a comment on what your proposed behavior would be. Did I catch all of the cases? And are my comments about what has changed correct? If so, I think this seems right.

```python
app.layout = html.Div([
    dcc.Input(id={'type': 'input', 'index': 1}),
    html.Div(id='output-container'),
    html.Div(id='output-on-page')
])

@app.callback(Output('output-container', 'children'), ...)
def display_output(...):
    return html.Div([
        dcc.Input(id={'type': 'input', 'index': 2}),
        html.Div(id='output-2')
    ])

@app.callback(
    Output('output-2', ...),
    [Input({'type': 'input', 'index': ALL})])
def update_dynamic_output_pattern(...):
    # When this is triggered because output-2 was rendered,
    # nothing has changed


@app.callback(
    Output('output-2', 'value'),
    [Input({'type': 'input', 'index': 2})])
def update_dynamic_output_single(value):
    # When this is triggered because output-2 was rendered,
    # nothing has changed


@app.callback(
    Output('output-on-page', 'value'),
    [Input({'type': 'input', 'index': 1})])
def update_output_on_page(value):
    # When this is triggered on page load,
    # nothing has changed


@app.callback(
    Output('output-on-page', 'value'),
    [Input({'type': 'input', 'index': ALL})])
def update_output_on_page_pattern(value):
    # When this triggered on page load,
    # nothing has changed
    # When dcc.Input(id={'type': 'input', 'index': 2})
    # is rendered (from display_output)
    # then `{'type': 'input', 'index': 2}` has changed
```

chriddyp:
(And that could serve as the minimal example in the documentation)

dash/testing/plugin.py
```diff
@@ -49,6 +49,12 @@ def pytest_addoption(parser):
         help="set this flag to control percy finalize at CI level",
     )
 
+    dash.addoption(
+        "--pause",
+        action="store_true",
+        help="pause using pdb after opening the test app, so you can interact with it",
```
alexcjohnson, line 55: FYI - this is a feature I've been thinking about for a while now - very often I want to run an app from the test suite and live-debug it, I would end up copying it into a temp file, fiddling with the indentation and imports to get it to run standalone... now you can just do eg:
```
pytest -k cbwc001 --pause
```
And you're in a Python debugger - where, after advancing a few times, you can run selenium / `dash_duo` commands - but you also have the app from the test `cbwc001` open in a browser window and can interact with it and debug in JS.

Marc-Andre-Rivet: Oh! Very nice :shipit:

alexcjohnson:
The issue @chriddyp mentioned https://github.com/plotly/dash/pull/1103#issuecomment-606897721 from the docs, that the title never goes back from "Updating...", is still present. Investigating...

alexcjohnson:
Ah got it - we're creating and throwing some errors that I forgot to trap, so I could clear the associated callback from the queue and surface the error in the devtools.

The specific error we're hitting in the docs is (in case someone wants to fix it):

> Uncaught (in promise) ReferenceError: A nonexistent object was used in an `Input` of a Dash callback. The id of this object is `memory-button` and the property is `n_clicks`. The string ids in the current layout are: [memory, memory-output, local, session, wait-for-layout, location, banner, logo-home, backlinks-top, chapter, backlinks-bottom, pagemenu]

alexcjohnson:
After https://github.com/plotly/dash/pull/1103/commits/7c47b0577abec4f15f74e6617658c49fc02f8de1 (running docs with devtools on so we see the errors... with devtools off the title is still correct and the errors only appear in the js console) 

Screen Shot 2020-04-07 at 5 18 58 PM

MM-Lehmann:
whats that rationale behind the restriction that I cannot combine `MATCH` with other (static) outputs? I would like to have a `MATCH` callback on some GUI elements which save their current state into a `dcc.Store` object and retrieve it later from there. Is it because in the background multiple callbacks are created for each match of objects and those can't have duplicate outputs between each other?

alexcjohnson:
That’s essentially it - the value to return into that single output would be ill-defined, as it would be trying to get a separate value from each matched input.

I see how specifically with `Store` it would be possible to construct the callback in such a way as to make it self-consistent. But that kind of construction would be tricky and fragile. Keeping MATCH keys the same across all outputs makes this much simpler for everyone to reason about (including us writing the rendering engine!) but in return you sometimes have to handle your data a little differently.

In this case I would suggest either making a separate `Store` for each MATCH value, or a separate callback that combines the inputs with ALL and returns the single `Store` value.

lhsusan1016:
When I used two outputs, one of which uses MATCH and the other has nothing to do with 'index':
@app.callback(
    [Output({'type': 'dynamic-output', 'index': MATCH}, 'children'),
     Output('title_button', 'children')],
    [Input({'type': 'dynamic-dropdown', 'index': MATCH}, 'value')],
    [State({'type': 'dynamic-dropdown', 'index': MATCH}, 'id')],
)


 it raised an error as follows:

In the callback for output(s):
  {"index":MATCH,"type":"dynamic-output"}.children
  title_button.children
Output 1 (title_button.children)
does not have MATCH wildcards on the same keys as
Output 0 ({"index":MATCH,"type":"dynamic-output"}.children).
MATCH wildcards must be on the same keys for all Outputs.
ALL wildcards need not match, only MATCH.


Why is it necessary to require all the outputs with MATCH, and how can I solve this problem?
Thanks!!!

alexcjohnson:
@lhsusan1016 as described in https://github.com/plotly/dash/pull/1103#issuecomment-624058974 if there were two different elements that satisfy your `MATCH` pattern, then there would need to be two different invocations of this callback, one for each. Then this would be two different values for `title_button.children` - so which one would we use? It’s ambiguous and therefore an error. 

You can solve this by either switching to an `ALL` pattern to create all the dynamic outputs at once, or making a separate callback for `title_button.children`

I happened to see your comment, but old closed PRs are not great places to start conversations. This kind of question is better at https://community.plotly.com

luggie:
I can't run any of the minimal examples on https://dash.plotly.com/pattern-matching-callbacks without some js errors.
Which version of dash and components are they meant for? I'm on 1.16.3
Screenshot from 2020-10-18 13-52-01

PR #1111, opened 2020-02-05
Adds DASH_HOST env variable

OwenMatsuda:
This PR adds the ability to set a `DASH_HOST` environment variable to change the host value when running the server. 

## Testing

 - Check to see if injecting an environment variable changes the host
No Env

 - Make sure that it defaults to `127.0.0.1` if no environment variable
Env

 - If the host parameter is in `app.runserver()`, then it will take preference over the environment variable. In this case, the argument was `host=127.0.0.1`
Env & Parameter


OwenMatsuda:
@alexcjohnson Please review

josegonzalez:
DashR now uses `HOST` for this https://github.com/plotly/dashR/pull/167

OwenMatsuda:
> DashR now uses HOST for this plotly/dashR#167

Should I just essentially copy the DashR change then and add PORT too?

OwenMatsuda:
```python
def run_server(
        self,
        host=os.getenv('HOST', "127.0.0.1"),
        port=os.getenv('PORT', 8050),
        ...
)
```

josegonzalez:
I'd guess so but will leave that decision to others :)

OwenMatsuda:
Moved to https://github.com/plotly/dash/pull/1134 because I accidentally rebased off of master instead of dev.

PR #1123, opened 2020-02-14
[WIP / POC] Ability to prevent initial callback from firing

chriddyp:
Via a new property in the `app.callback`, `prevent_initial_call=True`

```
@app.callback(Output('another-output', 'children'), [Input('another', 'value')], prevent_initial_call=True)
```

This will prevent it for both initial renders in `app.layout` as well as new component renders from a callback.

WIP / POC for now. Would need lots of tests.


christianwengert:
+1 I believe this would be very useful. With the wildcards it can happen very quickly, that I have like one hundred initial callbacks which make the app very unresponsive for a couple of seconds. By preventing all wildcards to fire this would be much faster and better

alexcjohnson:
superseded by #1225

PR #1130, opened 2020-02-25
Added words to make the sentence valid.

ashwinpn:
I added a few words to make the sentence semantically correct.

 *....the prop values will not be stored with components without an ID,

## Contributor Checklist

- [Done] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))

### DIFF
--- a/dash-renderer/src/persistence.js
+++ b/dash-renderer/src/persistence.js
@@ -3,8 +3,9 @@
  *
  * When users input new prop values, they can be stored and reapplied later,
  * when the component is recreated (changing `Tab` for example) or when the
- * page is reloaded (depending on `persistence_type`) Storage is tied to
- * component ID and will not on with components without an ID.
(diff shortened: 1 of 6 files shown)


dash-renderer/src/persistence.js
```diff
@@ -3,8 +3,9 @@
  *
  * When users input new prop values, they can be stored and reapplied later,
  * when the component is recreated (changing `Tab` for example) or when the
- * page is reloaded (depending on `persistence_type`) Storage is tied to
- * component ID and will not on with components without an ID.
+ * page is reloaded (depending on `persistence_type`). Storage is tied to
+ * component ID, and the prop values will not be stored with components
```
alexcjohnson:
```suggestion
 * component ID, and the prop values will not be stored with components
```

PR #1132, opened 2020-02-26
Allow compress to be set with env variable

chriddyp:
For when Dash apps are deployed on servers that do gzip compression faster than Python


chriddyp:
ugh wrong branch

PR #1133, opened 2020-02-26
allow gzip to be set via env

chriddyp:


chriddyp:
for apps that are deployed on servers that have their own gzip that is faster than python

PR #1134, opened 2020-02-27
Adds DASH and PORT env vars

OwenMatsuda:
This PR adds the ability to set a DASH_HOST environment variable to change the host value when running the server.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1841,7 +1841,8 @@ class Dash(object):
 
     def run_server(
         self,
-        port=8050,
+        host=os.getenv("HOST", "127.0.0.1"),
+        port=os.getenv("PORT", "8050"),
         debug=False,
         dev_tools_ui=None,
         dev_tools_props_check=None,
@@ -1860,7 +1861,12 @@ class Dash(object):
         If a parameter can be set by an environment variable, that is listed
         too. Values provided here take precedence over environment variables.
 
+        :param host: Host IP used to serve the application
+            env: ``HOST``
+        :type host: string
+
         :param port: Port used to serve the application
+            env: ``PORT``
         :type port: int
 
         :param debug: Set Flask debug mode and enable dev tools.
@@ -1933,9 +1939,18 @@ class Dash(object):
             dev_tools_prune_errors,
         )
 
+        # Verify port value
+        try:
+            port = int(port)
+            assert port in range(1, 65536)
+        except Exception as e:
+            e.args = [
+                "Expecting an integer from 1 to 65535, found port={}".format(repr(port))
+            ]
+            raise
+
         if self._dev_tools.silence_routes_logging:
             # Since it's silenced, the address doesn't show anymore.
-            host = flask_run_options.get("host", "127.0.0.1")
             ssl_context = flask_run_options.get("ssl_context")
             self.logger.info(
                 "Running on %s://%s:%s%s",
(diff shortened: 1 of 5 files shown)


alexcjohnson:
Alright, after way too much discussion with @josegonzalez and @rpkyle I'm on board with this change. Looks like you just have to adjust the usage re the test failure:
```
dash/dash.py:1913:13: W1508: os.getenv default type is builtins.int. Expected str or None. (invalid-envvar-default)
```
And let's give it some helpful behavior if you fail to provide an int or a string that can be coerced to an int. Then we just need a CHANGELOG entry for the new options and I think this will be ready to 💃

OwenMatsuda:
> And let's give it some helpful behavior if you fail to provide an int or a string that can be coerced to an int.

@alexcjohnson What kind of helpful behavior do you think? Should it just default to the `8050` port, or should it throw an error?

alexcjohnson:
Throw an error like `"Expecting an integer from 1 to 65535, found {}".format(repr(port))` (only including the range if you actually verify that, which would be a nice touch)

OwenMatsuda:
I think this is ready for review. My only potential concern is that for port, when the user inputs it as a parameter within `app.runserver(port=8050)`, port is an int. However, when grabbing it from an environment variable, port is a string that is later coerced into an int within the try/except block.

dash/dash.py
```diff
@@ -1928,7 +1929,12 @@ def run_server(
         If a parameter can be set by an environment variable, that is listed
         too. Values provided here take precedence over environment variables.
 
+        :param host: Host IP used to serve the application
+            env: ``HOST``
+        :type host: string
+
         :param port: Port used to serve the application
+            env: ``PORT``
         :type port: int
```
alexcjohnson: Just mark `port` here as either an int or a string containing an int, and we should be covered re: data types. Your coercion and try/except below looks great!

Can you please add a couple of tests of the failure cases, and maybe one of setting an integer string port and verifying that an app shows up there? These can probably all go in https://github.com/plotly/dash/blob/dev/tests/unit/test_configs.py

The failures can just be something like:
```py
app = Dash()
app.layout = html.Div()
with pytest.raises(...) as excinfo:
    app.run_server(port="garbage")
assert excinfo.value == ...
```

And the pass case can be something like:
```py
app = Dash()
app.layout = html.Div("hi", id="out")
dash_duo.run_server(app, port="12345")
assert dash_duo.server_url == "https://127.0.0.1:12345"
dash_duo.wait_for_text_to_equal("out", "hi")
```

OwenMatsuda:
I'm trying to use pytest's monkeypatch to mock env vars, but I'm running into a bug: https://github.com/pytest-dev/pytest/issues/6858

alexcjohnson:
There's a fixture `empty_environ` at the top of the `test_configs` file - does that help?

OwenMatsuda:
I saw that, but that would mean adding the env vars `HOST` AND `PORT` to `dash/_configs.py`. At first, I thought that would make more sense than the setup I have now, but then I noticed [a problem](https://github.com/plotly/dash/blob/735480bd9796f504cfb0cfdd8b567505901503f1/dash/_configs.py#L47-L49):
```python
    env = load_dash_env_vars().get('DASH_{}'.format(name.upper()))
    if env is None:
        return default
```

Unless we wanted to add another function specific to `HOST` and `PORT`, we would have to rename them to `DASH_HOST`, and `DASH_PORT`, respectively, which would be misaligned with https://github.com/plotly/dashR/pull/167.

Along those same lines, I was wondering what the "parity" with Dash for Python was since the `HOST` and `PORT` didn't previously exist.

PR #1138, opened 2020-02-27
Reduce test warnings

scjody:
This is an attempt to reduce warnings printed during tests. Currently whenever an exception is raised during `text_to_equal()`, the stacktrace is printed. Since failures are expected fairly often (for example if the element being examined doesn't exist yet), the test output of even a successful run of the Dash Enterprise automated integration tests are filled with them.

This PR simply removes the exception logging since I doubt it's useful to anyone.

## Contributor Checklist

   -  [x] Test with the Dash Enterprise automated integrated tests

### DIFF
--- a/dash/testing/wait.py
+++ b/dash/testing/wait.py
@@ -64,7 +64,6 @@ class contains_text(object):
                 elem.get_attribute("value")
             )
         except WebDriverException:
-            logger.exception("contains_text encountered an exception")
             return False
 
 
@@ -82,7 +81,6 @@ class text_to_equal(object):
                 or str(elem.get_attribute("value")) == self.text
             )
         except WebDriverException:
-            logger.exception("text_to_equal encountered an exception")
             return False
 
 
@@ -99,5 +97,4 @@ class style_to_equal(object):
             logger.debug("style to equal {%s} => expected %s", val, self.val)
             return val == self.val
         except WebDriverException:
-            logger.exception("style_to_equal encountered an exception")
             return False


dash/testing/wait.py
```diff
@@ -92,8 +100,7 @@ def __call__(self, driver):
                 or str(elem.get_attribute("value")) == self.text
             )
         except WebDriverException:
-            logger.exception("text_to_equal encountered an exception")
```
scjody: We could get rid of the whole `try/except` here instead, and just let `WebDriverException` be raised.

alexcjohnson:
@scjody sounds like you guys have some work to do to figure out if this will exactly suit your needs - so let us know how that goes - but from my perspective this looks very nice!

I'll let you sort out the linter issues, but the test failures will be resolved if you merge in `dev`.

scjody:
@alexcjohnson @Marc-Andre-Rivet We're happy with this now. I changed my original design and the existing behaviour of `until()` and `until_not()` is preserved unless the `wait_cond` class is changed to enable the new behaviour. (See the top comment for a full explanation.)

If you're curious, [here's a test run with the current behaviour](https://circleci.com/gh/plotly/dash-deployment-server/42695) and [here's one with the new behaviour](https://circleci.com/gh/plotly/dash-deployment-server/42718?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) (the second one may fail due to a problem destroying the instance, but everything up until that point was successful and no longer has the warnings).

Can you please review this or let me know the next steps?

alexcjohnson:
@scjody the functionality looks great! The only thing I'd like is a test that (1) indeed there are no unwanted log entries and (2) failures still cause the test to fail.

For (1) seems like the easiest would be to have a slow callback that creates a new element, trigger the callback and call `wait_for_text_to_equal` on the initially missing element:

```py
app.layout = html.Div([
    html.Button(id="btn"),
    html.Div(id="out")
])

@app.callback(Output("out", "children"), [Input("btn", "n_clicks")])
def slow_click(n):
    time.sleep(1)
    return html.Div(n, id="out-inner")

dash_duo.start_server(app)
dash_duo.multiple_click("#btn", clicks=2)
dash_duo.wait_for_text_to_equal("#out-inner", "2")
assert _no_unwanted_logs_
```

Then for (2) a check (in the same test?) that a real failure of either flavor (exception or just false) still fails as expected (does pytest let you do it this way?)

```py
with pytest.raises(AssertionError):
    dash_duo.wait_for_text_to_equal("#out-inner", "3", timeout=1)
assert _correct_logs_

with pytest.raises(AssertionError):
    dash_duo.wait_for_text_to_equal("#nonexistent", "3", timeout=1)
assert _correct_logs_
```

dunno where this should go, maybe we need a new `tests/integration/test_testing.py`?

scjody:
@alexcjohnson Testing the test tooling seems reasonable (especially since it's used by code other than `dash` itself) but I don't see how the [`text_to_equal` callable class](https://github.com/plotly/dash/blob/74594b0d00714e223f4a6bd01af94b08be481b29/dash/testing/wait.py#L79) is supposed to work and it isn't used anywhere in `dash` or `dash-deployment-server`. (The [`wait_for_text_to_equal` method](https://github.com/plotly/dash/blob/74594b0d00714e223f4a6bd01af94b08be481b29/dash/testing/browser.py#L270) is used in many places but that doesn't have any relationship to the code I'm changing.)

I think the best way forward here is for me to add unit tests of `until()` and `until_not()` that don't depend on browser-based testing, and remove the changes to `text_to_equal`, `contains_text`, and `style_to_equal` in this PR since we have no way of testing them.

Does this sound OK to you?

scjody:
To explain a bit more about the [`text_to_equal` callable class](https://github.com/plotly/dash/blob/74594b0d00714e223f4a6bd01af94b08be481b29/dash/testing/wait.py#L79):
* When called, a `driver` argument must be passed in.
* `until()` and `until_not()` offer no way to pass in a `driver` argument when calling a `wait_cond`.

It may be possible to turn `text_to_equal` into something suitable for use with `until()` with currying, but I really doubt that's what's intended here. It's more likely that `text_to_equal` was an attempt to write something generic that ended up not being used.

alexcjohnson:
It's a bit convoluted, but [`wait_for_text_to_equal` *does* use `text_to_equal`](https://github.com/plotly/dash/blob/dev/dash/testing/browser.py#L264) (and likewise with `style_to_equal` and `contains_text`). You're right that these are not particularly generalizable, so maybe shouldn't be part of the public API... but that doesn't need to be your problem now.

So I think the success condition of `text_to_equal` is adequately tested by our existing test suite, I just want to make sure that it can still fail in the expected way, hence part (2) [above](https://github.com/plotly/dash/pull/1138#issuecomment-594836253) - I don't care as much about the other two, especially since you're making the same changes to all of them, though ideally we'd do similar failure-case tests of them as well.

Unit tests of `until` and `until_not` sound great.

scjody:
@alexcjohnson Thanks for pointing that out and sorry I missed it.

It still doesn't look like `text_to_equal()` and friends are ever used with `until()` or `until_not()` so I don't think they should be part of this PR. Instead they're used with Selenium's [`WebDriverWait`](https://selenium-python.readthedocs.io/waits.html#explicit-waits), which takes care of passing in the `driver` argument.

Are you OK with me removing the changes to `text_to_equal()` and friends from this PR and just adding unit tests of `until()` and `until_not()`?

alexcjohnson:
If that meets the needs of your team then sure, that's a fine approach.

scjody:
I took another look at this and it's actually a `text_to_equal()` / `wait_for_text_to_equal()` call that's causing most of the noise for us. I don't see a good way to apply my approach there since it uses Selenium's `WebDriverWait`.

I modified this PR to remove the exception logging from `text_to_equal()` since it's never useful for us. @Marc-Andre-Rivet @alexcjohnson What do you think of this? If it's useful for Dash of course I support keeping it in. In that case I'll probably just add a message to our tests advising that the exceptions can be ignored, because I don't think it's worth having a DDS-specific `text_to_equal()` and `wait_for_text_to_equal()`.

alexcjohnson:
I don't see a use for this logging, it's expected to get these exceptions when you're looking for an element that doesn't exist yet. In fact I'd go a little further: if we're going to remove it from `text_to_equal`, we should remove it from the other two (`style_to_equal` and `contains_text`) as well.

scjody:
Makes sense. I'll do that tomorrow and re-test.

scjody:
OK, the new commit removes all 3 exception logging calls, and works for us.

PR #1171, opened 2020-03-31
Include license in the sdist of dash-renderer

scarabeusiv:
This will ensure the pkg has a license shipped even if people fetch the package from pypi.

### DIFF
--- a/dash-renderer/MANIFEST.in
+++ b/dash-renderer/MANIFEST.in
@@ -2,3 +2,4 @@ include package.json
 include digest.json
 include dash_renderer/*.js
 include dash_renderer/*.map
+include LICENSE


alexcjohnson:
Does this work? We don't have a `LICENSE` file in the `dash-renderer` dir, only at the repo root.

scarabeusiv:
Hmm True I just blinly copied it out from other package I was doing this for during that day. Maybe it should contain ../LICENSE as to deal with the path.

scarabeusiv:
I am now on mobile and could fix it later next week, but I gave dev access to the branch or as it is oneliner just do it from scratch and revoke this one :)

alexcjohnson:
Is there a use case for fetching `dash-renderer` from PyPI? Normally it's just used internally by `dash`, and we've been discussing simply folding `dash-renderer` into the `dash` package. Originally I think we kept them separate in order to facilitate using the `dash-renderer` front end with different back ends, but the other back ends that I'm aware of just grab the built renderer bundle from this repo one way or another.

That said we also publish `dash-renderer` to [`npm`](https://www.npmjs.com/package/dash-renderer) and will continue to do so for the `serve_locally=False` case, that alone probably warrants a LICENSE and a lightweight README.

alexcjohnson:
As of #1611, the `dash-renderer` pypi package is no longer separate from `dash`, so this issue is moot.

PR #1172, opened 2020-04-01
add warn when apiThunk receive not application/json response

sherry0429:
dash use default flask server, code in dash.py return flask.Response(mimetype='application/json')

but when developer try to use custom server, like python SimpleHttpServer, and handle request itself, when miss 'Content-Type application/json',  no warning.  developer maybe confused.

Now many python project depends on flask/django/other server, and use nginx/uwsgi etc. this pr hope reduce debug time when developer try use custom server handle dash_render.js requests.

## Contributor Checklist

- [ ✅] I have broken down my PR scope into the following TODO tasks
   -  [ ✅] add warn when apiThunk receive not application/json response
- [ ✅] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))


dash-renderer/src/actions/api.js
```diff
@@ -53,6 +56,7 @@ export default function apiThunk(endpoint, method, store, id, body) {
                         return json;
                     });
                 }
+                logWarningOnce("Response miss content-type application/json")
```
alexcjohnson, line 59:
```suggestion
                logWarningOnce('Response is missing header: content-type: application/json');
```

sherry0429: Thanks for review

sherry0429:
> Thank you @sherry0429 - agreed, this could save a lot of headaches! I tweaked the wording and adjusted the syntax to (I think) appease our linter. Will merge once tests pass 🎉 💃

Thanks, I see also some test mistake (lint-unit), should I still change something ?

dash-renderer/src/actions/api.js
```diff
@@ -53,6 +56,7 @@ export default function apiThunk(endpoint, method, store, id, body) {
                         return json;
                     });
                 }
+                logWarningOnce('Response is missing header: content-type: application/json');
```
alexcjohnson, line 59:
```suggestion
                logWarningOnce(
                    'Response is missing header: content-type: application/json'
                );
```

alexcjohnson:
Looks like I got it this time 😅 
In case you make another PR, there's a handy command `npm run format` (run from the `dash-renderer`) directory that will automatically fix any linting issues in the JS files.

sherry0429:
@alexcjohnson Sorry, I have remeber it, =.=, thanks !

alexcjohnson:
Not to worry, I forget to lint sometimes too. We appreciate your contribution!

sherry0429:
Ok, Thanks !

Issue #1176, opened 2020-04-03
[Feature Request] Improve callback graph


jjaraalm:
Right now the callback graph is somewhat useful for identifying logical flow issues, but I think it could be greatly improved if it could be dynamic rather than static.

## Possible improvements
- [ ] Zoom in/out functionality for large/complex callback chains
- [ ] Callback graph accessible from a dedicated url (maybe this already exists?) in addition to an overlay so that you can do test interactions while examining the callback graph
- [ ] Distinguish between clientside and serverside callbacks using different node colors.
- [ ] Report profiling information
    - [ ] Number of times a callback was invoked per node
    - [ ] Time spent (total or average) in the callback. Maybe link to the node color?
    - [ ] Label each edge with the number of times that input triggered a callback
- [ ] Live debugging
    - [ ] Highlight currently executing callbacks and modified outputs
    - [ ] Probe the last input/output by hovering/focusing an edge

For live debugging, I'm thinking something similar to the Labview execution highlighting if you've ever used it. When you turn on debugging mode, callbacks intentionally get a delay injected into them to slow everything down so you can visually watch and inspect the logical flow. Like this:

![](https://labviewwiki.org/w/images/d/d8/Dataflow_-_Nodes_execute_in_parallel.gif)

gvwilson:
Hi - we are tidying up stale issues and PRs in Plotly's public repositories so that we can focus on things that are most important to our community. If this issue is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. (Please note that we will give priority to reports that include a short reproducible example.) If you'd like to submit a PR, we'd be happy to prioritize a review, and if it's a request for tech support, please post in our community forum. Thank you - @gvwilson

PR #1179, opened 2020-04-04
Improve Callback Graph

jjaraalm:
Attempt to improve and significantly expand the features available in the callback graph to improve its usefulness in debugging and testing. Will try to cover the ideas given in https://github.com/plotly/dash/issues/1176. Comments/suggestions would be welcome, especially for styling.

## Current Status 
captured (1)

captured (2)

## Changes
- Replaced `viz.js` with `cytoscape.js` via `react-cytoscape`. This results in some minor visual and layout changes.
- `CallbackGraphContainer` now subscribes to `layout` and `paths` so it can introspect the current state.
- Added new reducers to compute profiling information and report all changed props (not just those that are inputs).

## Issues
- State layout is finicky. `dagre` was giving horrible results even when cycles were pruned away. Using `breadthfirst` for now, but layout properties may need to be tailored to viewport size instead of being fixed.
- Outputs that return `no_update` are still being flashed during execution highlighting. No way to detect them right now.

## Contributor Checklist

    - [X] Zoom in/out functionality for large/complex callback chains
    - [ ] Callback graph accessible from a dedicated url.
    - [X] Distinguish between clientside and serverside callbacks using different node colors.
    - [x] Add state information to the callback graph
    - [x] Report profiling information
        - [x] Number of times a callback was invoked per node
        - [x] Time spent (total or average) in the callback. Maybe link to the node color?
        - [ ] Label each edge with the number of times that input triggered a callback
    - [ ] Live introspection
        - [X] Components, properties, and callback inputs/outputs are inspect-able in real time.
        - [ ] Component properties are modifiable.
    - [ ] Live debugging
        - [x] Highlight currently executing callbacks and modified outputs
        - [ ] Probe the last input/output by hovering/focusing an edge
        - [ ] Manually emit a property changed event to trigger callback paths

### optionals


chriddyp:
Love where this is going @jjaraalm !

nicolaskruchten:
Amazing work! I’m happy to see GraphVis go in favour of a more modern alternative that enables rich interactions :) thanks for doing this!

jjaraalm:
@chriddyp @nicolaskruchten Thanks! It's a fun project, and I think it could be useful too!

chriddyp:
That profiling screenshot is so good!

It would be cool if we could time the callback execution on the backend and pass that time delta up with the `_update-component` API call. Then, we could distinguish between "total time", "time processing backend", "time over network", "time to render".Then, folks could see if simple callbacks on the backend are causing a delay because they're sending a tremendous amount of data over the wire (which would be a great candidate for a clientside callback!) vs a complex backend task that sends a simple payload. I think we'd have to expand the `_update-component` API response to include extra meta information beyond just the component props to do this.

jjaraalm:
Agreed, a better breakdown of the various timings would be helpful. Especially since right now, I'm only looking at the delay between request and response. Even simple callbacks (repeatedly clicking a button) can look expensive if they happen often and the queue gets backed up. This shouldn't be too hard to do and I'll look into it. Glad you all are interested!

chriddyp:
Awesome stuff @jjaraalm ! Feel free to ping me if you have any questions 👍

chriddyp:
(FYI I am assigning myself just to denote to the rest of our team that I'll be your ambassador for this pull request)

jjaraalm:
@chriddyp sounds good. 

I've added support for reporting server-side timings using  `Server-Timing`  headers. This way (in modern browsers) we get timing reported both in dev-tools on a per-request basis and through introspection as an aggregate total.

Users can tag and report timings for resources using a new API like:

```python
@app.callback(Output('slow-store', 'data'), [Input('style-output', 'style'), Input('out', 'children')])
def combine_data(style, value):

    time.sleep(0.1)
    dash.callback_context.record_timing('task_1', 0.1, 'The first task.')

    time.sleep(0.7)
    dash.callback_context.record_timing('task_2', 0.7)

    time.sleep(0.2)
    dash.callback_context.record_timing('task_3', 0.2, 'Cleanup task.')
    
    return dict(style=style, value=value)
```
which translates into the HTTP headers:

```
Server-Timing: dash_total;dur=1002
Server-Timing: task_1;desc="The first task.";dur=100
Server-Timing: task_2;dur=700
Server-Timing: task_3;desc="Cleanup task.";dur=200
```

The `dash_total` resource is automatically computed on all requests.

Distinguishing between "queue time" and "network time" seems difficult and I'm not quite sure how to do it. The difference between `totalTime - totalComputeTime` should be the combination of both though. I'll take a look into rendering time next.

devtools-timing
resource_introspection

chriddyp:
> I've added support for reporting server-side timings using Server-Timing headers.

Brilliant idea! I was not aware of these headers before now, this is very nice.

Re `dash_total`: this is very nice! Let's make this configurable via an argument like `dev_tools_clientside_profiling` so that it is only included if `debug=True` and then turned off when deployed with e.g. `gunicorn`. This is for security reasons as it's unadvised to share any information about the backend to the front-end in production.

***

> Distinguishing between "queue time" and "network time" seems difficult and I'm not quite sure how to do it. The difference between totalTime - totalComputeTime should be the combination of both though. 

Yeah good point, this is a tricky one to record but also to communicate to our users. 

I suppose that there are two queues:
1. Browser queue: The browser will only send around 6 (https://stackoverflow.com/questions/561046/) requests to the server at a time, so if a user has a Dash app with 60 charts, 10 batches of requests will be made.
2. The server itself will queue things up, whether that's being done with `flask` or `gunicorn` or a load balancer.

So, when we display `totalTime - totalComputeTime` to our users, I suppose we label this like "Network & Request Queue Time".

In some cases, the queuing time will dominate the time. In most cases, I think the payload size itself will dominate. So, another metric we could display could be the size of the request. We can get this from the `content-length` header. It would be nice to display this in shorthand like `kb` or `mb`
image

***

>  I'll take a look into rendering time next.

Thinking a bit about this now, this one actually might be pretty tricky. I think the main components that a user might care about is the graph component as will handle the most amount of data on the page. We do asynchronous rendering here, so even if we knew when React "finished" rendering the component, we wouldn't be able to capture the actual rendering time.

This might be a rabbit hole, so feel free to stay out of it for now, I think the other performance breakdowns will be the most valuable for our users!

***

Playing around with this locally, another thing that comes to mind is conditionally re-rendering the callback tree based off of which components are visible on the page. So, if you have a large multi-page application, you would only see the callbacks that are associated with outputs or inputs that are currently rendered on the page. This data is available in the `paths` object in the store.

***

We merged & released some deeper changes to the `dash-renderer` code in #1103. It looks like this introduced some conflicts in this PR, sorry about that!

***

This is looking really great otherwise and I know the community will love it. Are you interested in continuing to add new features to this or would you be looking to try to merge this in soon and follow up with more features? Either way is fine!

jjaraalm:
@chriddyp I'm in no rush to get this merged in. I'd love to add more features, but just need to find time to  work on it at some point. I didn't realize that `Graphs` were async! Good point, I was just going to try and monitor to see when react reported everything complete. It would be nice to know, but sounds like we should postpone it for now.

Re: #1103, I've been waiting on that so I'm more than happy to accommodate. Interestingly, I noticed that #1103 (or some other change in dev) messes with the `viz.js` callback graph and makes it very unstable for me. Looks like it's being continually rebuilt and there's some non-deterministic code (probably in `viz.js`) that jiggles the layout. Not sure if anyone's opened a bug report on it.

alexcjohnson:
Yep, that was me in #1103 mucking with the existing callback graph implementation - had to tweak it a bit in order to not totally choke on the dict IDs. Sorry about the merge conflicts! Feel free to file a bug report about the instability you're seeing, but my personal feeling is what you have here is already such an improvement that we should work to get it merged without adding any more features than it already has; then we can add all these other ideas in future PRs.

chriddyp:
> but my personal feeling is what you have here is already such an improvement that we should work to get it merged without adding any more features than it already has; then we can add all these other ideas in future PRs.

I agree! The work done in this PR already improves things 10x, I would love to see the community start to use this new version 👍

nicolaskruchten:
I'd love to see a screenshot of what this looks like when multiple props are involved in a single component, similar to the `cities-dropdown` here: https://github.com/nicolaskruchten/dash_callback_chain/blob/master/example.png

jjaraalm:
Ok, sounds good. I'm working on merging in dev right now. Losing the request_queue means I'll need to rework some stuff. I hope to have a working version soon.

I haven't played around with very complex chains yet. There will be some layout/graphical differences:
- `viz.js` seems to prefer rounded edges and have a better edge avoidance system. `cytoscape` looks like it will only auto-round edges if there are multiple edges connecting the same nodes, which can only happen if a property is both`Input` and `State` or both `Output` and `State`. 
- Initial layout and sizing is finicky, and I think this needs to get resolved before we can merge this in. It should hopefully be easy to fix this, it just hasn't been a high priority

nicolaskruchten:
I don't think curved edges are a must, personally, and viz.js always kind of had odd curves IMO :) So long as we can avoid edges overlapping on nodes I think we'll be OK.

alexcjohnson:
@jjaraalm can I help with the merge here? I'm happy to do it but don't want to interfere with any unpushed WIP you have.

nicolaskruchten:
I'm so excited about this PR :) Maybe it makes sense to try to tackle it in chunks? I.e. replacing Viz.js/adding in client-side and states in one PR, then layering on this interactive stuff in another?

alexcjohnson:
I tend to think it’ll be easier to keep what’s here and just finish it without any more features... that’s what I’ll try anyway.

jjaraalm:
Sorry, I haven't been able to work on this recently. If you want to take it over that would be fine with me. I have some changes I haven't pushed that make it (mostly) work with wildcards. I'll push what I have alter today.

josegonzalez:
@jjaraalm Let us know if there is anything we can do to help (I've personally always wanted a nicer debugger in Dash)! Can't wait to see what you have in the pipeline in regards to wildcard support here :)

dash-renderer/src/components/error/CallbackGraph/CallbackGraphContainer.react.js
```diff
@@ -5,13 +5,22 @@ import stylesheet from './CallbackGraphContainerStylesheet';
...
-import Dagre from 'cytoscape-dagre';
 import CytoscapeComponent from 'react-cytoscapejs';
-Cytoscape.use(Dagre);
 
 import PropTypes from 'prop-types';
 
 
+// Set the layout method.
+// NOTE: dagre should be nicer for DAG, but even with manual
+//       cycle pruning, it gives terrible results with State.
```
alexcjohnson, line 23: @jjaraalm can you share some examples where dagre doesn't look good? I'm testing a few things and they generally look good to me - and much better than breadthfirst.

dash-renderer/src/components/error/CallbackGraph/CallbackGraphContainer.react.js
```diff
@@ -283,13 +292,23 @@ function CallbackGraph() {
...
 
+    const {graphLayout} = profile;
+    const cyLayout = graphLayout || {
+        name: 'dagre',
+        padding: 10,
+        spacingFactor: 0.8,
+        // after initial layout, just use this again on later draws
+        // but we'll also save the layout whenever users interact with it
+        ready: setPresetLayout,
+    }
```
alexcjohnson: The first render uses Dagre - after that we just save the actual layout and regurgitate it when rendering again. That means you can close and reopen the graph, and it'll be faster the second time (only really noticeable on a huge graph like dash-docs, but there it's 4-5 seconds the first time, half a second after that), as well as saving any zoom / pan / moved nodes from the last time you looked at the graph.

A bit hacky, I just stash the layout with the profiling info in the redux store.

Dagre seems to do a good job for all the apps I've found to play with. On the biggest one, dash-docs, it does make a few silly errors - 5 or 6 out of the hundreds of nodes are wildly out of place, with an obvious way to fix them. I played around with Dagre's options a bit but could only make this worse, and anyway the vast majority of the graph looks good. Other more reasonable graphs look fine AFAICT. But I'd love it if other folks could try it out and let me know what you think!

dash-renderer/package.json
```diff
@@ -6,10 +6,10 @@
   "scripts": {
     "prepublishOnly": "rm -rf lib && babel src --extensions=\".ts,.tsx,.js,.jsx\" --out-dir lib --copy-files",
     "private::format.js-eslint": "eslint --quiet --fix .",
-    "private::format.js-prettier": "prettier --config .prettierrc --write \"src/**/*.js\"",
+    "private::format.js-prettier": "prettier --config .prettierrc --write \"src/**/*.{js,ts}\"",
```
Marc-Andre-Rivet, line 9: Would modify these to also include `tsx` extensions

dash-renderer/src/StoreObserver.ts
```diff
@@ -48,33 +42,31 @@ export default class StoreObserver<TStore> {
...
 
     private __init__ = (store?: Store<TStore>) => {
         this._store = store;
         if (store) {
             this._unsubscribe = store.subscribe(this.notify);
         }
 
-        forEach(o => o.lastState = null, this._observers);
-    }
+        forEach(o => (o.lastState = null), this._observers);
```
Marc-Andre-Rivet, line 60: That just feels weird but ok..

dash-renderer/src/actions/dependencies_ts.ts
```diff
@@ -14,20 +14,31 @@ import {
     pickBy,
     props,
     reduce,
-    zipObj
+    zipObj,
```
Marc-Andre-Rivet, line 17: Could we consider using `trailingComma: "none"`?

dash-renderer/src/components/error/CallbackGraph/CallbackGraphContainer.react.js
```diff
@@ -1,89 +1,415 @@
...
+        elements.push({
+            data: {
+                source: source,
+                target: target,
+                type: type,
+            },
+        });
+    }
+
+    (graphs.callbacks || []).map((callback, i) => {
```
Marc-Andre-Rivet, line 74: Should be a `forEach`, we're not mapping anything

dash-renderer/src/actions/dependencies_ts.ts
alexcjohnson: Yeah I'm on board with no trailing commas - this way is designed to minimize diffs, but I agree that a bigger diff is justified in the name of better readability.

dash-renderer/tslint.json
```diff
@@ -1,7 +1,8 @@
 {
     "defaultSeverity": "error",
     "extends": [
-        "tslint:recommended"
+        "tslint:recommended",
+        "tslint-config-prettier"
```
Marc-Andre-Rivet, line 5: I think "quotemark" should either be removed or updated to:
        "quotemark": [true, "single", "avoid-escape"],

So that tslint doesn't disagree with prettier if wrapping a " ' ' ' " or ' " " " ' depending on settings.

alexcjohnson: yeah we should probably remove a whole bunch of these overrides... I just got it far enough that they didn't fight with each other in the configuration I saw. For now I'll just 🔪 `quotemark`.

rpkyle:
> @chriddyp sounds good.
> 
> I've added support for reporting server-side timings using `Server-Timing` headers. This way (in modern browsers) we get timing reported both in dev-tools on a per-request basis and through introspection as an aggregate total.
> 
> ```
> Server-Timing: dash_total;dur=1002
> Server-Timing: task_1;desc="The first task.";dur=100
> Server-Timing: task_2;dur=700
> Server-Timing: task_3;desc="Cleanup task.";dur=200
> ```
> 
> The `dash_total` resource is automatically computed on all requests.

@jjaraalm Thanks for all your efforts to improve the callback graph implementation -- the enhanced context provided by this PR will be very helpful for debugging purposes. 

One brief question about `dash_total`; I see it in `Server-Timing` above, but I'm unsure where it's added to the headers. Will this happen elsewhere, or might we want to include something like the following just before `for name, info ...`?

```python
response.headers.add("Server-Timing", 'dash_total;dur={}'.format(dash_total["dur"]))
```

alexcjohnson:
> I see it in Server-Timing above, but I'm unsure where it's added to the headers.

@rpkyle the current implementation actually has this called `__dash_server` - it's created [here](https://github.com/plotly/dash/pull/1179/files#diff-228c9aefac729977642672aa1416bacaR1409), finalized [here](https://github.com/plotly/dash/pull/1179/files#diff-228c9aefac729977642672aa1416bacaR1414), and processed along with the custom times [here](https://github.com/plotly/dash/pull/1179/files#diff-228c9aefac729977642672aa1416bacaR1425).

dash/testing/browser.py
```diff
@@ -148,7 +153,43 @@ def percy_snapshot(self, name="", wait_for_callbacks=False):
...
+                    i.height = c.height;
+                    i.setAttribute('style', c.getAttribute('style'));
+                    i.className = c.className;
+                    i.setAttribute('data-canvasnum', stash.length);
+                    stash.push(c);
+                    c.parentElement.insertBefore(i, c);
+                    c.parentElement.removeChild(c);
+                });
+            """
+            )
```
alexcjohnson, line 173: New feature for `dash.testing` - `percy_snapshot` adds a kwarg `convert_canvases` that turns every `<canvas>` into an `` with matching contents, then puts the canvases back afterward. This lets us capture Cytoscape, among other things (cc @xhlulu)
Screen Shot 2020-08-29 at 11 39 51 PM
Sizing isn't quite as expected, but I assume that's just Percy rendering as a different screen size from what's used in circleci.

dash-renderer/src/components/error/CallbackGraph/CallbackGraphContainer.react.js
```diff
@@ -141,6 +141,26 @@ function flattenInputs(inArray, final) {
...
+        padding: 10,
+        nodeSep: 0,
+        rankSep: 80,
+        rankDir: 'LR'
+    },
+    force: {
+        name: 'cose',
+        padding: 10,
+        animate: false
+    }
```
alexcjohnson, line 161: @nicolaskruchten I couldn't come up with a single layout that always seemed better than the others, so made a dropdown to choose among three. Another side benefit of this is if you drag nodes around to tweak the layout, and you want to get back to the original, you can change to a different algo and then back to the original.

tests/integration/devtools/test_callback_timing.py
```diff
@@ -0,0 +1,38 @@
+from time import sleep
+import requests
+
+import dash_html_components as html
+import dash
+from dash.dependencies import Output, Input
+
+
+def test_dvct001_callback_timing(dash_thread_server):
```
alexcjohnson, line 9: @rpkyle this test should cover parity between Python and R for the `Server-Timing` header, including `record_timing`.

PR #1185, opened 2020-04-09
Sort also the directory name for proper ordering between css and js plugin folders

winnerNick:
This a small fix.
I have sorted the os.walk function to achieve ordering in directories naming.
Many times you need to include plugins in the assets folders (e.g. 10_jquery/..., 20_bootstrap/...), but some plugins depends on others, so you need a way to load then first.
Until now, sorting is only available for filenames in each folder, which cannot cover the above case.
The workaround would be to have all files in one folder, but this will disable the ability to structure properly your assets folder.


alexcjohnson:
Seems reasonable to me, thanks! Just needs a CHANGELOG entry and it's ready to go.

PR #1186, opened 2020-04-09
Assorted fixes required for CRAN submission

rpkyle:
This PR proposes several enhancements and minor fixes in the R package generator, based on requests from CRAN maintainers:
- ~support for arbitrary file extensions in component library assets (required by plotly/dashR#186)~ superseded by plotly/dash#1078
- [x] fixes for wildcard props handling (ensures these are generated within components)
- [x] adjust line wrapping in R online help to satisfy CRAN upload requirements (wrap at 60 chars instead of 80)
- [x] add all authors and copyright holders to `Authors@R` field within `DESCRIPTION`
- [x] add `\value` field to all `.Rd` files for exported methods and "explain function results". for functions returning no value, document as `\value{None}`.
- [x] where copyrights held by an entity other than package authors, indicate via `cph` role attributes in the `Authors@R` field, or by using a `Copyright` field. 

Specifically, two new YAML keys are supported in `dash-info.yaml`:
- `pkg_copyright`: when present, will add in a `Copyright:` line including the string value (see https://cran.r-project.org/web/packages/policies.html)
- `pkg_authors`: when present, `dash-generate-components` will use this string as-is for `Authors@R:`, which CRAN prefers to `Author:` and `Maintainer:` elements within `DESCRIPTION`. When this string is not present, the generator will create it based on the `author` key within `package.json`, and assume that the author is the maintainer.


dash/development/_r_components_generation.py
```diff
@@ -422,7 +422,7 @@ def write_help_file(name, props, description, prefix, rpkg_data):
             funcname=funcname,
             name=name,
             default_argtext=textwrap.fill(default_argtext,
-                                          width=80,
+                                          width=60,
```
Marc-Andre-Rivet, line 443: 60, 70, 80? is `60` wrong?

dash/development/_r_components_generation.py
```diff
@@ -259,7 +262,8 @@ def generate_js_metadata(pkg_data, project_shortname):
     sys.path.insert(0, os.getcwd())
     mod = importlib.import_module(project_shortname)
 
-    alldist = getattr(mod, "_js_dist", []) + getattr(mod, "_css_dist", [])
+    alldist = getattr(mod, "_js_dist", []) + getattr(mod, "_css_dist", []) \
+        + getattr(mod, "_asset_dist", [])
```
Marc-Andre-Rivet, line 266: We've discussed an `_asset_dist` bucket that would generalize on the existing `_js_dist` and `_css_dist` ones. Is this work towards that direction or are the arbitrary file extensions required to be in `_asset_dist`? The Python PR does not have an equivalent for this atm.

.circleci/config.yml
```diff
@@ -124,7 +124,7 @@ jobs:
   build-core-27:
       <<: *build-core
       docker:
-        - image: circleci/python:2.7-stretch-node-browsers
+        - image: circleci/python:2.7.16-stretch-node-browsers
```
Marc-Andre-Rivet, line 127: From the previous CI runs, this seems unwarranted. The tests seemed to be passing fine and not requiring a specific version override atm.

dash/development/_r_components_generation.py
rpkyle: `60` is correct, can't change text of commit message easily with rebase given the merge commits in the way -- figured it will get squashed in the end, and disappear. Sorry, should've explained offline.

dash/development/_r_components_generation.py
rpkyle: I think @wbrgss suggested we discuss this with you, but I'm fine with `_asset_dist` as a general bucket if it works elsewhere. If we should use another structure, just let me know -- either works for me.

.circleci/config.yml
rpkyle: Ah, OK, will revert.

rpkyle: fixed in c39de33

rpkyle:
This PR is held until some refactoring is completed by @Marc-Andre-Rivet related to asset handling within Dash, seems probable that a "legacy mode" using `_js_dist` and `_css_dist` will remain, and a revised handler with just `_assets` and a better interface for defining them will be implemented.

chriddyp:
> related to asset handling within Dash

Is there an issue that discusses this?

Marc-Andre-Rivet:
@rpkyle I think we should drop `_asset_dist` and just use `_js_dist` for now. While the `_asset_dist` is an interesting addition, I think we should rework css/js/asset in depth and with all Dash flavors in mind before committing to supporting additional props. `_js_dist` is not perfect but it covers all usages with `_css_dist`

This PR would be superseded by https://github.com/plotly/dash/pull/1078.

dash/development/_r_components_generation.py
rpkyle: > @rpkyle I think we should drop `_asset_dist` and just use `_js_dist` for now. While the `_asset_dist` is an interesting addition, I think we should rework css/js/asset in depth and with all Dash flavors in mind before committing to supporting additional props. `_js_dist` is not perfect but it covers all usages with `_css_dist`
> 
> ~This PR~ Portions of this PR would be superseded by #1078.

OK, I'll remove the elements of this pull request which are addressed by #1078. The remaining two items are still worth merging in for the sake of the CRAN release.

rpkyle:
@alexcjohnson @Marc-Andre-Rivet This is ready for your review, and should resolve the CRAN maintainer concerns. I'd like to merge this today if possible, so Marc-André can use it for package builds starting with the upcoming Dash release. 🙏

rpkyle:
> 💄 There are a bunch of places here the indentation is just changed from 4 to 8 spaces inside parens, either function args or `if` conditions - please put back the original 4 spaces. Also there's a `._r_components_generation.py.swp` file committed -> should be `.gitignore`'d?
> 
> Aside from that these changes look good to me!

Sure, not entirely aware what happened, but the linter demanded more spaces, so I added them. 🤷‍♂️  I can try removing them, but if it fails the check I'll have to determine what led to the (new) warnings.

Will add the `vim`-generated swapfile, should probably `.gitignore` all `.swp` files, never any reason to commit them.

rpkyle:
> Also there's a `._r_components_generation.py.swp` file committed -> should be `.gitignore`'d?
> 

fixed in ae28bac

rpkyle:
> 💄 There are a bunch of places here the indentation is just changed from 4 to 8 spaces inside parens, either function args or `if` conditions - please put back the original 4 spaces.

OK, restored in 06dbc8da. Will try to sort out why that happened, might be a difference in configuration between `pylint` on my machine vs. CircleCI.

rpkyle:
> This PR should be retargeted to `dev` now that #1078 is merged, yes? Anyway after that and the Windows build is sorted out I think it's ready to go! 💃

I've taken your advice on removing the spaces within the `--r-suggests` statement inside of `package.json` in `plotly/dash-core-components`, and the workaround enables the test to pass, but we still haven't sorted out why it fails (and we should).

I've opened https://github.com/plotly/dash-core/issues/184 so we don't lose track, and 019df65 will add back the spaces we have now omitted when necessary.

PR #1197, opened 2020-04-15
Initial support for Dash.jl component generation

### LINKED ISSUE #1189: Support for generating component libraries for use with Dash.jl
Currently, we generate Dash component libraries for Python and R dynamically from JavaScript, using a common `dash-generate-components` interface. This allows us to release updates to libraries like `dash-core-components` simultaneously for all supported backends, with essentially identical help string content, tailored to the idioms of each language.

As for [Python](https://github.com/plotly/dash/blob/dev/dash/development/_py_components_generation.py) and [R](https://github.com/plotly/dash/blob/dev/dash/development/_r_components_generation.py), we should provide a [Dash.jl](https://github.com/plotly/Dash.jl) component generator, tentatively entitled `_jl_components_generation.py`, which integrates seamlessly with our existing component generation pipeline.

rpkyle:
This PR proposes to contribute modifications to `dash-generate-components` which enable generation of components for use with the Julia version of the Dash framework.

- [x] Support for generating individual Julia components with docstrings
- [x] Changes to the `dash-generate-components` argument parser
- [x] Add Julia reserved keywords list to Dash, and omit any properties which match an element in this list
- [x] Generation of module metadata (e.g. `Project.toml`)
- [x] Updating component template to reflect any changes to the API required for general parity with Dash for Python & Dash for R

For the moment, the proposed syntax to generate Julia components includes a prefix, as with the R component generator (there are proposals to allow importing Julia modules with aliases, as in Python, though I don't believe this is available yet, based on https://github.com/JuliaLang/julia/issues/1255).

```
dash-generate-components ./src/components dash_html_components -p package-info.json --r-prefix 'html' --jl-prefix 'html'
```

Closes #1189.

@Marc-Andre-Rivet @waralex


mbauman:
There's nothing particularly special about `Project.toml` — and you don't need to use Julia to create it.  The only challenge is the creation of the package UUID.  It must remain constant between versions and must be distinct from all other UUIDs.  I think what you could do is something like:

```python
import uuid
u = uuid.UUID('c3754af0-4ebb-44ab-a1a8-173fddf15295')
uuid.UUID(hex=u.hex[:-12] + hex(hash(__name__))[-12:])
```

That is, use a pre-computed (and shared) UUID, but replace the last 12 hex digits with some bits from the hashed package name.

mbauman:
> For the moment, the proposed syntax to generate Julia components includes a prefix, as with the R component generator (there are proposals to allow importing Julia modules with aliases, as in Python, though I don't believe this is available yet, based on JuliaLang/julia#1255)

Yeah, we don't yet have a nice `import DashCoreComponents as DCC` syntax, but it's as easy as:

```
import DashCoreComponents
const DCC = DashCoreComponents
```

Personally, I'd be happy doing a `using DashCoreComponents` and just using `dropdown`s and `input`s directly (without qualification).

rpkyle:
Thanks @mbauman, these responses are most helpful.
 
With regard to `dropdown` and `input`, my only thought is that there may be name collisions between generated component names and existing Julia functions. 

Realistically, I'm not as comfortable with this language as I am with R, but perhaps my concerns are misplaced. For now, it seems workable, curious what others think.

@Marc-Andre-Rivet @chriddyp @alexcjohnson

waralex:
> In terms of referencing things like input, my main concern is the potential for collisions between existing Julia functions and the current component interface (particularly those from imported modules).

It seems to me that the more significant problem is the name conflict in different modules. For example `input` is present in `html` and `bootstrap` component modules and I think not only in them

rpkyle:
> > In terms of referencing things like input, my main concern is the potential for collisions between existing Julia functions and the current component interface (particularly those from imported modules).
> 
> It seems to me that the more significant problem is the name conflict in different modules. For example `label` is present in `html` and `bootstrap` component modules and I think not only in them

I believe we'd address this by declaring `html_label` and `dbc_label` (assuming `dbc` is the abbreviated form of `DashBootstrapComponents.jl` as it often is when importing `dash-bootstrap-components` in Python), but maybe I'm missing something here.

rpkyle:
@waralex is currently in the process of updating the component generator templates for Julia, and he'll commit them to this branch when ready for review.

dash/development/_jl_components_generation.py
```diff
@@ -0,0 +1,400 @@
...
+    # now copy over all JS dependencies from the (Python) components dir
+    # the inst/lib directory for the package won't exist on first call
+    # create this directory if it is missing
+    if os.path.exists("deps"):
+        shutil.rmtree("deps")
+
+    os.makedirs("deps")
+
+    for javascript in glob.glob("{}/*.js".format(project_shortname)):
+        shutil.copy(javascript, "deps/")
```
mbauman, line 394: Instead of simply putting these into `deps`, it'd be really awesome to register these files (js/css/maps) as [Artifacts](https://julialang.github.io/Pkg.jl/v1/artifacts/).  The main motivation for me would be that it'd allow the Julia PackageCompiler — that is the tool that can compile Julia code into redistributable binaries — to find and include them in its binaries.  It's really helpful for deployments.

waralex: Does it make sense to create Artifacts for data that can only be used with a certain package and are meaningless for everyone else? I haven't worked with Artifacts yet, but from the documentation, it seemed to me that this is a solution for data that is independent of concrete packages.

waralex: Theoretically, we may not register components as packages at all. Instead, we can create artifacts with resources and a JSON file with component descriptions. And we could add a function or macros to Dash that generates functions for components based on the artifact name and adds resources to the index page. But this makes the Dash interface different from the interface in Python and R.

rpkyle: Hi @mbauman Does "deployments" here refer to deploying a Dash.jl web application? Or just deploying collections of Julia-related data into containerized environments?

@waralex Having read through https://julialang.org/blog/2019/11/artifacts/ briefly, I could envision scenarios in which registering data for Dash component libraries as artifacts does make sense. 

For example, the `dashBio` package in R has sample data, which I think would be worth registering as an artifact instead of downloading/rebundling within the Juila package itself. (Assuming I understand what Matt is getting at, of course.)

mbauman: Yes, I mean deploying the Dash.jl web app to a service.  For example, in my [POC demo](GitHub.com/mbauman/covid19), I used PackageCompiler to compile `Dashboards.jl` in order to improve its performance on the slow free Heroku dynos... but I had to keep the source around since that's where these assets lived.  If these were artifacts, I could essentially create a much lighter-weight binary executable.

And yes, we're actively working on workflows for improving data-as-artifacts as it can then provide complete reproducibility (since the artifacts themselves are content-hashed and versioned) — and allow you to remove large datasets from Git repos.

dash/development/_jl_components_generation.py
```diff
@@ -0,0 +1,523 @@
...
+                    )
+                    for prop_name, prop in list(type_object["value"].items())
+                )
+            ),
+        )
+
+    return dict(
+        array=lambda: "Array",
+        bool=lambda: "Bool",
+        number=lambda: "Float64",
```
alexcjohnson, line 158: I'm not sure if there are functional consequences of this, but at least from a documentation perspective might it be better to include more number types? `Float64 | Int64` looks like it covers the output of `JSON2.read`, ie values to be received by callback functions. But values you can send back as callback returns are broader, likely `Real`.

dash/development/_jl_components_generation.py
```diff
@@ -0,0 +1,523 @@
...
+                description=prop["description"],
+                indent_num=0,
+            )
+            for p, prop in list(filter_props(props).items())
+        ),
+    )
+
+
+# pylint: disable=too-many-arguments
+# pylint: disable=too-many-arguments
```
alexcjohnson, line 300: 🔪 - both of them in fact 😅 Doesn't look necessary in the Python or R variants anymore either...

dash/development/_jl_components_generation.py
```diff
@@ -0,0 +1,523 @@
...
+# lower case names for the resulting functions; if a prefix
+# is supplied, leave it as-is
+def format_fn_name(prefix, name):
+    if prefix:
+        return "{}_{}".format(prefix, name.lower())
+    return name.lower()
+
+
+def generate_metadata_strings(resources, metatype):
+    def noting_or_string(v):
```
alexcjohnson, line 364:
```suggestion
    def nothing_or_string(v):
```

dash/development/_jl_components_generation.py
waralex: This is only used in documentation. Yes, it makes sense to replace types with abstract ones. 
Abstract types for numbers looks like this:
```
abstract type Number end
abstract type Real     <: Number end
abstract type AbstractFloat <: Real end
abstract type Integer  <: Real end
abstract type Signed   <: Integer end
abstract type Unsigned <: Integer end
```

waralex: This is only used in documentation. Yes, it makes sense to replace types with abstract ones.
Abstract types for numbers looks like this:
```
abstract type Number end
abstract type Real     <: Number end
abstract type AbstractFloat <: Real end
abstract type Integer  <: Real end
abstract type Signed   <: Integer end
abstract type Unsigned <: Integer end
```

dash/development/_jl_components_generation.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+def generate_toml_file(project_shortname, pkg_data):
+    package_author = pkg_data.get("author", "")
+    project_ver = pkg_data.get("version")
+    package_name = jl_package_name(project_shortname)
+    u = uuid.UUID(jl_dash_uuid)
+    package_uuid = uuid.UUID(hex=u.hex[:-12] + hex(hash(package_name))[-12:])
+
+    authors_string = 'authors = ["{}"]'.format(package_author) if package_author else ""
```
alexcjohnson, line 413: minor: with no authors we'll end up with a funny-looking blank line in the output

dash/development/_jl_components_generation.py
alexcjohnson: OK good, let's go with `Real`.

mbauman: There's also the `any=lambda: "Bool | Float64 | String | Dict | Array"` below.  I remember finding these annotations very helpful when looking at the online Python docs.  And we can express this directly in Julia with `Union{...}`s.

In fact, it looks like we can use the `union` lambda below to print out real Julian `Union{}`s using your existing architecture.  I'm afraid it'd take further restructuring to make [an options block like this](https://dash.plotly.com/dash-core-components/checklist) format like the below for Julia, but this would be closer to how we'd naturally document these things:

> * `label::Union{String, Number}` (required): The checkbox's label
> * `value::Union{String, Number}` (required): The value of the checkbox. This value corresponds to the items specified in the value property.
> * `disabled::Bool` (optional): If true, this checkbox is disabled and can't be clicked on.

All that said, this sort of strict notation doesn't work as well for arrays and dicts, though, since we support far more than just `Array` and `Dict`.  The builtin `Array` is just one of many possible `AbstractArray`s, and JSON2 also supports serializing `Tuple`s and `Set`s the same way.  Similarly, the builtin `Dict` is just one of many dictionary-like things; JSON2 also supports NamedTuples and even your own structs.  That said, I don't want to lean too heavily on what the particular JSON library does.  This gets into the nitty gritty, but JSON2 should probably serialize all `AbstractDict`s as JSON dicts; it [doesn't right now](https://github.com/quinnj/JSON2.jl/issues/35). It may be good to move to JSON3 in the future which has a [far more robust manner of flagging these types](https://quinnj.github.io/JSON3.jl/stable/) and is even speedier, so tying yourself to a particular implementation isn't the best.

One way we distinguish between the concrete implementations (like `Array` and `Dict`) and their general behaviors in our own documentation is just to use lowercase without code formatting, simply: array and dictionary.  And if I were using these sorts of more relaxed terms, I wouldn't use the `Union{}` notation at all.

So all that to say, maybe don't uppercase any of these guys?  Any maybe keep the (x | y | z) union printing?

waralex: The point is that these types are actually mapping what JS expects and the question of linking them to Julia types is fairly conditional. I don't know how to write this in the documentation to make it clear. There is no verification of these types at the Julia level. To json dict also mapped NamedTuple, for example and Tuple mapped to json array.

waralex: >So all that to say, maybe don't uppercase any of these guys? Any maybe keep the (x | y | z) union printing?

I support it. With an indication in the documentation that these are types at the js level, not at the Julia level

dash/development/_jl_components_generation.py
mbauman: You could simply swap these two lines so the blank line falls between sections:

https://github.com/plotly/dash/pull/1197/files#diff-4de66a02832ac93535450311a2298fdeR87-R88

alexcjohnson: or just move one of the newline characters from `jl_projecttoml_string ` to `authors_string`

dash/development/_jl_components_generation.py
alexcjohnson: > these are types at the js level, not at the Julia level

Yes, but the Julia app developer needs to know two things:
- What type(s) can they expect to receive in callbacks?
- What type(s) can they provide as prop values, either as callback return values or in layout components?

The more precision we can provide around these two issues the better. The first is easier, as the set of possible inputs from a JSON string is fairly limited. And then I assume that JSON2 is at least symmetric in the sense that `JSON2.write(JSON2.read(s))==s` - ie if you provide a return value consistent with the argument value we would provide, it will always work.

Given that, I'd prefer that we base our component documentation (ie this generated code) on this more restricted set of types you get from de-serializing JSON, and then somewhere in our general `dash.jl` docs we can describe the other types that you can provide as prop values - and we can frame this explicitly as "Anything that JSON2 (or JSON3 if we switch to that) will serialize the same way as type X."

In Dash for Python we've found that folks get confused when these types don't match - for example the DCC date pickers can be initialized using Python `datetime` objects, because they have a standard JSON serialization, but when you create a callback it will receive the date as a string. So we document this as a string type and explicitly stringify it in our [main example](https://dash.plotly.com/dash-core-components/datepickersingle):
```py
date=str(dt(2017, 8, 25, 23, 59, 59))
```
(though I notice in later examples you'll see a `datetime` passed in directly - we should probably switch them all to strings, and/or make an explicit note about serializing `datetime` objects!)

waralex: You should also keep in mind that each argument is converted to JSON using `Front.to_dash`. By default, this is an identical conversion. But it can be overloaded with third-party packages. For example you can use it to pass a PlotlyBase object instead of dcc_graph

dash/development/_jl_components_generation.py
rpkyle: fixed in 1a15907e

dash/development/_jl_components_generation.py
rpkyle: fixed in 5b2edcdc

dash/development/_jl_components_generation.py
rpkyle: @alexcjohnson changed to `Real` in fd24e508

dash/development/_jl_components_generation.py
rpkyle: @alexcjohnson nice catch! fixed in 2e813bf1

alexcjohnson: As @mbauman mentioned, we should also update this in `any`

dash/development/_jl_components_generation.py
alexcjohnson: As @mbauman mentioned we should also update this in `any`

rpkyle: fixed in 7eb50221

rpkyle:
@alexcjohnson @waralex My understanding is that we're holding this PR until we settle on a final configuration for the core packages, and a suitable method for ensuring users always have the right combination of Dash + core component libraries, and that installing Dash installs them at the same time -- as described in more detail in https://github.com/plotly/Dash.jl/issues/22.

dash/development/_jl_components_generation.py
staticfloat: Hello!  I'm the author of the Artifacts system; @mbauman asked me to come in and discuss artifacts with you all.

> I haven't worked with Artifacts yet, but from the documentation, it seemed to me that this is a solution for data that is independent of concrete packages.

This is true in some senses, but not true in others.  The data is independen" in that it is stored in a separate directory, but it is not independent in that the code of your package can be as reliant on the artifact data as you want.  Examples of artifact usage in the ecosystem right now include:

* Precompiled dynamic libraries
* Large data files that shouldn't live inside a git repository
* Small pieces of data that many packages all need to access

Note that Artifacts are immutable, we don't allow writing to them.  This is part of our master plan to increase reliability and reproducibility within the Julia package ecosystem; we want to eventually set package directories to be read-only by default, and disallow packages the ability to modify themselves.  To this day, a disheartening percentage of the bug reports we get have to do with local state getting wedged somehow, and package authors not designing their packages such that it can recover from a half-finished tarball extraction into their `deps` folder, for example.  To that end, whenever we see new packages downloading things into `deps/` we immediately jump to Artifacts as a good solution.

A side-benefit, as @mbauman pointed out, is that Artifacts implicitly perform their file path calculating at runtime, which avoids embedding file paths into precompilation files.  This allows a compiled form of your code to be relocatable; and enables users to create and distribute applications that bundle Dash.

I'm happy to help out if there are any questions/concerns about using Artifacts.

PR #1201, opened 2020-04-17
1193 multipage validation

### LINKED ISSUE #1193: [BUG] dash 1.11 breaks Dynamic Multipage App Validation
**Describe your context**

`pip list | grep dash` 
```
dash                               1.11.0              
dash-bootstrap-components          0.9.1               
dash-core-components               1.9.1               
dash-dangerously-set-inner-html    0.0.2               
dash-flow-example                  0.0.5               
dash-html-components               1.0.3               
dash-renderer                      1.4.0               
dash-table                         4.6.2               
```

**Describe the bug**

Copy example *Dynamically Create a Layout for Multi-Page App Validation* from https://dash.plotly.com/urls

<details>
  <summary>Click to expand!</summary>

```python
import dash

import flask

app = dash.Dash(
    __name__,
    external_stylesheets=['https://codepen.io/chriddyp/pen/bWLwgP.css']
)

url_bar_and_content_div = html.Div([
    html.Div(id='page-content')
])

layout_index = html.Div([
    dcc.Link('Navigate to "/page-1"', href='/page-1'),
    html.Br(),
    dcc.Link('Navigate to "/page-2"', href='/page-2'),
])

layout_page_1 = html.Div([
    html.H2('Page 1'),
    dcc.Input(id='input-1-state', type='text', value='Montreal'),
    dcc.Input(id='input-2-state', type='text', value='Canada'),
    html.Button(id='submit-button', n_clicks=0, children='Submit'),
    html.Div(id='output-state'),
    html.Br(),
    dcc.Link('Navigate to "/"', href='/'),
    html.Br(),
    dcc.Link('Navigate to "/page-2"', href='/page-2'),
])

layout_page_2 = html.Div([
    html.H2('Page 2'),
    dcc.Dropdown(
        id='page-2-dropdown',
        options=[{'label': i, 'value': i} for i in ['LA', 'NYC', 'MTL']],
        value='LA'
    ),
    html.Div(id='page-2-display-value'),
    html.Br(),
    dcc.Link('Navigate to "/"', href='/'),
    html.Br(),
    dcc.Link('Navigate to "/page-1"', href='/page-1'),
])


def serve_layout():
    if flask.has_request_context():
        return url_bar_and_content_div
    return html.Div([
        url_bar_and_content_div,
        layout_index,
        layout_page_1,
        layout_page_2,
    ])


app.layout = serve_layout


# Index callbacks
@app.callback(Output('page-content', 'children'),
              [Input('url', 'pathname')])
def display_page(pathname):
    if pathname == "/page-1":
        return layout_page_1
    elif pathname == "/page-2":
        return layout_page_2
    else:
        return layout_index


# Page 1 callbacks
@app.callback(Output('output-state', 'children'),
              [Input('submit-button', 'n_clicks')],
              [State('input-1-state', 'value'),
               State('input-2-state', 'value')])
def update_output(n_clicks, input1, input2):
    return ('The Button has been pressed {} times,'
            'Input 1 is "{}",'
            'and Input 2 is "{}"').format(n_clicks, input1, input2)


# Page 2 callbacks
@app.callback(Output('page-2-display-value', 'children'),
              [Input('page-2-dropdown', 'value')])
def display_value(value):
    print('display_value')
    return 'You have selected "{}"'.format(value)


if __name__ == '__main__':
    app.run_server(debug=True)
```

</details>

**Expected behavior**

No errors.

**Screenshots**

```json
🛑 Errors (4)
⛑️
ID not found in layout
10:53:31 AM
Attempting to connect a callback Input item to component:
  "page-2-dropdown"
but no components with that id exist in the layout.

If you are assigning callbacks to components that are
generated by other callbacks (and therefore not in the
initial layout), you can suppress this exception by setting
`suppress_callback_exceptions=True`.
This ID was used in the callback(s) for Output(s):
  page-2-display-value.children
```
once per ID:
<details>
  <summary>Click to expand!</summary>

```json
⛑️
ID not found in layout
10:53:31 AM
Attempting to connect a callback Input item to component:
  "submit-button"
but no components with that id exist in the layout.

If you are assigning callbacks to components that are
generated by other callbacks (and therefore not in the
initial layout), you can suppress this exception by setting
`suppress_callback_exceptions=True`.
This ID was used in the callback(s) for Output(s):
  output-state.children
⛑️
ID not found in layout
10:53:31 AM
Attempting to connect a callback Output item to component:
  "page-2-display-value"
but no components with that id exist in the layout.

If you are assigning callbacks to components that are
generated by other callbacks (and therefore not in the
initial layout), you can suppress this exception by setting
`suppress_callback_exceptions=True`.
This ID was used in the callback(s) for Output(s):
  page-2-display-value.children
⛑️
ID not found in layout
10:53:31 AM
Attempting to connect a callback Output item to component:
  "output-state"
but no components with that id exist in the layout.

If you are assigning callbacks to components that are
generated by other callbacks (and therefore not in the
initial layout), you can suppress this exception by setting
`suppress_callback_exceptions=True`.
This ID was used in the callback(s) for Output(s):
  output-state.children
```

</details>

alexcjohnson:
Fixes #1193 - the `flask.has_request_context` trick inside an `app.layout` function can be used again to supply a special layout just for validation purposes.

In addition, added a new attribute `app.validation_layout` that does the same thing more explicitly. 

## Contributor Checklist

    -  [x] https://github.com/plotly/dash-docs/pull/857 updates the dash docs


tests/integration/devtools/test_callback_validation.py
```diff
@@ -730,11 +730,7 @@ def multipage_app(validation=False):
     layout_page_2 = html.Div(
         [
             html.H2("Page 2"),
-            dcc.Dropdown(
```
alexcjohnson: `dcc.Dropdown` is giving React deprecated lifecycle method warnings ATM... I thought we had already updated all of these in the 16.13 push? Anyway switched to `Input` to get my `assert not dash_duo.get_logs()` clean console test to pass.

dash/dash.py
```diff
@@ -421,18 +422,43 @@ def layout(self):
...
+                }
+                if props.get("children", children):
+                    props["children"] = children or []
+                return cls(**props)
+
+            layout_value = self._layout_value()
+            _validate.validate_layout(value, layout_value)
+            self.validation_layout = simple_clone(
+                # pylint: disable=protected-access
+                layout_value, [simple_clone(c) for c in layout_value._traverse_ids()]
```
alexcjohnson, line 459: Make this validation layout as simple as possible, to minimize its overhead on the page: - strip out components with no ID
- strip out most prop values, at least in Py3 where we can fully inspect the component constructor signature so we know what props are required.

This step is only run once when you set the layout, and even then only when you set it to a function AND you haven't already set `app.validation_layout`

dash-renderer/src/actions/dependencies.js
```diff
@@ -464,9 +464,17 @@ function wildcardOverlap({id, property}, objs) {
 }
 
 export function validateCallbacksToLayout(state_, dispatchError) {
-    const {config, graphs, layout, paths} = state_;
-    const {outputMap, inputMap, outputPatterns, inputPatterns} = graphs;
+    const {config, graphs, layout: layout_, paths: paths_} = state_;
     const validateIds = !config.suppress_callback_exceptions;
+    let layout, paths;
+    if (validateIds && config.validation_layout) {
+        layout = config.validation_layout;
```
chriddyp, line 471: Is this because `validation_layout` is pruned down, so faster to crawl?

dash/dash.py
chriddyp: If we're just using this to validate IDs w.r.t. callbacks, why do we need the required values?

alexcjohnson: We also validate props against the component type, so we can't just create dummy components, they have to be the same type as the real components. And unless I were to recreate the JSON serialization for components, the easiest way to do that is to actually instantiate the same component class. But doing so requires that we include required props, because the generated components enforce the required props during construction.

dash-renderer/src/actions/dependencies.js
alexcjohnson: This is just saying *if* we have a separate validation layout, that's the one we need to check the callbacks against. The fact that it's pruned should make `computePaths` (next line) faster, but the main reason for pruning was to reduce the size of `_dash-config` on the index page - I was imagining it could become enormous, if you have multiple tables and graphs on different pages all concatenated together.

chriddyp: Ah I see, because we're validating the components & component props in the front-end as well now, not just the callback signatures?

chriddyp: Whereas previously we were validating the components's properties in the render cycle I think?

dash/dash.py
chriddyp: Ohhhhhh by "props against the component type" do you mean during callback validation with messages like e.g. "You registered a callback to update a Div's childrenz property. However, Div does not have a "children" property."?

dash/dash.py
```diff
@@ -421,18 +422,43 @@ def layout(self):
...
     @layout.setter
     def layout(self, value):
         _validate.validate_layout_type(value)
-        self._cached_layout = None
+        self._layout_is_function = isinstance(value, patch_collections_abc("Callable"))
         self._layout = value
 
+        # for using flask.has_request_context() to deliver a full layout for
+        # validation inside a layout function - track if a user might be doing this.
+        if self._layout_is_function and not self.validation_layout:
```
chriddyp, line 435: I wonder if this is something we should turn off if `debug=False` too, or introduce a new `dev_tools_validate_layout` function. Just to reduce the performance hit if there are many small components (e.g. a `dash_html_components.Table`)

dash/dash.py
alexcjohnson: > e.g. "You registered a callback to update a Div's childrenz property. However, Div does not have a "childrenz" property."?

exactly.

dash/dash.py
alexcjohnson: Good point. Those mostly wouldn't have IDs, so wouldn't be sent to the front end anyway, but with debug off we don't need it at all. In fact I guess we should be able to short-circuit all the front-end validation in that case too.

alexcjohnson: I added clauses to skip this when `suppress_callback_validation` is used, but didn't remove it entirely when `debug=False`. There's more to think about here - on the one hand, without this validation the app behavior can change, and on the other there's a lot more validation we could be skipping if that's what we want to do. I've created #1206 to follow up.

eddy-geek:
I confirm the fix works on my app (tested with `dev` branch, after remembering to rebuild dash_renderer ;-)). Thanks for the responsiveness!

PR #1203, opened 2020-04-20
Component nested artifacts + Test

Marc-Andre-Rivet:
Based on the community PR #1125:

* Adds sanity testing for component with nested (and standard non-nested) artifacts
* Updates the R component generator to include all nested files in the target folder instead of only the files directly in the current folder
* Allows component authors to use the same target name for both R and Python generation, whatever the artifacts folder structure
* Replace dash-render-test-components.UncontrolledInput usage with dcc.Input
---

- [ ] DashR test https://github.com/plotly/dashR/pull/191
- [ ] Archive https://github.com/plotly/dash-renderer-test-components


dash/development/_r_components_generation.py
```diff
@@ -502,14 +501,19 @@ def write_js_metadata(pkg_data, project_shortname, has_wildcards):
...
+                continue
 
-    for sourcemap in glob.glob("{}/*.map".format(project_shortname)):
-        shutil.copy(sourcemap, "inst/deps/")
+            target_dirname = os.path.join(os.path.join("inst/deps/", rel_dirname))
+
+            if not os.path.exists(target_dirname):
+                os.makedirs(target_dirname)
+
+            shutil.copy(os.path.join(rel_dirname, filename), target_dirname)
```
Marc-Andre-Rivet, line 516: So this is the real change - everything else is just for the nested/standard file structure test.

Instead of globing files w/o recursion, glob with recursion (using os.walk because glob.glob does not support the `recursive` named param in py27)

tests/integration/test_generation.py
```diff
@@ -0,0 +1,20 @@
...
+        [
+            MyStandardComponent(id="standard", value="Standard"),
+            MyNestedComponent(id="nested", value="Nested"),
+        ]
+    )
+
+    dash_duo.start_server(app)
+
+    assert dash_duo.wait_for_element("#standard").text == "Standard"
+    assert dash_duo.wait_for_element("#nested").text == "Nested"
```
Marc-Andre-Rivet, line 20: Check if both dummy standard and nested components load correctly

.pylintrc
```diff
@@ -61,7 +61,8 @@ disable=fixme,
         old-style-class,
         superfluous-parens,
         bad-continuation,
-        line-too-long
+        line-too-long,
+        wrong-import-order
```
alexcjohnson, line 65: Is this necessary? I kind of like the rule to keep imports organized...

@plotly/dash-generator-test-component-standard/.Rhistory
```diff
@@ -0,0 +1,6 @@
+library(dash)
+library(dashHtmlComponents)
+library(dashCoreComponents)
+library(dashTable)
+library(dashGeneratedComponentStandard)
+quit()
```
alexcjohnson, line 6: 🔪 and ignore - this and `.Rdata`

package.json
```diff
@@ -8,15 +8,17 @@
     "private::lint.black": "if [[ $PYLINTRC != '.pylintrc' ]]; then black dash tests --exclude metadata_test.py --check; fi",
     "private::lint.flake8": "flake8 --exclude=metadata_test.py dash tests",
     "private::lint.pylint-dash": "PYLINTRC=${PYLINTRC:=.pylintrc37} && pylint dash setup.py --rcfile=$PYLINTRC",
-    "private::lint.pylint-tests": "pylint tests/unit tests/integration -d all -e C0410,C0411,C0412,C0413,W0109",
+    "private::lint.pylint-tests": "PYLINTRC=${PYLINTRC:=.pylintrc37} && pylint tests/unit tests/integration -d all --rcfile=$PYLINTRC",
     "private::lint.renderer": "cd dash-renderer && npm run lint",
+    "private::test.setup-nested": "cd \\@plotly/dash-generator-test-component-nested && npm ci && npm run build && pip install -e .",
+    "private::test.setup-standard": "cd \\@plotly/dash-generator-test-component-standard && npm ci && npm run build && pip install -e .",
```
alexcjohnson, line 14: Is your plan - the reason these packages are up in `@plotly` rather than in `tests` - to publish them for use in the R (& Julia) tests?

@plotly/dash-generator-test-component-standard/.Rhistory
Marc-Andre-Rivet: 🙈

package.json
Marc-Andre-Rivet: No. Don't plan on publishing them at this point. I simply wanted to isolate them from other concerns. The `package.json` file contains `"private": true` which should prevent accidental publication.

.pylintrc
Marc-Andre-Rivet: I was having issues with pylint 1.9.4 (py27) vs pylint 2.3.1 (py37) ordering expectations when using `from x import y as z`, but not doing that anymore, attempting revert.

PR #1212, opened 2020-04-26
Missing inputs

### LINKED ISSUE #1200: dash 1.11 suppresses unrelated callbacks on component-load if ReferenceError detected
**Describe your context**
```
dash                               1.10.0     
dash-core-components               1.9.0      
dash-html-components               1.0.3      
dash-renderer                      1.3.0         
dash-table                         4.6.2 
```

**Describe the bug**

When components are added to the layout of an app in dash 1.11 (e.g in a multipage app), all the callbacks associated with those new components are triggered in python but they won't update the app on the front end if a ReferenceError was detected (i.e if an output to a callback was added without the corresponding inputs).

**Expected behavior**

In previous versions of dash, all the well defined callbacks associated with the components added to the dash layout would still update the app

```python
import dash

app = dash.Dash(__name__, suppress_callback_exceptions=True)

app.layout = html.Div([
    html.Button("Button", id="input1", n_clicks=0),
    html.Div(id='output1', children='initial value for output1'),
    html.Div(id='partial-output-initial-layout'),
    html.Div(id='content'),
])


def render_layout():
    return html.Div([
        dcc.Input(id='input2'),
        html.Div(id='output2', children='initial value for output2'),
        html.Div(id='input3'),
        html.Div(id='output3', children='initial value for output3'),
        html.Div(id='partial-output-secondary-layout'),
    ])


@app.callback(dash.dependencies.Output('content', 'children'),
              [dash.dependencies.Input('url', 'pathname')])
def display_content(url):
    if url:
        return render_layout()


@app.callback(
    Output('output1', 'children'),
    [Input('input1', 'n_clicks')])
def button_callback(n_clicks):
    print('callback with input1')
    return 'callback with input1'


@app.callback(
    Output('output2', 'children'),
    [Input('input2', 'value')]
)
def dcc_input_callback(value):
    print('callback with input2')
    return 'callback with input2'



@app.callback(
    Output('output3', 'children'),
    [Input('input3', 'children')]
)
def html_div_callback(value):
    print('callback with input3')
    return 'callback with input3'


@app.callback(
    Output('partial-output-secondary-layout', 'children'),
    [Input('non-existant-input', 'value')]
)
def non_existant_input_callback_secondary_layout(value):
    return 'this will fail'


@app.callback(
    Output('partial-output-initial-layout', 'children'),
    [Input('non-existant-input', 'value')]
)
def non_existant_input_callback_primary_layout(value):
    return 'this will fail'

if __name__ == '__main__':
    app.run_server(debug=True, port=8091)
```

alexcjohnson:
Fixes #1200 - regression in Dash 1.11 with callbacks that have outputs but no inputs.

### optionals

    -  [ ] at some point we should add an in-depth exploration of these cases to the docs for callbacks. Maybe not needed right away though, as we're primarily just fixing a regression.
    -  [ ] we'll follow up in the [v1.11 forum post](https://community.plotly.com/t/dash-v1-11-0-release-introducing-pattern-matching-callbacks/37592/26) by linking it to the v1.12 post.

### DIFF
--- a/dash-renderer/src/actions/index.js
+++ b/dash-renderer/src/actions/index.js
@@ -136,37 +136,33 @@ function moveHistory(changeType) {
 }
 
 function unwrapIfNotMulti(paths, idProps, spec, anyVals, depType) {
+    let msg = '';
+
     if (isMultiValued(spec)) {
-        return idProps;
+        return [idProps, msg];
     }
+
     if (idProps.length !== 1) {
         if (!idProps.length) {
-            if (typeof spec.id === 'string') {
-                throw new ReferenceError(
-                    'A nonexistent object was used in an `' +
-                        depType +
-                        '` of a Dash callback. The id of this object is `' +
-                        spec.id +
-                        '` and the property is `' +
-                        spec.property +
-                        '`. The string ids in the current layout are: [' +
-                        keys(paths.strs).join(', ') +
-                        ']'
-                );
-            }
-            throw new ReferenceError(
+            const isStr = typeof spec.id === 'string';
+            msg =
                 'A nonexistent object was used in an `' +
-                    depType +
-                    '` of a Dash callback. The id of this object is ' +
-                    JSON.stringify(spec.id) +
-                    (anyVals ? ' with MATCH values ' + anyVals : '') +
-                    ' and the property is `' +
-                    spec.property +
-                    '`. The wildcard ids currently available are logged above.'
-            );
-        }
-        throw new ReferenceError(
-            'Multiple objects were found for an `' +
+                depType +
+                '` of a Dash callback. The id of this object is ' +
+                (isStr
+                    ? '`' + spec.id + '`'
+                    : JSON.stringify(spec.id) +
+                      (anyVals ? ' with MATCH values ' + anyVals : '')) +
+                ' and the property is `' +
+                spec.property +
+                (isStr
+                    ? '`. The string ids in the current layout are: [' +
+                      keys(paths.strs).join(', ') +
+                      ']'
+                    : '`. The wildcard ids currently available are logged above.');
+        } else {
+            msg =
+                'Multiple objects were found for an `' +
                 depType +
                 '` of a callback that only takes one value. The id spec is ' +
                 JSON.stringify(spec.id) +
@@ -174,10 +170,10 @@ function unwrapIfNotMulti(paths, idProps, spec, anyVals, depType) {
                 ' and the property is `' +
                 spec.property +
                 '`. The objects we found are: ' +
-                JSON.stringify(map(pick(['id', 'property']), idProps))
-        );
+                JSON.stringify(map(pick(['id', 'property']), idProps));
+        }
     }
-    return idProps[0];
+    return [idProps[0], msg];
 }
 
 function startCallbacks(callbacks) {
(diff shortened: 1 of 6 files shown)


tests/integration/callbacks/test_missing_inputs.py
```diff
@@ -0,0 +1,484 @@
...
+    dash_duo.find_element("#btn").click()
+    dash_duo.wait_for_text_to_equal("#out1", "item 0, item 1")
+    dash_duo.wait_for_text_to_equal("#out2", "item 0, item 1 - 2")
+
+    assert not dash_duo.get_logs()
+
+
+def test_cbmi007_all_multi_wildcards_some_outputs(dash_duo):
+    # same as above (cbmi006) but multi-output, some outputs present some missing.
+    # Again we DO NOT fire the callback
```
alexcjohnson, line 300: This is a case I didn't think of until writing related tests, but I think it makes sense, because if there's any missing output and no inputs, that output will probably show up when the inputs do.

This is basically the same as test `cbmi004` just for the case when all the inputs have multi-value wildcards so we do _try_ to fire the callback, and we need to bail out at a later step when we collect all the outputs.

alexcjohnson:
The first 4 of these tests (modified for the changed internals) pass when run against Dash v1.10 - so this clears the regression as far as I understand it. The other 5 tests all use wildcards so are not relevant to v1.10. cc @chriddyp @Marc-Andre-Rivet

dash-renderer/src/actions/index.js
```diff
@@ -411,8 +442,45 @@ function fillVals(paths, layout, cb, specs, depType) {
...
+            const inputs = fillInputs(inputList, i);
+            if (isMultiValued(specs[i]) && !inputs.length) {
+                emptyMultiValues++;
+            }
+            return inputs;
+        } catch (e) {
+            if (e instanceof ReferenceError) {
+                errors.push(e);
+                return null;
+            }
```
Marc-Andre-Rivet, line 458: Wondering what would be required to be able to write this (and https://github.com/plotly/dash/pull/1212/files#diff-15692ba768eeab56e95aa409ab7c19b7R279) so that we don't use exceptions as flow control.

dash-renderer/src/actions/index.js
```diff
@@ -411,8 +442,45 @@ function fillVals(paths, layout, cb, specs, depType) {
...
+            // (if all inputs are multivalued and all missing we still return
+            // them as normal, and fire the callback.)
+            return null;
+        }
+        // If we get here we have some missing and some present inputs.
+        // That's a real error, so rethrow the first missing error.
+        // Wildcard reference errors mention a list of wildcard specs logged
+        // TODO: unwrapped list of wildcard ids?
+        // eslint-disable-next-line no-console
+        console.log(paths.objs);
```
Marc-Andre-Rivet, line 479: Is this meant to be `console.error`?

dash-renderer/src/actions/index.js
alexcjohnson: > don't use exceptions as flow control

yeah I suppose that's cleaner now that they aren't always errors -> https://github.com/plotly/dash/pull/1212/commits/b44bd54a8cd6c2d3e0e89e73c2cbc190b499ec5a

PR #1218, opened 2020-04-29
[WIP] Improve performance: What's the case for a revert on a boundary exception?

Marc-Andre-Rivet:
Running against dash-docs `/all`, seeing errors in the DashCanvas component being caught by the renderer's TreeContainer error boundary causing a history `revert` to be applied. In this specific case, the revert is done against a `setProps` call the dcc.Location component did to align its props with window.location. This brings about an interesting loop..

On render, dcc.Location identified correctly its `pathname` was misaligned with window.location (undefined vs `/all`), triggering a setProps for `pathname: '/all'`. This triggers a callback for the page content.

On render, DashCanvas fails, triggering a revert, causing the last know action to be cancelled. Namely, pathname is now `undefined`.

dcc.Location renders and identifies correctly, again, that its `pathname` is misaligned with window.location, triggering a setProps for `pathname: '/all'`. This triggers a callback for page content.

This goes on for a while...

This scenario causes ~95 additional callbacks to be triggered and many additional component renders.

---

I am uncertain as to the usefulness of reverting on an error and proposing here to remove the `revert` code entirely. The DevTools UI uses `undo` and `redo` so that feature would be unaffected.

---

It seems there's little more to be gained in terms of pure performance with the `/all` case that couldn't be investigated with a simpler use case or looking at optimizing `shouldComponentUpdate` for the larger components. In truth, a lot of the performance issues with this page have to do with components that are very expensive to render  (dcc.Graph, DashCanvas, various DashBio components). For example, there is a ~2 minutes "wall of silence" during rendering of the page, after the initial callback for the layout, where no additional ApiController render occurs, no callbacks are fired, and only a few components are re-rendered, as the expensive components described above get rendered - and deal with various DOM-related errors that somehow seem to be related to the weight of the page -- taking the example pages individually, these errors disappear.

### DIFF
--- a/dash-renderer/src/actions/index.js
+++ b/dash-renderer/src/actions/index.js
@@ -111,7 +111,6 @@ function triggerDefaultState(dispatch, getState) {
 
 export const redo = moveHistory('REDO');
 export const undo = moveHistory('UNDO');
-export const revert = moveHistory('REVERT');
 
 function moveHistory(changeType) {
     return function(dispatch, getState) {
--- a/dash-renderer/src/components/error/ComponentErrorBoundary.react.js
+++ b/dash-renderer/src/components/error/ComponentErrorBoundary.react.js
@@ -2,7 +2,7 @@ import {connect} from 'react-redux';
 import {Component} from 'react';
 import PropTypes from 'prop-types';
 import Radium from 'radium';
-import {onError, revert} from '../../actions';
+import {onError} from '../../actions';
 
 class UnconnectedComponentErrorBoundary extends Component {
     constructor(props) {
@@ -28,7 +28,6 @@ class UnconnectedComponentErrorBoundary extends Component {
                 info,
             })
         );
-        dispatch(revert);
     }
 
     componentDidUpdate(prevProps, prevState) {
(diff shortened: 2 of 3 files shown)


alexcjohnson:
I guess the original intent of reverting here was to put the page back to the last known self-consistent state, so the user would be able to continue using the app as if they had never done the action that generated the error. That still seems like a useful feature, with important implications for how robust real apps will feel in production: if we successfully revert, the user can keep going with the app, just effectively won't have the broken option available to them. If we DON'T revert, the app may be stuck in a dead state and require reloading entirely to function again.

But I guess I see two issues with this: First is what you're seeing here, where we never reached a self-consistent state in the first place. Second, it's not clear to me that `revert` is really targeting self-consistency. In particular when there's a callback chain, won't this just be reverting one step in the chain? If so, we're reverting to a state you could never get to without the error, and who knows what that means.

So perhaps what's really needed is to set a checkpoint when we reach a stable state - ie nothing in `pendingCallbacks` and no `setProps` in progress (if that's something we can determine, given late updates like you're seeing from `Location`) - and have `revert` target that state if such a state exists, otherwise do nothing?

Marc-Andre-Rivet:
> So perhaps what's really needed is to set a checkpoint when we reach a stable state

That makes sense. Additionally, components that wrap up third-party components that can misbehave should implement their own error boundary to prevent "normal" errors from triggering the render's `revert`. In effect, adding a rendering try..catch, swallowing known errors that are not caused by bad props, just bugs in implementation.

PR #1219, opened 2020-04-29
Include transpiled source files in published package

Marc-Andre-Rivet:


dash-renderer/package.json
```diff
@@ -22,6 +23,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/polyfill": "7.8.7",
+    "@plotly/dash-component-plugins": "^1.2.0",
```
Marc-Andre-Rivet, line 26: Moved from devDeps because it is required for building the package and this way on `npm install dash-renderer` this dependency will be fetched.

dash-renderer/package.json
```diff
@@ -37,11 +39,11 @@
     "viz.js": "2.1.2"
   },
   "devDependencies": {
+    "@babel/cli": "^7.8.4",
```
Marc-Andre-Rivet, line 42: To run the `babel` command in the scripts.

dash-renderer/package.json
```diff
@@ -66,5 +68,9 @@
     "webpack-dev-server": "^3.10.3",
     "webpack-serve": "^3.2.0",
     "whatwg-fetch": "^3.0.0"
-  }
+  },
+  "files": [
+    "/dash_renderer/*{.js,.map}",
+    "/lib/**"
+  ]
```
Marc-Andre-Rivet, line 75: Cleaner package content: include the build artifacts and the transpiled code.

PR #1220, opened 2020-04-30
Missing pattern-matching outputs

### LINKED ISSUE #1216: [BUG] pattern-matching callbacks with empty or partly-empty Output
As reported in the [dash 1.11 forum post](https://community.plotly.com/t/dash-v1-11-0-release-introducing-pattern-matching-callbacks/37592/27) we currently have problems when a pattern-matching callback matches no items in its `Output`(s). This particular case is straightforward, the callback simply shouldn't fire and no error should occur. But there are multiple variants that deserve scrutiny:

1. `Output({key1: ALL})` (the case posted on the forum - the `ALL` can match no items)
2. `Output({key1: ALL, key2: MATCH})` (same thing but the pattern has a `MATCH` too - again the callback should not fire with no matching items)
3. `[Output({key1: ALL}), Output({key2: ALL})]` (two `ALL` patterns - if either one matches items the callback should fire, but if neither matches anything it should not)
4. `[Output({key1: ALL}), Output(str)]` (an `ALL` and a simple string ID - the callback should always fire)
5. `[Output({key1: ALL, key2: MATCH}), Output({key2: MATCH})]` (an `ALL+MATCH` and a `MATCH` - the callback should always fire)

### Should removing items from an `Output(ALL)` trigger the callback?

This I'm not so sure about. Removing items from an `Input(ALL)` certainly should and does trigger the callback, and adding items to `Output(ALL)` triggers the callback as an "initial call" for the new item.

In practice this currently mainly happens when you remove the entire category - but that will change when we implement array operation callbacks #968.

The argument I can think of for not triggering is that the inputs to the callback haven't changed, and if the callback is expensive you may not want to rerun it just because the user deleted something from the page. Also if it's clear to users that this is expected, they can force the callback to run on item removal by adding `Input({key1: ALL}, "id")` matching the `Output({key1: ALL}, prop)`

The argument I see in favor of triggering is that these outputs often depend on each other, and the list of outputs you're required to return can itself be thought of as an input. Indeed, this is how I discovered the issue, as I made a test app one way to cover case (1) and then tried to modify it for case (4) and was surprised that it mostly worked, until I cleared the `ALL` match.

Here's the original version of the test app (only works on the [missing-outputs branch](https://github.com/plotly/dash/compare/missing-outputs?expand=1)):
```py
import dash
from dash.dependencies import Input, Output, ALL


app = dash.Dash(__name__)

    html.Button("items", id="btn-1"),
    html.Button("values", id="btn-2"),
    html.Div(id="content"),
    html.Div("Output init", id="output"),
])


@app.callback(Output("content", "children"), [Input("btn-1", "n_clicks")])
def content(n1):
    return [html.Div(id={"i": i}) for i in range((n1 or 0) % 4)]


@app.callback(Output({"i": ALL}, "children"), [Input("btn-2", "n_clicks")])
def content_inner(n2):
    n1 = len(dash.callback_context.outputs_list)
    if not n1:
        raise ValueError("should not be called with no outputs!")
    return [n2 or 0] * n1


@app.callback(Output("output", "children"), [Input({"i": ALL}, "children")])
def out2(contents):
    return sum(contents)


if __name__ == "__main__":
    app.run_server(debug=True)
```

And the modification, combining the last two callbacks into one:
```py
@app.callback(
    [Output({"i": ALL}, "children"), Output("output", "children")],
    [Input("values", "n_clicks")]
)
def content_and_output(n2):
    n1 = len(dash.callback_context.outputs_list[0])
    content = [n2 or 0] * n1
    return content, sum(content)
```
Which works great until 4 clicks of the `items` button, when `content` gets cleared.

You can fix it either by adding `items` as an input and recalculating `n1` from that, or by adding `Input({"i": ALL}, "id")`. The question is whether that should be necessary or if I should treat this as a bug.

alexcjohnson:
Fixes #1216 

Pattern-matching callbacks had some issues with `ALL` wildcards in their `Output`, reported by user `@jaser` on the [forum](https://community.plotly.com/t/dash-v1-11-0-release-introducing-pattern-matching-callbacks/37592/27). Everything worked well as long as the `ALL` matched something, but if it didn't we would throw an error.

There are a number of additional cases to consider, involving various combinations of `ALL` and `MATCH` wildcards and multiple outputs, as enumerated in #1216 - I will comment on the tests which of these cases each covers. And as discussed in #1216, we will NOT trigger a callback simply because one or more output items was removed. If you want that to happen, include the output component set as inputs as well - our suggestion is to just change the prop to `'id'` to avoid circular dependencies, and as a side benefit you get a clean list of the IDs involved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [#1078](https://github.com/plotly/dash/pull/1078) Permit usage of arbitrary file extensions for assets within component libraries
 
 ### Fixed
+- [#1220](https://github.com/plotly/dash/pull/1220) Fixes [#1216](https://github.com/plotly/dash/issues/1216), a set of related issues about pattern-matching callbacks with `ALL` wildcards in their `Output` which would fail if no components matched the pattern.
 - [#1212](https://github.com/plotly/dash/pull/1212) Fixes [#1200](https://github.com/plotly/dash/issues/1200) - prior to Dash 1.11, if none of the inputs to a callback were on the page, it was not an error. This was, and is now again, treated as though the callback raised PreventUpdate. The one exception to this is with pattern-matching callbacks, when every Input uses a multi-value wildcard (ALL or ALLSMALLER), and every Output is on the page. In that case the callback fires as usual.
 - [#1201](https://github.com/plotly/dash/pull/1201) Fixes [#1193](https://github.com/plotly/dash/issues/1193) - prior to Dash 1.11, you could use `flask.has_request_context() == False` inside an `app.layout` function to provide a special layout containing all IDs for validation purposes in a multi-page app. Dash 1.11 broke this when we moved most of this validation into the renderer. This change makes it work again.
 
--- a/dash-renderer/src/actions/dependencies.js
+++ b/dash-renderer/src/actions/dependencies.js
@@ -397,12 +397,12 @@ function findInOutOverlap(outputs, inputs, head, dispatchError) {
 }
 
 function findMismatchedWildcards(outputs, inputs, state, head, dispatchError) {
-    const {anyKeys: out0AnyKeys} = findWildcardKeys(outputs[0].id);
-    outputs.forEach((out, outi) => {
-        if (outi && !equals(findWildcardKeys(out.id).anyKeys, out0AnyKeys)) {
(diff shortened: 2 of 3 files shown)


alexcjohnson:
Thanks @Marc-Andre-Rivet - this part of the code probably still is more convoluted than it would need to be, though I was pleased to see that this particular change actually slightly reduced LOC in `src`. I'm glad you're poking around in it, as we continue updating things I'd appreciate any suggestions you have to further clean it up!

Issue #719, opened 2019-05-14
Allow extracting meta data from typescript components


andrenanninga:
We would like to use `dash-generate-components` on some Typescript components to allow our Python developers to use them in their Dash code. We have a custom pipeline to transpile the Typescript components to a bundle but we're currently blocked by this line of code:

https://github.com/plotly/dash/blob/ec1576c3652cbb90ffb064abe2753a17e13f3c51/dash/extract-meta.js#L63

Extracting metadata from components is currently only allowed for `.js` and `.jsx` files. However react-docgen has [no issues extracting metadata from Flow or Typescript](https://github.com/reactjs/react-docgen#flow-and-typescript-support). 

Would you support the addition of `.ts` and `.tsx` files to this filter. This would allow us to continue using Typescript.

## Example typescript code 

- `Dev.jsx`

(Currently I am saving this as `Dev.jsx` to allow the extracting of metadata.)

```typescript
import React from 'react';

type Props = {
    /**
     * my description of what a label is
     */
    label?: string,
};

export default class Test extends React.Component<Props> {
    render() {
        const {label} = this.props;
        return <h1>{label}</h1>;
    }
}
```

- `Test.py`

```python
# AUTO GENERATED FILE - DO NOT EDIT

from dash.development.base_component import Component, _explicitize_args


class Test(Component):
    """A Test component.


Keyword arguments:
- label (string; optional): my description of what a label is"""
    @_explicitize_args
    def __init__(self, label=Component.UNDEFINED, **kwargs):
        self._prop_names = ['label']
        self._type = 'Test'
        self._namespace = 'something'
        self._valid_wildcard_attributes =            []
        self.available_properties = ['label']
        self.available_wildcard_properties =            []

        _explicit_args = kwargs.pop('_explicit_args')
        _locals = locals()
        _locals.update(kwargs)  # For wildcard attrs
        args = {k: _locals[k] for k in _explicit_args if k != 'children'}

        for k in []:
            if k not in args:
                raise TypeError(
                    'Required argument `' + k + '` was not specified.')
        super(Test, self).__init__(**args)
```

alexcjohnson:
Good idea! @Marc-Andre-Rivet do you see any blockers to allowing this? We already support flow, I can imagine us running into problems with complex types but presumably as long as people stay within reason we should be able to handle it.

Marc-Andre-Rivet:
[...]

PR #1227, opened 2020-05-04
Support referencing nested components

mjpsyapse:
Our team is creating dash components for [bizcharts](https://bizcharts.net/products/bizCharts).  Some of the components are nested inside of other components (Ex: [Guide.Arc](https://github.com/alibaba/BizCharts/blob/master/src/addepter/Guide.tsx#L19-L26)).  In order to make it easier to map the python component classes to the react components, I'm adding the ability to supply dot notation strings to access nested components.

Ex: 
```python
Component._type = "TestComponent.NestedComponent"
```

## Contributor Checklist

   -  [x] Update registry js file to use dot notation to reference components

### optionals

### DIFF
--- a/dash-renderer/src/registry.js
+++ b/dash-renderer/src/registry.js
@@ -1,3 +1,5 @@
+import { path } from 'ramda';
+
 export default {
     resolve: component => {
         const {type, namespace} = component;
@@ -5,8 +7,9 @@ export default {
         const ns = window[namespace];
 
         if (ns) {
-            if (ns[type]) {
-                return ns[type];
+            const c = path(type.split('.'), ns);
+            if (c) {
+                return c;
             }
 
             throw new Error(`Component ${type} not found in ${namespace}`);
--- a/dash-renderer/src/registry.js
+++ b/dash-renderer/src/registry.js
@@ -1,4 +1,4 @@
-import { path } from 'ramda';
+import {path} from 'ramda';
 
 export default {
     resolve: component => {
--- a/dash-renderer/tests/registry.test.js
+++ b/dash-renderer/tests/registry.test.js
@@ -1,0 +1,44 @@
+import registry from '../src/registry';
+
+describe('registry', () => {
+
+    afterEach(() => {
+        delete window.TestNamespace;
+    });
+
+    describe('resolving nested components', () => {
+        beforeEach(() => {
+            window.TestNamespace = {
+                TestComponent: {
+                    NestedComponent: 'test',
+                },
+            };
+        });
+
+        test('referencing nested components', () => {
+            const component = {
+                type: 'TestComponent.NestedComponent',
+                namespace: 'TestNamespace',
+            };
+            const actual = registry.resolve(component);
+            expect(actual).toEqual('test');
+        });
+    });
+
+    describe('resolving standard components', () => {
+        beforeEach(() => {
+            window.TestNamespace = {
+                TestComponent: 'test',
+            };
+        });
+
+        test('referencing nested components', () => {
+            const component = {
+                type: 'TestComponent',
+                namespace: 'TestNamespace',
+            };
+            const actual = registry.resolve(component);
+            expect(actual).toEqual('test');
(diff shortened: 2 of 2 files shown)


alexcjohnson:
Interesting, can you say more about the motivations here? Is `Guide.Arc` an `Arc` that's always a child of a `Guide` component? Or is `Arc` a _type_ of guide?

There must also be changes needed on the build side, in order for the Python component to be aware that it's nested without having to modify the generated component class?

mjpsyapse:
@alexcjohnson 
Thanks for reviewing!

**Is `Guide.Arc` an `Arc` that's always a child of a `Guide` component? Or is `Arc` a _type_ of guide?**
Just wanna be clear on this one, but yes `Guide.Arc` is used as a child element of the `Guide` element. Ex:
```jsx
<Guide>
  <Guide.Arc {...props} />
  <Guide.Line {...otherProps} />
</Guide>
```
I guess conceptually it can be considered a type of guide, too, but not in the type theory sense of programming.  But anyhow!  Hopefully the example above clears things up.

**RE: There must also be changes needed on the build side, in order for the Python component to be aware that it's nested without having to modify the generated component class?**
We are creating dash components manually for each React component, so it turns out this isn't an issue. The ideal case would be to [add support for typescript in the dash component generator](https://github.com/plotly/dash/issues/719), but that seemed like more work than it was worth for our team.

**More on the why...**
We would like to continue using the [cdn hosted bizcharts file](https://unpkg.com/bizcharts@3.2.4/umd/BizCharts.min.js), so we have to be able to reference the components nested within each other as shown above. Otherwise, we would have to publish our own distribution of bizcharts after "flattening" all the component classes.  I.e. publish bizcharts with `Guide.Arc` transformed to `GuideArc` etc...

mjpsyapse:
@alexcjohnson Sorry to bother you, but do you have any other feedback on this PR?

alexcjohnson:
@mjpsyapse thanks for the nudge. I know this is a very simple change but there are a few things we're concerned about. Not saying the answer is necessarily no, but there are a few questions we need to consider.

First, the access `window[namespace][component]` is itself not intended to be part of the public API, but part of the internal contract between `dash` components and `dash-renderer`, and there are a few reasons we may want to change this in the future. Maybe we want to support multiple versions of the same component suite on the same page; or maybe we will want to move component package loading entirely into the renderer. I'm not sure, and of course we'd need to support the existing behavior for backward compatibility with non-core components, but this PR would lock us a little tighter to that model.

Second, building the Python components manually means other languages (R, Julia soon) can't benefit from your package. Even if it's not important for your team, we want to encourage patterns that expand the ecosystem as much as possible.

So I wonder if there isn't a solution for your use case that's a similar level of effort to creating the Python wrappers manually, but done in JS?
- You don't need a totally new bizcharts distribution, just a wrapper that imports components from the CDN bundle as an external dep and re-exports the components you care about.
- Just put the CDN bundle as the first entry in `_js_dist` and your wrapper second, that should ensure it's available when the wrapper executes.
- Re-exporting would make a flat structure, yes, but this also lets you do as much or as little `PropTypes` validation as you want (doing this all on the Python side you won't get any validation unless you write it yourself, AFAIK)
- And then if you ever decide to take these components further - adding interactivity back to the rest of Dash for example, that's just a matter of extending the wrapper components; whereas if you're referencing the CDN components directly it would be a complete rewrite to make an extension.
- Also if TypeScript -> PropTypes support matures to the point that we can support it in the component generators, this would also be an incremental change rather than a completely new architecture.

Does that make sense? Do you think a strategy like this would work for your purposes?

mjpsyapse:
@alexcjohnson   Thanks for your time, and I appreciate the feedback.  I'll take those options into consideration.

alexcjohnson:
Thanks for understanding @mjpsyapse - we absolutely want to help you make new components, so if you hit any snags, or if the strategies I've suggested don't work for whatever reason, please reach out either here or on community.plotly.com.

PR #1228, opened 2020-05-04
Prevent initial call

### LINKED ISSUE #1225: Ability to prevent initial callback from firing
@chriddyp started prototyping this in #1123 and I'm preparing a new PR that will supersede #1123 with a more complete implementation and tests. But a question about how this should work regarding later callbacks:

Currently if all of the inputs to callback C are themselves outputs to other callbacks A and B, we effectively do not treat C as an "initial callback" - that is, it won't trigger without one of its inputs changing, so if A and B both raise `PreventUpdate` on page load (or layout chunk load), C will not be called.

The question is how this should apply to `prevent_initial_call` - if A and B both have `prevent_initial_call=True`, is that equivalent to `PreventUpdate` so C should not fire? Or is it as though A and B are simply not considered initially and the initialization chain should start at C?

My initial reaction is that this case is not the same as `PreventUpdate` and C **should** fire. My reasoning:
- Presumably in this situation you've set the initial values for A and B as part of the layout - so it's not so much that you're preventing an update, it's just that for performance reasons you've already made the update happen.
- If you do want C to fire and we automatically prevent it, you'd have to do something super hacky like add an extra input. On the other hand, if we allow it to fire and you don't want that, all you have to do is set `prevent_initial_call=True` on C as well.

Anyone care to argue the other side? @Marc-Andre-Rivet @chriddyp @christianwengert do you agree?

alexcjohnson:
Closes #1225 
Supersedes #1123 

Adds the ability to control whether or not callbacks fire when the page or layout chunk first loads. This can simplify a lot of code like `if n_clicks is None: raise PreventUpdate`, and it can improve loading performance, particularly for pattern-matching callbacks, where a page can easily generate hundreds of callback invocations, since there is no longer even a server call for these callbacks.

This option is off by default, but can either be turned on for each callback independently:
```py
@app.callback(outputs, inputs, prevent_initial_call=True)
def cb(...):
    ...

app.clientside_callback(func, outputs, inputs, prevent_initial_call=True)
```

or app-wide, in which case it can be turned _off_ for individual callbacks:
```py
app = dash.Dash(prevent_initial_callbacks=True)

@app.callback(outputs, inputs, prevent_initial_call=False)
def cb(...)
    ...
```

- [x] Docs follow-up issue: https://github.com/plotly/dash-docs/issues/874

### DIFF
--- a/dash-renderer/src/actions/dependencies.js
+++ b/dash-renderer/src/actions/dependencies.js
@@ -1238,6 +1238,9 @@ export function getCallbacksInLayout(graphs, paths, layoutChunk, opts) {
                 ) {
                     // This callback should trigger even with no changedProps,
                     // since the props that changed no longer exist.
+                    // We're kind of abusing the `initialCall` flag here, it's
+                    // more like a "final call" for the removed inputs, but
+                    // this case is not subject to `prevent_initial_call`.
                     if (flatten(cb.getOutputs(newPaths)).length) {
                         cb.initialCall = true;
                         cb.changedPropIds = {};
@@ -1255,10 +1258,13 @@ export function getCallbacksInLayout(graphs, paths, layoutChunk, opts) {
                 const cb = getCallbackByOutput(graphs, paths, id, property);
                 if (cb) {
                     // callbacks found in the layout by output should always run
+                    // unless specifically requested not to.
                     // ie this is the initial call of this callback even if it's
                     // not the page initialization but just a new layout chunk
-                    cb.initialCall = true;
-                    addCallback(cb);
+                    if (!cb.callback.prevent_initial_call) {
+                        cb.initialCall = true;
+                        addCallback(cb);
+                    }
                 }
             }
         }
@@ -1270,14 +1276,13 @@ export function getCallbacksInLayout(graphs, paths, layoutChunk, opts) {
             if (chunkPath) {
                 handleThisCallback = cb => {
                     if (
-                        all(
+                        !all(
                             startsWith(chunkPath),
                             pluck('path', flatten(cb.getOutputs(paths)))
                         )
                     ) {
-                        cb.changedPropIds = {};
+                        maybeAddCallback(cb);
                     }
-                    maybeAddCallback(cb);
                 };
             }
             for (const property in inIdCallbacks) {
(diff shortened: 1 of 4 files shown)


dash-renderer/src/actions/dependencies.js
```diff
@@ -1270,14 +1276,13 @@ export function getCallbacksInLayout(graphs, paths, layoutChunk, opts) {
             if (chunkPath) {
                 handleThisCallback = cb => {
                     if (
-                        all(
+                        !all(
                             startsWith(chunkPath),
                             pluck('path', flatten(cb.getOutputs(paths)))
                         )
                     ) {
-                        cb.changedPropIds = {};
```
alexcjohnson, line 1278: I'm not sure what I was thinking here with clearing out `changedPropIds`, but in this case (we've found a callback by its inputs due to a layout change, and just determined that all its outputs are in the same layout chunk) we know the callback is already included in the queue via its outputs so the best thing to do is simply not re-include it at all. This extra `addCallback` was harmless before `prevent_initial_call`, just a slight bit of extra work; but now that we have `prevent_initial_call` it's possible this callback would *not* be added based on its outputs.

If this condition isn't met, there are outputs outside the new layout chunk and it's not in fact an initial call anyway - so it's appropriate to add the callback and that's what the new logic does.

tests/integration/callbacks/test_prevent_initial.py
```diff
@@ -0,0 +1,335 @@
...
+from dash.exceptions import PreventUpdate
+
+# parametrize all the tests, but no need to do all 8 combinations, 5 is enough!
+flavors = [
+    {"clientside": False, "content": False, "global": False},
+    {"clientside": True, "content": False, "global": False},
+    {"clientside": False, "content": True, "global": False},
+    {"clientside": True, "content": True, "global": True},
+    {"clientside": False, "content": False, "global": True},
+]
```
alexcjohnson, line 16: `pytest` does a nice job logging dictionary test parametrizations (`flavor0`, `flavor1` etc):
Screen Shot 2020-05-04 at 5 15 44 PM

maartenbreddels:
Initial impression is really good! Together with #832 we can avoid any unneeded calls, and have a very fast page load.

PR #1238, opened 2020-05-09
Add R build test to CircleCI

rpkyle:
This PR proposes a set of tests to assess the health of the entire Dash component library build pipeline for R, as we currently do with Dash for Python.

There have been a few instances of package builds in the past using `dash-generate-components` which appeared to succeed, but failed to install or work properly due to minor misconfiguration issues. A build test would help identify these inconsistencies prior to a Dash release.

Late last year, @Marc-Andre-Rivet described a desirable workflow for testing R builds upon pushing to `dev`:

- [x] trigger a rebuild of Dash for R within CircleCI, and all Dash Core packages using the versions at the head of their respective `dev` branches
- [x] install the rebuilt packages within CircleCI
- [x] launch unit tests to assess basic package functionality in R 
- [x] then, if unit tests are successful, runs a basic integration test to demonstrate that the `dashr-on-premise-sample-app` is functional and interactive (such a test is proposed in https://github.com/plotly/dashR/pull/199)

I suggest adding the following integrity checks to the above list:
- [x] remove non-R artifacts from each of the cloned repositories (these can cause errors/warnings and lengthen the time required to tarball/check packages)
- [x] run `R CMD build` to ensure that packages are properly formatted, can be built into tarballs, and are ready for `R CMD check`
- [x] run `R CMD check --as-cran --no-manual` to ensure that changes have not violated CRAN requirements  

The current draft appears to work generally as intended.

@Marc-Andre-Rivet @alexcjohnson @josegonzalez


rpkyle:
@alexcjohnson @Marc-Andre-Rivet Not sure what's going on with the test failures, but I think this is ready for review otherwise.

.circleci/config.yml
```diff
@@ -213,6 +213,112 @@ jobs:
...
+          path: test-reports
+      - store_test_results:
+          path: test-reports
+      - store_artifacts:
+          path: /tmp/dash_artifacts
+
+      - run:
+          name: 🦔 percy finalize
+          command: npx percy finalize --all
+          when: always
```
alexcjohnson, line 319: might need this to be `when: on_fail`, and have the `percy_finalize` job require this one in addition to `test-37`? That's still not quite the right logic, as we'll finalize the Percy job twice if either `build-dashr` or `test-37` fails - cutting off any further snapshots in the one that didn't fail. But it's probably the best we can do given that CircleCI doesn't have "require these jobs to be done and execute even if they failed"

rpkyle: Ah, interesting. I hadn't considered that these tests could *both* modify when the Percy job is finalized. It does seem like `on_fail` is at least preferable to `always`, though.

rpkyle: fixed in 68417096

.circleci/config.yml
```diff
@@ -213,6 +213,112 @@ jobs:
...
+          name: ️️🏭 clone and npm build core for R
+          command: |
+            python -m venv venv
+            . venv/bin/activate
+            git clone --depth 1 https://github.com/plotly/dash.git -b ${CIRCLE_BRANCH} dash-main
+            cd dash-main && pip install -e .[dev,testing] --progress-bar off && cd ..
+            git clone --depth 1 https://github.com/plotly/dashR.git -b dev dashR
+            git clone --depth 1 https://github.com/plotly/dash-html-components.git -b dev
+            git clone --depth 1 https://github.com/plotly/dash-core-components.git -b dev
+            git clone --depth 1 https://github.com/plotly/dash-table.git -b dev
```
Marc-Andre-Rivet, line 237: No need for `-b dev` as the default branch is `dev` for html/dcc/table

.circleci/config.yml
```diff
@@ -213,6 +213,112 @@ jobs:
...
+      - checkout
+
+      - run:
+          name: ️️🏭 clone and npm build core for R
+          command: |
+            python -m venv venv
+            . venv/bin/activate
+            git clone --depth 1 https://github.com/plotly/dash.git -b ${CIRCLE_BRANCH} dash-main
+            cd dash-main && pip install -e .[dev,testing] --progress-bar off && cd ..
+            git clone --depth 1 https://github.com/plotly/dashR.git -b dev dashR
```
Marc-Andre-Rivet, line 234: Don't you want to be running the tests against whatever the branch of DashR is in PR `X`?

.circleci/config.yml
rpkyle: fixed in 74083cea

.circleci/config.yml
```diff
@@ -229,7 +229,7 @@ jobs:
           command: |
             python -m venv venv
             . venv/bin/activate
-            git clone --depth 1 https://github.com/plotly/dash.git -b dev dash-main
+            git clone --depth 1 https://github.com/plotly/dash.git -b ${CIRCLE_BRANCH} dash-main
```
alexcjohnson, line 232: oh hmm - I hadn't really quite paid attention to what's going on here before. This will work (clearly, since it works 😏) so if you're itching to merge and move on that's fine. But if you do leave it as is, please make an issue to come back to this:
- at the very least we already have this code via `checkout` so could use that instead of cloning again?
- but I wonder if we can't piggyback on the `build-core-37` job that already does most of the cloning and building you're doing here, the same way that `test-37` does.
- `build-core-37` also has a line `sed -i '/dash-/d' requires-install.txt` that prevents bringing in a second copy of dcc, html, and table - if we *don't* use the results of `build-core-37` we may want to do the same here.

.pylintrc
```diff
@@ -270,7 +270,7 @@ ignore-docstrings=yes
 ignore-imports=no
 
 # Minimum lines number of a similarity.
-min-similarity-lines=10
+min-similarity-lines=20
```
alexcjohnson, line 273: There's also `.pylintrc37` - and I was wondering why this didn't seem to matter there. I've also forgotten why we have two different ones, and we should perhaps figure that out and whether it's still important because... we don't seem to be using `.pylintrc37`. It's currently referenced in the wrong place, in `build-core-37`:

https://github.com/plotly/dash/blob/74083cea1322aa7682f7e427223fd25b33bcb5ea/.circleci/config.yml#L83-L85

rather than in `lint-unit-37` where it belongs:

https://github.com/plotly/dash/blob/74083cea1322aa7682f7e427223fd25b33bcb5ea/.circleci/config.yml#L31-L32

Not a problem that started here, but we should at least propagate this change to `.pylint37`, and then either use that where it was intended to be used or make an issue to fix this later.

.circleci/config.yml
rpkyle: @alexcjohnson I wasn't able to reuse the `dash` directory from an earlier run, though I can't recall why -- I did try that.

FWIW these lines are essentially pulled straight from https://github.com/plotly/dashR/blob/9c5a30760e3be985c866f122e9c2bd6ce2a516a6/.circleci/config.yml#L41-L44, though it's a bit different there since we aren't running *all* the Dash tests in that configuration, and we do clone the repo earlier in this workflow.

.pylintrc
rpkyle: I'll make the change in `.pylint37`.

rpkyle: fixed in 0c242625 and f8521442

.circleci/config.yml
alexcjohnson: Sure, that's fine - just make an issue to come back to this.

rpkyle: Since these are separate jobs, the paths are not shared by default, as far as I am aware. We could `persist_to_workspace` the `dash` package directory from a preceding run, so I'll open an issue to suggest this. We already do this for package tarballs, though not the repos themselves.

https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-artifacts-among-jobs

I am not certain that we want to reuse the component library directories after deleting most of their content for the R test run, but maybe this is a non-issue if we don't subsequently `persist_to_workspace` those paths. I'm not sure, I'm pretty new to configuring CircleCI.

rpkyle: @alexcjohnson https://github.com/plotly/dash/issues/1264

alexcjohnson:
I don't understand why Percy doesn't think it's finalized... something about the parallel jobs confusing it?

alexcjohnson:
OK there's something very odd going on with our percy config - looking at a passing percy build from another branch we have 7 builds for the same commit, with only the last one containing the snapshots. Apparently it all looks OK since the last one passes but it's a mess, should be only one.
Screen Shot 2020-05-21 at 9 50 50 PM
Whereas on this branch there's one more empty build, and now it comes *after* the real build so now it all fails?
Screen Shot 2020-05-21 at 9 53 56 PM

alexcjohnson:
dcc has a similar test job structure and only has one percy build per commit. The only difference I see is the env var `PERCY_PARALLEL_TOTAL` set in both of the snapshot-generating jobs:
```
        environment:
            PYTHON_VERSION: py37
            PERCY_PARALLEL_TOTAL: -1
```

rpkyle:
> dcc has a similar test job structure and only has one percy build per commit. The only difference I see is the env var `PERCY_PARALLEL_TOTAL` set in both of the snapshot-generating jobs:
> 
> ```
>         environment:
>             PYTHON_VERSION: py37
>             PERCY_PARALLEL_TOTAL: -1
> ```

Apparently this tells Percy _we're not sure how many parallel threads this run will have, but we're done when the orb sends `finalize_all`.

I'm not sure what that means, since whenever I hear "orb" I think Little Fluffy Clouds. Should we modify this setting? Is Percy not reporting the finalized state consistently?

rpkyle:
> I don't understand why Percy doesn't think it's finalized... something about the parallel jobs confusing it?

Seems possible, could it be related to our `parallelism` setting also?

PR #1240, opened 2020-05-10
Issue-1213 clientside callback_context Initial Commit

mrdawson:
## Clientside callback_context
Here’s a first pass at Issue-1213, adding callback_context to clientside callbacks. Before I get too much further into the weeds, I wanted to get a PR going to discuss.

The idea was to mimic the behavior of the serverside dash.callback_context as closely as possible. I think this is probably the way to go for consistency’s sake, but I am of course open to suggestions.
From within a clientside callback function, we can now use the trick with “triggered” to figure out which input triggered the callback:

`triggered = dash_clientside.callback_context.triggered.map(t => t[“prop_id”]);`

The first new test `test_clsd009_clientside_callback_context` is currently passing.

## Contributor Checklist

- [x] TODO
   -  [x] Add property `states`
   -  [x] Add property `states_list`
   -  [ ] ~~Add property `response` (need to check if this even makes sense in the clientside context)~~
   -  [x] Flush out tests for each of the context properties
- [x] All tests passing
----
#### Two quick off-topic things:
- I get the following warning when setting up `pip install -e .[testing,dev]`:
 
`requests 2.21.0 has requirement urllib3<1.25,>=1.21.1, but you'll have urllib3 1.25.9 which is incompatible.`

- I had to comment out all of /test/integration/test_generation.py to gets pytest to work.
I’m on Windows, Python 3.7

### DIFF
--- a/dash-renderer/src/actions/index.js
+++ b/dash-renderer/src/actions/index.js
@@ -531,6 +531,9 @@ function inputsToDict(inputs_list) {
     // returns an Object (map):
     //  keys of the form `id.property` or `{"id": 0}.property`
     //  values contain the property value
+    if (!inputs_list){
+        return {};
+    }
     const inputs = {};
     for (let i = 0; i < inputs_list.length; i++) {
         if (Array.isArray(inputs_list[i])) {
@@ -539,11 +542,11 @@ function inputsToDict(inputs_list) {
                 const id_str = `${JSON.stringify(inputsi[ii].id)}.${
                     inputsi[ii].property
                 }`;
-                inputs[id_str] = inputsi[ii].value;
+                inputs[id_str] = inputsi[ii].value ? inputsi[ii].value : null;
             }
         } else {
             const id_str = `${inputs_list[i].id}.${inputs_list[i].property}`;
-            inputs[id_str] = inputs_list[i].value;
+            inputs[id_str] = inputs_list[i].value ? inputs_list[i].value : null;
         }
     }
     return inputs;
@@ -578,9 +581,10 @@ function handleClientside(clientside_function, payload) {
                 prop_id => ({prop_id: prop_id, value: input_dict[prop_id]})
             );
         }
-        dc.callback_context.inputs_list = input_dict;
-        dc.callback_context.inputs = inputs;
-        // TODO: add the rest: states, states_list, response(??)
+        dc.callback_context.inputs_list = inputs;
+        dc.callback_context.inputs = input_dict;
+        dc.callback_context.states_list = state;
+        dc.callback_context.states = inputsToDict(state);
 
         const {namespace, function_name} = clientside_function;
         let args = inputs.map(getVals);
--- a/dash-renderer/src/actions/index.js
+++ b/dash-renderer/src/actions/index.js
@@ -525,6 +525,23 @@ const getVals = input =>
 
 const zipIfArray = (a, b) => (Array.isArray(a) ? zip(a, b) : [[a, b]]);
 
+function inputsToDict(inputs_list) {
+    // Ported directly from _utils.py, inputs_to_dict
+    // takes an array of inputs (some inputs may be an array)
+    // returns an Object (map):
+    //  keys of the form `id.property` or `{"id": 0}.property`
+    //  values contain the property value
+    let inputs = {};
+    for (let i = 0; i < inputs_list.length; i++) {
+        let inputsi = Array.isArray(inputs_list[i]) ? inputs_list[i] : [inputs_list[i]];
+        for (let ii = 0; ii < inputsi.length; ii++) {
+            let id_str = `${JSON.stringify(inputsi[ii].id)}.${inputsi[ii].property}`;
+            inputs[id_str] = inputsi[ii].value;
+        }
+    }
+    return inputs;
+}
+
 function handleClientside(clientside_function, payload) {
     const dc = (window.dash_clientside = window.dash_clientside || {});
     if (!dc.no_update) {
@@ -540,6 +557,23 @@ function handleClientside(clientside_function, payload) {
     }
 
     const {inputs, outputs, state} = payload;
+
+    if (!dc.callback_context) {
+        Object.defineProperty(dc, 'callback_context', {
+            value: {'triggered': [{'prop_id': '.', 'value': null}],
+                    'inputs_list': inputsToDict(payload.inputs),
+                    'inputs': payload.inputs,
+                    'outputs_list': payload.outputs}
+                    // TODO: add the rest: states, states_list, response(??)
+        })
+    }
+    else{
+        let input_dict = inputsToDict(inputs);
+        dc.callback_context.triggered = payload.changedPropIds.map(
+            prop_id => ({'prop_id': prop_id, 'value': input_dict[prop_id]}));
+        dc.callback_context.inputs_list = input_dict;
+        dc.callback_context.inputs = inputs;
+    }
 
     let returnValue;
 
--- a/dash-renderer/src/actions/index.js
+++ b/dash-renderer/src/actions/index.js
@@ -531,7 +531,7 @@ function inputsToDict(inputs_list) {
     // returns an Object (map):
     //  keys of the form `id.property` or `{"id": 0}.property`
     //  values contain the property value
-    if (!inputs_list){
+    if (!inputs_list) {
         return {};
     }
     const inputs = {};
--- a/dash-renderer/src/actions/index.js
+++ b/dash-renderer/src/actions/index.js
@@ -539,14 +539,16 @@ function inputsToDict(inputs_list) {
         if (Array.isArray(inputs_list[i])) {
             const inputsi = inputs_list[i];
             for (let ii = 0; ii < inputsi.length; ii++) {
-                const id_str = `${JSON.stringify(inputsi[ii].id)}.${
+                const id_str = `${stringifyId(inputsi[ii].id)}.${
                     inputsi[ii].property
                 }`;
-                inputs[id_str] = inputsi[ii].value ? inputsi[ii].value : null;
+                inputs[id_str] = inputsi[ii].value ?? null;
             }
         } else {
-            const id_str = `${inputs_list[i].id}.${inputs_list[i].property}`;
-            inputs[id_str] = inputs_list[i].value ? inputs_list[i].value : null;
+            const id_str = `${stringifyId(inputs_list[i].id)}.${
+                inputs_list[i].property
+            }`;
+            inputs[id_str] = inputs_list[i].value ?? null;
         }
     }
     return inputs;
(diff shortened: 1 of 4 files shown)


alexcjohnson:
@mrdawson this looks great! Just a few comments:

Rather than checking for its existence, I'd just set up `dc.callback_context` before the callback is executed, then delete it when the callback is done. And if we do that, I'm not even sure it's necessary to use `Object.defineProperty` - the only reason to do that with `no_update` and `PreventUpdate` was so they wouldn't be accidentally overwritten, but that wouldn't be a concern anymore.

> All tests passing

`npm run format` from the `dash-renderer` dir can fix the JS lint issues, and for the Py2 test unfortunately we'll have to avoid f-strings. Yes, we still have some users on Py2...

> requests 2.21.0 has requirement urllib3<1.25,>=1.21.1, but you'll have urllib3 1.25.9 which is incompatible.

Looks like our `requests` requirement is out of date. We do update them fairly regularly but if this causes any problems the safest thing to do is run tests in a venv.

> I had to comment out all of /test/integration/test_generation.py to gets pytest to work.
I’m on Windows, Python 3.7

Hmm, good to know! None of our core team works on Windows normally but this is of course an important environment to support. We do have a Windows test run on CI `build-windows-37`, which passes, but perhaps we should be running a few more of our tests in that environment. Would you mind creating a new issue for this and detailing the errors you saw?

mrdawson:
- The integration tests were all working on my end (Py 2.7 and 3.7). Anyone know why they are failing on CirleCI? 
- `response` is not included at this point. If someone wants to clarify why/how to add it, I can do so.
- @alexcjohnson Why delete the `dc.callback_context` after the callback is executed? Is it just best practice to free up the space?

alexcjohnson:
> The integration tests were all working on my end (Py 2.7 and 3.7). Anyone know why they are failing on CirleCI?

Yeah we really need a better error message from `wait_for_text_to_equal` and its friends, to show what the text *is*, not just what it *isn't*... My guess would be somehow the browser is different and `Object.keys` or `JSON.stringify` picked the keys in a different order. `Object.keys` you can just sort, `JSON.stringify` is trickier but here's one that should suffice:

```js
function stringifySorted(v) {
    if (Array.isArray(v)) {
        return '[' + v.map(stringifySorted).join(',') + ']';
    }
    if (v && typeof v === 'object') {
        const parts = Object.keys(v)
            .sort()
            .map(k => JSON.stringify(k) + ':' + stringifySorted(v[k]));
        return '{' + parts.join(',') + '}';
    }
    return JSON.stringify(v);
}
```

If that doesn't work, in order to debug you can just put in a few seconds `sleep` and do `assert dash_duo.get_element('#your-id').text == expected` - which will show the actual string found in the error message.

> `response` is not included at this point. If someone wants to clarify why/how to add it, I can do so.

Not useful for client-side callbacks. This is mostly to support server-side callbacks setting cookies.

> Why delete the `dc.callback_context` after the callback is executed? Is it just best practice to free up the space?

Maybe not super important, but we do this server side so that if someone tries to use `callback_context` when they're not in a callback it will fail rather than give them stale, meaningless data.

dash-renderer/src/actions/index.js
```diff
@@ -525,6 +525,33 @@ const getVals = input =>
...
+        if (Array.isArray(inputs_list[i])) {
+            const inputsi = inputs_list[i];
+            for (let ii = 0; ii < inputsi.length; ii++) {
+                const id_str = `${JSON.stringify(inputsi[ii].id)}.${
+                    inputsi[ii].property
+                }`;
+                inputs[id_str] = inputsi[ii].value ? inputsi[ii].value : null;
+            }
+        } else {
+            const id_str = `${inputs_list[i].id}.${inputs_list[i].property}`;
```
alexcjohnson, line 548: This can get object IDs too, with the `MATCH` wildcard.

For both this and the array case above lets use the [`stringifyId` function](https://github.com/plotly/dash/blob/dev/dash-renderer/src/actions/dependencies.js#L118) which handles both strings and objects, and ensures sorted object keys.

dash-renderer/src/actions/index.js
```diff
@@ -525,6 +525,33 @@ const getVals = input =>
...
+            inputs[id_str] = inputs_list[i].value ? inputs_list[i].value : null;
```
alexcjohnson, line 549: Some falsy values are real, it's just `undefined` we care about - perhaps we should use the [null-coalescing operator](https://github.com/plotly/dash-table/pull/729#discussion_r404411715)
```js
inputs[id_str] = inputs_list[i].value ?? null;
```
Also above.

dash-renderer/src/actions/index.js
```diff
@@ -544,12 +571,29 @@ function handleClientside(clientside_function, payload) {
     let returnValue;
 
     try {
+        // setup callback context
+        const input_dict = inputsToDict(inputs);
+        dc.callback_context = {};
+        if (payload.changedPropIds.length === 0) {
+            dc.callback_context.triggered = [{prop_id: '.', value: null}];
```
alexcjohnson, line 578: I don't think we want this dummy value on the client side. On the server it's there for backward compatibility as versions before 1.11 would always have at least one entry in `triggered` (and in fact almost always had exactly one entry) so people were in the habit of just looking at `triggered[0]["prop_id"]`. In JS there's no such history to work around. Also in Python we have the flexibility to make this length-1 array [look falsy](https://github.com/plotly/dash/blob/439e13b673e8e8f3d4223e905d44c08c9f11eef2/dash/_callback_context.py#L21) which we can't do in JS.

dash-renderer/src/actions/index.js
mrdawson: Oh cool. Is this what you mean? This should still work for wildcard object IDs.
``` JavaScript
const id_str = `${stringifyId(inputs_list[i].id)}.${inputs_list[i].property}`;
```
(same for the array case)

dash-renderer/src/actions/index.js
mrdawson: Sweet. Yeah this is a bit better.

dash-renderer/src/actions/index.js
mrdawson: Sounds good. I can just remove the `if` part of that conditional.

alexcjohnson:
Love it! Thanks for the changes, and your tests are fantastic. Glad to see they work without a fancy custom stringify after all 🎉 

The only last thing I'd ask for is a changelog entry. Just add `## [UNRELEASED]` at the top and describe this feature under `### Added`. Then we'll be ready to merge!

mrdawson:
🎉Thanks for your quick feedback!

BaconEggsRL:
Hi, beginner here having trouble getting this to work.
I get "undefined" for
window.dash_clientside.callback_context;

window.dash_clientside is definied but has no property "callback_context"

Thanks

BaconEggsRL:
> hi @BaconEggsRL Can you please share more information on what's not working? Can you share the code that is creating this error?

Hi Adam--not sure why but when I put the code in the .js file in the assets folder it does not work. I just get "cannot read property of undefined."
But when I put the same code in an inline app.callback it works. It doesn't seem to be finding the function when I put it in the "dashAgGridComponentFunctions.js" file.

.js code:
`window.dash_clientside = Object.assign({}, window.dash_clientside, {
    clientside: {
        test_function: function(btn-1, btn-2) {
            return "Hello from someone"
        }
    }
});`

python code:
`from dash import Dash, html, dcc, Output, Input, clientside_callback, ClientsideFunction

app = Dash(__name__)

app.layout = html.Div(
    [
        dcc.Markdown(f"Log (inline function)"),
        html.Div(id='log-clientside-inline', style={'display': 'block'}),  # none | block

        dcc.Markdown(f"Log (dashAgGridComponentFunctions.js function)"),
        html.Div(id='log-clientside-js', style={'display': 'block'}),  # none | block

        html.Button('btn-1', id="btn-1"),
        html.Button('btn-2', id="btn-2"),
    ]
)

# Clientside callback context (inline)
clientside_callback(
    """
    function(btn1, btn2) {
        // console.log('inline test')
        const triggered = dash_clientside.callback_context.triggered.map(t => t.prop_id);
        return "Hello from [" + triggered + "]"
    }
    """,
    Output("log-clientside-inline", "children"),
    Input("btn-1", "n_clicks"),
    Input("btn-2", "n_clicks"),
    prevent_initial_call=True,
)

# Clientside callback context (.js)
clientside_callback(
    ClientsideFunction(
        namespace='clientside',
        function_name='test_function'
    ),
    Output('log-clientside-js', 'children'),
    Input('btn-1', 'n_clicks'),
    Input('btn-2', 'n_clicks'),
    prevent_initial_call=True,
)


if __name__ == "__main__":
    app.run(debug=True)
`

BaconEggsRL:
I have similar issue as these folks:
https://community.plotly.com/t/how-to-use-two-functions-for-clientside-callbacks-within-js-files/71084/7
https://community.plotly.com/t/clientside-callback-example-with-js-function-throws-cannot-read-property-apply-of-undefined-error/44411/5

I can only get the function to work as an inline function. When I try to put it in .js file I get undefined error.

assets\dashAgGridComponentFunctions.js
```
window.dash_clientside = Object.assign({}, window.dash_clientside, {
    clientside: {
        test_function: function(btn-1, btn-2) {
            return "Hello from someone"
        }
    }
});
```

test_clientside_callback.py
```
from dash import Dash, html, dcc, Output, Input, clientside_callback, ClientsideFunction

app = Dash(__name__)

app.layout = html.Div(
    [
        dcc.Markdown(f"Log (inline function)"),
        html.Div(id='log-clientside-inline', style={'display': 'block'}),  # none | block

        dcc.Markdown(f"Log (dashAgGridComponentFunctions.js function)"),
        html.Div(id='log-clientside-js', style={'display': 'block'}),  # none | block

        html.Button('btn-1', id="btn-1"),
        html.Button('btn-2', id="btn-2"),
    ]
)

# Clientside callback context (inline)
clientside_callback(
    """
    function(btn1, btn2) {
        // console.log('inline test')
        const triggered = dash_clientside.callback_context.triggered.map(t => t.prop_id);
        return "Hello from [" + triggered + "]"
    }
    """,
    Output("log-clientside-inline", "children"),
    Input("btn-1", "n_clicks"),
    Input("btn-2", "n_clicks"),
    prevent_initial_call=True,
)

# Clientside callback context (.js)
clientside_callback(
    ClientsideFunction(
        namespace='clientside',
        function_name='test_function'
    ),
    Output('log-clientside-js', 'children'),
    Input('btn-1', 'n_clicks'),
    Input('btn-2', 'n_clicks'),
    prevent_initial_call=True,
)


if __name__ == "__main__":
    app.run(debug=True)
```

BaconEggsRL:
I think I fixed it just now.

Two issues:
 - 1st problem: Apparently JavaScript variables cannot have "-" characters. So that caused an error.
 - 2nd problem: I moved my test scripts into another folder inside the main project directory. So they can longer see the assets folder or any of the functions. I created a new assets folder for the test scripts and that solved the issue.

It does not seem to matter the name of the .js file as long as it is in the assets folder.

Thank you

PR #1254, opened 2020-05-15
Callback chain refactoring and performance improvements

Marc-Andre-Rivet:
Part of a larger performance rework. This PR proposes to break up the callback chain into smaller constituents by rewriting it using (1) a state-machine model where callbacks transition through various states (eg. requested, prioritized, executing, watched, executed, stored) that offer more limited logic each step of the way and hopefully easier optimization paths. (2) by handling derived store props and state transitions in (1) through the use of store observers.

- remove `pendingCallbacks` from the store and replace it with more easily used data structures for evaluating if:
  a. the app is ready (no callbacks to process)
  b. the component (or one if its children, for loading components) is awaiting the result of a callback
- remove `connect` from TreeContainer component
  a. Resolve `loading_state` in parent
  b. Resolve a `loading state hash` that can be used to decide if the parent needs to be re-rendered for its children
  c. Use `context` to provide shared data that do not affect re-renders (eg. graphs, etc.) - this replaces the `connect` usage

Companions
DCC: https://github.com/plotly/dash-core-components/pull/810 (testing only)
Table: https://github.com/plotly/dash-table/pull/785 (contains a tabe fix)
Docs: https://github.com/plotly/dash-docs/pull/888 (testing only)

There are some issues in the companion PRs but they seem related to CI only / the tests themselves are passing.

- [x] Percy diff for table in dash-docs
https://percy.io/plotly/dash-docs/builds/5455921/view/319601375/1280?mode=diff&browser=firefox&snapshot=319601375 (missing number of pages)

---

Performance

Gains increase as the number of displayed elements increases and can be explained by (1) less renders, (2) less expensive checks for states (eg. is_loading, shouldComponentUpdate).

1. dash.plotly.com/ vs branch (few components, none heavy, no callbacks)
JS time: 625ms vs 592ms (20 runs, p=0.08)
Total time: 1198ms vs 888ms (20 runs, p<=0.001)

2. dash.plotly.com/datatable/interactivity (few components, heavy: table/graph, w/ callbacks)
No significant difference after 10 runs.
JS time: 2488ms vs 2435ms (10 runs, p ~ 0.70)
Total time: 3430ms vs 3369ms (10 runs, p ~ 0.70)

3. HTML Table with 1000 rows (dev vs. branch) (lots of lightweight components, no callbacks)
JS time: 2706ms vs 1072ms (5 runs, p<=0.001)
Total time: 3460ms vs 1817ms (5 runs, p<=0.001)

4. HTML Table with 5000 rows (dev vs. branch) (lots of lightweight components, no callbacks)
JS time: 8412ms vs 3835ms (5 runs, p<=0.001)
Total time: N/A, devtools terminated test too early for dev and numbers were no significant - JS times were obtained by starting/ending the tests manually


dash-renderer/@Types/modules.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'cookie' {
+    const value: {
+        parse: (cookie: string) => {
+            _csrf_token: string
+        }
+    };
+
+    export default value;
+}
```
Marc-Andre-Rivet, line 9: Need to shim third-parties that don't have TS module definitions

dash-renderer/jest.config.js
```diff
@@ -85,7 +85,7 @@ module.exports = {
   // notifyMode: "always",
 
   // A preset that is used as a base for Jest's configuration
-  // preset: null,
+  preset: "ts-jest/presets/js-with-babel",
```
Marc-Andre-Rivet, line 88: For Jest to work with TS

dash-renderer/src/APIController.react.js
```diff
@@ -34,6 +47,18 @@ const UnconnectedContainer = props => {
...
+    propsRef.current = props;
+
+    const provider = useRef({
+        fn: () => ({
+            _dashprivate_config: propsRef.current.config,
+            _dashprivate_dispatch: propsRef.current.dispatch,
+            _dashprivate_graphs: propsRef.current.graphs,
+            _dashprivate_loadingMap: propsRef.current.loadingMap,
+        }),
+    });
```
Marc-Andre-Rivet, line 60: These props never affect the decision to re-render a TreeContainer. Instead of passing them and running a check with `shouldComponentUpdate`, provide through a context instead. The context value is a function that gives access to the `current` value of a subset of props coming from the store.

dash-renderer/src/APIController.react.js
```diff
@@ -65,11 +82,24 @@ const UnconnectedContainer = props => {
...
-            />
+            <DashContext.Provider value={provider.current}>
+                <TreeContainer
+                    _dashprivate_error={error}
+                    _dashprivate_layout={layout}
+                    _dashprivate_loadingState={getLoadingState(
+                        layout,
+                        [],
+                        loadingMap
+                    )}
```
Marc-Andre-Rivet, line 95: The loadingState is calculated by the parent component instead of through the store `connect` - also when not loading, the value is `false` instead of `{ is_loading: false }` - this is handled in the TreeContainer to remain consistent with the existing contract.

dash-renderer/src/APIController.react.js
```diff
@@ -65,11 +82,24 @@ const UnconnectedContainer = props => {
...
+                    _dashprivate_loadingStateHash={getLoadingHash(
+                        [],
+                        loadingMap
+                    )}
```
Marc-Andre-Rivet, line 99: Since loadingState only tells the component itself if it is loading or not, we need another value to act on the state of the component itself and its entire sub-tree. The loadingState hash is a value that represents the state of the entire layout tree under this component. It will be updated as pending callbacks change.

dash-renderer/src/APIController.react.js
```diff
@@ -65,11 +82,24 @@ const UnconnectedContainer = props => {
...
+                    _dashprivate_path={JSON.stringify([])}
```
Marc-Andre-Rivet, line 100: The path of this component inside the layout. As this is built on each render, it will never be the same as for the previous render. Stringify it. It will be handled in the TreeContainer. This with the above modifications allows the TreeContainer to be a PureComponent.

dash-renderer/src/APIController.react.js
```diff
@@ -157,6 +187,7 @@ UnconnectedContainer.propTypes = {
     graphs: PropTypes.object,
     layoutRequest: PropTypes.object,
     layout: PropTypes.object,
+    loadingMap: PropTypes.any,
```
Marc-Andre-Rivet, line 190: The loadingMap replaces the `pendingCallbacks` in that it allows knowing the state of all components in the layout tree and in conjunction with `isLoading` (not used here), it allows to know if the app is ready.

dash-renderer/src/AppProvider.react.tsx
```diff
@@ -0,0 +1,49 @@
+import PropTypes from 'prop-types';
```
Marc-Andre-Rivet, line 1: rewritten js -> tsx

dash-renderer/src/StoreObserver.ts
```diff
@@ -0,0 +1,113 @@
...
+    private _store?: Store<TStore>;
+    private _unsubscribe?: Unsubscribe;
+
+    private readonly _observers: IStoreObserverState<Store<TStore>>[] = [];
+
+    constructor(store?: Store<TStore>) {
+        this.__init__(store);
+    }
+
+    observe = (
```
Marc-Andre-Rivet, line 36: Registers a new observer on the store

dash-renderer/src/StoreObserver.ts
```diff
@@ -0,0 +1,113 @@
...
+        forEach(o => o.triggered = true, triggered);
+
+        forEach(
+            o => {
+                o.lastState = store.getState();
+                o.observer(store);
+                o.triggered = false;
+            },
+            triggered
+        );
```
Marc-Andre-Rivet, line 104: Notifies all observers. As observers can cause store side-effects (sync or async), it is necessary to keep track of when/how the observers got triggered to ensure (1) minimizing useless observer executions, (2) making sure all required observers are run.

dash-renderer/src/store.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface IStoreState {
+    callbacks: ICallbacksState;
+    isLoading: IsLoadingState;
+    loadingMap: LoadingMapState;
+    [key: string]: any;
+}
+
+let store: Store<IStoreState>;
+const storeObserver = new StoreObserver<IStoreState>();
```
Marc-Andre-Rivet, line 17: - js -> ts
- added export for the store observer

dash-renderer/src/observers/loadingMap.ts
```diff
@@ -0,0 +1,77 @@
...
+import { setLoadingMap } from '../actions/loadingMap';
+import { IStoreObserverDefinition } from '../StoreObserver';
+import { IStoreState } from '../store';
+
+const observer: IStoreObserverDefinition<IStoreState> = {
+    observer: ({
+        dispatch,
+        getState
+    }) => {
+        const {
```
Marc-Andre-Rivet, line 19: This observer sets the store value `loadingMap` and from the TreeContainer's perspective, replaces `pendingCallbacks` as the mean to determine the loading state of a component and its sub-layout.

This is done by constructing a "tree" where each node corresponds to a path-fragment. Each node contains child nodes for continuing path-fragments and a list of all the (id, prop) tuples of callback outputs that could affect the node itself or one of its children.

For example,
Callback #1 may affect (id=a,prop=value) at ["props", "children", 0]
Callback #2 may affect (id=b,prop=value) at ["props", "children", 1] and (id=b,prop=text) at ["props", "children", 1]

The resulting tree will be
```
{
  "__dashprivate__idprop__": [(id=a,prop=value),(id=b,prop=value),(id=b,prop=text)]
  "props": {
      "__dashprivate__idprop__": [(id=a,prop=value),(id=b,prop=value),(id=b,prop=text)]
    "children": {
        "__dashprivate__idprop__": [(id=a,prop=value),(id=b,prop=value),(id=b,prop=text)]
      0: {
          "__dashprivate__idprop__": [(id=a,prop=value)]
      },
      1: {
          "__dashprivate__idprop__": [(id=b,prop=value),(id=b,prop=text)]
      }
    }
  }
}
```

dash-renderer/src/StoreObserver.ts
```diff
@@ -0,0 +1,113 @@
...
+    filter,
+    forEach,
+    map,
+    path
+} from 'ramda';
+
+import { Store, Unsubscribe } from 'redux';
+
+type Observer<TStore> = (store: TStore) => void;
+type UnregisterObserver = () => void;
```
Marc-Andre-Rivet, line 12: Could use a third-party library to do this.. but this was simple enough to implement and sufficient for our need.. 🤷

dash-renderer/src/observers/executedCallbacks.ts
```diff
@@ -0,0 +1,219 @@
...
+
+            const { data, error, payload } = executionResult;
+
+            if (data !== undefined) {
+                forEach(([id, props]: [any, { [key: string]: any }]) => {
+                    const parsedId = parseIfWildcard(id);
+                    const { graphs, layout: oldLayout, paths: oldPaths } = getState();
+
+                    // Components will trigger callbacks on their own as required (eg. derived)
+                    const appliedProps = applyProps(parsedId, props);
```
Marc-Andre-Rivet, line 113: Implementation here is adapted from https://github.com/plotly/dash/pull/1254/files#diff-15692ba768eeab56e95aa409ab7c19b7L308

dash-renderer/src/observers/isLoading.ts
```diff
@@ -0,0 +1,28 @@
...
+            callbacks,
+            isLoading
+        } = getState();
+
+        const pendingCallbacks = getPendingCallbacks(callbacks);
+
+        const next = Boolean(pendingCallbacks.length);
+
+        if (isLoading !== next) {
+            dispatch(setIsLoading(next));
```
Marc-Andre-Rivet, line 22: App isLoading if there's at least one non-stored callback.

dash-renderer/src/observers/prioritizedCallbacks.ts
```diff
@@ -0,0 +1,104 @@
...
+        dispatch,
+        getState
+    }) => {
+        const { callbacks: { executing, watched }, config, hooks, layout, paths } = getState();
+        let { callbacks: { prioritized } } = getState();
+
+        const available = Math.max(
+            0,
+            6 - executing.length - watched.length
+        );
```
Marc-Andre-Rivet, line 46: callbacks that `can` be executed are promoted from `requested` to `prioritized` - from `prioritized` callbacks will be executed - conditions can be applied to limit the number of concurrent callbacks or bundle them, etc.

dash-renderer/src/observers/prioritizedCallbacks.ts
```diff
@@ -0,0 +1,104 @@
...
+    observer: async ({
+        dispatch,
+        getState
+    }) => {
+        const { callbacks: { executing, watched }, config, hooks, layout, paths } = getState();
+        let { callbacks: { prioritized } } = getState();
+
+        const available = Math.max(
+            0,
+            10 - executing.length - watched.length
```
Marc-Andre-Rivet, line 45: Most browsers don't allow more than 6 concurrent connections to the same domain. By over-provisioning, attempting to grossly balance useless callback execution (eg. subsequently dropped) vs. suffering from callback starvation.

dash-renderer/src/observers/prioritizedCallbacks.ts
```diff
@@ -0,0 +1,104 @@
...
+        const ids = reduce((res, [cb]) => ([
+            ...res,
+            ...cb.getInputs(paths),
+            ...cb.getState(paths)
+        ]), [] as ICallbackProperty[], callbacks);
+
+        /*
+            Make sure the app is ready to execute callbacks impacting `ids`
+        */
+        await isAppReady(layout, paths, uniq(pluck('id', ids)));
```
Marc-Andre-Rivet, line 80: As before, determine all impacted ids and make sure the components are ready. Optimizations here could include *not* a passthrough for callbacks linked to ready components.

dash-renderer/src/observers/prioritizedCallbacks.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        /*
+            Make sure to only execute callbacks that are still in the `prioritized` list (isAppReady is async - state could have changed)
+        */
+        const { callbacks: { prioritized: updatedPrioritized } } = getState();
+        const [remainingCallbacks] = partition(
+            ([cb]) => includes(cb, updatedPrioritized),
+            callbacks
+        );
```
Marc-Andre-Rivet, line 89: `await` is async - the prioritized callbacks could have been cleaned up since - make sure to only promote the remaining prioritized callbacks

dash-renderer/src/observers/prioritizedCallbacks.ts
```diff
@@ -0,0 +1,104 @@
...
+            return executeCallback(cb, config, hooks, paths, layout, stash);
+        });
+
+        dispatch(aggregateCallbacks([
+            remainingCallbacks.length ? removePrioritizedCallbacks(map(([cb]) => cb, remainingCallbacks)) : null,
+            executingCallbacks.length ? addExecutingCallbacks(executingCallbacks) : null
+        ]));
+
+    },
+    inputs: ['callbacks.prioritized', 'callbacks.completed']
```
Marc-Andre-Rivet, line 101: Because of the async / double pruning handling of prioritized callbacks, it would be possible for no callbacks to be promoted and for the callback chain to get stuck - this is why this observer also re-evaluates prioritized callbacks when `completed` changes.

dash-renderer/src/observers/requestedCallbacks.ts
```diff
@@ -0,0 +1,263 @@
...
+
+const observer: IStoreObserverDefinition<IStoreState> = {
+    observer: ({
+        dispatch,
+        getState
+    }) => {
+        const { callbacks, callbacks: { prioritized, executing, watched, executed, stored }, paths } = getState();
+        let { callbacks: { requested } } = getState();
+
+        const pendingCallbacks = getPendingCallbacks(callbacks);
```
Marc-Andre-Rivet, line 55: This observer promotes `requested` callbacks to `prioritized` for later execution. It handles most pruning cases upfront although some additional pruning does happen in the prioritized and executing observers.

Reasons for removing/modifying callbacks are numerous and documented below.

dash-renderer/src/observers/requestedCallbacks.ts
```diff
@@ -0,0 +1,263 @@
...
+            - prune them out of `requested`
+        */
+        const rCircular = (
+            !readyCallbacks.length &&
+            !prioritized.length &&
+            !executing.length &&
+            !watched.length &&
+            !executed.length &&
+            requested.length
+        ) ? requested : [];
```
Marc-Andre-Rivet, line 163: This handling of circular dependencies may need to change as it is different from the current undocumented behavior. I would suggest a potential approach of pruning out callback definitions in a circular dependency from the graph entirely, allowing for no ambiguity.

dash-renderer/src/observers/storedCallbacks.ts
```diff
@@ -0,0 +1,69 @@
...
+import { getPendingCallbacks } from '../utils/callbacks';
+import { IStoreObserverDefinition } from '../StoreObserver';
+
+const observer: IStoreObserverDefinition<IStoreState> = {
+    observer: ({
+        dispatch,
+        getState
+    }) => {
+        const { callbacks } = getState();
+        const pendingCallbacks = getPendingCallbacks(callbacks);
```
Marc-Andre-Rivet, line 32: Stored callbacks are kept to ensure callbacks with dependencies have access to the required information (eg. layout update). Callbacks without an execution group are dropped immediately while callbacks with an execution group are dropped out when no other callbacks remain pending for the group.

dash/testing/dash_page.py
```diff
@@ -36,13 +36,7 @@ def redux_state_paths(self):
...
         return self.driver.execute_script(
             """
-            return window.store.getState().pendingCallbacks.map(function(cb) {
-                var out = {};
-                for (var key in cb) {
-                    if (typeof cb[key] !== 'function') { out[key] = cb[key]; }
-                }
-                return out;
-            })
+            return !window.store.getState().isLoading;
```
Marc-Andre-Rivet, line 39: There's no `pendingCallbacks` anymore, changing usage for `isLoading`, which is equivalent to `|pendingCallbacks|!=0`

tests/integration/renderer/test_multi_output.py
```diff
@@ -133,10 +133,8 @@ def set_bc(a):
         dev_tools_hot_reload=False,
     )
 
-    # the UI still renders the output triggered by callback.
-    # The new system does NOT loop infinitely like it used to, each callback
-    # is invoked no more than once.
-    dash_duo.wait_for_text_to_equal("#c", "X")
+    # The new system does NOT trigger callbacks in circular dependencies
+    dash_duo.wait_for_text_to_equal("#c", "")
```
Marc-Andre-Rivet, line 137: This needs to be looked into. Circular dependencies are pruned out entirely now. They used to execute partially on load and unpredictably on prop updates.

dash-renderer/src/observers/prioritizedCallbacks.ts
chriddyp: > subsequently dropped

the case where callbacks are returned out of order and then their responses ignored?

chriddyp: I’d imagine that it would be faster for the browser to manage queuing and dispatching the extra requests would be faster than us managing and dispatching the requests, since there wouldn’t be any blocking JS in between request dispatches.

But I suppose the trade off is how many of these responses end up getting dropped on our end. So, when do sponsors get dropped? 

One case would be when there are multiple requests to the same callback, like typing data into a dcc.Input or dragging a slider. In that case, I suppose we would take the time to introspect the pending callbacks to see if there are any redundant updates and, if so, skip the requests that were triggered earlier? eg if a slider is dragged from 1 to 20, then we might make 5 requests with values from 1-5, and then before we start the next batch, we’d see that the remaining 15 pending requests are to the same callback so we’d drop requests 5-19 and only dispatch the request for 20. 

Am I thinking about this the right way? Are there other use cases where this comes into play besides multiple updates to the same component?

Marc-Andre-Rivet: > I’d imagine that it would be faster for the browser to manage queuing and dispatching the extra requests would be faster than us managing and dispatching the requests, since there wouldn’t be any blocking JS in between request dispatches.

Technically this is true but under certain scenarios, keep the flexibility to prune out the callbacks before the request is made can result in better real-world performance, with the caveat that we need to be careful about preventing sub-optimal usage of concurrent connections.

For example, when loading dash-docs `/all` with the `dev` version of the renderer, at some point ~155 callbacks will be "pending" - that is assigned for execution but blocked at the browser level from executing because of the concurrency rules.

> So, when do responses get dropped?

Callbacks can be pruned at any state up to when execution completes. Pruning prior to execution is by far preferable if possible as it doesn't use up available connections. Specifically pruning is done on each callback completion and request (https://github.com/plotly/dash/pull/1254/files#diff-178b2bd05a773877fd1b117eef8b1e8cR124) for all non-completed callbacks.

Let's imagine that the user clicks a link to go from dash-docs `/all` to `/something-else` - the `dccLocation.pathname` callback *needs* to wait for 150 of the 155 callbacks to complete before it's gets executed as they are already in the queue and out of our control.

While this is not currently part of the PR, doing something as simple as prioritizing callbacks that were triggered by user interactions (this should be a relatively simple addition), over component derived props, over layout change chains would make it possible to execute the user interaction callback first and drop the large majority of the 155 callbacks by layout/pruning, as most would not have been assigned already.

It is also possible to `abort` XHR requests once they've been requested but this is problematic for the scenario above. The callback's result is what causes the 155 previous callbacks to be invalidated, if these are already in queue, the new one can't take precedence.

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ jobs:
 
   artifacts:
     docker:
-      - image: circleci/python:3.7-stretch-node-browsers
+      - image: circleci/python:3.7.6-stretch-node-browsers
```
Marc-Andre-Rivet, line 13: Issues with Node v12.17 when installing optional dependencies - reverting back to previous docker images for all environments.

dash-renderer/src/APIController.react.js
alexcjohnson: I suppose your intent was to be explicit and consistent about always stringifying / parsing when you set / read this value... looks a little weird to me to vs just `_dashprivate_path="[]"` but OK.

dash-renderer/src/AppProvider.react.tsx
```diff
@@ -0,0 +1,49 @@
...
+import {Provider} from 'react-redux';
+
+import initializeStore, { observe } from './store';
+import AppContainer from './AppContainer.react';
+
+import executedCallbacks from './observers/executedCallbacks';
+import executingCallbacks from './observers/executingCallbacks';
+import isLoading from './observers/isLoading'
+import loadingMap from './observers/loadingMap';
+import prioritizeCallbacks from './observers/prioritizedCallbacks';
```
alexcjohnson, line 12: name mismatch?

dash-renderer/src/actions/dependencies_ts.ts
```diff
@@ -0,0 +1,323 @@
...
+    });
+    return matches;
+}
+
+/*
+ * Take a list of callbacks and follow them all forward, ie see if any of their
+ * outputs are inputs of another callback. Any new callbacks get added to the
+ * list. All that come after another get marked as blocked by that one, whether
+ * they were in the initial list or not.
+ */
```
alexcjohnson, line 88: obsolete comment

If I'm understanding this correctly: `priority` is an array, whose length is the number of the number of callbacks in the chain starting with this one (and including this one), that block other callbacks. The entries in that array are the number of callbacks directly blocked by each of those, but after the first one they're in a somewhat meaningless order. Is that right?

dash-renderer/src/observers/prioritizedCallbacks.ts
```diff
@@ -0,0 +1,124 @@
...
+    const lDiff = c2.priority.length - c1.priority.length;
+    if (lDiff) {
+        return lDiff;
+    }
+
+    for (let i = 0; i < c1.priority.length; ++i) {
+        const pDiff = c2.priority[i] - c1.priority[i];
+        if (pDiff) {
+            return pDiff;
+        }
```
alexcjohnson, line 46: Following on my previous comment/question: `sortPriority` first sorts by the total number of items in this array - ie not the total number of other callbacks you're blocking, but the total number of _blocking_ callbacks you're blocking - then by number of direct blocks of this callback, followed by number of direct blocks of blocked callbacks (again, in a somewhat meaningless order). Am I understanding that correctly?

This seems like a plausible sorting, and I guess with the information at hand there probably isn't a single best answer, just trying to understand what was done and why. Unless the specific logic here is important, it feels like there may be a simpler way to do this that would also collect all the logic in one place rather than separate `getPriority` and `sortPriority` each holding part of the logic. Like, have `getPriority` just reduce to a single score and sort descending on that score.

dash-renderer/src/actions/dependencies_ts.ts
```diff
@@ -0,0 +1,323 @@
...
+    map,
+    mergeWith,
+    partition,
+    pickBy,
+    props,
+    reduce,
+    zipObj
+} from 'ramda';
+import { ICallback, ICallbackProperty, ICallbackDefinition, ILayoutCallbackProperty, ICallbackTemplate } from '../types/callbacks';
+import { addAllResolvedFromOutputs, splitIdAndProp, stringifyId, getCallbacksInLayout, isMultiValued, idMatch } from './dependencies';
```
alexcjohnson, line 20: Is there a pattern to what got moved to `.ts` and what's still in the `.js` file? Or is it just the things you modified you moved to ts?

alexcjohnson: in particular `getLayoutCallbacks` vs `getCallbacksInLayout` is a bit confusing...

dash-renderer/src/checkPropTypes.js
```diff
@@ -21,7 +21,7 @@ export default function checkPropTypes(
     values,
     location,
     componentName,
-    getStack
+    getStack = null
```
alexcjohnson, line 24: ?

dash-renderer/src/actions/dependencies_ts.ts
alexcjohnson: oh no, looks like I misunderstood... the length is the depth of the whole callback tree and the entries are the number of callbacks blocked by each level of the tree?

dash-renderer/src/reducers/callbacks.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export interface ICallbacksState {
+    requested: ICallback[];
+    prioritized: ICallback[];
+    executing: IExecutingCallback[];
+    watched: IExecutingCallback[];
+    executed: IExecutedCallback[];
+    stored: IStoredCallback[];
+    completed: number;
+}
```
alexcjohnson, line 64: I gather this is the order that callbacks step through the process? I think I understand some of it:
- requested: something has caused this callback to need to fire
- prioritized: we've determined that this callback isn't waiting on anything other callback to complete
- executing: we have initiated the request for this callback

Is that correct? What are the other three (watched, executed, stored) for?

dash-renderer/src/reducers/callbacks.ts
```diff
@@ -0,0 +1,147 @@
...
+
+import {
+    ICallback,
+    IExecutedCallback,
+    IExecutingCallback,
+    IStoredCallback
+} from '../types/callbacks';
+
+export enum CallbackActionType {
+    AddApplied = 'Callbacks.AddApplied',
```
alexcjohnson, line 15: is `applied` in use?

dash-renderer/src/checkPropTypes.js
Marc-Andre-Rivet: When TS code calls JS code it tries to figure out what's going on anyway - without a default value on the parameter it thinks it's mandatory. Could make it `undefined` instead.

dash-renderer/src/actions/dependencies_ts.ts
Marc-Andre-Rivet: For a given callback with `[3, 4, 2]` it means that 3 callbacks can be triggered by the callback being evaluated, 4 callbacks can be triggered by those 3, and 2 callbacks by those 4. It's amounts to looking at the callback tree and determining how many callbacks exist at each depth in the tree.

dash-renderer/src/observers/prioritizedCallbacks.ts
Marc-Andre-Rivet: Priority sorting is done by first sorting by the depth of the callback tree and then by looking at the breadth of the callback tree, giving priority to a larger number of callbacks at smaller depths in the tree of a given callback.

Marc-Andre-Rivet: > Like, have getPriority just reduce to a single score and sort descending on that score.

Could probably make it a string of format `dd.dd.dd.dd` with each fragment containing the number of nodes, `0` padded and pruned at `99`. The first entry would be the depth itself. So for example `[4,3,2]` would become `04.04.03.02` and `[102,67]` would become `02.99.67`

dash-renderer/src/actions/dependencies_ts.ts
Marc-Andre-Rivet: It's a mixed bag. Some things are new, some things got moved to TS just to get a more precise function signature as it was used in TS and needed to be casted to `any`, etc.

dash-renderer/src/reducers/callbacks.ts
Marc-Andre-Rivet: `stored` holds on to `executed` callbacks after the response has been processed. Once there are no more callbacks associated with the execution group (layout and initialization callbacks) the stored callbacks for an execution group are removed.

`executed` callbacks have been processed by the server but their response has not been processed into the layout yet.

`watched` callbacks are probably redundant with `executing` callbacks - I'll see if I can 🔪  that step out entirely.

`executing`, `prioritized` and `requested` callbacks are as you describe above.

dash-renderer/src/reducers/callbacks.ts
Marc-Andre-Rivet: 🔪  https://github.com/plotly/dash/pull/1254/commits/ae38bf8d2e7cb125c3fb4a0944b93c4e121787eb

dash-renderer/src/AppProvider.react.tsx
Marc-Andre-Rivet: https://github.com/plotly/dash/pull/1254/commits/0da58103cdc60fdef78880ec96a0dc882752dda4#diff-7f8dd7ff81ab1823b478e9367574185cR14

dash-renderer/src/actions/dependencies_ts.ts
Marc-Andre-Rivet: > getLayoutCallbacks vs getCallbacksInLayout 

Agreed

dash-renderer/src/observers/prioritizedCallbacks.ts
Marc-Andre-Rivet: https://github.com/plotly/dash/pull/1254/commits/e50b7c2eff303b2093924f7903b47ad7c01f6874

dash-renderer/src/AppProvider.react.tsx
```diff
@@ -31,3 +30,5 @@ AppProvider.defaultProps = {
 };
 
 export default AppProvider;
+
+
```
alexcjohnson, line 34: Could redistribute these extra trailing newlines to some of the files that have none 😉 Somehow GH's new icon really screams at me...
Screen Shot 2020-06-02 at 9 42 20 AM

Marc-Andre-Rivet: https://github.com/plotly/dash/pull/1254/commits/10185dd8688baed6cd24f1db9b5e925f79937546

dash-renderer/src/actions/dependencies_ts.ts
Marc-Andre-Rivet: https://github.com/plotly/dash/pull/1254/commits/484a4e5d8a58d2cd661be80e1158b2e1698fccb2

Marc-Andre-Rivet:
@alexcjohnson Addressed concerns about circular callbacks pruning vs. execution in https://github.com/plotly/dash/pull/1254/commits/c93389bf59f77aac3bcddeb8f4af13ccc34057e2 - with this commit circular callbacks will not simply be pruned out but will instead be executed with an arbitrary entrypoint callback and will trigger the callbacks in the loop until such a time that it would re-executed a previously executed callback (through the new `predecessors` prop)

Marc-Andre-Rivet:
@alexcjohnson https://github.com/plotly/dash/pull/1254/commits/4396e7e8dc56db580cdb0927677d7a7c971d8b7e addresses a regression exposed by https://github.com/plotly/dash/commit/c93389bf59f77aac3bcddeb8f4af13ccc34057e2 - 

`ALLSMALLER` wildcard callbacks with lower ids were not blocking those high higher ids, each callback triggering new callbacks in higher ids, resulting in `n*(n+1)/2` callbacks instead of `n`. Also callbacks dependent on `ALL` on the same components as the `ALLSMALLER` callback above would be be executed too soon and result in incorrect results.

dash-renderer/src/TreeContainer.js
```diff
@@ -26,46 +23,16 @@ import {recordUiEdit} from './persistence';
...
-            _dashprivate_layout={component}
-            _dashprivate_path={path}
-        />
-    );
+import {
+    getLoadingHash,
+    getLoadingState,
+    validateComponent,
+} from './utils/TreeContainer';
+import {DashContext} from './APIController.react';
```
alexcjohnson, line 31: Apparently this works, but isn't it a circular dependency that `TreeContainer` imports from `APIController` which imports from `TreeContainer`?

Marc-Andre-Rivet: This is a really good question. It does look like a circular dependency, but it's not. ES6 imports granularity is defined at the module (export) level, not files. This was one of the key feature distinguishing it from "require" (`module.exports = ...`) when it was first developed.

In addition, modules need to be used immediately for the circular dependency to be problematic. For example this fails at runtime:

file1.js
```
import { A } from 'file2';

export const B = () => A();
B(); // A is undefined
```

file2.js
```
import { B } from 'file1';

export const A = () => B();
A(); // B is undefined
```

But this won't:
file1.js
```
import { A } from 'file2';

export const B = () => A();

(async () => {
    B(); // A is defined
})();
```

file2.js
```
import { B } from 'file1';

export const A = () => B();

(async () => {
    A(); // B is defined
})();
```

This behavior is probably the closest you can get to using sync `<script />` tags, but since there's no clear order for ES6 imports, both fail.

```
<script>
  function A() { B(); }
  A(); // B is undefined

  /* (async () => {
    A(); // B is defined
  })(); */
</script>
<script>
  function B() { A(); }
  B(); // A is defined!!! ... oups! stack overflow, here be dragons!

  /* (async () => {
    B(); // A is defined
  })(); */
</script>
```

alexcjohnson: OK - no need to alter anything now but I'm still a little wary of it. Maybe the tooling has improved in the meantime to prevent this but I know circular imports caused very confusing problems in the past in plotly.js, and it seems like it would only take a small and seemingly unrelated change to make the imports here problematic.

dash-renderer/src/utils/callbacks.ts
```diff
@@ -0,0 +1,8 @@
+import { omit, values } from 'ramda';
+
+import { ICallbacksState } from '../reducers/callbacks';
+import { ICallback } from '../types/callbacks';
+
+export const getPendingCallbacks = (state: ICallbacksState) => Array<ICallback>().concat(
+    ...values(omit(['stored', 'completed'], state))
```
alexcjohnson, line 7: I take it `Array().concat(...)` is effectively `flatten`?

dash-renderer/tslint.json
```diff
@@ -0,0 +1,58 @@
...
+            "node_modules/**",
+            "@Types/**",
+            "venv/**",
+            "**/*.js"
+        ]
+    },
+    "rules": {
+        "array-type": false,
+        "arrow-parens": [true, "ban-single-arg-parens"],
+        "ban-types": false, // to allow use of 'function' type
```
alexcjohnson, line 20: Can you actually use comments here (and below)? Normally not allowed by JSON.

Marc-Andre-Rivet: In VSCode I have this file handled as a `JSON with Comments`, removing the comments https://github.com/plotly/dash/pull/1254/commits/a5f9c3148c6c05342dff1054bed2c320c12403dc

dash-renderer/src/utils/callbacks.ts
Marc-Andre-Rivet: `concat` allows for n-arguments, either objects or arrays - I guess 99% of the time we don't use it that way https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

dash-renderer/src/TreeContainer.js
Marc-Andre-Rivet: Looked into https://github.com/aackerman/circular-dependency-plugin but it's a bit cumbersome. It identifies circular dependencies at the file level not for each individual export.. typically we should easily catch sync circular dependencies as they will result in console errors of the type `Can't ?? of undefined` during initialization, and we check for those in most if not all tests in this repo. But I agree, there's some fuzziness here.

dash/testing/dash_page.py
alexcjohnson: This looks like it's inverting the logic - previously a truthy `redux_state_rqs` meant "we're waiting for something," now it's "we're NOT loading anything."

Given that this is part of our public `dash.testing` API, it seems like we should probably keep "no requests" as `[]`, though I don't think it matters (for backward compatibility) what we put in the array when there _are_ requests. Could we reduce it to a list of output strings? ie the `"..id.prop...id2.prop2.."` form for multi-output or `"id.prop"` for single - that could be useful for debugging failures.

There was also the question about maintaining backward compatibility with older versions of dash, in case someone wants to use this framework to test running apps that may include various dash versions, or perhaps for testing a version compatibility matrix. I already broke that for versions prior to 1.11 when I switched from `requestQueue` to `pendingCallbacks` - it would be nice to un-break that eventually, and in the meantime at least not break it again. Let's just check which of the relevant items exists in the store and return some list based on that, with the "nothing in progress" state being and empty list.

tests/integration/renderer/test_multi_output.py
```diff
@@ -133,9 +133,7 @@ def set_bc(a):
         dev_tools_hot_reload=False,
     )
 
-    # the UI still renders the output triggered by callback.
-    # The new system does NOT loop infinitely like it used to, each callback
-    # is invoked no more than once.
```
alexcjohnson, line 138: I gather we're back to the behavior in this deleted comment now, not the new comment "The new system does NOT trigger callbacks in circular dependencies?"

Marc-Andre-Rivet: yes, need to 🔪  the 🔪

alexcjohnson:
> @alexcjohnson [4396e7e](https://github.com/plotly/dash/commit/4396e7e8dc56db580cdb0927677d7a7c971d8b7e) addresses a regression exposed by [c93389b](https://github.com/plotly/dash/commit/c93389bf59f77aac3bcddeb8f4af13ccc34057e2) -
> 
> `ALLSMALLER` wildcard callbacks with lower ids were not blocking those high higher ids, each callback triggering new callbacks in higher ids, resulting in `n*(n+1)/2` callbacks instead of `n`. Also callbacks dependent on `ALL` on the same components as the `ALLSMALLER` callback above would be be executed too soon and result in incorrect results.

Ah, I see - that's the reason for the test failures in c93389b - but in principle I can imagine fixes for this issue that would still give the correct final state, just inefficiently. Would it be worth 🔒 that a little tighter by asserting the correct number of callback invocations in one or two of the existing [wildcard tests](https://github.com/plotly/dash/blob/dev/tests/integration/callbacks/test_wildcards.py)?

Beyond that, just my latest comment about `redux_state_rqs` https://github.com/plotly/dash/pull/1254#discussion_r439525925 and the reverted test comment above https://github.com/plotly/dash/pull/1254#discussion_r439527964 and I think we're ready to go!!!

tests/integration/renderer/test_multi_output.py
Marc-Andre-Rivet: https://github.com/plotly/dash/pull/1254/commits/ca02d5578f13eb39acb111a33d1f4f1f84477282

Marc-Andre-Rivet:
@alexcjohnson 
> Would it be worth 🔒 that a little tighter by asserting the correct number of callback invocations in one or two of the existing wildcard tests?

https://github.com/plotly/dash/pull/1254/commits/be36ab47ae5977e71a2a225175b353647e5f7531

Also, `redux_state_rqs` now supports both legacy and current implementation for Dash matrix testing with https://github.com/plotly/dash/pull/1254/commits/afcdba90b68c178303ec68140ae7a07005d9c4c3 and https://github.com/plotly/dash/pull/1254/commits/63f14afffe1c91d6a00b8a688da2a62f804a46e5

PR #1262, opened 2020-05-21
Do not append wildcards into R function formals

rpkyle:
An earlier PR provided some 🌴 for the R component generator, but failed to excise a statement that duplicated the `...` argument within generated R component formals. This PR removes the two offending lines, and components (as well as their corresponding `.Rd` pages) are generated as required now.

@Marc-Andre-Rivet @alexcjohnson


rpkyle:
Following on offline conversation with @alexcjohnson, will fix by way of `add-rbuild-test`. 🙈

Issue #1264, opened 2020-05-21
Use persist_to_workspace to DRY up CircleCI config for repo clones


rpkyle:
The proposed CircleCI config which implements the R build test in #1238 repeats several repo clone steps performed in previous jobs. We could consider using `persist_to_workspace` to recycle these resources, and avoid having to request them again.

We already use the workspace feature, but currently we store package tarballs rather than the entire repositories.

More information is available here:

https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-artifacts-among-jobs

alexcjohnson:
Can we consider this solved by #1280? cc @Marc-Andre-Rivet

gvwilson:
Hi - we are tidying up stale issues and PRs in Plotly's public repositories so that we can focus on things that are most important to our community. If this issue is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. (Please note that we will give priority to reports that include a short reproducible example.) If you'd like to submit a PR, we'd be happy to prioritize a review, and if it's a request for tech support, please post in our community forum. Thank you - @gvwilson

PR #1151, opened 2020-03-12
Git hooks husky

Marc-Andre-Rivet:
This PR proposes many changes to the Dash monorepo toolchain.

- Add `black`
- Add `husky`
- Update CI to use wrapped scripts for both Python and JS linting / formatting

Py linting
- run `flake8` against all of `dash` and `tests`, no exceptions (more coverage)
- run `pylint` against all of `dash` and a subset of rules against `tests` (same as before)
- run `black` against all of `dash` and `tests`, no exceptions (new)

Js linting
- run the dash-renderer's npm `lint` command -- whatever it may contain

Format auto-runs black and the renderer's `format` script -- whatever it may contain.

Husky
`Husky` is a JS package that automagically runs commands on git hooks. For this repo it will run the `lint` script and expect no errors before allowing a commit to go through.

For the hooks to be present, the developer will need to run `npm ci` or `npm install` locally when cloning the repo.

CI config has been modified to use the lint/format scripts in the root folder of the repo instead of explicitly running flake8/pylint/black commands.

PR #1276, opened 2020-05-25
✏️ Fill in missing steps in CONTRIBUTING.md

dannysepler:
Looks like the CONTRIBUTING.md file was missing a few steps with npm?

NB: the black pre-commit failed on three files (`test_hot_reload.py`, `testing/plugin.py`, and `application_runners.py`). As a result, I committed with `--no-verify`. These pre-commit errors should be resolved here: https://github.com/plotly/dash/pull/1277

## Contributor Checklist

This checklist feels N/A cause there's no code change.


package.json
```diff
@@ -12,7 +12,7 @@
     "private::lint.renderer": "cd dash-renderer && npm run lint",
     "private::test.setup-nested": "cd \\@plotly/dash-generator-test-component-nested && npm ci && npm run build && pip install -e .",
     "private::test.setup-standard": "cd \\@plotly/dash-generator-test-component-standard && npm ci && npm run build && pip install -e .",
-    "private::test.unit-dash": "PYTHONPATH=~/dash/tests/assets pytest tests/unit",
```
dannysepler, line 15: was breaking for me since i didn't clone dash into `~/`

alexcjohnson: heh nor did I - looks like this was originally hard-coded in our circleci setup, then moved out of there when we created a dash-level `package.json` in #1151 but not generalized. TBH locally I've always run the desired tests directly via `pytest` but this is certainly better 🎉 

Actually it would be cleaner to do something inside `unit/test_app_runners.py` rather than rely on `PYTHONPATH` on the command line. It'd be nice if you could just run `pytest` and have it work. Apparently that was never the case for this test... I wonder if we could just `chdir` into `tests/assets`? `sys.path` won't work because the app gets started in a new process.

Anyway @dannysepler not something you need to worry about unless you're motivated to try. I'm just thinking out loud.

alexcjohnson:
@rpkyle looks like we may need to address #1264 sooner rather than later - unless there's a simple way we can generalize `${CIRCLE_BRANCH}` for PRs from a fork? That appears to be the cause of the test failures here.

package.json
dannysepler: totally agreed! big fan of just running `pytest` -- it's a lot more intuitive for new peeps like me. unfortunately i _still_ can't get the tests working lol (attached an issue about it), so i'm scared to do anything beyond the super surface-level

CONTRIBUTING.md
```diff
@@ -116,6 +119,8 @@ Note: *You might find out that we have more integration tests than unit tests in
 
 We introduced the `dash.testing` feature in [Dash 1.0](https://community.plotly.com/t/announcing-dash-testing/24868). It makes writing a Dash integration test much easier. Please read the [tutorial](http://dash.plotly.com/testing) and add relevant integration tests with any new features or bug fixes.
 
+To run the integration tests, you may have to [install circleci](https://circleci.com/docs/2.0/local-cli/).
```
alexcjohnson, line 122: Oh - I see, because the npm script `private::test.integration-dash` uses the circleci commands for splitting tests across multiple containers... @Marc-Andre-Rivet I see why you did it this way but in retrospect baking circleci into `package.json` seems a bit unfortunate... maybe we should mark the top-level scripts as meant for CI (so `test.unit`->`citest.unit` etc?), and make a new script `test` just for running locally (more for documentation purposes than actually to be run, but it'd be nice if it DID run!). Let me see if I can fix the `unit/test_app_runners.py` issue discussed previously, then I think this would be something very simple like:
```
"test": "pytest && cd dash-renderer && npm run test"
```

alexcjohnson: Fixed `test_app_runners.py` in dc837bd and c865257, and updated the scripts in a481b94 - @Marc-Andre-Rivet are you OK with this?

Then I took out this now-obsolete note about the circleci CLI and made a few more updates to CONTRIBUTING in bce1a37

FWIW if I run the entire `pytest` suite locally I currently get three errors, repeatably: `cbcx001`, `rddd001`, and `rdif001`. Haven't looked into it, and all three of these pass on CI.

dannysepler: thanks for all your help here alex! a further point to why i'm happy you made a new target, the `circleci tests glob` method requires "circleci-agent" available as a local bash command which was... hard to find documentation for? this is all to say this is MUCH better now!

as far as the three tests: i also have 3 failing tests but they're given by test name for me? `test_browser_smoke`, `test_process_server_smoke`, and `test_threaded_server_smoke` fail on `<<Webdriver not initialized correctly>>` -- is this what you're referring to as well?

alexcjohnson: hmm no, `cbcx001` for example is `test_cbcx001_modified_response`, from `tests/integration/callbacks/test_callback_context.py`

We should really extend the test-case-ID pattern to the unit tests as well - I think they were left out of that pattern mainly because the unit tests run so much faster than integration tests, but it'd still be nice to be able to easily target unit tests, and just for uniformity.

If those ones are failing, sounds like a problem with chromedriver or something? But if the integration tests pass then things are _generally_ working fine, so I'm not sure what the problem could be.

dannysepler: Gotcha, sounds like a me problem then :) thanks for explaining!

Issue #1265, opened 2020-05-21
Need to add setup-tests entries in package.json for R


rpkyle:
Currently, `package.json` has entries for Python test setup directives:

```json
    "private::test.setup-nested": "cd \\@plotly/dash-generator-test-component-nested && npm ci && npm run build && pip install -e .",
    "private::test.setup-standard": "cd \\@plotly/dash-generator-test-component-standard && npm ci && npm run build && pip install -e .",
```

and

```json
    "setup-tests": "run-s private::test.setup-*",
```

We should add a similar entry for R after determining how best to make available the R binary within the Python image's environment, i.e.

```json
    "private::test.setup-r-standard": "cd \\@plotly/dash-generator-test-component-standard && npm ci && npm run build && sudo R CMD INSTALL . && cd ../../..",
    "private::test.setup-r-nested": "cd \\@plotly/dash-generator-test-component-nested && npm ci && npm run build && sudo R CMD INSTALL . && cd ../../.."
```

and

```json
    "setup-tests-r": "run-s private::test.setup-r-*",
```

@Marc-Andre-Rivet

PR #1277, opened 2020-05-25
✅ Working pre-commit

dannysepler:
- 3 test files had not been blacked yet
- A pylint error is firing. I am ignoring it, doesn't seem particularly dangerous

## Contributor Checklist

- [👎 ] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))

:( I have issues running the tests. Going to open an issue for the integration tests! Perhaps either the documentation needs an extra section, or there's a way to magically standardize the "chromedriver", the "geckodriver", etc etc. (Don't know selenium that well haha)

PR #1280, opened 2020-05-26
CI on forked repo and other tweaks

Marc-Andre-Rivet:
- Fix issue running `dashr` CI job from a fork
- Fix https://github.com/plotly/dash/issues/1265


dash/testing/application_runners.py
```diff
@@ -222,8 +222,7 @@ def stop(self):
                 logger.info("proc.terminate with pid %s", self.proc.pid)
                 self.proc.terminate()
                 if self.tmp_app_path and os.path.exists(self.tmp_app_path):
-                    logger.debug("removing temporary app path %s",
-                                 self.tmp_app_path)
+                    logger.debug("removing temporary app path %s", self.tmp_app_path)
```
alexcjohnson, line 225: Surprised these bits of formatting aren't caught by our CI linting job - does this PR fix that? Looks like you're making largely the same changes @dannysepler made in #1277

Marc-Andre-Rivet: Yeah, I'm also a bit confused at this. The husky commit hook should be running essentially the same thing as CI. Maybe there are still some dependencies whose versions aren't fixed that would account for the different behavior?

package.json
```diff
@@ -10,16 +10,21 @@
...
     "private::test.unit-dash": "PYTHONPATH=~/dash/tests/assets pytest tests/unit",
     "private::test.unit-renderer": "cd dash-renderer && npm run test",
     "private::test.integration-dash": "TESTFILES=$(circleci tests glob \"tests/integration/**/test_*.py\" | circleci tests split --split-by=timings) && pytest --headless --nopercyfinalize --junitxml=test-reports/junit_intg.xml ${TESTFILES}",
     "format": "run-s private::format.*",
     "initialize": "run-s private::initialize.*",
     "lint": "run-s private::lint.*",
-    "setup-tests": "run-s private::test.setup-*",
-    "test.integration": "run-s setup-tests private::test.integration-*",
+    "setup-tests.py": "run-s private::test.py.deploy-*",
+    "setup-tests.R": "run-s private::test.R.deploy-*",
```
rpkyle, line 26: @Marc-Andre-Rivet Will this work within `3.7-stretch-node-browsers`? 

I would have thought you'd either need to layer in an R image or install the binary in order to run R within the Python base image.

Issue #1285, opened 2020-05-28
[BUG] Cannot run debug=True when sub classing html or dcc components 


gioxc88:
```
python 3.7.7
dash                 1.11.0
dash-core-components 1.9.1
dash-html-components 1.0.3
dash-renderer        01.4.0
dash-table           4.6.2
```

Whenever I try to subclass a component I cannot use the app in debug mode.
Unfortunately it becomes unusable without debug mode when developing.

This is the full traceback

```
Traceback (most recent call last):
  File "C:\Users\gioxc\AppData\Local\Programs\Python\Python37\lib\pkgutil.py", line 493, in find_loader
    spec = importlib.util.find_spec(fullname)
  File "C:\Users\gioxc\AppData\Local\Programs\Python\Python37\lib\importlib\util.py", line 114, in find_spec
    raise ValueError('{}.__spec__ is None'.format(name))
ValueError: __main__.__spec__ is None

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "apptest.py", line 108, in <module>
    app.run_server(debug=True)
  File "C:\Users\gioxc\AppData\Local\Programs\Python\Python37\lib\site-packages\dash\dash.py", line 1475, in run_server
    dev_tools_prune_errors,
  File "C:\Users\gioxc\AppData\Local\Programs\Python\Python37\lib\site-packages\dash\dash.py", line 1282, in enable_dev_tools
    for x in list(ComponentRegistry.registry) + ["dash_renderer"]
  File "C:\Users\gioxc\AppData\Local\Programs\Python\Python37\lib\site-packages\dash\dash.py", line 1277, in <listcomp>
    os.path.dirname(package.path)
  File "C:\Users\gioxc\AppData\Local\Programs\Python\Python37\lib\site-packages\dash\dash.py", line 1282, in <genexpr>
    for x in list(ComponentRegistry.registry) + ["dash_renderer"]
  File "C:\Users\gioxc\AppData\Local\Programs\Python\Python37\lib\pkgutil.py", line 499, in find_loader
    raise ImportError(msg.format(fullname, type(ex), ex)) from ex
ImportError: Error while finding loader for '__main__' (<class 'ValueError'>: __main__.__spec__ is None)

```

alexcjohnson:
Oh interesting - I guess the subclass is defined in the main app file itself? Not a use case we've encountered before but I guess we can filter out `__main__` from this list. I wonder if there are any other cases `pkgutil.find_loader` will fail?

[...]

PR #1288, opened 2020-05-30
Debug in same module

dannysepler:
Resolves https://github.com/plotly/dash/issues/1285. "pkgutil.find_loader" should return None on `__main__`, but doesn't on many versions of python (and still might not? It's failing for me on 3.7).

We should be able to just ignore `__main__` for the hot reloader. I tried it on the file linked here https://community.plotly.com/t/debug-mode-breaks-when-subclassing-dash-core-component/38498 and it worked

## Contributor Checklist

- [**No**] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))

Tests still don't work for me locally. Hopefully the CI tests suffice.

### optionals


dash/dash.py
```diff
@@ -1338,14 +1338,19 @@ def enable_dev_tools(
             _reload = self._hot_reload
             _reload.hash = generate_hash()
 
+            # find_loader should return None on __main__ but doesn't
+            # on some python versions https://bugs.python.org/issue14710
+            packages = [
+                pkgutil.find_loader(x)
+                for x in list(ComponentRegistry.registry) + ["dash_renderer"]
+                if x not in ("__main__")
```
alexcjohnson, line 1346:
```suggestion
                if x != "__main__"
```

Or `if x not in ("__main__",)` but we might as well do simple `!=` until we find more cases we need to exclude.

alexcjohnson:
Thanks @dannysepler - there's something wrong with our CI tests for everyone right now, but once we resolve that (and the little comment above) this looks good!

dash/dash.py
dannysepler: thanks!

PR #1302, opened 2020-06-12
1189 julia components

waralex:
Changes in generator to reverse dependencies. 
Now the main components depend on DashBase, and the rest depend on Dash. I also took out the logic for checking parameters to DashBase and removed the generation of additional constructors if the component does not have the children property. 
Corresponding Dash PR: https://github.com/plotly/Dash.jl/pull/42


dash/development/_jl_components_generation.py
```diff
@@ -23,59 +26,48 @@
 export {funcname}
 
 """
-    {funcname}(;kwargs...)
-    {funcname}(children::Any;kwargs...)
-    {funcname}(children_maker::Function;kwargs...)
+    {funcname}(;kwargs...){childrens_signatures}
```
alexcjohnson, line 29: children is already plural, let's just call this `children_signatures` (also `children_definitions`)

dash/development/_jl_components_generation.py
```diff
@@ -23,59 +26,48 @@
...
 const resources_path = realpath(joinpath( @__DIR__, "..", "deps"))
 const version = "{version}"
 
 {component_includes}
 
 function __init__()
-    Dash.register_package(
-        Dash.ResourcePkg(
+    DashBase.register_package(
+        DashBase.ResourcePkg(
```
alexcjohnson, line 62: Not new here, but is it important that this be two nested calls? Could it be refactored to just:
```julia
DashBase.register_package(
    "{project_shortname}",
    resources_path,
    version = version,
    [
        {resources_dist}
    ]
)
```

dash/development/_jl_components_generation.py
```diff
@@ -23,59 +26,48 @@
...
-                throw(ArgumentError("Invalid property $(string(prop)) for component " * "{funcname}"))
-            end
-
-            push!(result.props, prop => Front.to_dash(value))
-        end
-
-    return result
+        available_props = Symbol[{component_props}]
+        wild_props = Symbol[{wildcard_symbols}]
+        return Component("{funcname}", "{element_name}", "{module_name}", available_props, wild_props; kwargs...)
```
alexcjohnson, line 36: FWIW on the Python side we could have used only `kwargs` like this but we chose to insert all the prop names explicitly in the signature (except wildcards of course) because it helps with autocompletion in IDEs. I have no idea if similar functionality exists in Julia, and anyway even if we do want to do this we can add it later.

alexcjohnson:
@waralex this and https://github.com/plotly/Dash.jl/pull/42 look great, pending a few minor comments. I thought we had solved the CI build problem you're seeing here, but @Marc-Andre-Rivet probably knows precisely what's going on (aside from the linter complaints, which you should be able to address easily via `flake8` / `black` when you're ready)

dash/development/_jl_components_generation.py
waralex: @alexcjohnson fixed in 650b9d3

dash/development/_jl_components_generation.py
alexcjohnson: @waralex points out we can do this later - not a major issue anyway.

alexcjohnson:
merging into #1197, will sort out the tests and give a final review there.

PR #1310, opened 2020-06-24
fix multiple concurrent loading states

alexcjohnson:
Fixes https://github.com/plotly/dash-core/issues/195
Regression from #1254 - only one loading state could be shown at a time.

### DIFF
--- a/dash-renderer/src/observers/loadingMap.ts
+++ b/dash-renderer/src/observers/loadingMap.ts
@@ -1,7 +1,6 @@
 import {
     equals,
     flatten,
-    forEach,
     isEmpty,
     map,
     reduce
@@ -43,26 +42,22 @@ const observer: IStoreObserverDefinition<IStoreState> = {
         const nextMap: any = isEmpty(loadingPaths) ?
             null :
             reduce(
-                (res, path) => {
+                (res, {id, property, path}) => {
                     let target = res;
-                    const idprop = {
-                        id: path.id,
-                        property: path.property
-                    };
+                    const idprop = {id, property};
 
                     // Assign all affected props for this path and nested paths
                     target.__dashprivate__idprops__ = target.__dashprivate__idprops__ || [];
                     target.__dashprivate__idprops__.push(idprop);
 
-                    forEach(p => {
-                        target = (target[p] =
-                            target[p] ??
-                                p === 'children' ? [] : {}
-                        )
+                    path.forEach((p, i) => {
+                        target = (target[p] = target[p] ??
+                            (p === 'children' && typeof path[i + 1] === 'number' ? [] : {})
+                        );
 
                         target.__dashprivate__idprops__ = target.__dashprivate__idprops__ || [];
                         target.__dashprivate__idprops__.push(idprop);
-                    }, path.path);
+                    });
 
                     // Assign one affected prop for this path
                     target.__dashprivate__idprop__ = target.__dashprivate__idprop__ || idprop;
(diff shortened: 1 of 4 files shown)


dash-renderer/src/observers/loadingMap.ts
```diff
@@ -43,26 +42,22 @@ const observer: IStoreObserverDefinition<IStoreState> = {
...
+                    const idprop = {id, property};
 
                     // Assign all affected props for this path and nested paths
                     target.__dashprivate__idprops__ = target.__dashprivate__idprops__ || [];
                     target.__dashprivate__idprops__.push(idprop);
 
-                    forEach(p => {
-                        target = (target[p] =
-                            target[p] ??
-                                p === 'children' ? [] : {}
```
alexcjohnson, line 60: Turns out `??` has higher precedence than `?` in JS 🙈 The only necessary part of this fix is the extra parentheses around this line.

Added the `typeof path[i + 1]` clause so that the `[]` vs `{}` distinction is correct - though it doesn't seem to actually cause problems when the container has the wrong type 🤷

chriddyp:
Do we want to test loading states for chained callbacks too? Dropdown A ➡️  Dropdown B ➡️  Graph. If Dropdown A changes, do we show a loading state for Graph before Dropdown B is updated?

I'm not sure what the existing behavior is but seems like it could be another piece of logic that could've changed during the refactor.

alexcjohnson:
> Do we want to test loading states for chained callbacks too?

Sure, I can add a test for that later tonight.

chriddyp:
Here is a test case for chained callbacks & loading states:
```python
import dash
import time

app = dash.Dash(__name__)

app.layout = html.Div([
    html.Button(id='button', children='Start', n_clicks=0),

    html.Div(
        dcc.Loading(html.Div(id='output-1', style={
            'width': 200,
            'height': 200,
            'backgroundColor': 'hotpink',
        })),
        style={'display': 'inline-block'}
    ),

    html.Div(
        dcc.Loading(html.Div(id='output-2', style={
            'width': 200,
            'height': 200,
            'backgroundColor': 'rebeccapurple',
        })),
        style={'display': 'inline-block'}
    ),

    html.Div(
        dcc.Loading(html.Div(id='output-3', style={
            'width': 200,
            'height': 200,
            'backgroundColor': 'green',
        })),    
        style={'display': 'inline-block'}
    ),

    html.Div(
        dcc.Loading(html.Div(id='output-4', style={
            'width': 200,
            'height': 200,
            'backgroundColor': '#FF851B',  # orange
        })),
        style={'display': 'inline-block'}
    )
])

@app.callback(Output('output-1', 'children'), [Input('button', 'n_clicks')])
def update_output_2(n_clicks):
    time.sleep(2)
    return 'Output 1: {}'.format(n_clicks)


@app.callback(Output('output-2', 'children'), [Input('output-1', 'children')])
def update_output_2(children):
    time.sleep(2)
    return 'Output 2: {}'.format(children)


@app.callback([Output('output-3', 'children'), Output('output-4', 'children')], [Input('output-2', 'children')])
def update_output_2(children):
    time.sleep(2)
    return [
        'Output 3: {}'.format(children),
        'Output 4: {}'.format(children)
    ]



if __name__ == '__main__':
    app.run_server(debug=True)
```

In 1.12.0 we don't display a loading state for grandchildren, only immediate children. Looks like the 1.13 behavior is the same right now.

Here's what this looks like in 1.12.0

alexcjohnson:
Thanks @chriddyp - added that app as a test case in https://github.com/plotly/dash/pull/1310/commits/a2259221e386bd5c7aee62d8b4c45a4f85196bf5

Issue #856, opened 2019-08-07
[Feature Request] make DocumentTitle changes adjustable


philipphutterer:
Hello there,

as already "discussed" [in the community forum](https://community.plot.ly/t/15379) some time ago, I suggest to have a possibility that allows you to disable the annoying change of the Document Title to "Updating..." every time there is a data update - what might happen very often since you can choose any arbitrary small interval.

In my opinion it should be possible to either
- dynamically set the title during the data update in the callback or
- be a static value what can be set up initially

code file: `DocumentTitle.react.js`

Thanks in advance and best regards,
Philipp

CaseGuide:
Do the features described in #286 take care of this? It seems like overwriting the HTML title tag might fix it.

philipphutterer:
I tried to override the HTML title tag as suggested in #286 and it did not work. The updating script is still intercepting this.

chriddyp:
This feature is now available in `dash==1.14.0` and is documented in http://dash.plotly.com/external-resources. Look for the `title=` and `update_title=`. See the "[Customizing Dash's Document or Browser Tab Title]( http://dash.plotly.com/external-resources)", " [Update the Document Title Dynamically based off of the URL or Tab]( http://dash.plotly.com/external-resources)", and "[Customizing or Removing Dash's "Updating..." Message](http://dash.plotly.com/external-resources)" sections.

Issue #732, opened 2019-02-04
Ability to configure / change the loading state in the tab: "Updating..."


chriddyp:
Right now, `dash-renderer` updates the page tab whenever there is a request in flight.

For fast updating callbacks, e.g. `dcc.Interval`, this becomes pretty annoying: https://community.plot.ly/t/how-to-avoid-updating-html-browser-title-when-using-interval-event/15379

This is one of the few "hardcoded" features in Dash that we should make configurable:
1. We could make this a `config` option to turn `on`/ `off` or change the text
2. We could remove it from `dash-renderer` and attempt to roll it in to loading states. However, loading states are currently configurable only _per-component_, not across the entire app.

Seems like (1) would be the way to go right now. It could be something like 
```
update_page_title_during_requests = true / false / 'Loading ...'
```

fl4p:
It would also be nice to simply have a CSS class added while loading is happening, i.e. `<body class="loading">`

stlehmann:
This would be a great improvement especially when it comes to displaying live data with small intervals (like 100ms). I would like to offer my help implementing this. Maybe someone of the core developers could give a hint where to start. I still couldn' t figure out where in the code the title is changed. I really like the idea of @chriddyp to add an attribute in the config:

> Seems like (1) would be the way to go right now. It could be something like
> 
> update_page_title_during_requests = true / false / 'Loading ...'

stlehmann:
The title is set in DocumentTitle.react.js:

```
import {connect} from 'react-redux';
import {Component} from 'react';
import PropTypes from 'prop-types';

class DocumentTitle extends Component {
    constructor(props) {
        super(props);
        this.state = {
            initialTitle: document.title,
        };
    }

    UNSAFE_componentWillReceiveProps(props) {
        if (props.pendingCallbacks.length) {
            document.title = 'Updating...';
        } else {
            document.title = this.state.initialTitle;
        }
    }

    shouldComponentUpdate() {
        return false;
    }

    render() {
        return null;
    }
}
```

chriddyp:
[...]

PR #1315, opened 2020-06-28
Configure document-title during updates

stlehmann:
This PR addresses #856 and #732. It adds an `update_title` parameter to `Dash.__init__()`which can be set to a String or None. If this parameter is set to a String this will be the document title that is shown if Dash is updating. If `update_title` is set to None or "" the document title will not change at all. This is especially useful for applications with short update intervals to prevent flickering of the document title.

I have not yet added any tests because I didn't know where to put them. If you should request some tests I am more than willing to provide some :)


dash-renderer/src/components/core/DocumentTitle.react.js
```diff
@@ -5,14 +5,16 @@ import PropTypes from 'prop-types';
 class DocumentTitle extends Component {
     constructor(props) {
         super(props);
+        const {update_title} = props.config;
         this.state = {
             initialTitle: document.title,
+            update_title: update_title,
```
stlehmann, line 11: I was not sure whether to use camel- or snake-case here. If you prefer camel-case I can change it right-away.

alexcjohnson: Snake is good. Then this can in fact be simplified:
```suggestion
            update_title,
```

dash-renderer/src/components/core/DocumentTitle.react.js
```diff
@@ -29,8 +31,10 @@ class DocumentTitle extends Component {
 
 DocumentTitle.propTypes = {
     isLoading: PropTypes.bool.isRequired,
+    update_title: PropTypes.string,
```
alexcjohnson, line 34: Dunno if you can see the test results https://circleci.com/gh/plotly/dash/24509, but they helpfully flagged this one: should be `config` with `update_title` nested inside a `PropTypes.shape`.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [1.13.4] - 2020-06-25 
 ### Fixed
+- [#1315](https://github.com/plotly/dash/pull/1315) Add `update_title` parameter to set or disable the "Updating...." document title during updates. Closes [#856](https://github.com/plotly/dash/issues/856) and [#732](https://github.com/plotly/dash/issues/732)
```
alexcjohnson, line 7: Can you put this in a new section `## Unreleased` above `## [1.13.4]`?

alexcjohnson:
@stlehmann thanks, this looks great! Re tests: How about in [`test_loading_states.py`](https://github.com/plotly/dash/blob/dev/tests/integration/renderer/test_loading_states.py)? The same pattern we used there should work for this purpose, with `Lock` to pause the callback until you've seen what you want in the browser.

dash-renderer/src/components/core/DocumentTitle.react.js
stlehmann: I had a look at the implementation of the other core components and changed the line accordingly. I hope that is OK.

stlehmann:
@alexcjohnson 
> How about in [`test_loading_states.py`](https://github.com/plotly/dash/blob/dev/tests/integration/renderer/test_loading_states.py)? The same pattern we used there should work for this purpose, with `Lock` to pause the callback until you've seen what you want in the browser.

Thanks for the hint. I'll get into it.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+### Fixed
```
alexcjohnson, line 6:
```suggestion
### Added
```

dash-renderer/src/components/core/DocumentTitle.react.js
```diff
@@ -29,8 +31,10 @@ class DocumentTitle extends Component {
 
 DocumentTitle.propTypes = {
     isLoading: PropTypes.bool.isRequired,
+    config: PropTypes.object,
```
alexcjohnson, line 34:
```suggestion
    config: PropTypes.shape({update_title: PropTypes.string}),
```

`PropTypes.object` will work, but since we know the piece we want from inside `config` it's nice to specify it. Note I'm suggesting `shape`, not `exact`, because `exact` would require us to list _all_ of config and here we only care about this one part.

stlehmann: I see...this is much more explicit. Thanks for suggesting the change. I didn't know that one could make such explicit specificifications for PropTypes.

stlehmann:
@alexcjohnson I added tests to test_loading_states.py. I tested for update_title being set to default, None, empty and a custom string. Hopfefully the test implementation is acceptable. Otherwise just let me know :)

tests/integration/renderer/test_loading_states.py
```diff
@@ -167,3 +167,111 @@ def find_text(spec):
 
     find_spinners()
     find_text({1: 1, 2: 1, 3: 1, 4: 1})
+
+
+def test_update_title_default(dash_duo):
```
alexcjohnson, line 172:
```suggestion
def test_rdls003_update_title_default(dash_duo):
```

`rdls` comes from the path of this file, renderer->loading_states, and `003` is serial within the file. This lets you run this whole file `pytest -k rdls` or just this test case `pytest -k rdls003`

tests/integration/renderer/test_loading_states.py
```diff
@@ -167,3 +167,111 @@ def find_text(spec):
...
+        [Input("button", "n_clicks")]
+    )
+    def update(n):
+        with lock:
+            return n
+
+    with lock:
+        dash_duo.start_server(app)
+        dash_duo.find_element("#button").click()
+        assert dash_duo.driver.title == "Updating..."
```
alexcjohnson, line 196: Very nice tests! One thing though: your callback will execute on page load, right? So holding the lock while starting the server means we're testing the title during the initialization sequence. That's a good thing to test, but I think in at least one of these tests (and maybe all of them) we should test:
- during startup, we see the correct updating title
- when the callback is done we see the regular title
- when the callback is triggered interactively we see the updating title again

Also rather than a simple `assert`, for robustness I think it'd be better to do a `wait.until`, in case things don't happen quite in the expected order:

```py
from dash.testing.wait import until

until(lambda: dash_duo.driver.title == "Updating...", timeout=1)
```

And one last thing, if you're feeling adventurous: These test cases are nearly identical so they could be parametrized, something like:
```py
import pytest

@pytest.mark.parametrize(
    "kwargs,page_updating_title",
    [
        ({}, "Updating..."),
        ({"updating_title": None}, "Dash"),
        ({"updating_title": ""}, "Dash"),
        ({"updating_title": "Hello World"}, "Hello World")
    ]
)
def test_rdls003_update_title_default(dash_duo, updating_title, page_updating_title)
```

stlehmann: I tried to implement your suggestions. I omitted an explicit test for the correct update title on startup because the callback should be called on startup afaik. I hope this is OK. The parametrize decorator is a great thing. Didn't know that until now, thanks :) Hope I did it all right with the locks. I'm not that sure about it.

tests/integration/renderer/test_loading_states.py
```diff
@@ -167,3 +169,46 @@ def find_text(spec):
...
+    )
+
+    @app.callback(
+        Output("output", "children"),
+        [Input("button", "n_clicks")]
+    )
+    def update(n):
+        with lock:
+            # check for update-title while processing callback
+            until(lambda: dash_duo.driver.title == expected_update_title, timeout=1)
```
alexcjohnson, line 202: Interesting... this appears to work but I'm kind of surprised it does - it might actually be waiting for the timeout period and then throwing an error silently. That's because the server and the browser side are running in different processes (hence the need for `Lock`).

I'd remove the title check inside the callback, and just put it at the very end of the test, after `#button.click()` (but still inside the `with lock:` block)

stlehmann: @alexcjohnson 
> I'd remove the title check inside the callback, and just put it at the very end of the test, after #button.click() (but still inside the with lock: block)

This was my first shot but that didn't work at that time because I missed the lock inside the callback. Now this works like a charm. I also added the check on startup again.

stlehmann:
You are very welcome. Thanks again for the review and the great mentoring. 🙇

dash-renderer/src/components/core/DocumentTitle.react.js
```diff
@@ -5,14 +5,16 @@ import PropTypes from 'prop-types';
...
+        const {update_title} = props.config;
         this.state = {
             initialTitle: document.title,
+            update_title,
         };
     }
 
     UNSAFE_componentWillReceiveProps(props) {
-        if (props.isLoading) {
-            document.title = 'Updating...';
```
alexcjohnson, line 15: @Marc-Andre-Rivet as we were discussing the other day - the renderer no longer holds this default value at all. That means when the R and Julia implementations update to include this functionality in their renderer, they will *need* to add this option to the back end at the same time or the default behavior will change to no title changes ever.

chriddyp:
_Update to those passing through GitHub: This feature is now documented in http://dash.plotly.com/external-resources. Look for the `title=` and `update_title=`. See the "[Customizing Dash's Document or Browser Tab Title]( http://dash.plotly.com/external-resources)", " [Update the Document Title Dynamically based off of the URL or Tab]( http://dash.plotly.com/external-resources)", and "[Customizing or Removing Dash's "Updating..." Message](http://dash.plotly.com/external-resources)" sections._

PR #1319, opened 2020-07-01
Removing circularity and other fixes to CircleCI build

noisycomputation:
After forking dash and dash-table and trying to get these packages to build using CircleCI, I noticed some strange behavior. It seemed that dash and dash_render needed to be installed before either of these packages could be built, and at the specific version numbers being built.

This PR represents the minimal set of changes to the CircleCI build script that allows dash, dash_render, dash-table, dash-core-components, and dash-html-components to build from scratch, without being available for install from PyPi. It also allows version numbers to be bumped and all packages to be built without loosening and re-tightening version pinning for the various dash packages, which is how these packages are currently built upstream. My intent is to *start* a discussion about improving the build process, something with which I am happy to help.

*Current upstream build process*. The `build-core-37` job requires the versions of `dash` and `dash_renderer` specified in `requires-install.txt` file to be available for install from PyPi. The only reason for these dependencies is that the `renderer` tool is required to install dash_renderer. Strictly speaking, building dash requires dash, dash_render, dash-table, dash-core-components, and dash-html-components to be available at PyPi. What is more, when bumping version numbers, the *new* versions must already be in PyPi.

The current upstream workaround to these problems during version bumps is to relax version pinning in `requires-install.txt` while setting the new version numbers in the various `setup.py` and `package.json` files. This build succeeds in CircleCI. The version pinning in `requires-install.txt` is then re-tightened, which causes the next build to fail since the new versions are not yet available at PyPi. This build error is fixed by pushing a blank commit, which triggers a new CircleCI build which succeeds, but only because CircleCI restores the build cache (i.e. the virtual environment directory `venv`) from the lint stage of the previous build instead of trying to install the necessary dash packages from PyPi. The packages resulting from this build are then presumably uploaded to PyPi.

*What this PR changes*. The core change is that all lines with dash packages in `requires-install.txt` are commented out before any packages are installed at any step; this change is then reversed before saving the build cache. This breaks the dependency on any dash packages in PyPi. To make the `renderer` tool available, `dash_render` has to be built first with a blank \_\_init\_\_.py, since that file is generated by the `renderer` tool itself, and Python v2 will not import the dash_renderer package without an \_\_init\_\_.py file (Python v3 will import is as a namespace package, and this is enough to let `renderer` work). [Here](https://circleci.com/gh/noisycomputationorg/dash) is a link to the CircleCI build page.

I am a little conflicted about relying on `renderer` for the build, at least as implemented, which requires the entire dash and dash_renderer packages to be imported before the tool can be run. Maybe this module can be made independent, or be implemented as an in-tree build backend pursuant to PEP 517.

I am sure the build script can be improved further. The build cache handling comes to mind. But I wanted to see if there was any appetite before spending any more time on it. Also, please note that I haven't done anything with the R build, as that is well outside my wheelhouse.

**Note:** The CircleCI builds are configured to run only on the circleci branch.

## Contributor Checklist

- [x] I have broken down my PR scope into the following TODO tasks (all pertain to .circleci/config.yaml)
   -  [x] Create reusable commands to avoid code repetition
   -  [x] Revise build jobs to avoid necessity of installing dash packages from PyPi
- [x] I have run the tests on [CircleCI](https://circleci.com/gh/noisycomputationorg/dash) and they passed 

### optionals

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,14 +29,16 @@ commands:
     steps:
       - run:
           name: Create virtual environment in ./venv
-          command: python -m venv venv || ( sudo pip install --upgrade virtualenv --progress-bar off && virtualenv venv )
+          command: |
+            # Try Linux Python v3 | Linux Python v2 | Windows Python v3 | Windows Python v2
+            python -m venv venv || ( sudo pip install --upgrade virtualenv --progress-bar off && virtualenv venv ) || ( py -m venv venv ) || ( pip install --upgrade virtualenv --progress-bar off && virtualenv env )
 
   pip_install_all_requires_cmd:
     steps:
       - run:
           name: Install dash and all dependencies in requires*.txt
           command: |
-            source venv/bin/activate || source venv/Scrips/activate
+            source venv/bin/activate || source venv/Scripts/activate
             pip install -e . --no-cache-dir -r requires-install.txt -r requires-dev.txt -r requires-testing.txt --progress-bar off
 
   build_core_dev_requirements_cmd:
@@ -49,7 +51,12 @@ commands:
           command: |
             # dash-renderer needs to be installed in the virtual env in order to run the `renderer` tool
             source venv/bin/activate || source venv/Scripts/activate
-            cd dash-renderer && npm ci && npm run build:js && pip install -e . --no-cache-dir --progress-bar off && cd ..
+            cd dash-renderer && npm ci && npm run build:js
+
+            # Python <3.3 will not import dash_renderer without __init__.py. The `renderer` tool
+            # generates it, but both dash and dash_renderer must be installed for `renderer` tool
+            # to run. For now, create a blank __init__.py, `renderer` will clean and build
+            touch dash_renderer/__init__.py && pip install -e . --no-cache-dir --progress-bar off && cd ..
       - requires_install_restore_dash_cmd
 
   build_dash_renderer_cmd:
@@ -236,11 +243,14 @@ jobs:
       - checkout
       - run: echo $PYVERSION > ver.txt
       - restore_cache_cmd
+      - attach_workspace:
+          at: ~/dash
+      - install_built_packages_cmd
       - requires_install_remove_dash_cmd
       - run:
           name: ️️🏗️ Build misc
           command: |
-              source venv/bin/activate && pip install --no-cache-dir --upgrade -e . --progress-bar off && mkdir packages
+              source venv/bin/activate && pip install --no-cache-dir --upgrade -e . --progress-bar off
 
               git clone --depth 1 https://github.com/noisycomputation/dash-core-components.git
               cd dash-core-components && npm ci && npm run build && python setup.py sdist && mv dist/* ../packages/  && cd ..
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,126 @@
 version: 2.1
+
+commands:
+  save_cache_cmd:
+    steps:
+      - save_cache:
+          key: dep-{{ checksum ".circleci/config.yml" }}-{{ checksum "ver.txt" }}-{{ checksum "requires-dev.txt" }}-{{ checksum "requires-install.txt" }}-{{ checksum "requires-testing.txt" }}
+          paths:
+              - venv
+
+  restore_cache_cmd:
+    steps:
+      - restore_cache:
+          key: dep-{{ checksum ".circleci/config.yml" }}-{{ checksum "ver.txt" }}-{{ checksum "requires-dev.txt" }}-{{ checksum "requires-install.txt" }}-{{ checksum "requires-testing.txt" }}
+
+  requires_install_remove_dash_cmd:
+    steps:
+      - run:
+          name: Remove dash from requires-install.txt
+          command: sed -i 's/^dash/#dash/' requires-install.txt
+
+  requires_install_restore_dash_cmd:
+    steps:
+      - run:
+          name: Restore dash to requires-install.txt
+          command: sed -i 's/^#dash/dash/' requires-install.txt
+
+  create_venv_cmd:
+    steps:
+      - run:
+          name: Create virtual environment in ./venv
+          command: python -m venv venv || ( sudo pip install --upgrade virtualenv --progress-bar off && virtualenv venv )
+
+  pip_install_all_requires_cmd:
+    steps:
+      - run:
+          name: Install dash and all dependencies in requires*.txt
+          command: |
+            source venv/bin/activate || source venv/Scrips/activate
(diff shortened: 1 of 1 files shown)


alexcjohnson:
Thanks @noisycomputation - there definitely are some quirks to this process today, I’m curious to hear what @Marc-Andre-Rivet thinks about your changes when he gets back online next week.

Marc-Andre-Rivet:
@noisycomputation Thanks for the suggestions. I think I generally agree with the spirit of the PR but want to clarify some points.

> dash and dash_render needed to be installed before either of these packages could be built

You're right. This is a known issue / dependency between the `dash-component-generator` and dash and dash-renderer. This dependency is artificial and we've been discussing removing it for a while now. 

> that allows dash, dash_render, dash-table, dash-core-components, and dash-html-components to build from scratch, without being available for install from PyPi

During normal development we do not typically bump the version of the packages. We simply tag all changes as unreleased in the changelog and continue to use the previous release version. We then use `-b <branch>` for the proposed changes to component repos as required (most of the time it's not). This has been sufficient for our development testing needs and we haven't really been running into misaligned versions issues except at release time.

This brings us to the relaxing/tightening of version requirements. You are right here too. This is tedious and annoying. There's still many manual steps in releasing Dash at the moment and those are error prone. So while it is extremely cumbersome, these checks are actually voluntary at this time. Removing the entries from `require-install.txt` makes the CI slightly more different from an actual release and has been a source of errors in the past.

---

Again, I really like here is the generalization and extraction of the common smaller tasks and my only worry is due to our current release procedure. With that in mind, I'd like to suggest:

- Making the [install](https://github.com/plotly/dash/pull/1319/files#diff-1d37e48f9ceff6d8030570cd36286a61R16) and [restore](https://github.com/plotly/dash/pull/1319/files#diff-1d37e48f9ceff6d8030570cd36286a61R22) steps conditional - the file will be edited `if` the branch is neither `dev` or `master` -- this will help improve the general development experience but still allow for the sanity check we want to keep for now on those branches
- Removing the entries specific to the `circleci` [branch](https://github.com/plotly/dash/pull/1319/files#diff-1d37e48f9ceff6d8030570cd36286a61R426-R428)

noisycomputation:
> * Making the [install](https://github.com/plotly/dash/pull/1319/files#diff-1d37e48f9ceff6d8030570cd36286a61R16) and [restore](https://github.com/plotly/dash/pull/1319/files#diff-1d37e48f9ceff6d8030570cd36286a61R22) steps conditional - the file will be edited `if` the branch is neither `dev` or `master` -- this will help improve the general development experience but still allow for the sanity check we want to keep for now on those branches

If we omit the step of commenting out the dash packages in `requires-install.txt` before running `pip install -e .` from the root `dash` directory, then `pip` will attempt to pull those in from PyPi if not already installed, which will cause the circularity / inability to bootstrap issue (including the version loosening/tightening dance during a release).

My intent was to remove installing pre-built dash packages (from PyPi or otherwise) as a prerequisite for building dash. This allows the build system itself to be tested during the build. More so than installing the build system from a previous build, I would think.

Once the packages are built (in the `build-core` and `build-misc` jobs), the freshly built packages are installed, followed by a `pip -e .` against the original, unaltered `requires-install.txt`, and then the `lint` and `test` jobs are run.

I suppose that this leaves open the possibility that an older `renderer` version could fail to build the current code, but is this really a concern? Local build instructions could specify that, just as in the official CI build, `dash` and `dash_renderer` must be installed first from source to make the `renderer` tool available before `dash` can be built as a source distribution package. While it's messy to require a user to follow the CI workflow in order to build `dash`, it's really the only way to guarantee the same result as the CI build.

**Frankly, maybe I am focusing on the wrong thing.** The only reason for the circularity/bootstrapping issue is the `renderer` tool, which is an entry point into `dash/development/build_process.py`. It goes like this:
 * `dash_renderer` requires `renderer` to build it
 * `renderer` is an entry point defined by the `dash` `setup.py`, which points to `dash/development/build_process.py`.
 * To use this `renderer` entry point, `dash` must be installed. Let's comment out the `dash` dependencies in `requires-install.txt` because we don't want to install any pre-existing `dash` packages from PyPi. `dash` installs fine. But running `renderer` fails because doing so imports `dash`, which in turn imports `dash_renderer`. But the whole point of this exercise is that we want to build `dash_renderer`, it can't be imported because it hasn't been built yet.

OK, so let's just run `build_process.py` directly. It imports three functions from `dash/_utils.py`, but these functions are not used elsewhere in the codebase, so we can just place them in `build_process.py` and add a few imports (hashlib, io, shlex, subprocess, functools.wraps). This works when running from `dash/development/` due to some hard-coded relative paths in Renderer.\_\_init\_\_(), but that's a simple enough fix. Then maybe we place `build_process.py` in a new root directory called `tools`, call it `build_renderer.py` and change the CI *and* `CONTRIBUTING.md` instructions to run this in lieu of `renderer`. 

Or something like that. It's a bit clumsy any way you slice it due to having both Python and Node components to the build, but doing it this way would not make `.circleci/config.yml` or the `CONTRIBUTING.md` instructions any more complicated, and would solve the circularity issue, that is permit the project to bootstrap. The build would go something like this:
 * Use the `renderer` tool to build the Node side of `dash_renderer`
 * Build the `dash_renderer` sdist
 * Build the `dash` sdist
 * Clone and build dash-core-components, dash-table, and dasah-html-components
 * Install the packages (dash last, since it depends on the others)
 * Done

This would avoid having to mess with the `requires-install.txt` file, since the individual packages would be installed before `dash`, so `pip` would not attempt to fetch from PyPi. During releases, the new versions could appear in `requires-install.txt` from the start, because this file would only be referenced when finally installing `dash`, after the new versions of all the dependencies were already installed. If linting and testing succeeds with the just-built packages, the packages could be uploaded to PyPi.

Thoughts?

> * Removing the entries specific to the `circleci` [branch](https://github.com/plotly/dash/pull/1319/files#diff-1d37e48f9ceff6d8030570cd36286a61R426-R428)

Yes, this is only there as a placeholder and will be removed when this is ready to be merged. Also, I noticed that I linked to the wrong CircleCI build in my initial PR description. I ran out of build credits on my individual account and had to set up an org account to take advantage of their open source offer of extra credits, the correct build is at https://app.circleci.com/pipelines/github/noisycomputationorg (but I suppose you have to either take it on faith or verify that the two forks are identical). Would you prefer that I open a new PR from that fork and account now, or can we continue the discussion here and only switch to a new PR once we have agreed on an approach?

I do like keeping the individual `noisycomputation` builds separate from the `-org` builds. The former is for my customized version of dash (other than the occasional bug fix, my only real motivation is to expose the active cell highlight color in `dash-table` as its own CSS property so I can change it or disable it via CSS), the latter is for working on the build system for the upstream version.

gvwilson:

gvwilson:
@noisycomputation thank you again for this, but we feel it has gone stale (our fault) and we're hoping to do a major overhaul of our entire build process later this year or early 2025.

PR #1322, opened 2020-07-06
Add a CHANGELOG entry for PR #1320

jbampton:
refs #1320


alexcjohnson:
Thanks @jbampton - we appreciate the fixes, but I don't think a change like this needs a changelog entry, as it doesn't affect any features or functionality, just the documentation.

PR #1332, opened 2020-07-09
Removing side effect

qlieumontadv:
Just removing the side effects on the logging package.

## Contributor Checklist

   -  [x] task 1: use % instead of letting (maybe) logging do it.

### optionals

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1607,7 +1607,7 @@ class Dash(object):
             else:
                 display_url = (protocol, host, ":{}".format(port), path)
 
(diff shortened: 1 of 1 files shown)


alexcjohnson:
@qlieumontadv apologies for taking a long time to look at this - can you say more what you mean by side effect? The standard `pylint` config expressly requires the existing form over explicit %-formatting.

qlieumontadv:
Maybe I doesn't do something right but I have some errors about an invalid formatter, pycharm point at this line, this change solve the issue...
But if pylint mark it wrong, just change it on my computer  for now...

PR #1343, opened 2020-07-23
don't override document.title if set to None

chriddyp:
As requested by a [Dash Enterprise](https://plotly.com/dash) customer.

This allows users to override `document.title` during runtime with a
clientside callback or a special component
(https://github.com/plotly/dash-core-components/issues/833)

Example of modifying during runtime:
```python
import dash

app = dash.Dash(__name__, update_title=None)


app.layout = app.layout = html.Div([
    dcc.Input(id='input'),
    html.Div(id='output'),

    html.Div(id='dummy'),
    dcc.Tabs(id='tabs-example', value='tab-1', children=[
        dcc.Tab(label='Tab one', value='tab-1'),
        dcc.Tab(label='Tab two', value='tab-2'),
    ]),
    html.Div(id='tabs-example-content')
])


app.clientside_callback(
    """
    function(tab_value) {
        console.log(tab_value);
        document.title = tab_value;
        return null; // dummy output
    }
    """,
    Output('dummy', 'children'),
    [Input('tabs-example', 'value')]
)


@app.callback(Output('output', 'children'), [Input('input', 'value')])
def update_output(value):
    return value

if __name__ == '__main__':
    app.run_server(debug=True)
```

Example with custom table title via index string:
```python
import dash

app = dash.Dash(__name__, update_title=None)

app.index_string = '''
<!DOCTYPE html>
<html>
    <head>
        {%metas%}
        <title>Hello World</title>
        {%favicon%}
        {%css%}
    </head>
    <body>
        <div>My Custom header</div>
        {%app_entry%}
        <footer>
            {%config%}
            {%scripts%}
            {%renderer%}
        </footer>
        <div>My Custom footer</div>
    </body>
</html>
'''


app.layout = app.layout = html.Div([
    dcc.Input(id='input'),
    html.Div(id='output'),

    html.Div(id='dummy'),
    dcc.Tabs(id='tabs-example', value='tab-1', children=[
        dcc.Tab(label='Tab one', value='tab-1'),
        dcc.Tab(label='Tab two', value='tab-2'),
    ]),
    html.Div(id='tabs-example-content')
])


@app.callback(Output('output', 'children'), [Input('input', 'value')])
def update_output(value):
    return value

if __name__ == '__main__':
    app.run_server(debug=True)
```

 This is a follow up of https://github.com/plotly/dash/pull/1315/.

### DIFF
--- a/dash-renderer/src/components/core/DocumentTitle.react.js
+++ b/dash-renderer/src/components/core/DocumentTitle.react.js
@@ -7,13 +7,19 @@ class DocumentTitle extends Component {
         super(props);
         const {update_title} = props.config;
         this.state = {
+            title: document.title,
             update_title,
         };
     }
 
     UNSAFE_componentWillReceiveProps(props) {
-        if (this.state.update_title && props.isLoading) {
-            document.title = this.state.update_title;
+        if (props.isLoading) {
+            this.setState({title: document.title});
+            if (this.state.update_title) {
+                document.title = this.state.update_title;
+            }
+        } else {
+            document.title = this.state.title;
         }
     }
 
--- a/dash-renderer/src/components/core/DocumentTitle.react.js
+++ b/dash-renderer/src/components/core/DocumentTitle.react.js
@@ -7,7 +7,6 @@ class DocumentTitle extends Component {
         super(props);
         const {update_title} = props.config;
         this.state = {
-            initialTitle: document.title,
             update_title,
         };
     }
@@ -15,8 +14,6 @@ class DocumentTitle extends Component {
     UNSAFE_componentWillReceiveProps(props) {
         if (this.state.update_title && props.isLoading) {
             document.title = this.state.update_title;
-        } else {
-            document.title = this.state.initialTitle;
         }
     }
 
--- a/dash-renderer/src/components/core/DocumentTitle.react.js
+++ b/dash-renderer/src/components/core/DocumentTitle.react.js
@@ -13,13 +13,21 @@ class DocumentTitle extends Component {
     }
 
     UNSAFE_componentWillReceiveProps(props) {
+        if (!this.state.update_title) {
+            // Let callbacks or other components have full control over title
+            return;
+        }
         if (props.isLoading) {
             this.setState({title: document.title});
             if (this.state.update_title) {
                 document.title = this.state.update_title;
             }
         } else {
-            document.title = this.state.title;
+            if (document.title === this.state.update_title) {
+                document.title = this.state.title;
+            } else {
+                this.setState({title: document.title});
+            }
         }
     }
 
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -227,6 +227,15 @@ class Dash(object):
         with a ``plug`` method, taking a single argument: this app, which will
         be called after the Flask server is attached.
     :type plugins: list of objects
+
+    :param title: Default ``Dash``. Configures the document.title
+    (the text that appears in a browser tab).
+
+    :param update_title: Default ``Updating...``. Configures the document.title
+    (the text that appears in a browser tab) text when a callback is being run.
+    Set to None or '' if you don't want the document.title to change or if you
+    want to control the document.title through a separate component or
+    clientside callback.
     """
 
     def __init__(
@@ -252,6 +261,7 @@ class Dash(object):
         prevent_initial_callbacks=False,
         show_undo_redo=False,
         plugins=None,
+        title='Dash',
         update_title="Updating...",
         **obsolete
     ):
@@ -320,6 +330,9 @@ class Dash(object):
             "Invalid config key. Some settings are only available "
             "via the Dash constructor"
         )
+
+        # keep title as a class property for backwards compatability
+        self.title = title
 
         # list of dependencies - this one is used by the back end for dispatching
         self.callback_map = {}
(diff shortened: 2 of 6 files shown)


dash-renderer/src/components/core/DocumentTitle.react.js
```diff
@@ -7,16 +7,13 @@ class DocumentTitle extends Component {
...
         this.state = {
-            initialTitle: document.title,
             update_title,
         };
     }
 
     UNSAFE_componentWillReceiveProps(props) {
         if (this.state.update_title && props.isLoading) {
             document.title = this.state.update_title;
-        } else {
```
Marc-Andre-Rivet, line 18: While this allows the clientside callback to set the value, it seems to break existing tests / behavior. Would a better verification be to online revert back to `inititalTitle` if the value is `update_title` instead?

chriddyp:
Huh, odd that pylint would fail. I didn't change any python files.
image

dash-renderer/src/components/core/DocumentTitle.react.js
chriddyp: Thanks! I totally misinterpreted what `update_title` meant. I believe https://github.com/plotly/dash/pull/1343/commits/e190cc941b89275677f0b3b995d21d5220e0209b is the logic we're looking for

alexcjohnson: I think it'd be safer to bail out entirely (before even the `props.isLoading` test) if `!update_title`. Then this entire component would be a noop, as it should be in that case; callbacks can do whatever they want with `document.title` and we won't interfere.

When you _do_ have an `update_title`, I may agree with @Marc-Andre-Rivet that there needs to be a test when we're reverting back to the non-updating title. Consider a callback chain, where the first callback changes `document.title`, but we're still in a loading state so `DocumentTitle` doesn't receive new props and we don't pick up the changed title into `this.state`. Then after the second callback we reset `document.title` to `this.state.title`, losing the callback-provided title.

Seems to me the logic you have right now is fine during loading, but when reverting to `this.state.title` below it would be better to do something like:
```js
if (document.title === this.state.update_title) {
    document.title = this.state.title;
} else {
    this.setState({title: document.title});
}
```

alexcjohnson:
Tangentially related to this PR - you can set `app.title` to control the regular page title (which I totally forgot about when reviewing #1315). We use this internally, but I don't see it documented anywhere, and it feels inconsistent with the `update_title` kwarg. Perhaps we should support `title` as a kwarg too? Also `app.config.update_title` works, so it would be nice if `app.config.title` did too. We could go full symmetry and allow `app.update_title`, but it might be cleaner to deprecate `app.title` and encourage both of these to go through kwargs or `app.config`.

alexcjohnson:
> Huh, odd that pylint would fail.

weird, perhaps we somehow got `isort` v5? Maybe we need to pin it to `4.3.21` in `requires_dev.txt`? `pylint` tries to handle that itself, but perhaps someone else brings it in unpinned first?

dash-renderer/src/components/core/DocumentTitle.react.js
chriddyp: Thanks @alexcjohnson @Marc-Andre-Rivet ! I believe I have addressed your comments and added the appropriate tests now.

chriddyp:
> Perhaps we should support title as a kwarg too? 

I added this as well and kept the `app.title` for backwards compatibility. Also added a unit test here.

chriddyp:
Huh, seeing these re-introduced snapshots on Percy

image

dash/dash.py
```diff
@@ -725,7 +739,9 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
         config = self._generate_config_html()
         metas = self._generate_meta_html()
         renderer = self._generate_renderer()
-        title = getattr(self, "title", "Dash")
+
+        # use self.title instead of app.config.title for backwards compatibility
```
alexcjohnson, line 743: In principle we could make title into a `@property` that syncs `app.title` with `app.config.title` - but not a big deal. Most people should just use the kwarg regardless.

alexcjohnson:
Dunno what the deal is with the "reintroduced" percy snapshots - but I'd just ignore it this time and look into it if we see that again.

OverLordGoldDragon:
Isn't below simpler? 

```python
def interpolate_index(self, **kwargs):
    kwargs['title'] = 'Custom Title'
    return dash.Dash.interpolate_index(self, **kwargs)
app.interpolate_index = interpolate_index.__get__(app)
```

alexcjohnson:
I'm not sure what you mean @OverLordGoldDragon - simpler than `app = dash.Dash(title='Custom Title')`?

OverLordGoldDragon:
@alexcjohnson Surely not, but your code blob isn't shown in this PR nor in Releases (it was unclear where the "`title` parameter" should go) - thanks for clarifying.

alexcjohnson:
OIC - yeah, this PR expanded a bunch from its original purpose, so we never showed the simple examples - sorry for the confusion. We've documented the new behavior at https://dash.plotly.com/external-resources (though @chriddyp it feels to me a bit hidden on that page - maybe time for a dedicated `dash.Dash` class API reference page?)
As we've discussed, this is a big argument in favor of merging the docs into this repo - then https://github.com/plotly/dash-docs/pull/931 would have been part of this PR.

hivemall:
VERY much appreciated feture + doc well, awesome

PR #1355, opened 2020-08-04
Move Dash logger.setLevel to init + remove now redundant startup message

### LINKED ISSUE #1354: [Feature Request] Possibility to configure Dash logger level

These lines, https://github.com/plotly/dash/blob/7b189ad7d66b2979d8c9cf9d7338763c3b44c6fe/dash/dash.py#L1626-L1632 introduced in #1289, gives by default the following output to the terminal
```
INFO:     Dash is running on http://127.0.0.1:8050
INFO:     Warning: This is a development server. Do not use app.run_server
INFO:     in production, use a production WSGI server like gunicorn instead.
```
which could be useful for Python developers that at some point want to take their app from `localhost` to production/shared server.

However, for other use cases, e.g. when using Dash for building command-line programs with end-user consumption and visualization on `localhost` only, the messages could for some users be confusing (or even scary 😱🙂), since they are non-programmers and see `Warning`, `Do not` together with unknown terminology.

The current log level is set here https://github.com/plotly/dash/blob/7b189ad7d66b2979d8c9cf9d7338763c3b44c6fe/dash/dash.py#L1354 in the function https://github.com/plotly/dash/blob/7b189ad7d66b2979d8c9cf9d7338763c3b44c6fe/dash/dash.py#L1251 which is called regardless of `debug` value in `app.run_server(debug=False/True)`.

An easy fix could probably simply be to move https://github.com/plotly/dash/blob/7b189ad7d66b2979d8c9cf9d7338763c3b44c6fe/dash/dash.py#L1354 to where the `logger` is initialized in the first place, https://github.com/plotly/dash/blob/7b189ad7d66b2979d8c9cf9d7338763c3b44c6fe/dash/dash.py#L378-L379 since `INFO` is the final level regardless of `debug` value. That would enable Python programs using `dash` to use something like
```
app = dash.Dash()
app.logger.setLevel(logging.WARNING)
```
to set appropriate log level before calling `app.run_server()`.

Could also add a new argument to `app.run_server` or `dash.Dash.__init__()` with wanted Dash log level.

**Additional context**

There is debug log potentially printed here https://github.com/plotly/dash/blob/7b189ad7d66b2979d8c9cf9d7338763c3b44c6fe/dash/dash.py#L707-L713 which would then be called during `__init__` through `init_app()`: https://github.com/plotly/dash/blob/7b189ad7d66b2979d8c9cf9d7338763c3b44c6fe/dash/dash.py#L378-L386 Is this relying on the user to set root log level to `<= DEBUG` before running Dash to get these logs? If so, the change mentioned above could still support this simply by moving `setLevel` to after `init_app`:
```python
        self.logger = logging.getLogger(name)
        self.logger.addHandler(logging.StreamHandler(stream=sys.stdout))

        if isinstance(plugins, patch_collections_abc("Iterable")):
            for plugin in plugins:
                plugin.plug(self)

        if self.server is not None:
            self.init_app()

```

anders-kiaer:
Resolves #1354.

## Contributor Checklist

   -  [X] Move `setLevel` to `__init__`
   -  [X] Remove two now redundant messages.

### Optionals

- [X] I have added entry in the `CHANGELOG.md` ~_(not added, this PR probably is not regarded as a "notable change", let me know if an entry should be added)_~.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [UNRELEASED]
+### Added
+- [#1355](https://github.com/plotly/dash/pull/1355) Removed redundant log message and consolidated logger initialization. You can now control the log level - for example suppress informational messages from Dash with `app.logger.setLevel(logging.WARNING)`.
+
 ## [1.14.0] - 2020-07-27
 ### Added
 - [#1343](https://github.com/plotly/dash/pull/1343) Add `title` parameter to set the
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -385,6 +385,8 @@ class Dash(object):
         if self.server is not None:
             self.init_app()
 
+        self.logger.setLevel(logging.INFO)
+
     def init_app(self, app=None):
         """Initialize the parts of Dash that require a flask app."""
         config = self.config
@@ -1351,8 +1353,6 @@ class Dash(object):
         if dev_tools.silence_routes_logging:
             logging.getLogger("werkzeug").setLevel(logging.ERROR)
 
-        self.logger.setLevel(logging.INFO)
-
         if dev_tools.hot_reload:
             _reload = self._hot_reload
             _reload.hash = generate_hash()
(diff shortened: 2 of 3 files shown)


alexcjohnson:
Nice! Actually I do think this warrants a changelog entry since it enables users to do something new. Something like:

> Removed redundant log message and consolidated logger initialization. You can now control the log level - for example suppress informational messages from Dash with `app.logger.setLevel(logging.WARNING)`.

(I don't think it's even necessary to say when you can use `setLevel`, since it's anytime after you have an app object and before you run the server)

anders-kiaer:
:+1: Agree - new changelog entry added :scroll: :bookmark:

PR #1371, opened 2020-08-18
Add support for calculating CSP hashes of inline scripts

### LINKED ISSUE #630: DashRenderer initialization and CSP
Dash is a [CSP friendly](https://en.wikipedia.org/wiki/Content_Security_Policy) framework, enabling writing XSS safe applications, however this had a setback in #367.

With `pip install 'dash==0.38.0' dash_html_components flask-talisman` you could test quite strict CSP settings in Dash locally with success:
```python
import dash
from flask_talisman import Talisman

app = dash.Dash(__name__)

csp = {
       'default-src': '\'self\'',
       'script-src': '\'self\'',
       'style-src': '\'self\''
      }

Talisman(app.server, content_security_policy=csp, force_https=False)
app.layout = html.Div(children=['Hello Dash!'])

if __name__ == '__main__':
    app.run_server()
```

With `dash==0.39.0` however this fails due to the new inline script https://github.com/plotly/dash/blob/05545461214e43707e7d4e36485935dc166be987/dash/dash.py#L174

A work-around could be to add hash of the current Dash generated inline script (`sha256-jZlsGVOhUAIcH+4PVs7QuGZkthRMgvT2n0ilH6/zTM0=`) to the server CSP script src headers, however that would be in need of update when the dash renderer string/configuration changes.

Not sure if it is feasible/fits the overall framework/plans, but could one suggestion be to follow the same concept as when the user e.g. wants to override the default `favicon.ico` (i.e. a file `favicon.ico` is placed in the `assets` folder - here in this case it could be e.g. a file `dash-renderer-config.js`). In addition to following the same concept in terms of overriding default assets and continue not having inline scripts in Dash core, it perhaps also better facilitates separation of Python- and JavaScript code.

anders-kiaer:
Closes #630. 

With the tests in this PR, we also increase the probability of the core `dash` framework remaining [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) friendly. 🔒👮 

---

## Contributor Checklist

   -  [X] Added a new `dash.Dash` method which calculates hashes of inline scripts for the [`script-src` CSP directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src).

### optionals


dash/dash.py
```diff
@@ -1128,6 +1130,33 @@ def _serve_default_favicon():
             pkgutil.get_data("dash", "favicon.ico"), content_type="image/x-icon"
         )
 
+    def csp_hashes_inline_scripts(self, hash_algorithm="sha256"):
+        """Calculates CSP hashes (sha + base64) of all inline scripts, such that
+        one of the biggest benefits of CSP (disallowing general inline scripts)
+        can be utilized together with Dash clientside callbacks (inline scripts).
```
alexcjohnson, line 1136: @anders-kiaer this looks great, and fantastic tests - I love `does_not_raise`, hadn't seen that trick before but I can see it coming in handy elsewhere.

The name is a little long, would it be ambiguous if we just call it `csp_hashes` or `script_hashes`?

Can you add a bit of usage example into the docstring here? Something like:
```
Calculate these hashes after all callbacks are defined, and add them to your CSP headers
immediately before starting the server, for example:

flask_talisman.Talisman(app.server, content_security_policy={
    "default-src": "'self'",
    "script-src": ["'self'"] + app.csp_hashes_inline_scripts()
})
```

Am I understanding correctly that this is what you'd normally do?

I haven't used `flask_talisman` before, but I presume that it works just as well (and with no changes to your code) with `gunicorn` as it does with `run_server`?

anders-kiaer: > The name is a little long, would it be ambiguous if we just call it `csp_hashes` or `script_hashes`?

Agree, it is currently a bit long. I think both of the two alternatives should be ok. If there is any ambiguity, that probably is:

- `csp_hashes`: CSP hashes are, generally, used both for inline `styles` and inline `scripts`. So from the name alone it is not clear if it is script hashes, style hashes or both. 

  However, in Dash framework context (since dynamic style attributes from Python side are set in a CSP compatible way through React, see https://github.com/plotly/dash-core-components/pull/753#issuecomment-583737160), only `script` hashes should be necessary, which reduces the risk of the ambiguity. There are of course Dash components out there that needs to do changes similar to https://github.com/plotly/dash-core-components/pull/753, but that is on the different custom/standard Dash components to solve/fix in their `webpack` build setup, not something concerning the Dash framework.
- `script_hashes`: Solves the style vs script ambiguity, but it does not indicate that the hashes are in "CSP format" and ready to go (`'{hash algorithm}-{base 64 hash}'`, including the `''`).

I like `csp_hashes` best of the two 👍 If there appears e.g. a `style` hash use case later, for some reason, this can be solved with a natural non-breaking extension with proper default value, e.g. `app.csp_hashes(hash_algorithm, source="scripts")`.

> Am I understanding correctly that this is what you'd normally do?
> 
> I haven't used `flask_talisman` before, but I presume that it works just as well (and with no changes to your code) with `gunicorn` as it does with `run_server`?

Yes and yes 🙂 `gunicorn` passes by default the (CSP) headers through. E.g. with that example CSP configuration in the suggested docstring, Dash users will get 🔒 `A+` rating out of the box on [Mozilla observatory](https://observatory.mozilla.org/).

tests/integration/test_csp.py
```diff
@@ -0,0 +1,59 @@
...
+    "add_hashes, hash_algorithm, expectation",
+    [
+        (False, None, pytest.raises(NoSuchElementException)),
+        (True, "sha256", does_not_raise()),
+        (True, "sha384", does_not_raise()),
+        (True, "sha512", does_not_raise()),
+        (True, "sha999", pytest.raises(ValueError)),
+    ],
+)
+def test_csp_hashes_inline_scripts(dash_duo, add_hashes, hash_algorithm, expectation):
```
alexcjohnson, line 28:
```suggestion
def test_incs001_csp_hashes_inline_scripts(dash_duo, add_hashes, hash_algorithm, expectation):
```

Just adding a test case ID for local targeting :)

pascalgulikers:
Too bad inline style is still not CSP compatible:
```
app = dash.Dash(
    __name__,
    prevent_initial_callbacks=True,
)

server = app.server

talisman = Talisman(server, content_security_policy={
    "default-src": "'self'",
    "script-src": ["'self'"] + app.csp_hashes(),
    'style-src': ["'self'"],
})

app.layout = html.Div([
    html.H1("Hello, Plotly Dash with Talisman and Gunicorn!")
])
```
afbeelding (1)

PR #1375, opened 2020-08-20
support named callback args not wrapped in a list

### LINKED ISSUE #1366: [BUG] Regression on callbacks using name argument `output=Output(...)`
Caused by https://github.com/plotly/dash/pull/1180

Example DCC failing test: https://github.com/plotly/dash-core-components/blob/dev/tests/test_integration_1.py#L391

Removing (output=) makes it work:
```
Output(component_id='test-pathname', component_property='children'),
```

The problematic line seems to be https://github.com/plotly/dash/blob/dev/dash/dependencies.py#L163

Updating to 👇 makes the test pass:
```
if isinstance(outputs, (list, tuple)) and len(outputs) == 1:
```

Expected result:
- DCC tests pass
- Added test in Dash for named arguments

@alexcjohnson

alexcjohnson:
Fixes #1366 
I opted to allow any of the callback args to be provided as a list or not, rather than a narrower fix of just the `output=Output(...)` regression. My rationale is it's unambiguous what the user means, so there's no harm accepting it.

### DIFF
--- a/tests/integration/callbacks/test_validation.py
+++ b/tests/integration/callbacks/test_validation.py
@@ -144,3 +144,52 @@ def test_cbva003_list_single_output(dash_duo):
     dash_duo.start_server(app)
     dash_duo.wait_for_text_to_equal("#out1", "1: Hi")
     dash_duo.wait_for_text_to_equal("#out2", "2: Hi")
+
+
+@pytest.mark.parametrize("named_out", [True, False])
+@pytest.mark.parametrize("named_in", [True, False])
+@pytest.mark.parametrize("named_state", [True, False])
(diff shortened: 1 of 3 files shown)


tests/integration/callbacks/test_validation.py
```diff
@@ -144,3 +144,52 @@ def o2(i):
...
+        for ai, name, flag in zip(a, names, flags):
+            if flag:
+                kwargs[name] = ai
+            else:
+                args.append(ai)
+        return args, kwargs
+
+    args, kwargs = make_args(
+        Output("out1", "children"), Input("in", "children"), State("state", "children")
+    )
```
Marc-Andre-Rivet, line 177: 😀

PR #1376, opened 2020-08-21
PersistenceTransforms for date in datePickerRange and datePickerSingle

harryturr:
Fixes https://github.com/plotly/dash-core-components/issues/700. 

The persistence logic in DatePickerRange and DatePickerSingle have been modified to extract/apply just the `date` portion of the `date-time`. This will provide stability to date pickers (eg defined in callbacks) for the entire day they are defined.

The date logic is handled in https://github.com/plotly/dash-core-components/pull/848 by using the class property `PersistenceTransforms`, but `persistence.js` has to be modified to check for transformation on both `propPart` and `propName`.

## Contributor Checklist

   -  [x] Have renderer check for nested props in `PersistenceTransforms`
   -  [x] Strip time in datepickers w/ `PersistenceTransforms` https://github.com/plotly/dash-core-components/pull/848/files


.circleci/config.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           command: |
               . venv/bin/activate && pip install --no-cache-dir --upgrade -e . --progress-bar off && mkdir packages
               cd dash-renderer && renderer build && python setup.py sdist && mv dist/* ../packages/ && cd ..
-              git clone --depth 1 https://github.com/plotly/dash-core-components.git
+              git clone -b persistence-hg --depth 1 https://github.com/plotly/dash-core-components.git
```
harryturr, line 107: For testing with https://github.com/plotly/dash-core-components/pull/848.

dash-renderer/src/persistence.js
```diff
@@ -265,10 +265,22 @@ const noopTransform = {
     apply: (storedValue, _propValue) => storedValue,
 };
 
-const getTransform = (element, propName, propPart) =>
-    propPart
-        ? element.persistenceTransforms[propName][propPart]
-        : noopTransform;
```
harryturr, line 271: Old logic checks for just `propPart` for `persistenceTransforms`, but here we add logic to check if either `propName` or `propPart` is the persisted prop.

tests/integration/renderer/test_persistence.py
```diff
@@ -526,3 +532,189 @@ def set_out(val):
         dash_duo.wait_for_text_to_equal(".out", "")
 
         dash_duo.find_element("#btn").click()
+
+
```
harryturr, line 536: Test persistence for DatePickerSingle and DatePicker range in callbacks, in conjunction with https://github.com/plotly/dash-core-components/pull/848.

.circleci/config.yml
```diff
@@ -205,7 +205,7 @@ jobs:
           command: |
               source venv/Scripts/activate && pip install --no-cache-dir --upgrade -e . --progress-bar off && mkdir packages
               cd dash-renderer && renderer build && python setup.py sdist && mv dist/* ../packages/ && cd ..
-              git clone --depth 1 https://github.com/plotly/dash-core-components.git
+              git clone -b persistence-hg --depth 1 https://github.com/plotly/dash-core-components.git
```
Marc-Andre-Rivet, line 208: Should be updated to point to the new branch?
```suggestion
              git clone -b hg-700-persistence --depth 1 https://github.com/plotly/dash-core-components.git
```

.circleci/config.yml
```diff
@@ -234,7 +234,7 @@ jobs:
             pip install --no-cache-dir --upgrade -e .[dev,testing] --progress-bar off
             git clone --depth 1 https://github.com/plotly/dashR.git -b dev dashR
             git clone --depth 1 https://github.com/plotly/dash-html-components.git
-            git clone --depth 1 https://github.com/plotly/dash-core-components.git
+            git clone -b persistence-hg --depth 1 https://github.com/plotly/dash-core-components.git
```
Marc-Andre-Rivet, line 237: Should be updated to point to the new branch?
```suggestion
            git clone -b hg-700-persistence--depth 1 https://github.com/plotly/dash-core-components.git
```

dash-renderer/src/persistence.js
```diff
@@ -265,10 +265,22 @@ const noopTransform = {
...
+const getTransform = (element, propName, propPart) => {
+    if (typeof element.persistenceTransforms !== 'undefined') {
```
Marc-Andre-Rivet, line 269: Here you can simplify the top level check to:
```suggestion
    if (element.persistenceTransforms) {
```
any truth-y value will pass the test (https://dorey.github.io/JavaScript-Equality-Table/)

dash-renderer/src/persistence.js
```diff
@@ -265,10 +265,22 @@ const noopTransform = {
...
+        if (
+            Object.getOwnPropertyNames(element.persistenceTransforms).includes(
```
Marc-Andre-Rivet, line 271: Can be simplified to `if(element.persistenceTransforms[propName])` in this case

tests/integration/renderer/test_persistence.py
```diff
@@ -526,3 +532,189 @@ def set_out(val):
...
+        return dcc.DatePickerSingle(
+            id="dps2",
+            date=datetime.today(),
+            persistence=True,
+            persistence_type="session",
+        )
+
+    @app.callback(Output("dps1-p", "children"), [Input("dps1", "date")])
+    def display_dps1(value):
+        print(value)
```
Marc-Andre-Rivet, line 569: 🔪

tests/integration/renderer/test_persistence.py
```diff
@@ -526,3 +532,189 @@ def set_out(val):
...
+            id="dpr2",
+            start_date=datetime.today() - td(days=3),
+            end_date=datetime.today(),
+            persistence=True,
+            persistence_type="session",
+        )
+
+    @app.callback(Output("dpr1-p-start", "children"), [Input("dpr1", "start_date")])
+    def display_dps1(value):
+        print(value)
```
Marc-Andre-Rivet, line 654: 🔪

tests/integration/renderer/test_persistence.py
```diff
@@ -526,3 +532,189 @@ def set_out(val):
...
+        return value
+
+    @app.callback(Output("dpr1-p-end", "children"), [Input("dpr1", "end_date")])
+    def display_dps1(value):
+        print(value)
```
Marc-Andre-Rivet, line 659: 🔪

tests/integration/renderer/test_persistence.py
```diff
@@ -526,3 +532,189 @@ def set_out(val):
...
+        .send_keys(Keys.HOME)
+        .key_up(Keys.SHIFT)
+        .send_keys("01/01/2020")
+        .pause(0.2)
+        # .send_keys("01/02/2020")
+        .send_keys(Keys.END)
+        .key_down(Keys.SHIFT)
+        .send_keys(Keys.HOME)
+        .key_up(Keys.SHIFT)
+        .send_keys("01/02/2020")
```
Marc-Andre-Rivet, line 709: DRY 🌴🐫

tests/integration/renderer/test_persistence.py
```diff
@@ -526,3 +532,189 @@ def set_out(val):
...
+        .move_to_element(dps2)
+        .pause(0.2)
+        .click(dps2)
+        .send_keys(Keys.END)
+        .key_down(Keys.SHIFT)
+        .send_keys(Keys.HOME)
+        .key_up(Keys.SHIFT)
+        .send_keys(Keys.DELETE)
+        .send_keys("01/01/2020")
+        .send_keys(Keys.ENTER)
```
Marc-Andre-Rivet, line 607: DRY 🌴🐫

tests/integration/renderer/test_persistence.py
Marc-Andre-Rivet: While I think it's good to have at least one test that checks that persistence works as expected with and without a `part`, the tests validating the behavior of DPSingle and DPRange should be moved to DCC.

Marc-Andre-Rivet: We could create a new test component similar to https://github.com/plotly/dash/tree/dev/%40plotly/dash-generator-test-component-standard that has both (1) prop part, (2) no prop part for two different persisted props, and we can validate the behavior against that component instead.

tests/integration/renderer/test_persistence.py
harryturr: fe58b2a

tests/integration/renderer/test_persistence.py
harryturr: fe58b2a

tests/integration/renderer/test_persistence.py
harryturr: fe58b2a

tests/integration/renderer/test_persistence.py
harryturr: fe58b2a

tests/integration/renderer/test_persistence.py
harryturr: fe58b2a

dash-renderer/src/persistence.js
harryturr: 91c5324

dash-renderer/src/persistence.js
harryturr: 91c5324

tests/integration/renderer/test_persistence.py
harryturr: b063d60 to e4026e6

tests/integration/renderer/test_persistence.py
alexcjohnson: There's also `dash_duo.clear_input` that would simplify some of these (assuming there isn't some reason it fails for datepickers)

https://github.com/plotly/dash/blob/fe58b2a697bbb185fd58fd2d413f5022e66afb6f/tests/integration/renderer/test_persistence.py#L401

.circleci/config.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           command: |
               . venv/bin/activate && pip install --no-cache-dir --upgrade -e . --progress-bar off && mkdir packages
               cd dash-renderer && renderer build && python setup.py sdist && mv dist/* ../packages/ && cd ..
-              git clone --depth 1 https://github.com/plotly/dash-core-components.git
+              git clone -b hg-700-persistence --depth 1 https://github.com/plotly/dash-core-components.git
```
Marc-Andre-Rivet, line 107: These should be unnecessary now that we test against a custom-made component

@plotly/dash-test-components/src/components/MyPersistedComponent.js
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    UNSAFE_componentWillReceiveProps(nextProps) {
+        const {value} = this.input.current;
+        const valueAsNumber = convert(value);
+        this.setInputValue(
+            isNil(valueAsNumber) ? value : valueAsNumber,
+            nextProps.value
+        );
+        if (this.props.type !== 'number') {
```
Marc-Andre-Rivet, line 47: Stuff like number support can be 🔪 from the test component

@plotly/dash-test-components/src/components/MyPersistedComponent.js
```diff
@@ -0,0 +1,293 @@
...
+                        'debounce',
+                        'value',
+                        'n_submit',
+                        'n_submit_timestamp',
+                        'selectionDirection',
+                        'selectionEnd',
+                        'selectionStart',
+                        'setProps',
+                        'loading_state',
+                    ],
```
Marc-Andre-Rivet, line 91: As we had discussed, this is copied over from `dcc.Input`. A lot of these props are not really useful for the test either. Try and keep the surface area for this component to the minimum you need to do the tests you need, nothing more.

@plotly/dash-test-components/src/components/MyPersistedComponentNested.js
```diff
@@ -0,0 +1,298 @@
+import React, {PureComponent} from 'react';
```
Marc-Andre-Rivet, line 1: I think this can be 🔪 altogether. Nested vs. non-nested components have to do with the generator only.

dash-renderer/src/persistence.js
```diff
@@ -265,10 +265,18 @@ const noopTransform = {
     apply: (storedValue, _propValue) => storedValue
 };
 
-const getTransform = (element, propName, propPart) =>
-    propPart
-        ? element.persistenceTransforms[propName][propPart]
-        : noopTransform;
+const getTransform = (element, propName, propPart) => {
+    if (element.persistenceTransforms) {
```
Marc-Andre-Rivet, line 269: Since both 
```
    if (element.persistenceTransforms) {
        if (element.persistenceTransforms[propName]) {
```
result in `return noopTransform, this can be simplified to a single `if`

With the newest [conditional chaining syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) this would become:
```
if (element.persistenceTransforms?.[propName]) { ... }
```
can also be be written with the more habitual
```
    if (element.persistenceTransforms && element.persistenceTransforms[propName]) {
```

@plotly/dash-test-components/src/components/MyPersistedComponent.js
harryturr: 37d454f

Marc-Andre-Rivet (approved):

PR #1384, opened 2020-08-27
Fix #1383 prevent_initial_call positional arg regression

### LINKED ISSUE #1383: prevent_initial_call in callback throws exception [BUG]

**Describe your context**

```
dash                      1.15.0
dash-bootstrap-components 0.9.2
dash-core-components      1.11.0
dash-html-components      1.1.0
dash-renderer             1.7.0
dash-table                4.10.0
```

**Describe the bug**

With the release of Dash 1.15.0. Setting prevent_initial_call in a callback now throws an exception.
It appears prevent_initial_call must be explicitly provided as a named parameter ex. prevent_initial_call=True. This was not previously the case if all parameters (Output, Input, State, and prevent_initial_call) were provided.

  File "C:\Program Files\Python36\lib\site-packages\dash\dash.py", line 980, in callback
    _args, _kwargs
  File "C:\Program Files\Python36\lib\site-packages\dash\dependencies.py", line 159, in handle_callback_args
    prevent_initial_call = args.pop()
AttributeError: 'tuple' object has no attribute 'pop' 

**Expected behavior**

An exception should not be thrown.

alexcjohnson:
Fixes #1383


tests/integration/callbacks/test_validation.py
```diff
@@ -193,3 +193,33 @@ def o2(i, s):
     dash_duo.start_server(app)
     dash_duo.wait_for_text_to_equal("#out1", "1: High")
     dash_duo.wait_for_text_to_equal("#out2", "2: High")
+
+
+def test_cbva005_tuple_args(dash_duo):
```
alexcjohnson, line 198: TIL - I thought this was going to be broken:

https://github.com/plotly/dash/blob/fed4a9123f5d0cf6213da157bb94f5e1900c3440/dash/dependencies.py#L165

Anyway now it's tested :)

but turns out even though `a = [] + ()` is an error, `a = []; a += ()` is fine https://python-forum.io/Thread-concatenating-list-with-tuple

requires-testing.txt
```diff
@@ -6,6 +6,7 @@ pytest-mock==2.0.0;python_version=="2.7"
 lxml==4.5.0
 selenium==3.141.0
 percy==2.0.2
+cryptography==3.0
```
alexcjohnson, line 9: Trying to fix https://app.circleci.com/pipelines/github/plotly/dash/1693/workflows/002acae1-f322-4389-928c-f2c14deaab1b/jobs/27184

Seems like there's something weird going on in the latest release (tonight!) of `cryptography`, which is pulled in by `requests` - see eg https://github.com/pyca/cryptography/pull/5412

PR #1385, opened 2020-08-27
Issue 1350 - Multi-input callback with sync event handling

### LINKED ISSUE #1350: [BUG] Concurrent callbacks broken in Dash 1.13+?
**Describe your context**
The MWE code below *does work* in this environment:
```
dash                              1.12.0
dash-core-components              1.10.0
dash-html-components              1.0.3
dash-leaflet                      0.0.19
dash-renderer                     1.4.1
dash-table                        4.7.0
```

The MWE code below *does **not** work* in this environment:
```
dash                              1.13.4
dash-core-components              1.10.1
dash-html-components              1.0.3
dash-leaflet                      0.0.19
dash-renderer                     1.5.1
dash-table                        4.8.1
```

**Describe the bug**

When there are nested elements with callbacks (e.g. a button inside a div and both have conflicting callbacks), the [documented method ](https://dash.plotly.com/advanced-callbacks) for differentiating between them (checking `ctx.triggered[0]['prop_id'].split('.')[0]`) does not work for Dash 1.13+ (it works with 1.12, though). I am not completely sure whether this was a deliberate change and an update to the documentation is needed or this is a bug. 

I have also asked in the [community forum](https://community.plotly.com/t/how-to-separate-parallel-concurrent-callbacks-nested-elements-in-dash-1-13/43163) first.

**Minimal working example code**

```python
import dash

app = dash.Dash()
app.layout = html.Div(id="div", style={"background": "yellow", "padding": "200px"}, children=[
    html.Button(id="button", children="show infotext"),
    html.Div(id="info", children="INFOTEXT", style={"display": "none"})
])


@app.callback(
    Output("info", "style"),
    [Input("button", "n_clicks"),
     Input("div", "n_clicks")])
def test(btn, div):
    ctx = dash.callback_context
    print("Callback", ctx.triggered)
    prop_ids = [x['prop_id'].split('.')[0] for x in ctx.triggered]
    if "button" in prop_ids:
        print("Action: show")
        return {"display": "block"}
    else:
        print("Action: hide")
        return {"display": "none"}

app.run_server(debug=True)
```

**Expected behavior**

When the button is pressed, the info box should be displayed. When the background is clicked, the infobox should be hidden. 

This code produces the expected behaviour in Dash 1.12. There, the callback is executed once only and `prop_ids ` is a list of all callbacks (`[“button”, “div”]` when clicking on the button and `[“div”]` when clicking on the div).

**Actual behavior**

Since version 1.13, dash behaves differently. The callback function is now executed twice, one time for the button callback and one time for the div. This makes it impossible to figure out which one was actually pressed.

Example output (Dash 1.13) for the above code after clicking one time on the button (the None callback is just the initial callback on load an can be ignored):

```
Callback [{'prop_id': '.', 'value': None}]
Action: hide
Callback [{'prop_id': 'button.n_clicks', 'value': 1}]
Action: show
Callback [{'prop_id': 'div.n_clicks', 'value': 1}]
Action: hide
```
So, the button callback is executed, the infobox is displayed, but then the div callback immediately hides the infobox again.

Marc-Andre-Rivet:
Fixes #1350


dash-renderer/src/observers/requestedCallbacks.ts
```diff
@@ -79,25 +87,35 @@ const observer: IStoreObserverDefinition<IStoreState> = {
...
+            Flatten all duplicated callbacks into a list for removal
+        */
+        const rDuplicates = flatten(rWithDuplicates);
+
+        /*
+            Merge duplicate groups into a single callback - merge by giving priority to newer callbacks
+        */
+        const rMergedDuplicates = map(group => assoc(
+            'changedPropIds',
+            mergeAll(pluck('changedPropIds', group)),
```
alexcjohnson, line 109: In order to ensure `DIRECT` doesn't turn into `INDIRECT` we need something like:

https://github.com/plotly/dash/blob/d9a8d9c2686fddd2a9a6884f9ed34bc44f1c767d/dash-renderer/src/actions/dependencies_ts.ts#L25

dash-renderer/src/observers/requestedCallbacks.ts
```diff
@@ -45,14 +49,18 @@ import {
...
 import { getPendingCallbacks } from '../utils/callbacks';
 import { IStoreObserverDefinition } from '../StoreObserver';
 
 const observer: IStoreObserverDefinition<IStoreState> = {
-    observer: ({
+    observer: async ({
         dispatch,
         getState
     }) => {
+        await wait(0);
```
Marc-Andre-Rivet, line 62: Adding this little wait apparently has wide-ranging consequences on the timing of various features. 
- the rendering `effect` is now incorrectly timed and causes `isAppReady` to never resolve at initialization as the event is triggered *before* the request is made, update here https://github.com/plotly/dash/pull/1385/files#diff-31f5093f6a582b2369ad2cd8216c4f67R69
- document title is now applied out of order and overridden by clientside callbacks with `document.title = ...`, updated here https://github.com/plotly/dash/pull/1385/files#diff-bd786cd26acb24c396ec355c2ff1c4a3R16
- timing of certain exceptions seems different enough to cause ordering issues in devtools, updated test here https://github.com/plotly/dash/pull/1385/files#diff-7b6da2b5053a169045af0651bfcb1b80R680

This list is probably not exhaustive yet. Some proposed fixes should make the overall structure more robust, others are patches or small fixes..

dash-renderer/src/observers/requestedCallbacks.ts
Marc-Andre-Rivet: @alexcjohnson Thanks for the heads up. Also distinguishing the initial callback from subsequent callbacks https://github.com/plotly/dash/pull/1385/commits/8b6ff26341acbd39f0f3d0fa1066bc89d2255eec#diff-178b2bd05a773877fd1b117eef8b1e8cR90

dash-renderer/src/APIController.react.js
```diff
@@ -63,8 +64,11 @@ const UnconnectedContainer = props => {
 
     useEffect(() => {
         if (renderedTree.current) {
-            renderedTree.current = false;
-            events.current.emit('rendered');
+            (async () => {
+                renderedTree.current = false;
+                await wait(0);
```
Marc-Andre-Rivet, line 69: I really dislike this artificial delay being required to counteract the additional processing delay in https://github.com/plotly/dash/pull/1385/files#diff-178b2bd05a773877fd1b117eef8b1e8cR63

dash-renderer/src/StoreObserver.ts
```diff
@@ -21,6 +21,7 @@ interface IStoreObserverState<TStore> {
 export interface IStoreObserverDefinition<TStore> {
     observer: Observer<Store<TStore>>;
     inputs: string[]
+    [key: string]: any;
```
Marc-Andre-Rivet, line 24: Just some flexibility when creating observers

dash-renderer/src/observers/documentTitle.ts
```diff
@@ -0,0 +1,54 @@
...
+            observer.config = config;
+            observer.mutationObserver?.disconnect();
+            observer.mutationObserver = new MutationObserver(() => updateTitle(getState));
+            observer.mutationObserver.observe(
+                document.querySelector('title'),
+                { subtree: true, childList: true, attributes: true, characterData: true }
+            );
+        }
+
+        updateTitle(getState);
```
Marc-Andre-Rivet, line 50: Replaces `<DocumentTitle />` and makes it arguably simpler to handle store changes and DOM mutations in a consistent manner as it doesn't need to deal with component lifecycle anymore.

Marc-Andre-Rivet: This solution should be more robust as it can react to both the DOM changing and the store changing in ways that are not dependant on exact execution timing (the additional wait broke the test/behavior)

dash-renderer/src/observers/requestedCallbacks.ts
```diff
@@ -78,17 +87,37 @@ const observer: IStoreObserverDefinition<IStoreState> = {
...
+                    rMergedDuplicates.push(mergeLeft({
+                        changedPropIds: reduce(mergeWith(Math.max), {}, pluck('changedPropIds', groupWithoutInitial)),
+                        executionGroup: filter(exg => !!exg, pluck('executionGroup', groupWithoutInitial)).slice(-1)[0]
+                    }, groupWithoutInitial.slice(-1)[0]) as ICallback);
+                }
+            }
+        }, values(
+            groupBy<ICallback>(
+                getUniqueIdentifier,
+                requested
```
Marc-Andre-Rivet, line 120: Merge callbacks, keep correct changePropIds DIRECT/INDIRECT and last know executionGroup if there is one. Always drop the initial callback if there is one and there are duplicates.

tests/integration/devtools/test_callback_validation.py
```diff
@@ -677,7 +677,7 @@ def c2(children):
 
     @app.callback([Output("a", "children")], [Input("c", "children")])
     def c3(children):
-        return children
+        return (children,)
```
Marc-Andre-Rivet, line 680: I assume that this is due to timing changes: The `not a tuple` error started appearing first instead of second and broke the test. Fixing it.

alexcjohnson:
@Marc-Andre-Rivet this PR appears to MOSTLY fix the issue @bpostlethwaite encountered 🎉 🏆 🍹 

Would you mind including the app below (and its corrected behavior) as a test case?

The premise: when prop A affects prop B, and both of them affect prop C, then the C callback should list both A and B as triggers. In the app below, there are in fact two primary inputs (A and A') affecting B, and the important thing - the reason we care about having both inputs appear as triggers - is for C to know whether it was A or A' that caused the callback to fire.

Ben's problem: only B was being listed as a trigger.

What I see on this branch:
- On initial load, only B seems to be a trigger, when it should be all three: A, A', and B. But I bet Ben doesn't care about this case ;)
- Later on, when the user changes A, both A and B appear as triggers 🎉 

Here's Ben's app:
```py
import dash


app = dash.Dash(__name__)
app.layout = html.Div(
    [
        html.Div(
            style={"display": "block"},
            children=[
                html.Div(
                    [
                        html.Label("ID: input-number-1"),
                        dcc.Input(id="input-number-1", type="number", value=0),
                    ]
                ),
                html.Div(
                    [
                        html.Label("ID: input-number-2"),
                        dcc.Input(id="input-number-2", type="number", value=0),
                    ]
                ),
                html.Div(
                    [
                        html.Label("ID: sum-number"),
                        dcc.Input(
                            id="sum-number", type="number", value=0, disabled=True
                        ),
                    ]
                ),
            ],
        ),
        html.Div(
            id="results",
            style={
                "border": "antiquewhite",
                "borderStyle": "dashed",
                "height": "200px",
                "width": "475px",
            },
        ),
    ]
)


@app.callback(
    Output("sum-number", "value"),
    [Input("input-number-1", "value"), Input("input-number-2", "value")],
)
def update_sum_number(n1, n2):
    return n1 + n2


@app.callback(
    Output("results", "children"),
    [
        Input("input-number-1", "value"),
        Input("input-number-2", "value"),
        Input("sum-number", "value"),
    ],
)
def update_results(n1, n2, nsum):
    ctx = dash.callback_context
    return [
        "{} + {} = {}".format(n1, n2, nsum),
        html.Br(),
        "ctx.triggered={}".format(ctx.triggered)
    ]


if __name__ == "__main__":
    app.run_server()
```

Marc-Andre-Rivet:
@alexcjohnson Dash version breakdown of callback context

`Dash v1.5.0` to `Dash v1.10.0` (pre PMC)
[{'prop_id': 'input-number-2.value', 'value': 0}]
[{'prop_id': 'sum-number.value', 'value': 0}]

`Dash v1.11.0` and above
[{'prop_id': '.', 'value': None}]
[{'prop_id': 'sum-number.value', 'value': 0}]

More to come.

Marc-Andre-Rivet:
@alexcjohnson Wondering why we override the "no trigger context" case with https://github.com/plotly/dash/blob/dev/dash/_callback_context.py#L31

alexcjohnson:
> After discussion, having initial callbacks without predecessors have no context feels like what was desired (triggered neither by user action of another callback).

Yep right, I was forgetting this behavior is *exactly* what we decided during the original callback refactor back in #1103 - so as of this PR we can consider Ben's issue completely resolved.

> Wondering why we override the "no trigger context" case with https://github.com/plotly/dash/blob/dev/dash/_callback_context.py#L31

See this discussion https://github.com/plotly/dash/pull/1103#issuecomment-581143860 and response https://github.com/plotly/dash/pull/1103#issuecomment-581271566 for the backward compatibility concerns that prompted this.

tests/integration/callbacks/test_callback_context.py
```diff
@@ -96,3 +99,247 @@ def report_triggered(n):
...
+            Input("btn0", "n_clicks"),
+            Input("btn1", "n_clicks"),
+            Input("btn2", "n_clicks"),
+        ],
+        prevent_initial_call=True,
+    )
+    def update(div1, div2, btn0, btn1, btn2):
+        global calls
+        global callback_contexts
+        global clicks
```
alexcjohnson, line 146: Rather than globals, most of our tests use `multiprocessing.Value` for this type of situation - a little cleaner I think, and doesn't depend on threading vs processes when the server and selenium parts branch.

https://github.com/plotly/dash/blob/92684802c94f0ff408c7747e3e8e650aaab43850/tests/integration/callbacks/test_multiple_callbacks.py#L18

tests/integration/callbacks/test_callback_context.py
```diff
@@ -96,3 +99,247 @@ def report_triggered(n):
...
+        clicks["div1"] = div1
+        clicks["div2"] = div2
+        clicks["btn0"] = btn0
+        clicks["btn1"] = btn1
+        clicks["btn2"] = btn2
+
+    def click(target):
+        ActionChains(dash_duo.driver).move_to_element_with_offset(
+            target, 5, 5
+        ).click().perform()
```
alexcjohnson, line 159: Just to be clear: the point here is to click in the padding of `target` so you're sure not to click any of its children?

Marc-Andre-Rivet: Exactly

tests/integration/callbacks/test_callback_context.py
Marc-Andre-Rivet: Since `callback_contexts` is an array of dicts and `clicks` is a dict, I don't think I can easily use `Value` as it expects simple types. Instead used a scoped `context` class that should not leak outside of the test function. https://github.com/plotly/dash/pull/1385/commits/90697b15c5600587b36c0e67ffc0a6fe89b03b4a

alexcjohnson: Well ok, I won't insist - it's probably fine, since we should always be using threads. I still like the `multiprocessing` pattern better as a general rule though, just to make it crystal clear that you're using this to communicate between the server thread and the selenium thread.

alexcjohnson:
This [test-27 failure](https://app.circleci.com/pipelines/github/plotly/dash/1742/workflows/6fc10d63-a4e3-46c4-824d-832240b8aa13/jobs/27901) is interesting - seems plausible that one of the callback invocations was aborted before being requested because another keystroke came in first, which isn't a bad thing at all. Which means if we even want to keep this assertion maybe we just loosen it?

PR #1386, opened 2020-08-27
Bump Flask due to https://github.com/pytest-dev/pytest/issues/5532

wbrgss:
Pins Flask in `requires-install.txt` to `v1.0.4` in order to avoid https://github.com/pytest-dev/pytest/issues/5532 (using `assert` in tests is broken). I suppose this could be in `requires-testing.txt` but I don't think it's too risky. Latest version is `v1.1.2` (https://flask.palletsprojects.com/en/1.1.x/changelog/)

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals


alexcjohnson (approved):
@wbrgss this seems like a good idea, though unlikely to have practical consequences for most people at this point. 💃

PR #1399, opened 2020-09-09
Fix DashBase compat borders in Julia components generator

waralex:
Package dependencies in Julia must have an upper bound on the compatible version. Despite the fact that DashBase is an internal package, the guys from the pkg-registration asked to fix this in the next versions of the core packages for Julia

### DIFF
--- a/dash/development/_jl_components_generation.py
+++ b/dash/development/_jl_components_generation.py
@@ -84,7 +84,7 @@ uuid = "{package_uuid}"
 
 [compat]
 julia = "1.2"
-{base_package} = ">=0.1"
+{base_package} = "0.1"
 """
 
 jl_component_include_string = 'include("{name}.jl")'


rpkyle (approved):
💃

alexcjohnson (approved):
If I'm understanding this correctly, `"0.1"` means [0.1.0, 0.2.0) - which is fine in the short term since both `Dash` and `DashBase` currently live in that range. Longer term it seems like `base_package = 'DashBase'` will want one thing (which could be anything from the exact version specified by `Dash` itself to an extremely broad value like `"0, 1"`, since the `DashBase` version will already be specified by `Dash` for any component using `DashBase` as its `base_package`) and `base_package = 'Dash'` will want something different once we get to the point of non-core component packages requiring specific minimum versions of `Dash`. But we can take care of that later.

PR #1404, opened 2020-09-10
Update CONTRIBUTING.md

emilhe:
Minor corrections to Contributor Guide

A few minor corrections to the commands that i needed to do for the build to work.


CONTRIBUTING.md
```diff
@@ -10,7 +10,7 @@ $ git clone https://github.com/plotly/dash
 $ cd dash
 $ python3 -m venv .venv/dev
 # activate the virtualenv (on windows .venv\dev\scripts\activate)
-$ . .venv/dev/bin/activate
+$ source .venv/dev/bin/activate
```
alexcjohnson, line 13: What OS / shell are you using? On my system (Mac / zsh) `.` and `source` seem to be equivalent, but [some notes](https://stackoverflow.com/a/20097303/9188800) suggest that `.` is the posix standard and `source` may not always be implemented.

Perhaps we need to use one and list the other in a comment?

CONTRIBUTING.md
```diff
@@ -23,7 +23,7 @@ $ pip install -e .
 # build and install components used in tests
 $ cd .. # should be back in dash/ root directory
 $ npm install
-$ npm run setup-tests
+$ npm run setup-tests.py # or npm run setup-tests.R
```
alexcjohnson, line 26: good catch, thanks!

CONTRIBUTING.md
alexcjohnson: Having chatted with some other linux users, it's not clear POSIX is particularly relevant here - we have your example where `source` works and `.` doesn't, but we have no examples of the opposite. So lets keep `source` as the primary but put `.` in a comment. Perhaps:
```
# activate the virtualenv
# on windows `.venv\dev\scripts\activate`
# on some linux / mac environments, use `.` instead of `source`
```

alexcjohnson: https://github.com/plotly/dash/pull/1404/commits/21dd1484914c9fd1da88d0bda395aaa7c2cdade5

alexcjohnson (approved):

PR #1408, opened 2020-09-15
persist callback graph layout when callbacks fire

### LINKED ISSUE #1401: Force Layout in Dev Tools Callback Graph changes when running a callback
force

### LINKED ISSUE #1402: Dev Tools Callback Graph structure's shape doesn't persist between callback updates
callback-structure

Safari

alexcjohnson:
Fixes #1401
Fixes #1402

The way I stored the callback graph layout is hacky: I just mutated the value in the redux store directly to insert it. I did this because I didn't want it to trigger a rerender of the callback graph (a potentially infinite loop), since I'm just recording changes that have already been drawn by cytoscape, so that they'll be available for the next time the graph is rendered. But I neglected to check that this info stuck around when new profile information is added - now it does. If there's a better pattern for this kind of use case I'd be happy to learn about it, but this works. I included a test specifically for #1402 but #1401 is also solved, it's the same issue.

## Contributor Checklist

### optionals

### DIFF
--- a/tests/integration/devtools/test_devtools_ui.py
+++ b/tests/integration/devtools/test_devtools_ui.py
@@ -110,13 +110,12 @@ def test_dvui003_callback_graph(dash_duo):
 
     # hide and redraw the callback graph so we get the new position
     dash_duo.find_element(".dash-debug-menu__button--callbacks").click()
-    dash_duo.find_element(".dash-debug-menu__button--callbacks").click()
-    sleep(2)
 
-    # fire callbacks so the callback graph redraws again
+    # fire callbacks so the profile state is regenerated
     dash_duo.find_element("#add").click()
+    dash_duo.find_element(".dash-debug-menu__button--callbacks").click()
     dash_duo.wait_for_text_to_equal("#totals", "0 of 1 items completed - 0%")
-    sleep(1)
+    sleep(2)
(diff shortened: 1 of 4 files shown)


tests/integration/devtools/test_devtools_ui.py
```diff
@@ -110,13 +110,12 @@ def test_dvui003_callback_graph(dash_duo):
 
     # hide and redraw the callback graph so we get the new position
     dash_duo.find_element(".dash-debug-menu__button--callbacks").click()
-    dash_duo.find_element(".dash-debug-menu__button--callbacks").click()
-    sleep(2)
 
-    # fire callbacks so the callback graph redraws again
+    # fire callbacks so the profile state is regenerated
     dash_duo.find_element("#add").click()
+    dash_duo.find_element(".dash-debug-menu__button--callbacks").click()
```
alexcjohnson, line 116: Worked fine locally but failed on CI before this commit, because the "Add" button was hidden by the callback graph. The bug still applies if the callbacks are fired while the graph is hidden, so that's what the test now does... but this brings up that it would be nice at some point to be able to shrink and/or move the callback graph so you can still interact with whichever parts of the app you want to while it's open.

Marc-Andre-Rivet (approved):
💃🐱💻

PR #1409, opened 2020-09-15
Clean up store initialization

Marc-Andre-Rivet:


alexcjohnson (approved):
Yep, that's definitely cleaner 💃

PR #1415, opened 2020-09-23
`setProps` and `rendered` event ordering causing unsent callbacks

Marc-Andre-Rivet:
This PR proposes to fix a timing issue causing callbacks not to be triggered under certain circumstances.

Possibly a regression in `1.16.0` caused by https://github.com/plotly/dash/pull/1385 timing changes.

The issue occurs when a callback depends on:
1. An unloaded (e.g. inside an unfocused dcc.Tab) async component (e.g. DataTable)
2. A dcc.Graph - or a component that delays `setProps` calls / events handling
3. All components above the DataTable and Graph in the hierarchy are not click reactive (e.g. no `n_clicks` prop)

---

Interestingly enough this may not be a (recent) regression, or a regression at all. Running any Dash `1.11+` against `test_basic_callbacks.py` tests fails on the new `cbsc009` test.

### DIFF
--- a/@plotly/dash-test-components/babel.config.js
+++ b/@plotly/dash-test-components/babel.config.js
@@ -3,4 +3,8 @@ const presets = [
     '@babel/preset-react'
 ];
 
-module.exports = { presets };
+const plugins = [
+    '@babel/plugin-syntax-dynamic-import'
+];
+
+module.exports = { presets, plugins };
--- a/@plotly/dash-test-components/src/components/AsyncComponent.js
+++ b/@plotly/dash-test-components/src/components/AsyncComponent.js
@@ -1,0 +1,19 @@
+import PropTypes from 'prop-types';
+import React, { Suspense } from 'react';
+import { asyncDecorator } from '@plotly/dash-component-plugins';
+import asyncComponentLoader from './../fragments/AsyncComponentLoader';
+
+const AsyncComponent = props => (<Suspense fallback={null}>
+    <RealAsyncComponent {...props} />
+</Suspense>);
+
+const RealAsyncComponent = asyncDecorator(AsyncComponent, asyncComponentLoader);
+
+AsyncComponent.propTypes = {
+    id: PropTypes.string,
+    value: PropTypes.string
+};
+
+AsyncComponent.defaultProps = {};
+
+export default AsyncComponent;
--- a/@plotly/dash-test-components/src/components/CollapseComponent.js
+++ b/@plotly/dash-test-components/src/components/CollapseComponent.js
@@ -1,0 +1,18 @@
+import PropTypes from 'prop-types';
+import React, { Fragment } from 'react';
+
+const CollapseComponent = props => (<Fragment>
+    {display ? props.children : null}
+</Fragment>);
+
+CollapseComponent.propTypes = {
+    children: PropTypes.node,
+    display: PropTypes.bool,
+    id: PropTypes.string
+};
+
+CollapseComponent.defaultProps = {
+    display: false
+};
+
+export default CollapseComponent;
--- a/@plotly/dash-test-components/src/components/DelayedEventComponent.js
+++ b/@plotly/dash-test-components/src/components/DelayedEventComponent.js
@@ -1,0 +1,18 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const DelayedEventComponent = ({ id, n_clicks, setProps }) => (<button
+    id={id}
+    onClick={() => setTimeout(() => setProps({ n_clicks: n_clicks + 1 }), 20)}
+/>);
+
+DelayedEventComponent.propTypes = {
+    id: PropTypes.string,
+    n_clicks: PropTypes.number
+};
+
+DelayedEventComponent.defaultProps = {
+    n_clicks: 0
+};
+
+export default DelayedEventComponent;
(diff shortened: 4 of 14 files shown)


@plotly/dash-test-components/src/components/AsyncComponent.js
```diff
@@ -0,0 +1,19 @@
+import PropTypes from 'prop-types';
```
Marc-Andre-Rivet, line 1: Adding a set of components with simple behaviors corresponding to various usage scenarios. This component is async.

@plotly/dash-test-components/src/components/DelayedEventComponent.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types';
```
Marc-Andre-Rivet, line 1: Adding a set of components with simple behaviors corresponding to various usage scenarios. This component delays `setProps` until later.

@plotly/dash-test-components/src/components/FragmentComponent.js
```diff
@@ -0,0 +1,15 @@
+import PropTypes from 'prop-types';
```
Marc-Andre-Rivet, line 1: Adding a set of components with simple behaviors corresponding to various usage scenarios. This is a top-level component without `n_clicks`

dash-renderer/src/TreeContainer.js
```diff
@@ -155,6 +147,14 @@ class BaseTreeContainer extends Component {
...
                 );
             }
+
+            // Always update this component's props
+            _dashprivate_dispatch(
+                updateProps({
+                    props: changedProps,
+                    itempath: _dashprivate_path
+                })
+            );
```
Marc-Andre-Rivet, line 157: Callbacks are processed after a `0ms` timeout - updating the props after triggering the notifications ensures the notifications are ready on re-render.

tests/integration/callbacks/test_basic_callback.py
```diff
@@ -404,3 +408,53 @@ def update_text(data):
...
+    )
+
+    dash_duo.wait_for_element("#n").click()
+    wait.until(
+        lambda: dash_duo.find_element("#output").text == '[1, 1, "A"]', 3,
+    )
+
+    dash_duo.wait_for_element("#d").click()
+    wait.until(
+        lambda: dash_duo.find_element("#output").text == '[1, 2, "A"]', 3,
```
Marc-Andre-Rivet, line 459: Simplified version of the original reproducible example involving the DataTable inside a dcc.Tab and the dcc.Graph, all inside a non-click reactive top-level component -- does not do away with all real components as it would be tedious but used custom made components for the key behavior.

Marc-Andre-Rivet: This test was run and fails against `dash-renderer==1.8.1`

@plotly/dash-test-components/babel.config.js
```diff
@@ -3,4 +3,8 @@ const presets = [
     '@babel/preset-react'
 ];
 
-module.exports = { presets };
+const plugins = [
```
Marc-Andre-Rivet, line 6: Add async support for the test components

@plotly/dash-test-components/src/fragments/AsyncComponentLoader.js
```diff
@@ -0,0 +1 @@
+export default () => import('./../fragments/AsyncComponent');
```
alexcjohnson, line 1:
```suggestion
export default () => import('./AsyncComponent');
```
?

Marc-Andre-Rivet: 😅

Marc-Andre-Rivet: A case of copypasta

dash-renderer/src/TreeContainer.js
alexcjohnson: Just to make sure I'm understanding this correctly: `notifyObservers` is what ultimately triggers the callback processing and attaches the `rendered` event listener. This is async (with a zero delay), but so is the rerender caused by `updateProps`, and by reversing the order here we ensure that the event listener is inserted in the queue before the rerender so we're in the clear.

Do I have that right?

It still feels a bit fragile, like what we ideally would want to do is attach the event listener synchronously here somehow and just let it be used later, in case another `0ms` delay got added unexpectedly. But that might be invasive, and anyway you've got a very clear test - very nice collection of new test components!

Marc-Andre-Rivet: Yes about the ordering change. TBH I have the same discomfort about the fragility or at least the fragmentation of the implementation between the APIController `useEffect`, https://github.com/plotly/dash/blob/dev/dash-renderer/src/actions/isAppReady.js#L38 DOM check and the async wait in `observers/requestedCallbacks`. What's certain is that we want the event to be triggered when we are certain the DOM has been updated after a React render, otherwise the DOM check might be moot, and that pretty much leaves only useEffect/componentDidUpdate/componentDidMount.

I've tested adding an additional `0ms` async delay prior to calling `notifyObservers` and it does fail the test, so a change in timing would probably be caught, which is comforting..

@plotly/dash-test-components/src/components/CollapseComponent.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types';
```
Marc-Andre-Rivet, line 1: Adding a set of components with simple behaviors corresponding to various usage scenarios. This component hides its content by default (simplified version of what dcc.Tabs/dcc.Tab do)

alexcjohnson (approved):
💃

PR #1416, opened 2020-09-24
Callback graph robustness

alexcjohnson:
The callback graph barfs on some apps, for two reasons:
- If there's a prop `width` the Dagre layout throws an error the way we constructed our graph elements. I just mangled the props in the graph element IDs to fix this.
- Even after this is fixed, for some apps the Dagre layout algorithm gets stuck in an apparently infinite loop. Switching the `ranker` from the default `network-simplex` to `tight-tree` appears to be more robust and gives generally comparable results.

In addition, I switched the force layout to `fcose`, which is more powerful than the built-in `cose`, I gave it stronger coupling between props in a single component to group them closer together, and I added a mode `force-loose` that skips that stronger coupling in case that works better in some circumstances.

I didn't add any tests of the new behavior, as I couldn't find a simple and public app that has problems previously.


@plotly/dash-test-components/src/components/WidthComponent.js
```diff
@@ -0,0 +1,17 @@
+import PropTypes from 'prop-types';
+import React, { Fragment } from 'react';
+
+const WidthComponent = props => (<Fragment>
```
Marc-Andre-Rivet, line 4: Callback graph fails against components with `width` prop.. simple component with `width`

tests/integration/devtools/test_devtools_ui.py
```diff
@@ -119,7 +121,36 @@ def test_dvui003_callback_graph(dash_duo):
...
+        use_debugger=True,
+        dev_tools_hot_reload=False,
+    )
+
+    dash_duo.find_element(".dash-debug-menu").click()
+    sleep(1)  # wait for debug menu opening animation
+    dash_duo.find_element(".dash-debug-menu__button--callbacks").click()
+    sleep(3)  # wait for callback graph to draw
+
+    assert dash_duo.get_logs() == []
```
Marc-Andre-Rivet, line 156: Some test with a callback involving a `width` prop. Fails against `dash-renderer==1.8.1`

Marc-Andre-Rivet (approved):
@alexcjohnson If you're 💃  with the test, I'm 💃  with the code change 😁

alexcjohnson (commented):
Looks great, thanks! 💃

PR #1423, opened 2020-10-06
CI / flask-compress `1.6.0`

Marc-Andre-Rivet:
`flask-compress` was updated to 1.6.0 yesterday and is breaking, amongst other things, `dash-table` builds. This is meant as (a) a stopgag while we figure out why this is problematic, (b) provide version locking in future versions of Dash, as this apparently has an impact.


alexcjohnson (approved):
💃

PR #1426, opened 2020-10-06
Issue 1424 - Latest flask-compress (1.6.0) breaks Dash

### LINKED ISSUE #1424: [BUG] Latest flask-compress (1.6.0) breaks Dash
`flask-compress` released version 1.6.0 on 2020-10-05 and updating to this version breaks component tests (confirmed in the table) in CI. Since `flask-compress` is a dash requirement and not version locked, this may affect negatively Dash installs in the wild.

- Determine the nature of the break
- Fix it..

Marc-Andre-Rivet:
Fixes #1424 

Proposes to override the default `COMPRESSION_ALGORITHM` to `['gzip']` so as to avoid the costly, and currently non-configurable, Brotli compression.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -280,6 +280,11 @@ class Dash(object):
         elif isinstance(server, bool):
             name = name if name else "__main__"
             self.server = flask.Flask(name) if server else None
+            if self.server is not None:
+                # flask-compress==1.6.0 changed default to ['br', 'gzip']
+                # and non-overridable default compression with Brotli is
+                # causing performance issues
+                self.server.config["COMPRESS_ALGORITHM"] = ["gzip"]
         else:
             raise ValueError("server must be a Flask app or a boolean")
 
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 
+
 import os
 import sys
 import collections
@@ -20,6 +21,7 @@ from future.moves.urllib.parse import urlparse
 import flask
 from flask_compress import Compress
 from werkzeug.debug.tbtools import get_current_traceback
+from pkg_resources import get_distribution, parse_version
 
 import plotly
 import dash_renderer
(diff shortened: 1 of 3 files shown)


dash/dash.py
```diff
@@ -280,6 +280,11 @@ def __init__(
         elif isinstance(server, bool):
             name = name if name else "__main__"
             self.server = flask.Flask(name) if server else None
+            if self.server is not None:
+                # flask-compress==1.6.0 changed default to ['br', 'gzip']
+                # and non-overridable default compression with Brotli is
+                # causing performance issues
+                self.server.config["COMPRESS_ALGORITHM"] = ["gzip"]
```
Marc-Andre-Rivet, line 287: Hesitant to override the values for servers passed as parameter https://github.com/plotly/dash/pull/1426/files#diff-228c9aefac729977642672aa1416bacaR277

alexcjohnson: Whenever we call `Compress(self.server)` we're effectively modifying the server's settings - and by default we do this even for user-provided servers. Perhaps we can check whether `server.config["COMPRESS_ALGORITHM"]` already has a value, and only set it if not?

Marc-Andre-Rivet: Turns out that running the code above against `1.5.0` fails and there is no `__version__` or `VERSION` on `flask_compress` so determining the version needs to be done through `pkg_resources`. Updating.

Marc-Andre-Rivet: https://github.com/plotly/dash/pull/1426#discussion_r500570299 - 👍

requires-install.txt
```diff
@@ -1,5 +1,5 @@
 Flask>=1.0.2
-flask-compress==1.5.0
+flask-compress
```
Marc-Andre-Rivet, line 2: unfix flask-compress

alexcjohnson (approved):
LGTM! 💃

PR #1442, opened 2020-10-22
Update react(-dom) from 16.13.0 to 16.14.0

antoinerg:
I think I'm getting console errors in my Dash app which may be fixed in 16.13.1 as reported in https://github.com/facebook/react/issues/18178#issuecomment-601425520


Marc-Andre-Rivet (approved):
:shipit:

PR #1447, opened 2020-10-28
Wait for proper non-stale element reference in WebDriverWait

### LINKED ISSUE #1164: [dash testing] - handle Stale
In https://github.com/plotly/dash-docs/pull/829 & https://app.circleci.com/pipelines/github/plotly/dash-docs/116/workflows/6647d2f6-4541-47fd-93fe-37d5f005f300/jobs/10757, I was getting the following error:

```
selenium.common.exceptions.StaleElementReferenceException: Message: stale element reference: element is not attached to the page document
```

This was hard to debug as I did not have this error locally. Some attempts at retrying didn't work, maybe I wasn't catching the right error (https://github.com/plotly/dash-docs/pull/829/commits/2c9fa1efbd76b8cec65bda34161290a210008fae, https://github.com/plotly/dash-docs/pull/829/commits/feb2482df0e9b8da311b517ca4a52db6631a3bdc)

In https://stackoverflow.com/a/44914767/4142536, there was a suggestion to use selenium's `ignored_exceptions` argument. Perhaps we can add that, and this common exception, to our `wait_for_text_to_equal` commands.

anders-kiaer:
Resolves #1164.

My experience is that the stale reference appears more frequently in multi-page apps using `dcc.Location` / `dcc.Link`, compared with `dcc.Tabs`. When it happens, it is difficult to debug, as it happens a bit by random (even in CI context, with same initial setup, multiple test runs on a Dash app can sometimes go fine, sometimes fail with `StaleElementReferenceException`).

This PR ensures that the `WebDriverWait` waits until a proper non-stale reference exists to the element, inspired by link to StackOverflow post provided in #1164. By [default `selenium.webdriver.support.wait.WebDriverWait` only ignores `NoSuchElementException`](https://selenium-python.readthedocs.io/api.html#module-selenium.webdriver.support.wait).

## Contributor Checklist

   -  [x] Add `StaleElementReferenceException` to list of ignored selenium errors while waiting on element.

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [UNRELEASED]
 ### Fixed
 - [#1434](https://github.com/plotly/dash/pull/1434) Fix [#1432](https://github.com/plotly/dash/issues/1432) for Julia to import non-core component packages without possible errors.
+- [#1447](https://github.com/plotly/dash/pull/1447) Fix [#1164](https://github.com/plotly/dash/issues/1164). Commands like `dash_duo.wait_for_element(some_search_pattern, timeout=some_timeout)` now waits until a _non-stale_ reference exists, and (still) raises `TimeoutException` if not satisfied within the timeout. Previously, `wait_for_element` could occasionally fail with `StaleElementReferenceException`.
 
 ## [1.16.3] - 2020-10-07
 ### Fixed
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -15,9 +15,11 @@ from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.common.action_chains import ActionChains
 
 from selenium.common.exceptions import (
-    WebDriverException,
-    TimeoutException,
     MoveTargetOutOfBoundsException,
+    NoSuchElementException,
+    StaleElementReferenceException,
+    TimeoutException,
+    WebDriverException,
 )
 
 from dash.testing.wait import text_to_equal, style_to_equal, contains_text, until
@@ -61,7 +63,15 @@ class Browser(DashPageMixin):
         self._driver = until(self.get_webdriver, timeout=1)
         self._driver.implicitly_wait(2)
 
-        self._wd_wait = WebDriverWait(self.driver, wait_timeout)
+        self._wd_wait_ignored_exceptions = (
+            NoSuchElementException,
+            StaleElementReferenceException,
+        )
+        self._wd_wait = WebDriverWait(
+            self.driver,
+            wait_timeout,
+            ignored_exceptions=self._wd_wait_ignored_exceptions,
+        )
         self._last_ts = 0
         self._url = None
 
@@ -232,7 +242,13 @@ class Browser(DashPageMixin):
     def _wait_for(self, method, args, timeout, msg):
         """Abstract generic pattern for explicit WebDriverWait."""
         _wait = (
-            self._wd_wait if timeout is None else WebDriverWait(self.driver, timeout)
+            self._wd_wait
+            if timeout is None
+            else WebDriverWait(
+                self.driver,
+                timeout,
+                ignored_exceptions=self._wd_wait_ignored_exceptions,
+            )
         )
         logger.debug(
             "method, timeout, poll => %s %s %s",


alexcjohnson:
@anders-kiaer this looks great, thanks! Does the `Draft` indicate you have more work in mind for the PR? I don't think it necessitates any new tests, since it isn't intended to add new functionality, just make the tests themselves more robust.

The test failure is likely not a problem, just a residual from the tests not running initially. We needed to adjust settings for fork PRs, after changing how we access docker images on CircleCI. If there's no more you intend to run here, an empty commit should resolve the failure.

anders-kiaer:
Thanks for the feedback @alexcjohnson.

Yes, I changed it to draft since I saw there was a chance I would probably want to change the `CHANGELOG.md` entry slightly to be more clear, and maybe consider if the PR was needed after all (when I looked more in detail in `dash.testing`).

My current understanding (might be wrong) is that e.g.
```
WebDriverWait(driver, timeout, ignored_exceptions=ignored_exceptions).until(some_condition)
```
will never fail on `StaleElementReferenceException` within `until()` if included in `ignored_exceptions`, but if the user does something like
```
el = WebDriverWait(driver, timeout, ignored_exceptions=ignored_exceptions).until(some_condition)
el.click()
```
during the `click()` method there is no such guarantee anymore (i.e. the `ignored_exceptions` is used while running `until()`). That is perhaps obvious for experienced `selenium` users, but could perhaps be stated clearly in the `CHANGELOG.md` entry.

The `.until(some_condition)` methods where ignoring exceptions are relevant should be methods where you find an element, and do something with it (e.g. extract `text`, `style` or `click` on it etc.). However, I see that those classes in Dash already have `try / except` on their own level, e.g. https://github.com/plotly/dash/blob/e2f7884ce1a587694d3272f6c2eecb9b13a32e9e/dash/testing/wait.py#L70-L84,
and that they typically catch the base class exception `WebDriverException` which a [broad range of other exceptions inherit from](https://selenium-python.readthedocs.io/api.html#module-selenium.common.exceptions).

I guess it is a matter of preference, if you want to have common exceptions to ignore within `WebDriverWait` initialization, and/or treat them on the level of the different functions with `try / except`. And then how broad you want to catch there (`WebDriverException` is quite broad by looking at the exception list in `selenium` documentation, might be that if some of those errors should occur you would prefer the actual error to be raised and not transformed into `TimeOutException` as I guess would be the current behaviour?)

---

As a side note, the use case that hit us randomly originally was `dash_duo.wait_for_element(...).click()` raising `StaleElementReferenceException` (then probably inheriting from the `click()` action). In order to not have everyone deal with `try/except` `StaleElementReferenceException` on `click` actions, maybe we (as a separate issue) could add something like `dash_duo.wait_click_when_clickable` to accompany `wait_for_text_to_equal`, `wait_for_contains_text`...?

alexcjohnson:
If we can reduce spurious failures by ignoring more exceptions, without impacting real errors that users will care about, then absolutely let's do it! I'm not aware of any other exceptions in the `WebDriverException` group arising during these waits, but I haven't really looked into it.

I wouldn't be opposed to a new method - perhaps `dash_duo.wait_for_and_click(selector)` - if that's the cleanest from a user standpoint. I'm still a bit unclear where the problem is coming from though. Is it that the element renders, so the `WebDriverWait` returns, but then before the `click` is executed another callback returns, which causes the element to be deleted and recreated thereby invalidating the original reference?

If so perhaps we should make `dash_duo._wait_for_callbacks()` an official part of the API (ie remove the leading `_`)? This one is a little tricky as it looks at the internal state of the Dash renderer, via its redux store, to determine whether there are still callbacks in progress. But we do use it internally and have maintained backward compatibility of that method when we've changed those internals (so you can use a newer version of `dash.testing` against an app that's running an older version of `dash` - not a problem for `dash_duo` but if you use `dash_br` to access an already-running app this could come up)

anders-kiaer:
Sorry for late response on this - have not been working on the Dash based project for some months :baby_bottle:

I tested adding your suggestion @alexcjohnson with `dash_duo._wait_for_callbacks()` prior to the line `el.click()` randomly giving problems with `StaleElementReferenceException`. The frequency of observed test failures with stale reference dropped immediately in our large Dash app from ~1/3 to 0/7. :tada: 

If making that function an official part of the API is still on the table, I think it sounds like a good idea. Another use case of it could perhaps be when someone has a large Dash app with callbacks that take more time than the `wait_timeout` - the test code can properly wait for callbacks to finish (instead of the less elegant solution of increasing general `wait_timeout`?).

gvwilson:
thanks again for this @anders-kiaer - closing as stale (both Dash and Selenium have moved on).

PR #1448, opened 2020-10-29
Improve error message

almarklein:
I was struggling with an error in an app I was working on. I got an error, but it did not tell me much. This catches and re-raises that error to add a more meaningful message, which would have saved me some time :)

## Contributor Checklist

 -  [x] Update code
 -  [x] I'm assuming this error is most likely to occur in the situation that I was in. If this is not the case, the message might only add to the confusion.


alexcjohnson:
Ah, very interesting - thanks for bringing this up @almarklein !

It's a little confusing to understand how forgetting the `@` gets us into this state, so let me detail it a little. The callback map entry is created when `app.callback` is called - and this happens whether you mark it as a decorator or not:

https://github.com/plotly/dash/blob/e2f7884ce1a587694d3272f6c2eecb9b13a32e9e/dash/dash.py#L998

But the callback function only gets attached to the callback map when the result of `app.callback` is evaluated with its function, and THIS part only happens when `app.callback` is marked as a decorator (or equivalently if you call it as a higher-order function `app.callback(...)(func)`):

https://github.com/plotly/dash/blob/e2f7884ce1a587694d3272f6c2eecb9b13a32e9e/dash/dash.py#L1046

Which explains how we got to a situation where a callback is dispatched without a function to dispatch to!

---

OK, with that out of the way, I think you're right that omitting the `@` is about the only way to encounter this particular error. My first reaction was to see if we could catch this error earlier, on startup rather than when we finally try to dispatch this callback. It's a bit tricky to figure out when that should happen though - it would need to be after all the callbacks have been defined, but also it shouldn't depend how you run the server (`run_server` vs `gunicorn` etc)... So yeah, let's keep the check where you have it.

The only thing I'd add to this is to include `output` - the "callback ID" that describes all the component IDs and props returned by this callback - in the message, so that if you have a big app with many callbacks it'll be easy to find the offending one.

almarklein:
> The only thing I'd add to this is to include output [...] in the message

Done!

alexcjohnson (approved):
💃 thanks!

almarklein:
Thanks for the quick and clear review :)

PR #1453, opened 2020-11-07
Generate Julia components with nested structure

tcbegley:
This PR allows generation of Julia components with nested structure, it is based on #1203 which did the same for R components. This change will allow me to bring Julia support to dash-bootstrap-components.

As far as I could see there wasn't an obvious place to add tests for this. Happy to take a stab at that if you point me to where I should do that.


tcbegley:
Hello, and happy new year!

Just want to check in on this as it's been a while. No particular rush to get anything merged, but it would be helpful to at least know in principle whether you are willing to make a change like this or if I should look for workarounds? I figured since it is simply bringing parity to R and Julia component generation it shouldn't be too controversial?

alexcjohnson (approved):
Happy New Year @tcbegley! Thanks for the PR and apologies for missing this earlier. We have a project lingering in the background to redo component generation in JS (node) rather than Python, which may be why this PR dropped off our radar. But that effort has been put on hold for a while so let's move ahead with this PR. I'm Assuming you've used it on your project and the Julia package works as expected? If so I don't think we need any explicit tests at this time. 💃

tcbegley:
> apologies for missing this earlier

No worries at all

> We have a project lingering in the background to redo component generation in JS (node) rather than Python

Oh cool, sounds like an interesting idea!

> I'm Assuming you've used it on your project and the Julia package works as expected?

Yep, seems to be all good this end.

PR #1489, opened 2020-12-09
Revert "Bump Flask due to https://github.com/pytest-dev/pytest/issues/5532"

wbrgss:
Reverts plotly/dash#1386


Marc-Andre-Rivet (changes requested):
@wbrgss No need to revert. We've "rolled with it" and are releasing with the change.

PR #1493, opened 2020-12-10
improve dash import test

### LINKED ISSUE #1143: [BUG] dash doesn't get imported when a file named "org.py", "dash.py", or "test.py" with specific content is present in the current directory // "AttributeError: module 'dash' has no attribute 'Dash'"
**Describe your context**

```
dash (1.9.1)
dash-core-components (1.8.1)
dash-html-components (1.0.2)
dash-renderer (1.2.4)
dash-table (4.6.1)

```

**Describe the bug**

If a file named ``org.py`` is present in the current directory with the following content:

```
```

then dash doesn't import and I get the following message:
```
>>> import dash
Dash was not successfully imported. Make sure you don't have a file named
'dash.py' in your current directory.
```

**Expected behavior**
dash should import without any error.

**Additional info**
- The org.py is never imported
- If I rename the file to a different name dash get imported without any problem.
- The problem is shown also with ``import dash_html_components as html```
- The problem is shown either on Windows and in Linux
- Tested with python3.4, python3.6, python3.8

**Steps to replicate the problem on Linux**
```
$ mkdir mytest
$ cd mytest
$ echo "import dash_core_components as dcc" > org.py
$ python3 -m venv venv
$ . venv/bin/activate
(venv) $ pip install dash
(venv) $ python
Python 3.4.6 (default, Mar 01 2017, 16:52:22) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dash
Dash was not successfully imported. Make sure you don't have a file named
'dash.py' in your current directory.
(venv) $
```

if I rename the file the import works:
```
(venv) $ mv org.py othername.py
(venv) $ python
Python 3.4.6 (default, Mar 01 2017, 16:52:22) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dash
>>>
```

alexcjohnson:
Fixes #1143 - with the help of companion PRs in the various component repos. (In principle we should do this for all component repos as well as the boilerplate)
- https://github.com/plotly/dash-core-components/pull/903
- https://github.com/plotly/dash-html-components/pull/169
- https://github.com/plotly/dash-table/pull/854

Through a confusing set of circumstances described in #1143, having a Python file with one of several common names in your directory (but NOT `dash.py`) which imports any dash component package causes a circular import of dash itself, which manifests as failing the "make sure you don't have a `dash.py` file in your directory" check in the `__init__.py` of the component package, because `dash` hasn't finished importing so doesn't have the attributes we're looking for.

I'm tempted to delete these checks entirely - if we did that, the circumstance described above would work fine, but when you do make a `dash.py` you'd see the error message:
```
  File "/Users/alex/plotly/dash-core-components/dash_core_components/__init__.py", line 30, in <module>
    from ._imports_ import *  # noqa: F401, F403
  File "/Users/alex/plotly/dash-core-components/dash_core_components/_imports_.py", line 1, in <module>
    from .Checklist import Checklist
  File "/Users/alex/plotly/dash-core-components/dash_core_components/Checklist.py", line 3, in <module>
ModuleNotFoundError: No module named 'dash.development'; 'dash' is not a package
```
Which is essentially what you'd see if you did this with any other Python package - `import pandas` when you have a file `pandas.py` for example.

And if we ever end up combining packages this whole business will be moot, as you'll never get to the point of generating this error. Still, for now these changes should solve the issue.

## Contributor Checklist


.circleci/config.yml
```diff
@@ -122,7 +122,7 @@ jobs:
           command: |
               . venv/bin/activate && pip install --no-cache-dir --upgrade -e . --progress-bar off && mkdir packages
               cd dash-renderer && renderer build && python setup.py sdist && mv dist/* ../packages/ && cd ..
-              git clone --depth 1 https://github.com/plotly/dash-core-components.git
+              git clone --depth 1 -b import-test https://github.com/plotly/dash-core-components.git
```
alexcjohnson, line 125: Remove this branch after merging https://github.com/plotly/dash-core-components/pull/903

Marc-Andre-Rivet (approved):
:shipit: after 🔪 https://github.com/plotly/dash/pull/1493/files#r539834762

PR #1501, opened 2020-12-15
dedent error messages

chriddyp:
Our error messages are printed in the terminal & in the dash dev tools console. In both cases, we treat white space literally in order to display nice tracebacks. However, some of our error messages are indented. This PR dedents these error messages.

Example of behavior: 

image

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [UNRELEASED]
 ### Fixed
 - [#1493](https://github.com/plotly/dash/pull/1493) Fix [#1143](https://github.com/plotly/dash/issues/1143), a bug where having a file with one of several common names (test.py, code.py, org.py, etc) that imports a dash component package would make `import dash` fail with a cryptic error message asking whether you have a file named "dash.py"
-- [#1501](https://github.com/plotly/dash/1499) Dedented error messages.
+- [#1501](https://github.com/plotly/dash/pull/1501) Dedented error messages.
 
 
 ## [1.18.1] - 2020-12-09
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [UNRELEASED]
 ### Fixed
 - [#1493](https://github.com/plotly/dash/pull/1493) Fix [#1143](https://github.com/plotly/dash/issues/1143), a bug where having a file with one of several common names (test.py, code.py, org.py, etc) that imports a dash component package would make `import dash` fail with a cryptic error message asking whether you have a file named "dash.py"
+- [#1501](https://github.com/plotly/dash/1499) Dedented error messages.
```
rpkyle, line 8:
```suggestion
- [#1501](https://github.com/plotly/dash/pull/1499) Dedented error messages.
```

alexcjohnson:
I'm puzzled as to why this is necessary - we tried to dedent all the Dash exceptions in the base class.

https://github.com/plotly/dash/blob/cd2ff7c25c459d86bd0f60a578f89910425b94e1/dash/exceptions.py#L4-L6

If you manage to figure out the test failures, they seem to be the same as @almarklein sees in https://github.com/plotly/dash-core-components/pull/905 - something broke in the `dash-html-components` build process? If nothing else I'll try and sort this out later this afternoon.

alexcjohnson:
In the specific example you show above, `dedent` is not going to fix the problem because it will only remove a consistent number of spaces from each line, and the lines `[0] Tab` and `[2] function` have no indentation. This comes from when `path` is inserted in the string here:

https://github.com/plotly/dash/blob/cd2ff7c25c459d86bd0f60a578f89910425b94e1/dash/_validate.py#L180-L184

because `path` is a result of `'\n'.join(...)` in various places, meaning it has unindented lines within it.

Seems to me the solution is to indent subsequent lines of `path, something like `path.replace('\n', ' ' * 12 + '\n'), and use that pattern anywhere else we have multiline args interpolated in these messages.

alexcjohnson:
superseded by #1530

PR #1506, opened 2020-12-17
Loosen testing requirements & improve testing

### LINKED ISSUE #1466: Loosen or update cryptography requirement for dash[testing]

**Describe your context**
dash[testing] has Cryptography hardcoded to 3.0 (`cryptography==3.0`). Can we either get this loosened or updated to 3.2.1 as cryptography > 3.2 has a vulnerability.

**Describe the bug**

dash[testing] has Cryptography hardcoded to 3.0 (`cryptography==3.0`). 

**Expected behavior**

Can we either get this loosened or updated to 3.2.1 as cryptography > 3.2 has a vulnerability.

alexcjohnson:
Fixes #1466 - loosens all `dash[testing]` requirements to `>=` for better compatibility with outside projects.
In addition I made two other changes:
- Bumped various `dash[dev]` requirements. I had thought that this was only used by us internally, on CI, but in fact it's required for building components, and as such is referenced in the [component boilerplate](https://github.com/plotly/dash-component-boilerplate/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/requirements.txt). So we may need to revisit this.
- Something in that process broke running our old-style locally, though they still appear to run on CI. I took this opportunity to convert the rest of them to `dash.testing` and getting rid of the last `unittest` style tests.

## Contributor Checklist


requires-dev.txt
```diff
@@ -1,14 +1,15 @@
...
+pylint==2.6.0;python_version>="3.7"
+astroid==2.4.2;python_version>="3.7"
 black==19.10b0;python_version>="3.0"
-virtualenv==20.0.10;python_version=="2.7"
-fire==0.2.1
-coloredlogs==14.0
+virtualenv==20.2.2;python_version=="2.7"
+fire==0.3.1
+coloredlogs==15.0
+flask-talisman==0.7.0
```
alexcjohnson, line 15: Moved `flask-talisman` here from `requires-testing` because this is used in our tests themselves, not in `dash.testing`.

requires-testing.txt
```diff
@@ -1,13 +1,12 @@
...
+pytest>=6.0.2;python_version>="3.0"
+pytest>=4.6,<5;python_version=="2.7"
+pytest-sugar>=0.9.4
+pytest-mock>=3.2.0;python_version>="3.0"
+pytest-mock>=2.0.0,<3;python_version=="2.7"
+lxml>=4.6.2
+selenium>=3.141.0
+percy>=2.0.2
+requests[security]>=2.21.0
+beautifulsoup4>=4.8.2,<=4.9.3;python_version=="2.7"
```
alexcjohnson, line 10: `beautifulsoup4` has [stated](https://pypi.org/project/beautifulsoup4/) it's about to drop Py2 support, so I set a preemptive upper bound for Py2 at the current version.

tests/unit/test_resources.py
```diff
@@ -34,7 +36,13 @@ class StatMock(object):
 
 
 def test_external(mocker):
+    assert "dash_core_components" in ComponentRegistry.registry
+    assert "dash_html_components" in ComponentRegistry.registry
+    mocker.patch("dash.development.base_component.ComponentRegistry.registry")
+    ComponentRegistry.registry = {"dash_core_components", "dash_html_components"}
```
alexcjohnson, line 42: These tests would fail previously if run in tandem with the integration tests, because they import and register various other component packages. After this change, I can simply call `pytest` locally to run all the Python-based tests in this repo.

There's still one that fails locally for me: The iframe sandbox test [rdif001 fails on its last line](https://github.com/plotly/dash/blob/cd2ff7c25c459d86bd0f60a578f89910425b94e1/tests/integration/renderer/test_iframe.py#L60) - it can't see the log error it's supposed to have. I'm not sure what's going on here but it seems we've run afoul of some sort of security restriction: in the Chrome devtools I can't even see the DOM inside the iframe (though I can interact with this DOM via Selenium).
Screen Shot 2020-12-16 at 9 42 59 PM
Anyway this test still runs fine on CI, but if what I'm seeing is new behavior in the latest Chrome (v87) we may need to sort this out soon.

requires-dev.txt
```diff
@@ -1,14 +1,15 @@
...
 mock==3.0.5;python_version=="2.7"
-flake8==3.7.9
-PyYAML==5.3
-pylint==1.9.4;python_version<"3.7"
-pylint==2.3.1;python_version=="3.7"
-astroid==2.2.5;python_version=="3.7"
+flake8==3.8.4
+PyYAML==5.3.1
+pylint==1.9.5;python_version<"3.7"
+pylint==2.6.0;python_version>="3.7"
```
Marc-Andre-Rivet, line 9: Could we sanity test these flake8/pylint changes against the other core repos?

Marc-Andre-Rivet: They have an annoying habit of changing their rules..

tests/integration/callbacks/test_multiple_callbacks.py
```diff
@@ -305,3 +305,281 @@ def out(v):
...
+    }
+
+    @app.callback(Output("parent-a", "value"), Input("grandparent", "value"))
+    def update_parenta(value):
+        call_counts["parent-a"].value += 1
+        return "a: {}".format(value)
+
+    @app.callback(Output("parent-b", "value"), Input("grandparent", "value"))
+    def update_parentb(value):
+        time.sleep(0.2)
```
Marc-Andre-Rivet, line 384: Why the sleep? Is something in this test time senstive?

alexcjohnson: This was just copied over from `test_chained_dependencies_branched_lineage` in the now-deleted [`test_render.py`](https://github.com/plotly/dash/pull/1506/files#diff-cd7e55c505e7622e615c748b9d7ebd9f71d1360e357d9e09b5db946d5cb5a8c0L252) - some of the other tests that had delays embedded I converted to use `Lock`. This one wasn't so obvious to me how to do that, and honestly I didn't look into exactly what it's using the delay for, but I did cut the delays in half.

requires-dev.txt
alexcjohnson: Good call - they may well need their own updates to `.pylintrc` files, or syntax updates. PRs coming.

tests/integration/callbacks/test_multiple_callbacks.py
Marc-Andre-Rivet: 👌

tests/integration/renderer/test_due_diligence.py
```diff
@@ -99,3 +100,15 @@ def test_rddd001_initial_state(dash_duo):
 
     dash_duo.percy_snapshot(name="layout")
     assert dash_duo.get_logs() == [], "console has no errors"
+
+
+@pytest.mark.parametrize("child", [0, [0]])
+def test_rddd002_falsy_child(dash_duo, child):
+    app = dash.Dash(__name__)
+    app.layout = html.Div(id="falsy-wrapper", children=0)
```
Marc-Andre-Rivet, line 108: Shouldn't this be `children=child` ?

alexcjohnson: thanks - combined 2 tests into one here but missed that key step! https://github.com/plotly/dash/pull/1506/commits/c24b3829e1ec068c9f5494a9ba8f19bf09ad2004

tests/unit/test_resources.py
Marc-Andre-Rivet: Running this locally on Chrome 87, getting:
image

Running this locally on Chrome 89 (Canary), getting:
image

Moving from warning to warning+error.

Changing the iframe sandbox options to
```
<iframe src="{0}" sandbox="allow-scripts allow-same-origin">
```

- fixes the problem on Chrome 87
- fixes the warning on Chrome 89 but the error remains
image

Attempting to find additional information.

Marc-Andre-Rivet: Trying to access`window.localStorage` from the console from Chrome 89 provides a slightly better explanation
image
-- this is new

Marc-Andre-Rivet: In both Chrome 87 and 89 `localStorage` is not accessible from data URL but in 89 it's shown from the start. In both 87/89 running `window.localStorage` from the console gives
image

alexcjohnson: To be clear, the test _requires_ that a warning is logged so we're not trying to "fix" all the warnings. I'm not sure it's really important that we test for these logs, the main thing is that the app still works and doesn't error out trying to access cookies. So the easy answer here would be to simply remove the log check. But the "full" solution I guess would be to serve a real container page with an iframe in it, rather than a `data:` url, and keep the check as is.

For reference this feature was introduced in https://github.com/plotly/dash/pull/1080

requires-dev.txt
alexcjohnson: https://github.com/plotly/dash-core-components/pull/907, https://github.com/plotly/dash-html-components/pull/171, https://github.com/plotly/dash-table/pull/857 - All three, I started with this branch, got tests passing, then reverted to the dev branch of dash and tests still pass - All that was required (other than prohibiting the latest `xlrd`) was a couple of linting exclusions.

Marc-Andre-Rivet (approved):

PR #1507, opened 2020-12-17
Updating CI Python images

Marc-Andre-Rivet:
We've previously had issues moving to newer versions of these CI images, at least part of the issue seems to be that Chromedriver 75 changed a capabilities flag for logging.

https://chromedriver.chromium.org/downloads
```
Renamed capability loggingPrefs to goog:loggingPrefs, as required by W3C standard
```

This PR proposes to add the flag (while keeping the old one for non-core projects still using older CI images

---

Also changes expected behavior for `rdif001` as as of Chrome 87 running this test only results in warnings, now correctly not being picked up by `get_logs()`. As of Chrome 89 this behavior will change and at least 1 error will be present. This will require the test to be updated again and to have a different behavior in Python 2.7 vs. 3.x unless we manually update the version of Chrome used in that test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
 
   artifacts:
     docker:
-      - image: circleci/python:3.7.6-stretch-node-browsers
+      - image: circleci/python:3.7.9-stretch-node-browsers
         auth:
           username: dashautomation
           password: $DASH_PAT_DOCKERHUB
@@ -33,7 +33,7 @@ jobs:
   lint-unit-37: &lint-unit
     working_directory: ~/dash
     docker:
-      - image: circleci/python:3.7.6-stretch-node-browsers
+      - image: circleci/python:3.7.9-stretch-node-browsers
         auth:
           username: dashautomation
           password: $DASH_PAT_DOCKERHUB
@@ -73,7 +73,7 @@ jobs:
   lint-unit-36:
       <<: *lint-unit
       docker:
-        - image: circleci/python:3.6.9-stretch-node-browsers
+        - image: circleci/python:3.6.12-stretch-node-browsers
           auth:
             username: dashautomation
             password: $DASH_PAT_DOCKERHUB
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
   artifacts:
     docker:
-      - image: circleci/python:3.7.6-stretch-node-browsers
+      - image: circleci/python:3.7.9-stretch-node-browsers
```
Marc-Andre-Rivet, line 16: Bump to latest patch version for each Python version, side-effect is that this updates the version of Chrome

dash/testing/browser.py
```diff
@@ -431,6 +431,7 @@ def _get_chrome(self):
 
         capabilities = DesiredCapabilities.CHROME
         capabilities["loggingPrefs"] = {"browser": "SEVERE"}
+        capabilities["goog:loggingPrefs"] = {"browser": "SEVERE"}
```
Marc-Andre-Rivet, line 434: Add ChromeDriver 75+ correct capabilities. Keep the old one for compatibility.

tests/integration/renderer/test_iframe.py
```diff
@@ -57,4 +57,4 @@ def apply_cors(response):
     dash_duo.wait_for_element_by_id("btn").click()
     dash_duo.wait_for_element("#output-0").text == "0=1"
 
-    assert len(dash_duo.get_logs()) != 0
+    assert dash_duo.get_logs() == []
```
Marc-Andre-Rivet, line 60: This log entry in Chrome 87 is a warning and now not included.

.circleci/config.yml
alexcjohnson: Two thoughts:
1. should we switch to circleci's "next-gen" images? https://circleci.com/developer/images/image/cimg/python
2. Maybe a bigger task, but should we be switching from py3.7 to py3.9?

Marc-Andre-Rivet: @alexcjohnson I can certainly look into those. If possible I'd like to do these in a follow up as this work was initiated by an issue @harryturr encountered as part of his current work assignment.

alexcjohnson (approved):
💃

Issue #1403, opened 2020-09-10
add an 'x' button to close the error messages popup


chriddyp:
Currently you have to click on the devtools icon and then toggle errors button.

chriddyp:
In this UI
![image](https://user-images.githubusercontent.com/1280389/92668461-b48e7180-f2c3-11ea-9612-1c7e5875ec16.png)

Biswa5812:
Where is this particular UI available, under which folder?

chriddyp:
@Biswa5812 - It's in here: https://github.com/plotly/dash/tree/dev/dash-renderer/src/components/error/FrontEnd

chriddyp:
@Biswa5812 - Let us know if you end up taking a look at this one! This would be a good first issue to get started contributing to Dash.

Sable-20:
Quick update for anyone looking at this
https://github.com/plotly/dash/tree/dev/dash/dash-renderer/src/components/error/FrontEnd 
issue can be found here, there was a update adding /dash/

Alfred-Emmanuel:
Please has this been fixed? if no, could i be assigned to this?

gvwilson:
@Alfred-Emmanuel sure - if you would like to submit a PR we would be very happy to review it. thank you - @gvwilson

Alfred-Emmanuel:
Hi @gvwilson , I am running into some errors while trying to setup the project,  could you please let me know the prerequisites i need to have installed before being able to setup the project locally? i have nodejs and python installed already

gvwilson:
hi @Alfred-Emmanuel - are you using the instructions in `CONTRIBUTING.md`? If so, where is it failing for you? (If you can paste a log into a comment here that would help.)

Alfred-Emmanuel:
Yes, i am using the instructions in the CONTRIBUTING.md, everything was going well until i ran "npm run build", the log is quite long but this is it below

npm run build

> build
> run-s private::build.*


> private::build.components
> python dash/development/update_components.py 'all'

npx lerna exec --concurrency 3 --scope='{dash-core-components,dash-html-components,dash-table}' -- npm i
lerna notice cli v6.6.2
lerna info versioning independent
lerna notice filter including "{dash-core-components,dash-html-components,dash-table}"
lerna info filter [ '{dash-core-components,dash-html-components,dash-table}' ]
lerna info Executing command in 3 packages: "npm i"
[...]

PR #1508, opened 2020-12-19
Add an 'x' button to close the error messages box #1403

AnnMarieW:
This is a potential solution for https://github.com/plotly/dash/issues/1403
It works in my local environment, but needs better css for the x button

### DIFF
--- a/dash-renderer/src/components/error/FrontEnd/FrontEndError.css
+++ b/dash-renderer/src/components/error/FrontEnd/FrontEndError.css
@@ -6,19 +6,15 @@
     position: absolute;
     right: 0;
     top: 0;
-    color: #ccc;
+    color: #B9C2CE;
     font-size: 20px;
     cursor: pointer;
-<<<<<<< refs/remotes/origin/1403-x-button-to-close-error-box
-    margin-right: 5px
-=======
     margin-right: 10px
->>>>>>> One more css update to align the button  #1403
 }
 
 .dash-fe-error__icon-x:hover
 {
- color:black;
+ color:#a1a9b5;
 }
 
 
--- a/dash-renderer/src/components/error/FrontEnd/FrontEndError.css
+++ b/dash-renderer/src/components/error/FrontEnd/FrontEndError.css
@@ -9,7 +9,11 @@
     color: #ccc;
     font-size: 20px;
     cursor: pointer;
+<<<<<<< refs/remotes/origin/1403-x-button-to-close-error-box
     margin-right: 5px
+=======
+    margin-right: 10px
+>>>>>>> One more css update to align the button  #1403
 }
 
 .dash-fe-error__icon-x:hover
--- a/dash-renderer/src/components/error/FrontEnd/FrontEndError.css
+++ b/dash-renderer/src/components/error/FrontEnd/FrontEndError.css
@@ -9,6 +9,7 @@
     color: #ccc;
     font-size: 20px;
     cursor: pointer;
+    margin-right: 5px
 }
 
 .dash-fe-error__icon-x:hover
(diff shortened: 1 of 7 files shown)


alexcjohnson:
@AnnMarieW this works great! Re: styling, I'm thinking we can use a much lighter button right in the corner, something like this:
Screen Shot 2020-12-22 at 12 05 54 PM

Which I got from applying the following styling in the devtools, along with replacing the SVG icon with just a unicode `&times;` (or in fact just the unicode character ×)

```
<div style="
    position: absolute;
    right: 0;
    top: 0;
    color: #ccc;
    font-size: 20px;
    cursor: pointer;
">×</div>
```
Also to make `position: absolute;` work on this element, you have to define `position: relative;` on the `.dash-error-card__topbar` parent.

And the close button will want some `:hover` css to make it appear clickable.

AnnMarieW:
That looks so much better, Thanks!

dash-renderer/src/components/error/FrontEnd/FrontEndError.css
```diff
@@ -2,6 +2,22 @@
...
+    top: 0;
+    color: #ccc;
+    font-size: 20px;
+    cursor: pointer;
+    margin-right: 5px
+}
+
+.dash-fe-error__icon-x:hover
+{
+ color:black;
```
alexcjohnson, line 17: That's probably a stronger effect than we want - really we just need enough to be able to see a change in the X. Closest analog I see is the grey debug menu buttons, which starts out about the same as `#ccc` (just a tiny bit tilted toward blue)

https://github.com/plotly/dash/blob/e7f4984d0b58511cdd8d4c97ac5ecf963dd89c35/dash-renderer/src/components/error/menu/DebugMenu.css#L70

And on hover it gets just a bit darker

https://github.com/plotly/dash/blob/e7f4984d0b58511cdd8d4c97ac5ecf963dd89c35/dash-renderer/src/components/error/menu/DebugMenu.css#L84

I'm not sure where those exact colors came from, but we may be able to just copy them both.

AnnMarieW: I see, I'll make the change.  Thanks!

dash-renderer/src/components/error/FrontEnd/FrontEndErrorContainer.react.js
```diff
@@ -34,6 +34,12 @@ class FrontEndErrorContainer extends Component {
                         </strong>
                         ){connected ? null : '\u00a0 🚫 Server Unavailable'}
                     </div>
+                    <div
+                        className='dash-fe-error__icon-x'
+                        onClick={() => clickHandler()}
+                    >
+                        x
```
alexcjohnson, line 41:
```suggestion
                        ×
```
This is the `&times;` character, which makes a nicer button than the letter x.

dash-renderer/src/components/error/icons/OffIcon.svg
```diff
@@ -1,3 +1,3 @@
 <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 54 24">
-<path d="M18 2l18 18m0-18l-18 18" stroke-width="4" stroke="#fff"></path>
+<path d="M18 2l18 18m0-18l-18 18" stroke-width="6" stroke="#fff"></path>
```
alexcjohnson, line 2:
```suggestion
<path d="M18 2l18 18m0-18l-18 18" stroke-width="4" stroke="#fff"></path>
```
Not used in this PR anymore so let's put the original back.

alexcjohnson (approved):
Just two minor final edits then this is ready to go! 💃

Issue #1456, opened 2020-11-12
[Feature Request] persist data in url


bbtfr:
Thanks so much for your interest in Dash!

Before posting an issue here, please check the Dash [community forum](https://community.plotly.com/c/dash) to see if the topic has already been discussed. The community forum is also great for implementation questions. When in doubt, please feel free to just post the issue here :)

---

Dash support persistence, which is a simple-to-use feature for storing user choices in the UI.
For now, it only support persist data in memory / localStorage / sessionStorage.

It's useful if we can persist data in url, e.g.: localhost:8050/?my_key=my_value
With this, we can easily share a filled form, a tabs with specific tab opened, etc. to others by sending them a url.

ndrezn:
There is a solid implementation strategy proposed by @nicolaskruchten in https://community.plotly.com/t/boilerplate-starter-app-with-state-in-url-and-state-aware-hot-reloading/71821. But, agreed that it'd be wonderful to see a first-class way to support this without building this manually!

PR #1512, opened 2020-12-30
Add the LocationStore (window.location) to persistence types

flovouin:
This PR is a first draft to add `window.location` as a persistence store type. The motivation for this is that I've been searching around the internet for solutions to update the url as components are updated by the user, such that the user can then share the dashboard in a given state. I came across a few posts of people trying to do the same thing on Stackoverflow, [Plotly community forums](https://community.plotly.com/t/save-current-dashboard-state-as-url/23143), [GitHub](https://gist.github.com/jtpio/1aeb0d850dcd537a5b244bcf5aeaa75b) [gists](https://github.com/oegedijk/dash_querystrings), issues in this repo (#1456), etc.

I implemented a custom (Python) solution a while back that was working great, but was essentially a huge wrapper around Dash, and it didn't seem like the best solution.
I'm not sure this solution is the perfect one either, but after discovering `persistence`, it made sense to try it out. I've only made a very basic real life test of this, but I'm planning on pushing this a bit further to see if it's actually viable.

Before dedicating more time to this (adding tests, etc), could you please tell me if that's something you would consider adding to the code base? Also, I have limited frontend experience, so any feedback on the implementation is welcome. 🙂

Have a great holiday season,
Cheers,

## Contributor Checklist

   -  [x] Add `window.location` as a `persistence_type`

### optionals

    -  [ ] this GitHub [#PR number]() updates the [dash docs](https://github.com/plotly/dash-docs/blob/master/dash_docs/chapters/persistence/index.py)

### DIFF
--- a/dash-renderer/src/persistence.js
+++ b/dash-renderer/src/persistence.js
@@ -29,9 +29,11 @@
  *   stores `columns[i].name` for all columns `i`. Nested props also need
  *   entries in `persistenceTransforms` - see below.
  *
- * - `persistence_type`: one of "local", "session", or "memory", just like
- *   `dcc.Store`. But the default here should be "local" because the main use
- *   case is to maintain settings across reloads.
+ * - `persistence_type`: one of "local", "session", "memory", or "location",
+ *   just like `dcc.Store`. But the default here should be "local" because
+ *   the main use case is to maintain settings across reloads. "location" can
+ *   be used to store the values in the URL, such that the state can be
+ *   shared with other users.
  *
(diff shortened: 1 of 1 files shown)


ndrezn:
Closing this PR as the recommended pathway for interacting with Location to update components similar to the proposed API is using Dash pages: https://dash.plotly.com/urls#query-strings

We offer in Dash Enterprise a Deep Links functionality which offers an integrated UI and native support for persisting component ID/property pairs to the URL.

PR #1514, opened 2021-01-02
Refactor JSON encoding to use plotly.py JSON engine

jonmmease:
Update to make Dash use the active plotly.py JSON engine. The JSON engine system was introduced in plotly.py in https://github.com/plotly/plotly.py/pull/2955. 

Note: This PR adds a plotly.py lower version bound of 5.0.

plotly.py will use the faster orjson encoder by default if the orjson package is installed. But to make the orjson engine is in use, add this snippet:

```python
import plotly.io as pio
pio.json.config.default_engine = "orjson"
```


jonmmease:
There is one percy baseline change.  The plotly.py JSON engines always sort object keys in alphabetical order.  So the `devtools validation exception` snapshot changed.

#### Before:
label/value/typo

Screenshot_20210812_125657

#### After:
label/typo/value in alphabetical order

Screenshot_20210812_125709

jonmmease:
Ready for review

alexcjohnson (approved):
Add a changelog entry and this is great! 💃 

There are also a couple instances of `PlotlyJSONEncoder` in the tests, do you want to convert these too?

alexcjohnson:
Oh actually I don't see `orjson` listed as a dep, were you going to include it in `requires-dev.txt`? Can we perhaps pick one or two integration tests and run them with both `json` and `orjson`? I don't care which, they all make extensive use of JSON encodings :)

jonmmease:
Thanks! Yeah, I'll make sure both get used in some tests.

jonmmease:
ping @alexcjohnson or @chriddyp.  This is ready to merge, but I don't have permission to approve the percy diff [mentioned above](https://github.com/plotly/dash/pull/1514#issuecomment-897807922).  Could one of you approve that in percy?

chriddyp:
Approved, looks good!

jonmmease:
Thanks!

Issue #889, opened 2019-08-27
support circular dependencies


alexcjohnson:
Circular deps keep coming up. Most of the use cases I've seen are fairly simple, like 2 alternate ways to accomplish the same task, that you want to keep in sync. For example today on the community board someone asked the perfectly reasonable question of how to make the selected tab reflected in the url, and vice versa so you could load your app with the desired tab open. https://community.plot.ly/t/combining-dcc-tabs-and-dcc-location/27831

The code they tried seems like the natural way to do it, but fails as it requires a circular dependency between `Location.pathname` and `Tabs.value`. I don't see a simple alternative. We could make one using patterns we've used elsewhere (like adding a `*_timestamp` prop to one of these components, or having `Location.pathname` update `href` and vice versa, which it should arguably do) but this feels hacky and fragile - it's making an explicit circular ref into an *implicit* one that may make infinite loops that are hard to debug and even harder to fix.

---

My proposal: create a new `dash.dependencies` class, tentatively `CircularOutput`, by which users could specify that this particular output may be in a circular dependency chain. Then in the renderer we would:
- Prevent unchanged values from dispatching callbacks. I can't think of a reason *ever* to allow this, anyone want to argue with that? See https://github.com/plotly/dash/issues/883 where I first proposed this - also would help with things like https://github.com/plotly/dash-core-components/issues/608, and once we have this handled in the renderer we should be able to remove the corresponding checks from component code, making it simpler and easier to write robust components.
- Allow circular deps if somewhere in the chain there's a `CircularOutput`
- Keep track of the chain of props that led from a user action (or, more generally, a browser event, like for `Interval` components) to the output. This would be useful anyway to show with the stack trace when debugging errors.
- When a `CircularOutput` is changed, look back at this chain; if the same `CircularOutput` is already in the chain (meaning this would be its second change in response to the same stimulus) consider this a possible infinite loop and throw an error.

kmcmearty:
@chriddyp "There is also a case for using circular dependencies to sync up dcc.Store with controls so that you can access the the control values (via dcc.Store) in separate tabs or pages. It'd be circular so that when you'd render the initial page, dcc.Store would sync the value back to the original component."

[...]

PR #1525, opened 2021-01-13
Input output callback

chadaeschliman:
Allow circular callbacks in the special case of passing a component+property as an input and output of the same callback.  By combining this capability with `dash.callback_context` it is straightforward to have multiple components remain synchronized.  Other forms of circularity (involving multiple callbacks) are still detected and blocked.  This is related to: https://github.com/plotly/dash/issues/889 

Example code for a linked slider and numeric input:
```
import dash

external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']
app = dash.Dash(__name__, external_stylesheets=external_stylesheets)

    html.H1(children='Circular Test'),
    dcc.Slider(
        id='slider',
        min=0, max=20,
    ),
    dcc.Input(
        id='input',
        type='number',
        debounce=False,
        value=3
    ),
])

@app.callback(
    Output('input', 'value'),
    Output('slider', 'value'),
    Input('input', 'value'),
    Input('slider', 'value'),
)
def callback(iv, sv):
    ctx = dash.callback_context
    if not ctx.triggered:
        # default to input value at init
        if iv is not None:
            value = iv
        else:
            value = sv
    else:
        trigger_id = ctx.triggered[0]["prop_id"].split(".")[0]
        if trigger_id=='input':
            value = iv
        else:
            value = sv
    return value, value

app.run_server(debug=True)
```

Result:
circular_test

## Contributor Checklist

   -  [x] Disable checking for input/output overlap
   -  [x] Disable checking for circularity in the special case that an output is also an input
   -  [x] Make changes to ensure callbacks execute correctly (wasn't needed)
   -  [x] Modify tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 - [#1508](https://github.com/plotly/dash/pull/1508) Fix [#1403](https://github.com/plotly/dash/issues/1403): Adds an x button
 to close the error messages box.
+- [#1525] (https://github.com/plotly/dash/pull/1525): Add support for callbacks which have overlapping inputs and outputs. Combined with `dash.callback_context` this addresses many use cases which require circular callbacks.
 
 ### Changed
 - [#1503](https://github.com/plotly/dash/pull/1506) Fix [#1466](https://github.com/plotly/dash/issues/1466): loosen `dash[testing]` requirements for easier integration in external projects. This PR also bumps many `dash[dev]` requirements.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 - [#1508](https://github.com/plotly/dash/pull/1508) Fix [#1403](https://github.com/plotly/dash/issues/1403): Adds an x button
 to close the error messages box.
-- [#1525] (https://github.com/plotly/dash/pull/1525): Add support for callbacks which have overlapping inputs and outputs. Combined with `dash.callback_context` this addresses many use cases which require circular callbacks.
+- [#1525](https://github.com/plotly/dash/pull/1525) Adds support for callbacks which have overlapping inputs and outputs. Combined with `dash.callback_context` this addresses many use cases which require circular callbacks.
 
 ### Changed
 - [#1503](https://github.com/plotly/dash/pull/1506) Fix [#1466](https://github.com/plotly/dash/issues/1466): loosen `dash[testing]` requirements for easier integration in external projects. This PR also bumps many `dash[dev]` requirements.
--- a/dash-renderer/src/actions/dependencies.js
+++ b/dash-renderer/src/actions/dependencies.js
@@ -363,32 +363,21 @@ function findDuplicateOutputs(outputs, head, dispatchError, outStrs, outObjs) {
     });
 }
 
-function findInOutOverlap(outputs, inputs, head, dispatchError) {
-    outputs.forEach((out, outi) => {
+function checkInOutOverlap(out, inputs) {
         const {id: outId, property: outProp} = out;
-        inputs.forEach((in_, ini) => {
+    return inputs.some(in_ => {
             const {id: inId, property: inProp} = in_;
             if (outProp !== inProp || typeof outId !== typeof inId) {
-                return;
+            return false;
             }
             if (typeof outId === 'string') {
                 if (outId === inId) {
-                    dispatchError('Same `Input` and `Output`', [
-                        head,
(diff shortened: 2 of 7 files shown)


alexcjohnson:
Thanks @chadaeschliman - this is really interesting. I think you're correct that logically this is self-consistent: any time one of the inputs to this callback changes, we should execute the callback exactly once and whatever values it returns become the final values of the input/output props. This pattern could also be used for input validation.

@chriddyp @Marc-Andre-Rivet thoughts? Am I missing anything? Any reason we wouldn't want to allow this kind of circularity?

There are a number of cases we'll need to write CI tests for. Consider this set of callbacks where `a` and `b` are connected circularly and there's an extra input `e` and extra outputs `c` and (via another callback) `d`:
```py
cb1:
Output('a', 'value'), Output('b', 'value'), Output('c', 'value'),
Input('e', 'value'), Input('a', 'value'), Input('b', 'value')

cb2:
Output('d', 'value'),
Input('a', 'value')
```

If you change `a`, it's important that first `cb1` fires, changing `b` and `c` and possibly changing `a` too (maybe rounding, maybe validating it), and then `cb2` fires using the value of `a` returned by `cb1` and NOT the one the user entered (that was fed to `cb1`. `cb1` and `cb2` should each fire exactly once.

If you change `b` or `e`, the same should happen though in these cases it's more obvious that we'd use the final value of `a` since that's the only change in its value that could trigger `cb2`

chadaeschliman:
Thanks @alexcjohnson for considering this.  You were correct that depending on the order the callbacks were defined, sometimes cb2 would be called twice (first with the original value then with the new value).  There was a straightforward fix by modifying `getReadyCallbacks` to not consider overlapped inputs/outputs as outstanding outputs.  With this change, cb1 is initially the only readyCallback so it fires first followed by cb2.  I'm not sure how to make a CI test that verifies the order of callback execution.

alexcjohnson:
> With this change, cb1 is initially the only readyCallback so it fires first followed by cb2.

🎉 

> I'm not sure how to make a CI test that verifies the order of callback execution.

Make the circular callback actually *change* the values, and make the downstream callback return different values depending on which of the values it receives. Then all you have to do is inspect the final contents of each element, and count how many times each callback has been executed using a `multiprocessing.Value` like:

https://github.com/plotly/dash/blob/3402cb8f67466bcee8776b1d07dd3134d08bda76/tests/integration/callbacks/test_basic_callback.py#L54

If both of those come out right, we know the order was correct.

chadaeschliman:
Thanks @alexcjohnson for the tip.  I added test cbsc015 which checks this.  I'm not sure why percy/dash failed on the latest build.

alexcjohnson:
The new test looks great! Don't worry about the Percy failure, that came from some authorization issues we're working through with CircleCI.

It occurs to me it would be good to have the same test also with both callbacks clientside, since that can have different sequencing. That obviously can't use a `multiprocessing.Value` but it can stash a value in a `window` variable and read it back out with `dash_duo.driver.execute_script`.

And unless @Marc-Andre-Rivet has anything else to add, the only other thing I think we need is a changelog entry.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 - [#1508](https://github.com/plotly/dash/pull/1508) Fix [#1403](https://github.com/plotly/dash/issues/1403): Adds an x button
 to close the error messages box.
+- [#1525] (https://github.com/plotly/dash/pull/1525): Add support for callbacks which have overlapping inputs and outputs. Combined with `dash.callback_context` this addresses many use cases which require circular callbacks.
```
alexcjohnson, line 9:
```suggestion
- [#1525](https://github.com/plotly/dash/pull/1525) Adds support for callbacks which have overlapping inputs and outputs. Combined with `dash.callback_context` this addresses many use cases which require circular callbacks.
```

alexcjohnson (approved):
Yep, your clientside test looks great - hah, these days we're recommending putting the js function inline like this:

https://github.com/plotly/dash/blob/3402cb8f67466bcee8776b1d07dd3134d08bda76/tests/integration/callbacks/test_wildcards.py#L140-L149

But ironically we haven't included that form in `test_clientside` 🙄 

Looks great in the callback graph - here's your app plus one downstream callback:
Screen Shot 2021-01-15 at 7 43 55 PM

So I think we're good to go - thanks again for the PR! 💃

chadaeschliman:
Yeah! Thanks for your help!

nirvana-msu:
I have an exact opposite use-case to the prunning implemented here - I need a callback to actually fire in an infinite circular fashion - see #2139 for details. Would be great if we could have an option to avoid this infinite loop (of a single callback) instead of prunning as it is done here.

Issue #1527, opened 2021-01-15
[BUG] External assets not reflected in get_asset_url


karosc:
When serving assets from an external path by setting:

`serve_locally=False` 
and 
`assets_external_path=<url to external asset folder>` 

css, js, and favicon assets are pulled from the external asset directory when the app is initialized, but if I try to use `app.get_asset_url(<asset>)` it still pulls the asset from the dash directory rather than the external directory. 

Would developers be ammenable to a pull request altering the [get_asset_url function](https://github.com/plotly/dash/blob/3402cb8f67466bcee8776b1d07dd3134d08bda76/dash/dash.py#L1187) to look more like below?
```python
def get_asset_url(self, path):
    if self.config.assets_external_path:
        asset = get_asset_path(
            self.config.assets_external_path,
            path,
            self.config.assets_url_path.lstrip("/"),
        )
    else:
        asset = get_asset_path(
            self.config.requests_pathname_prefix,
           path,
            self.config.assets_url_path.lstrip("/"),
       )

    return asset

```
For reference [get_asset_path](https://github.com/plotly/dash/blob/3402cb8f67466bcee8776b1d07dd3134d08bda76/dash/_utils.py#L49) is defined in _utils.py as:
```python
def get_asset_path(requests_pathname, asset_path, asset_url_path):

    return "/".join(
        [
            # Only take the first part of the pathname
            requests_pathname.rstrip("/"),
            asset_url_path,
            asset_path,
        ]
    )
````

alexcjohnson:
Nice find, would you like to make a PR to fix this oversight?

PR #1528, opened 2021-01-15
update get_asset_url function

karosc:
This pull request addresses #1527 by adding and if statement to the `get_asset_url` function in dash.py instructing the app to use the external asset path if it is defined. 

I have tested this with an short I wrote and [published](https://github.com/karosc/Dash_Tester). If you run the app and examine the src attribute of the image, it references and image on my own site karosc.github.io.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1194,7 +1194,7 @@ class Dash(object):
         else:
             asset = get_asset_path(
                 self.config.requests_pathname_prefix,
-            path,
+                path,
                 self.config.assets_url_path.lstrip("/"),
             )
 
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1100,10 +1100,7 @@ class Dash(object):
     def _add_assets_resource(self, url_path, file_path):
         res = {"asset_path": url_path, "filepath": file_path}
         if self.config.assets_external_path:
-            res["external_url"] = "/".join([
-                self.config.assets_external_path.rstrip('/'), self.config.assets_url_path.strip('/'), url_path.lstrip('/')
-            ]
-            )
+            res["external_url"] = self.get_asset_url(url_path.lstrip("/"))
         self._assets_files.append(file_path)
         return res
 
@@ -1193,11 +1190,7 @@ class Dash(object):
         else:
             prefix = self.config.requests_pathname_prefix
 
-        asset = get_asset_path(
-            prefix,
-            path,
-            self.config.assets_url_path.lstrip("/")
-        )
+        asset = get_asset_path(prefix, path, self.config.assets_url_path.lstrip("/"))
 
         return asset
 
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1185,11 +1185,18 @@ class Dash(object):
         ]
 
     def get_asset_url(self, path):
-        asset = get_asset_path(
-            self.config.requests_pathname_prefix,
+        if self.config.assets_external_path:
+            asset = get_asset_path(
+                self.config.assets_external_path,
+                path,
+                self.config.assets_url_path.lstrip("/"),
+            )
+        else:
+            asset = get_asset_path(
+                self.config.requests_pathname_prefix,
             path,
-            self.config.assets_url_path.lstrip("/"),
-        )
+                self.config.assets_url_path.lstrip("/"),
+            )
 
         return asset
 
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -150,10 +150,12 @@ class Dash(object):
     :type assets_ignore: string
 
     :param assets_external_path: an absolute URL from which to load assets.
-        Use with ``serve_locally=False``. Dash can still find js and css to
-        automatically load if you also keep local copies in your assets
-        folder that Dash can index, but external serving can improve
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Very nice! Could be a little DRYer, as it's only the first arg to `get_asset_path` that needs to change, but that's not a big deal.

Can you add a couple of unit tests of `app.get_asset_url`? Should cover both the local and external cases, make an app and test a couple of asset urls for that app, but we don't need to run the app (so the external url need not be valid). `test_configs.py` looks like a good place to put it, and it would look something like these two:

https://github.com/plotly/dash/blob/3402cb8f67466bcee8776b1d07dd3134d08bda76/tests/unit/test_configs.py#L147-L179

After that, we'll just need a changelog entry and it'll be ready to go!

karosc:
does this seem more DRY? Not sure exactly what you were referring to.

```python
def get_asset_url(self, path):
        if self.config.assets_external_path:
            prefix = self.config.assets_external_path
        else:
	    prefix = self.config.requests_pathname_prefix

        asset = get_asset_path(
                     prefix,
                     path,
                     self.config.assets_url_path.lstrip("/"),
               )

        return asset
```

alexcjohnson:
Yep, that's about what I had in mind. But again, not a big deal.

karosc:
After building some tests, I am realizing that the third argument also needs to change. Conceptually `assets_external_path` is the full external path to the assets folder ([assuming this is the case based on the code here](https://github.com/plotly/dash/blob/3402cb8f67466bcee8776b1d07dd3134d08bda76/dash/dash.py#L1101), where the resource url is created by combining assets_external_path only with the asset path and not also with the asset folder path). 

Based on this, the function should look more like:

```python

    def get_asset_url(self, path):
        if self.config.assets_external_path:
            asset = get_asset_path(
                self.config.assets_external_path,
                path,
                '',
            )
        else:
            asset = get_asset_path(
                self.config.requests_pathname_prefix,
                path,
                self.config.assets_url_path.lstrip("/"),
            )

        return asset

```

tests/unit/test_configs.py
```diff
@@ -105,6 +105,28 @@ def test_pathname_prefix_assets(empty_environ, req, expected):
...
 
+@pytest.mark.parametrize(
+    "requests_pathname_prefix, assets_external_path, assets_url_path, expected",
+    [
+        (None    , None                        , "assets" , "/assets/reset.css"),
+        ("/app/" , None                        , "assets" , '/app/assets/reset.css'),
+        (None    , None                        , "css"    , "/css/reset.css"),
+        ("/app/" , None                        , "css"    , "/app/css/reset.css"),
+        (None    , "http:\\external.com/", "assets" , "http:\\external.com/assets/reset.css"),
+        ("/app/" , "http:\\external.com/", "css"    , "http:\\external.com/css/reset.css"),
```
alexcjohnson, line 116:
```suggestion
        (None    , "http://external.com/", "assets" , "http://external.com/assets/reset.css"),
        ("/app/" , "http://external.com/", "css"    , "http://external.com/css/reset.css"),
```

dash/dash.py
```diff
@@ -1098,8 +1100,9 @@ def _setup_server(self):
     def _add_assets_resource(self, url_path, file_path):
         res = {"asset_path": url_path, "filepath": file_path}
         if self.config.assets_external_path:
-            res["external_url"] = "{}{}".format(
-                self.config.assets_external_path, url_path
+            res["external_url"] = "/".join([
+                self.config.assets_external_path.rstrip('/'), self.config.assets_url_path.strip('/'), url_path.lstrip('/')
+            ]
```
alexcjohnson, line 1105: Hmm, maybe we could just call `get_asset_url` here?

alexcjohnson:
Looking good, just a couple of minor comments above, but also I think it needs linting (I sympathize with wanting aligned values in the test matrix but I don't think `black` will go for it 😅 ) - Probably just needs `npm run format`.

dash/dash.py
karosc: That works! Passing all tests.

karosc:
running npm format changed a bunch of other files, sorry for creating more review work...🤷‍♂️

alexcjohnson:
Ah, I'm guessing you're on a newer version of `black`. What version do you have installed? We've talked about the need to update to 20.8b1 (and maybe now is the right time to do that across Dash, cc @Marc-Andre-Rivet?) but currently we have 19.10b0

https://github.com/plotly/dash/blob/c5ba38f0ae7b7f8c173bda10b4a8ddd035f1d867/requires-dev.txt#L11

karosc:
oh shoot, you're right I am running 20.8b1. I can downgrade and reformat then if that's when you want to do.

alexcjohnson:
Downgrading and reformatting would be the easiest way to get this merged quickly 😄 We may upgrade but that's going to take a little work because we'll want to do it evenly across all the dash repos.

karosc:
Alright, I reset to before I ran the format routine, downgraded black to 19.10b0 and then redid the format. Things came out to be more expected since only my changes were reformatted by black.

alexcjohnson (approved):
Very nice - thank you! 💃

Issue #1205, opened 2020-04-21
Docstring generator is wrong for certain properties including `sort_by`


nicolaskruchten:
![image](https://user-images.githubusercontent.com/203523/79871002-c3769500-83b1-11ea-93df-124ca962cd90.png)

chriddyp:
While we're in the neighborhood, I'm sure there are many more fixes that we can make for markdown readability. Here are a few blocks that are pretty tough to digest:

Missing some newlines in here? Especially to split up `group` and `grouping`
![image](https://user-images.githubusercontent.com/1280389/80178785-ee303b80-85b3-11ea-8a3b-6bbdb469334b.png)

JS comment included in the default text
![image](https://user-images.githubusercontent.com/1280389/80178826-04d69280-85b4-11ea-8e47-45694bf5227f.png)

Nested dictionaries's lists aren't indendented
![image](https://user-images.githubusercontent.com/1280389/80178844-0f912780-85b4-11ea-984b-fb239ce38855.png)

Could use some more newlines + same nested dictionary issue
![image](https://user-images.githubusercontent.com/1280389/80178891-2768ab80-85b4-11ea-8146-901297f974a3.png)

`null` should be `None`
![image](https://user-images.githubusercontent.com/1280389/80178916-318aaa00-85b4-11ea-8590-2504602a4a54.png)

chriddyp:
Another related issue: https://github.com/plotly/dash-docs/issues/836

gvwilson:
Hi - we are tidying up stale issues and PRs in Plotly's public repositories so that we can focus on things that are most important to our community. If this issue is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. (Please note that we will give priority to reports that include a short reproducible example.) If you'd like to submit a PR, we'd be happy to prioritize a review, and if it's a request for tech support, please post in our community forum. Thank you - @gvwilson

PR #1531, opened 2021-01-17
Update docstrings to address #1205

AnnMarieW:
This PR updates how the docstrings are generated to address    [#1205](https://github.com/plotly/dash/issues/1205)

See more details in [dash-docs  #1063](https://github.com/plotly/dash-docs/pull/1063)


alexcjohnson:
As mentioned in https://github.com/plotly/dash-core-components/pull/914#issuecomment-762847279, I think it would be a good idea to escape double quotes in the generated docstrings - even if the added period handles most cases where this would cause problems, the triple-quote potential issue remains.

chriddyp:
This looks really good! Such a relief to see all of this fixed. Glad we have test cases here with all of the types. Looks like the double quote issue was fixed, any other concerns here @alexcjohnson ?

dash/development/_py_components_generation.py
```diff
@@ -551,7 +576,7 @@ def map_js_to_py_types_flow_types(type_object):
             else ""
         ),
         # React's PropTypes.shape
-        signature=lambda indent_num: "dict containing keys {}.\n{}".format(
+        signature=lambda indent_num: "dict with keys: {}.\n{}".format(
```
alexcjohnson, line 579: One last thing: Flow isn't used much but since we have it lets simplify it the same as react proptypes: remove the list of keys here since they're enumerated below, and remove the "Those keys have the following types" line.

alexcjohnson:
Oh, also we need a changelog entry. That and https://github.com/plotly/dash/pull/1531#discussion_r566837719 then I think this is ready to go!

alexcjohnson:
@harryturr this PR means the rest of the core will need to be included in the next release whether or not there's a changelog entry for each piece, so all the docstrings get rebuilt.

tests/unit/development/__init__.py
```diff
@@ -13,51 +13,81 @@ def has_trailing_space(s):
...
+    "- customProp (optional)",
+    "",
+    "- data-* (string; optional)",
+    "",
+    "- in (string; optional)",
+    "",
+    "- optionalAny (boolean | number | string | dict | list; optional)",
+    "",
+    "- optionalArray (list; optional):",
+    "    Description of optionalArray.  ",
```
alexcjohnson, line 34: Ah shoot, missed this before, but does this mean there is trailing whitespace in some of the docstrings? We should really avoid that. I see up above there's a `has_trailing_space` function, that gets used a couple of places in the tests like:
```py
assert not has_trailing_space(component_class_string)
```
Ideally any file we create in our code generators (or any string meant to be written to a file) should pass that test.

AnnMarieW: Well, I needed the trailing white spaces ("  \n" )to make the spacing work right in both Markdown and in the console.  Do the trailing white spaces in comments cause a problem?

alexcjohnson (approved):
💃 Very nice work - big improvement to our docstrings!

AnnMarieW:
Thanks for your kind words and assistance with this project!

PR #1550, opened 2021-02-08
Favicon new branding

dmt0:
Updated favicon as per the current Plotly branding :)

The icon is same format/dimension as the old one - ICO 48x48.


alexcjohnson (approved):
💃 Thanks @dmt0 🙇

PR #1551, opened 2021-02-08
Quieten browser init in test setup

scjody:
This code is shared by the DE automated end to end integration tests (e2e). The information logged here is never useful to us so this adds a significant amount of noise to our tests. If it's useful to you folks (in a non-debugging context), I'd be happy to make it configurable in the `Browser()` constructor! @plotly/dash-core Please let me know!

Example output [from a recent e2e run](https://app.circleci.com/pipelines/github/plotly/dash-deployment-server/5605/workflows/6ee1768b-b212-4fbc-ac36-15fefdc0c140/jobs/67254):

```
INFO     dash.testing.browser:browser.py:80 initialize browser with arguments
INFO     dash.testing.browser:browser.py:81   headless => False
INFO     dash.testing.browser:browser.py:82   download_path => /tmp/pytest-of-circleci/pytest-0/test_dds_snapshot_engine_pdf0/aws_ha_eks_download
INFO     dash.testing.browser:browser.py:83   percy asset root => /home/circleci/dds/tests/integration/tests/assets
```

These (or similar messages) are repeated a total of 192 times in the run above!

## Contributor Checklist

- [ ] I'll test this PR to make sure it works for us on Wednesday!

### DIFF
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -77,10 +77,10 @@ class Browser(DashPageMixin):
             )
             self.percy_runner.initialize_build()
 
-        logger.info("initialize browser with arguments")
-        logger.info("  headless => %s", self._headless)
-        logger.info("  download_path => %s", self._download_path)
-        logger.info("  percy asset root => %s", os.path.abspath(percy_assets_root))
+        logger.debug("initialize browser with arguments")
+        logger.debug("  headless => %s", self._headless)
+        logger.debug("  download_path => %s", self._download_path)
+        logger.debug("  percy asset root => %s", os.path.abspath(percy_assets_root))
 
     def __enter__(self):
         return self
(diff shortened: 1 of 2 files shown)


scjody:
PS the [CI failures](https://app.circleci.com/pipelines/github/plotly/dash/2053/workflows/2db13e80-769d-462b-b80a-527a58b1d92f/jobs/32092) are caused by a recent change to the cryptography library. Pinning it to an older version [like I'm doing here](https://github.com/plotly/dash-deployment-server/commit/575feb950092e22612ce882b514b7a7c006f3347) in DDS seems like a useful workaround.

alexcjohnson (approved):
💃 Thanks @scjody - those logs definitely don't need to be seen except in deep debugging.

In dash-land the `cryptography` issue only seems problematic on Py3.6 - presumably there's just no wheel published there? So for now I'll attempt to just restrict it before Py3.7 (which will affect 2.7 too but I'm done with Py2-specific code if at all possible!)

PR #1553, opened 2021-02-10
Adjust z-index of dash error menu

tcbegley:
Hello,

This very small PR adjusts the z-index of the dash error menu container. The motivation is that the current value (1001) is lower than some common Bootstrap components resulting in the error menu being obscured in some cases. The largest z-index in Bootstrap CSS is 1070 (used for tooltips), I rounded this up somewhat arbitrarily to 1100, but could reduce to 1071 I guess.

Here is a simple example app demonstrating the issue observed with a sticky navbar

```python
import dash

app = dash.Dash(external_stylesheets=[dbc.themes.BOOTSTRAP])

navbar = dbc.NavbarSimple(
    brand="dbc.NavbarSimple", sticky="top", color="primary", id="navbar"
)

app.layout = html.Div(navbar, id="layout")


# arbitrary callback to force error menu to appear
@app.callback(Output("layout", "children"), Input("navbar", "color"))
def raise_error(_):
    raise RuntimeError


if __name__ == "__main__":
    app.run_server(debug=True)
```

This was originally raised by a community member in facultyai/dash-bootstrap-components#515. I have verified that my change resolves the issue.

Happy to discuss or make any additional changes you would like.

## Contributor Checklist

   -  [x] Adjust the z-index of dash-error-menu
- [x] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 - [#1531](https://github.com/plotly/dash/pull/1531) Update the format of the docstrings to make them easier to read in the reference pages of Dash Docs and in the console.  This also addresses [#1205](https://github.com/plotly/dash/issues/1205)
+- [#1553](https://github.com/plotly/dash/pull/1553) Increase the z-index of the Dash error menu from 1001 to 1100 in order to make sure it appears above Bootstrap components.
 
 ### Fixed
 - [#1546](https://github.com/plotly/dash/pull/1546) Validate callback request `outputs` vs `output` to avoid a perceived security issue.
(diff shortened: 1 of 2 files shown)


alexcjohnson:
Thanks @tcbegley - I can't think of anything we want in front of the debug menu, other than temporary items like the [plotly.js drag cover slip that has a ridiculously high z](https://github.com/plotly/plotly.js/blob/f125c2482737ea37ac0b0ccdd294ce99804e3dc2/src/components/dragelement/index.js#L267) so we could even go higher if you think it might be relevant for other components. But absent any specific other value 1100 is great. @wbrgss anything from your side?

Please add a brief changelog entry indicating that this is for bootstrap compatibility and we'll be good to go!

wbrgss:
Looks right to me @alexcjohnson. Thanks for the contribution @tcbegley!

tcbegley:
Thanks both! Have added a line to the changelog.

alexcjohnson (approved):
💃 excellent, thanks again @tcbegley

PR #1558, opened 2021-02-22
Introduce dict callback

summerswallow-whi:
This is the dict callback I plan to introduce to plotly dash.

I believe it is more pythonic. Definitely easier to maintain code. I refactored it to support pattern matching as well.

I still need to write some tests up to test the various scenarios.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1052,6 +1052,135 @@ class Dash(object):
 
         return wrap_func
 
+    def dict_callback(self, *_args, **_kwargs):
+        """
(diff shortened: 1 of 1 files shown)


summerswallow-whi:
This was a mistake. Meant to PR to a our local copy for internal review first before submitting... Sorry

alexcjohnson:
@summerswallow-whi very interesting... in fact we have some similar functionality in the works as part of an extension to the existing `@app.callback`, hoping to roll it out sometime in the spring. cc @jonmmease

PR #1563, opened 2021-03-06
Support code injection in webpack 5

Marc-Andre-Rivet:
The webpack-dash-dynamic-import plugin was written with Webpack 4.x in mind. With Webpack 5.x the API for plugins changed in a way that makes the existing code injection produce invalid JS bundles. This PR makes the injection work with both 4.x and 5.x.

Related to https://github.com/plotly/dash-table/pull/875
Sanity check against Webpack 4.x project: https://app.circleci.com/pipelines/github/plotly/dash-core-components?branch=sanity-webpack-4and5

### DIFF
--- a/@plotly/webpack-dash-dynamic-import/LICENSE
+++ b/@plotly/webpack-dash-dynamic-import/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Plotly
+Copyright (c) 2021 Plotly, Inc
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/@plotly/webpack-dash-dynamic-import/LICENSE
+++ b/@plotly/webpack-dash-dynamic-import/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 Plotly
+Copyright (c) 2021 Plotly
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/@plotly/webpack-dash-dynamic-import/src/index.js
+++ b/@plotly/webpack-dash-dynamic-import/src/index.js
@@ -1,5 +1,4 @@
 const fs = require('fs');
-const webpack = require('webpack');
 
 function getFingerprint() {
     const package = fs.readFileSync('./package.json');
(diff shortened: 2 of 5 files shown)


alexcjohnson:
Nice find! Can you point to where this change is documented by Webpack? And I presume with this change we will support independently upgrading the renderer and individual component packages from 4 to 5?

@plotly/webpack-dash-dynamic-import/LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 Plotly
```
alexcjohnson, line 3: FYI, as I wrote in https://github.com/plotly/dash-core/issues/324

> I've seen some major packages like React drop the copyright year - that simplifies managing these and apparently doesn't have a strong downside, so I'd propose we do the same.

Marc-Andre-Rivet:
> Nice find! Can you point to where this change is documented by Webpack? And I presume with this change we will support independently upgrading the renderer and individual component packages from 4 to 5?

I can't find a reference to this in the plugins documentation but it seems to boil down to a change in the internal implementation with `Template.asString(<array>)` being resolved in 4.x with https://github.com/webpack/webpack/blob/444e59f8a427f94f0064cae6765e5a3c4b78596d/lib/Template.js#L164 and being resolved with `${array}` in 5.x https://github.com/webpack/tapable/blob/master/lib/SyncWaterfallHook.js#L17

@plotly/webpack-dash-dynamic-import/LICENSE
Marc-Andre-Rivet: Will bump to 2021 for now and until the discussion on this happens

alexcjohnson:
OK so if in 4.x the array was eventually evaluated as `source + '\n' + resolveImportSource()`, and now in 5.x we can't do that but `source + resolveImportSource()` works, is there one of those two flavors that works in both 4 and 5, so we don't need to branch?

Marc-Andre-Rivet:
> OK so if in 4.x the array was eventually evaluated as `source + '\n' + resolveImportSource()`, and now in 5.x we can't do that but `source + resolveImportSource()` works, is there one of those two flavors that works in both 4 and 5, so we don't need to branch?

Yes 😁

@plotly/webpack-dash-dynamic-import/src/index.js
```diff
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const webpack = require('webpack');
```
alexcjohnson, line 2: 🔪

alexcjohnson (approved):
Beautiful 💃 
Just one little 🔪 comment

PR #1567, opened 2021-03-09
Fix `dash-generate-components` for julia on case-insensitive filesystems

etpinard:
The problematic scenario occurs when

- the case-lowered component name equals the module name
- the `--prefix` option is left empty `''`

See example: https://github.com/etpinard/dash-textarea-autocomplete/tree/main/src

Fixes https://github.com/etpinard/dash-textarea-autocomplete/issues/7

---

The commit below prepends `comp_` in jl component file names. There are other possible solutions to this problem, but I thought this one was the least-invasive. Let me know!

## Contributor Checklist

### optionals


alexcjohnson:
Nice catch @etpinard 🍻  I wonder though if we can scope this to just the problematic case, ie a component whose name (prefixed or not) matches the module name?

Although to play devil's advocate, in principle you could also imagine someone making a module `MyThing` with components named `MyThing` and `comp_MyThing` or naming the package `comp_MyThing` with components `MyThing` and `comp_MyThing` 😈  Perhaps the most bulletproof solution would be not a filename prefix but a subdirectory to put all the generated components into, like `jl/mything.jl`? That would also be slightly nicer for devs working on the JS side, to not have as many `.jl` files to look past. Just have to be sure this would still work in case the user makes their own `src/jl/` dir and puts JS files into it.

cc @waralex

etpinard:
> Perhaps the most bulletproof solution would be not a filename prefix but a subdirectory to put all the generated components into, like `jl/mything.jl`?

Yep, that's a nice idea. Looking forward to hearing what @waralex has to say here. I'll make the changes accordingly once you have reach a consensus.  Cheers!

waralex:
> > Perhaps the most bulletproof solution would be not a filename prefix but a subdirectory to put all the generated components into, like `jl/mything.jl`?
> 
> Yep, that's a nice idea. Looking forward to hearing what @waralex has to say here. I'll make the changes accordingly once you have reach a consensus. Cheers!

Yes, placing files with components in a subdirectory looks like the most reasonable solution.

etpinard:
> Perhaps the most bulletproof solution would be not a filename prefix but a subdirectory to put all the generated components into, like jl/mything.jl?

Attempted in https://github.com/plotly/dash/pull/1567/commits/90dfd9f069aa8d9171b0aafda1dccc1e193cb3c1

I believe this commit makes this block

https://github.com/plotly/dash/blob/e8ac94919105a91c76a966c21aca2ec7b0297e22/dash/development/_jl_components_generation.py#L515-L518

obsolete. Alternatively, we could move the `os.makedirs("src/jl")` statement into `generate_module`, but then we would need to patch [`generate_components`](https://github.com/plotly/dash/blob/e8ac94919105a91c76a966c21aca2ec7b0297e22/dash/development/component_generator.py#L42) to make `src/jl/` available for `generate_struct_file` which is currently called  before `generate_module`.

> Just have to be sure this would still work in case the user makes their own src/jl/ dir and puts JS files into it.

I tested two cases:

- case with no `src/jl/` using [`dash-textarea-autocomplete#v1.0.0`](https://github.com/etpinard/dash-textarea-autocomplete/tree/v1.0.0)
- case with a js file in `src/jl/` imported into the component using [`dash-textarea-autocomplete#jl-comp-subdir`](https://github.com/etpinard/dash-textarea-autocomplete/tree/jl-comp-subdir)

with:

```
. venv/bin/activate
pip install -r requirements.txt 
pip install ../../dash
npm run build
julia --project -e 'import Pkg; Pkg.add("DashHtmlComponents")'
julia --project usage.jl
```

where the `build` script got generated from [`dash-component-boilerplate`](https://github.com/plotly/dash-component-boilerplate).

alexcjohnson (approved):
Agreed that we should be able to 🔪  `os.makedirs("src")` - then this looks good! 💃

PR #1576, opened 2021-03-25
Update toolchain

Marc-Andre-Rivet:
Fixes partially plotly/dash-core#203

py37 Percy run with all the changes, prior to bumping to py39: https://percy.io/plotly/dash/builds/9542243/unchanged/538975044?browser=firefox&viewLayout=side-by-side&viewMode=new&width=1280

- Followup: remove `py37` build checks in GH
- Update Webpack and Eslint (transition from deprecated Tslint)
- Port tests from Jest to Karma+Mocha: two pluses for this is that it now actually runs the tests in the browser instead of node and it's a more stable stack

PR #1578, opened 2021-03-27
Standardize and share Dash linting rules

Marc-Andre-Rivet:
Addition to https://github.com/plotly/dash/pull/1576 that allows standardizing `eslint` and `prettier` configurations across Dash projects.

- Adds `@plotly/eslint-config-dash` that exposes two base linting configurations for JavaScript and TypeScript.
- Adds `@plotly/prettier-config-dash` that exposes a basic Prettier configuration

These proposed configurations are based on the intersection of the Core's projects current configurations.

Both `@plotly/eslint-config-dash` and `@plotly/prettier-config-dash` will need to be published to NPM to be useable and to close this PR.

Percy visual tests will fail because of the un-merged upstream change from py37 to py39.


dash-renderer/.eslintrc
```diff
@@ -0,0 +1,12 @@
+{
+    "extends": [
+      "@plotly/eslint-config-dash/typescript",
+    ],
+    "rules": {
+      "no-constant-condition": 0,
+      "no-prototype-builtins": 0,
+      "no-unused-vars": 0,
```
Marc-Andre-Rivet, line 8: The renderer's configuration has some peculiarities that shouldn't be generalized to all projects. Also `no-unused-vars` just acknowledges the current state of the JavaScript code in this project not adhering to this rule.. there's a lot of dangling unused variables in JS files at the moment 😢

alexcjohnson: Hmm OK, would be nice to tighten up `no-unused-vars` at some point, and see whether the others actually have a legitimate reason to be violated. Care to make an issue to follow up on this later?

@plotly/prettier-config-dash/README.md
```diff
@@ -0,0 +1,3 @@
+# @plotly/prettier-config-dash
+
+Shareable Prettier configuration for Dash projects.
```
alexcjohnson, line 3: Not a big deal, but is it important that this is a separate package? Feels like it would be lighter to use if this were in the same package as the eslint config, and they're intended to be used in tandem.

alexcjohnson (approved):
LGTM, just a couple of small comments. 💃

@plotly/prettier-config-dash/README.md
Marc-Andre-Rivet: The prettier configuration could be added to the `eslint-...` one but it would need to be the default module export (so including "@plotly/eslint-config-dash" would be invalid and would return the Prettier configuration). It couldn't be the other around because eslint configs must be published under `@<scope>/eslint-config-<package>` or `eslint-config-<package>`

dash-renderer/.eslintrc
Marc-Andre-Rivet: I see, I missed a detail here: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md
```
// note you must disable the base rule as it can report incorrect errors
  "no-unused-vars": "off",
```

Also used the occasion to fix the `.eslintrc` config so that it is loaded correctly by VSCode when it referenced a tsconfig.

https://github.com/plotly/dash/pull/1578/commits/2fef7ed75398706233282d98cdaadb975af425b3

@plotly/prettier-config-dash/README.md
alexcjohnson: OK. I was hoping we could do something like `"prettier": "@plotly/eslint-config-dash/prettier"` but even if that works it does look a bit odd to have prettier explicitly nested under eslint.

Issue #1617, opened 2021-05-01
provide one-way binding


Marc-Andre-Rivet:
Proposing a new app callback definition that takes one input and one output and one-way binds a prop to another.

Would look like:

```
app.bind(Input('text', 'value'), Output('div', 'children'))
```

Would be run clientside.
- a non-children input would be wrapped in an array as required if the output prop is `children`
- allow MATCH / MATCH and ALL / ALL pattern matching callbacks

Could later support two-way binding with an additional argument and by allowing cycles of length 2 involving only bindings in the renderer, with callback looping prevention.

AnnMarieW:
The linked PR was closed without being merged, but this issue remained open. Closing it now.

PR #1619, opened 2021-05-01
Issue 1617 - One way binding

Marc-Andre-Rivet:
Basic implementation of #1617 demonstrating feasibility.

Adds client-side handling of one-way binding with support for:

- one input, one output
- MATCH input, MATCH output
- ALL input, ALL output

Allowed pattern-matching callbacks are not enforced in this PR and would be a subset of standard allowed patterns I'd think.

This is not meant as a full PR but rather a proof of concept for discussion.
Full implementation would need to enforce the allowed PMC patterns and possibly allow simple type conversions (e.g. children -> value; value -> children, stringify, format, etc.) either automagically or user-defined, or type inference from components prop types. This could be brought further and props could be assumed based on component type: for example a dcc.Slider linked to html.Div would most probably want to bind it's value to the div's children prop, allowing something like `app.bind('my_slider_id', 'my_div_id')` or inlined `Div(id='my_div').bind('my_slider')` (yes the language creator dislikes fluent interfaces -- https://mail.python.org/pipermail/python-dev/2003-October/038855.html)

Example app
```
import dash
from dash.dependencies import Input, Output, MATCH

slider_props = dict(min=0, max=20, step=1, value=5)
app = dash.Dash(__name__)

app.layout = html.Div([
    dcc.Slider(id='slider', **slider_props),
    dcc.Input(id='input'),
    dcc.Slider(id={'type': 'slider', 'index': '0'}, **slider_props),
    dcc.Input(id={'type': 'input', 'index': '0'}),
    dcc.Slider(id={'type': 'slider', 'index': '1'}, **slider_props),
    dcc.Input(id={'type': 'input', 'index': '1'})
])

app.bind(Input('slider', 'value'), Output('input', 'value'))
app.bind(Input({ 'type': 'slider', 'index': MATCH }, 'value'), Output({ 'type': 'input', 'index': MATCH }, 'value'))

app.run_server(debug=True)
```

Might become
```
import dash
from dash.dependencies import Input, Output, MATCH

slider_props = dict(min=0, max=20, step=1, value=5)
app = dash.Dash(__name__)

app.layout = html.Div([
    dcc.Slider(id='slider', **slider_props)
    dcc.Input(id='input').bind('slider'),
    dcc.Slider(id={'type': 'slider', 'index': '0'}, **slider_props),
    dcc.Input(id={'type': 'input', 'index': '0'}).bind({'type': 'slider', 'index': MATCH }),
    dcc.Slider(id={'type': 'slider', 'index': '1'}, **slider_props),
    dcc.Input(id={'type': 'input', 'index': '1'}).bind({'type': 'slider', 'index': MATCH })
])
app.run_server(debug=True)
```


alexcjohnson:
Another challenge with `app.layout = [..., Component().bind(other), ...]` is it'll fail if used in a callback, or even layout as a function. Adding (and removing) callbacks/bindings on the fly may be something we want to support someday, but let's consider that first and then come back to this syntax.

Re: `app.bind(Input, Output)` - seems like a nice shorthand, but I'm curious why we'd implement it in the renderer rather than just converting to the equivalent `app.clientside_callback` call? Seems like two-way binding could also be handled this way, perhaps just with the syntax `app.bind(Output, Output)` using the new limited circular callback functionality and `callback_context.triggered`.

Marc-Andre-Rivet:
> Another challenge with app.layout = [..., Component().bind(other), ...] is it'll fail if used in a callback, or even layout as a function. Adding (and removing) callbacks/bindings on the fly may be something we want to support someday, but let's consider that first and then come back to this syntax.

Fully agree. Although I don't see any obvious hard stopper for adding callbacks on `_dash-update-component`, only validating the new additions.

> Re: app.bind(Input, Output) - seems like a nice shorthand, but I'm curious why we'd implement it in the renderer rather than just converting to the equivalent app.clientside_callback call? Seems like two-way binding could also be handled this way, perhaps just with the syntax app.bind(Output, Output) using the new limited circular callback functionality and callback_context.triggered.

My thinking here is that not having to generate / inject JS here to be loaded is preferable with the dynamic scenario above. Also, it's much easier from within the renderer to do custom logic and interpolation as we have access to full prop-type and component type information, which one could arguably do in <flavor> language but would need to be duplicated for each flavor and would most likely be less robust.

alexcjohnson:
> I don't see any obvious hard stopper for adding callbacks on `_dash-update-component`, only validating the new additions.

We'd need to also delete these callbacks later if the chunk they came in with is modified. Perhaps these callbacks get annotated with a "mount point" ie the path to the chunk they were added with, and if that path or any of its parents is replaced in the layout, these callbacks are removed (or at least reconciled with any new callbacks coming in with the new chunk)

> My thinking here is that not having to generate / inject JS here to be loaded is preferable with the dynamic scenario above.

If we're thinking about the dynamic case I'd rather work toward full support, it would be strange to only allow it for this very narrow case. That said you're right that building in native renderer support doesn't seem like it's too hard.

> much easier from within the renderer to do custom logic and interpolation as we have access to full prop-type and component type information

You're thinking about default prop inference or something else? That feels a bit too magical to me, at least that's the conclusion we reached when iterating on dash-labs.

Marc-Andre-Rivet:
> We'd need to also delete these callbacks later if the chunk they came in with is modified.

I was thinking they would just become stale if they were missing from the layout. And reactivate if present again later.

> That said you're right that building in native renderer support doesn't seem like it's too hard.

Yes. I'd say the PR pretty much demonstrates that. 😸 

> You're thinking about default prop inference or something else? That feels a bit too magical to me, at least that's the conclusion we reached when iterating on dash-labs.

Yes. but if it's been discussed already then 🔪 that part. It does feel a bit magical, but magical is sometimes nice -- we would definitely want to explain what was done if there's an error on those callbacks.

> Seems like two-way binding could also be handled this way, perhaps just with the syntax app.bind(Output, Output) using the new limited circular callback functionality and callback_context.triggered.

I did see the PR but I had completely forgotten about it. If as you say that's an easy stretch, that would be nice.

gvwilson:

T4rk1n:
This is stale.
I guess this could be done entirely with a wrapper like:
```
def bind(input, output):
     clientside_callback(output, input, "value => value", prevent_initial_call)
```

Issue #1519, opened 2021-01-07
[BUG] Dependent callbacks are not waited and executed in order


fraghag:
**Describe your context**
I have a problem with chained callbacks where the last callback is called twice from one UI change. In the documentation I found this
https://dash.plotly.com/basic-callbacks
where it states under “Dash App With Chained Callbacks” that:

```
The final callback displays the selected value of each component. If you change the value of the countries RadioItems component, Dash will wait until the value of the cities component is updated before calling the final callback. This prevents your callbacks from being called with inconsistent state like with "America" and "Montréal" .
````

But when changing between America and Canada in the example on this page I can see that it first updates the text with the country and then the city. So when changing from America to Canada I can see “New York City is a city in Canada” very briefly which according to the documentation should not happen. I also ran this example locally with debugging and verified that the callback updating the text is run twice whenever I change the country.

```
dash                               1.18.0
dash-bootstrap-components          0.10.7
dash-core-components               1.14.0
dash-html-components               1.1.1
dash-renderer                      1.8.3
dash-ri-components                 2.1.0
dash-table                         4.11.1
```
**Describe the bug**
Callback A and callback B has resource A as a dependency, callback B is also dependent on the output of callback A.
When changing resource A, then callback A can be run before callback B is finished, and then runs again when callback A finishes since callback A updates another of callback B's input.

![image](https://user-images.githubusercontent.com/5681047/103886075-3e6b1180-50e1-11eb-9748-c08c8c5b470d.png)

**Expected behavior**

When chaining callbacks, callback B should wait for callback A to finish, if callback B is dependent on the output of callback A.

alexcjohnson:
Thanks @fraghag - I see it too and you're absolutely right that this is a bug. We'll have to poke around and see if this is a regression or simply a case that for some reason we've never handled correctly.

francescobodria:
I have the same problem with an app that I am developing.
I have two chained callbacks, one change the values of several sliders (10) together and the other one implements a function on these sliders values. The problem is that I need to evaluate the function when **all** the slider values have changed but I found that it fires up ten times, once every time a single slider value is updated. This causes a problem since it fires up with only partially changed values and raises errors.

eff-kay:
Hey everyone, I have started working on this issue. 

First of all can someone confirm by looking at the following screen recording if I am reproducing the issue correctly. 

https://www.loom.com/share/daaf5d30f76e455aa46705829fdd739d

eff-kay:
[...]

PR #1627, opened 2021-05-07
fix for callbacks waiting issue

eff-kay:
*This is a potential fix for [1519](https://github.com/plotly/dash/issues/1519), contributed only for discussion purposes. Not the final fix. The following checklist is not relevant for now*

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals

### DIFF
--- a/dash-renderer/src/APIController.react.js
+++ b/dash-renderer/src/APIController.react.js
@@ -117,6 +117,7 @@ const UnconnectedContainer = props => {
 };
 
 function storeEffect(props, events, setErrorLoading) {
+
     const {
         appLifecycle,
         dependenciesRequest,
--- a/dash-renderer/src/APIController.react.js
+++ b/dash-renderer/src/APIController.react.js
@@ -117,7 +117,6 @@ const UnconnectedContainer = props => {
 };
 
 function storeEffect(props, events, setErrorLoading) {
-
     const {
         appLifecycle,
         dependenciesRequest,
--- a/dash-renderer/src/TreeContainer.js
+++ b/dash-renderer/src/TreeContainer.js
@@ -127,7 +127,6 @@ class BaseTreeContainer extends Component {
             newProps
         );
 
-        console.log("props", newProps, "old", oldProps);
         if (!isEmpty(changedProps)) {
             // Identify the modified props that are required for callbacks
             const watchedKeys = getWatchedKeys(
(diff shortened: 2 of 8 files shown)


Marc-Andre-Rivet:
The behavior described is indeed what is happening :disappointed:. I was initially surprised at the behavior as it is pretty clear that the display children callback should wait on citites values update from the cities options callback before executing.

For the immediate scenario in the issue: https://github.com/plotly/dash/blob/dev/dash-renderer/src/observers/requestedCallbacks.ts#L234-L237 is too eager at declaring callbacks ready as it doesn’t check for subsequent calls a callback may trigger down the chain (https://github.com/plotly/dash/blob/20b67ab6d252125c129ed7ef84df833cb8f24082/dash-renderer/src/actions/dependencies_ts.ts#L145-L182) .

It seems to me the issue could be handled by ensuring that all subsequent callbacks of country values are guaranteed not to cause a change in an input displayed children uses instead of just checking against active callbacks — the logic to access this information already exists for the prioritized -> executing prioritization scheme which involves the depth of the callback tree for a given callback. Repurposing the logic of https://github.com/plotly/dash/blob/20b67ab6d252125c129ed7ef84df833cb8f24082/dash-renderer/src/actions/dependencies_ts.ts#L99 to extract all affected props in a callback’s tree.

If a prop change from a callback triggers a component initiated prop change it’s true the renderer can’t be aware of it. I agree there no clear way to solve this problem unless we make the renderer aware of these internal component links, which doesn’t seem desirable. In a way what we are saying with the current behavior is that user initiated changes and component initiated changes are equivalent — both are callbacks triggered by the UI are it’s as impossible to predict user interactions as it is to predict component interactions.

image

eff-kay:
I have pushed the new changes. In summary I am updating the getReadyCallbacks function.

https://github.com/plotly/dash/pull/1627/files#diff-9ef142ebb62ec253df4d4611f96d54124ca69347d9d27fd78c5ae6e1202f5ef4R249

The important changeset is the if block that starts from 281. Basically, now I am looking at all of the outputs potentially touched by the active callbacks and populating the outputMap from that. This seem to resolve the defect.

eff-kay:
I have added a test that I think would actually validate the issue reported. Let me know if you think of a scenario in which this would pass without the fix.

https://github.com/plotly/dash/pull/1627/files#diff-d94c2be667bd3581c7fbb922b85c16d0ce1ad2bdc25a47444c9eaf5650421b85R590

ndrezn:
This PR was superseded by: https://github.com/plotly/dash/pull/2344, which was released.

PR #1637, opened 2021-05-09
Expose  at top level and add optional  parameter

### LINKED ISSUE #1625: [Feature Request] Python function returning stringified dict component IDs
When you use pattern matching callbacks and component IDs of type dictionaries, Dash will automatically stringify the IDs before being sent to the browser/DOM.

As far as I know, even if you use the Dash pytest/selenium plugin, if you want to do certain tests on these components (and select them by ID), you would need to find out + assume how the dictionary ID → string ID conversion is done by Dash.

A python function callable from somewhere in the `dash` package which does the conversion, and thereby can safely be used in tests of Dash apps, across `dash` versions. E.g. something like:
```python
def stringify_id(dict_id: Dict[str, str]) -> str:
    ...


stringify_id({"id": "some_id", "type": "some_type"}) == '{\"id\"\:\"some_id\"\,\"type\"\:\"some_type\"\}'
```

**Additional context**
It looks like the conversion internally in Dash today is done on the JavaScript side:
https://github.com/plotly/dash/blob/597dc22c86de97530fb9b0a9bd9f7876bd0d7601/dash-renderer/src/actions/dependencies.js#L111-L123

One way of supporting the use case above would be to add a Python function doing the same stringify logic, and then add new selenium tests to the Dash test suite to ensure the Python-stringify = JavaScript-stringify. Can we do better? Or maybe there is already a solution for this use case (of testing elements with dict IDs) using e.g. `dash_duo`?

anders-kiaer:
Closes #1625 (except the `dash-docs` part - but I can take that after review and if accepted).

## Contributor Checklist

   -  [X] Expose `stringify_id` at top level.
   -  [X] Add `css_escape` function.

### optionals


tests/integration/test_css_select_by_id.py
```diff
@@ -0,0 +1,45 @@
...
+def test_found_by_css_selector(dash_duo, id):
+    # This test also indirectly checks that output from the JavaScript stringify_id
+    # function equals output from the corresponding Python function.
+    app = dash.Dash(__name__)
+
+    app.layout = html.Div(id=id)
+
+    dash_duo.start_server(app)
+
+    assert dash_duo.wait_for_element_by_id(dash.stringify_id(id, escape_css=True))
```
anders-kiaer, line 44: There is an easy possibility of extending
https://github.com/plotly/dash/blob/e8ac94919105a91c76a966c21aca2ec7b0297e22/dash/testing/browser.py#L277-L288
to check if input `element_id` is dict, and then do `dash.stringify_id(element_id, escape_css=True)` for the user in the `wait_for_element_by_id` function.

The user's test line here
```python
assert dash_duo.wait_for_element_by_id(dash.stringify_id(id, escape_css=True))
```
then becomes
```python
assert dash_duo.wait_for_element_by_id(id)
```
both in the case with string IDs, and dictionary IDs (when you use pattern-matching callbacks).

Not sure if that is too magical :crystal_ball: or a nice and easy to understand convenience.

gvwilson:
hi @T4rk1n - is this one still relevant or has it gone stale?

dash/_utils.py
```diff
@@ -192,10 +192,72 @@ def split_callback_id(callback_id):
     return {"id": id_, "property": prop}
 
 
-def stringify_id(id_):
-    if isinstance(id_, dict):
-        return json.dumps(id_, sort_keys=True, separators=(",", ":"))
-    return id_
+def css_escape(stringified_id):
```
T4rk1n, line 195: I would maybe rename to `escape_css_id`.

dash/__init__.py
```diff
@@ -3,6 +3,7 @@
 # must come before any other imports.
 __plotly_dash = True
 from .dash import Dash, no_update  # noqa: F401,E402
+from ._utils import stringify_id  # noqa: F401,E402
```
T4rk1n, line 6: Could also expose `css_escape`

tests/integration/test_css_select_by_id.py
T4rk1n: Good idea with extending the testing utilities, I think that is where most cases would be for escaping.

T4rk1n (commented):
Sorry this PR fell under the cracks, I think this is still good, just need to update with dev branch which include new code for [`stringify_id`](https://github.com/plotly/dash/blob/c847882b6bb8d5b86af5df6dad97cab61c0ae2c2/dash/_utils.py#L176-L183)

gvwilson:
thanks @T4rk1n - I'll add it to the pile for the next cycle (no need to work on it right now). Cheers - @gvwilson

Issue #1624, opened 2021-05-06
[BUG] Handle missing `timing_information` gracefully


remibaar:
**Describe your context**
```
dash                      1.17.0                  
dash-core-components      1.13.0             
dash-html-components      1.1.1              
dash-renderer             1.8.3              
```

**Describe the bug**
An error is raised when the devtools before request is not fully executed. 
```AttributeError: '_AppCtxGlobals' object has no attribute 'timing_information'``` 
In our case this happens because we register our own "before request" to the flask instance, to handle user logins. Flask will not execute any other `before_request_funcs` when a function returns something other than None. If an user is not logged in, we will throw a redirect to the login page.

So during the "before_request" phase the `timing_information` is not set. During the "after_request" phase, it expects `timing_information` was set. If not it will throw an error.

The error is caused by these lines: https://github.com/plotly/dash/blob/master/dash/dash.py#L1465-L1489

**Expected behavior**
It should check if the `timing_information` is set, and not expect it is. If it is not it gracefully should handle the error.

anders-kiaer:
Similar to #1475?

remibaar:
Indeed it is a duplicate of #1475
I will close this issue, and create a PR to fix it

Issue #1475, opened 2020-12-01
[BUG] `debug=True` and `@app.server.before_request` → `AttributeError: '_AppCtxGlobals' object has no attribute 'timing_information'`


anders-kiaer:
**Describe your context**
Please provide us your environment so we can easily reproduce the issue.

-  replace the result of `pip list | grep dash` below
```
dash                                  1.17.0
dash-core-components                  1.13.0
dash-html-components                  1.1.1
dash-renderer                         1.8.3
dash-table                            4.11.0
```

**Describe the bug**

With `debug=True`:

```
Traceback (most recent call last):
  File "[...]/python3.7/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "[...]/python3.7/site-packages/flask/app.py", line 2450, in wsgi_app
    response = self.handle_exception(e)
  File "[...]/python3.7/site-packages/flask/app.py", line 1867, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "[...]/python3.7/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "[...]/python3.7/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "[...]/python3.7/site-packages/flask/app.py", line 1953, in full_dispatch_request
    return self.finalize_request(rv)
  File "[...]/python3.7/site-packages/flask/app.py", line 1970, in finalize_request
    response = self.process_response(response)
  File "[...]/python3.7/site-packages/flask/app.py", line 2267, in process_response
    response = handler(response)
  File "[...]/python3.7/site-packages/dash/dash.py", line 1469, in _after_request
    dash_total = flask.g.timing_information["__dash_server"]
  File "[...]/python3.7/site-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
AttributeError: '_AppCtxGlobals' object has no attribute 'timing_information'
```

**What happens**

In our use case we use the Dash provided Flask server (`app.server`) to set up a `before_request`. On certain conditions (more specifically, certain `GET` parameters in the URL) we do a `flask` redirect. 

In `dash<1.16` the minimal example below works, however in `dash>=1.16` it fails with the above traceback.

**A minimal example**

```python
import flask
import dash
import dash_html_components as html


app = dash.Dash()

app.layout = html.Div("Hello world")

@app.server.before_request
def some_function():
[...]

PR #1640, opened 2021-05-13
[Bugfix] Handle missing `timing_information` gracefully

remibaar:
An error is raised when the devtools before request is not fully executed.
`AttributeError: '_AppCtxGlobals' object has no attribute 'timing_information'`

I've created a test to replicate, and a fix to handle the error gracefully.

More info in #1624 and #1475

## Contributor Checklist

   -  [x] Create test to replicate
   -  [x] Fix bug


remibaar:
Percy test is failing because of a new string "This is a Copy/Paste friendly version of the traceback." But that's (probably) not due to the changes I've made

dash/dash.py
```diff
@@ -1468,10 +1468,16 @@ def _before_request():
...
-                dash_total = flask.g.timing_information["__dash_server"]
-                dash_total["dur"] = round((time.time() - dash_total["dur"]) * 1000)
+                timing_information = flask.g.get("timing_information", None)
+                if timing_information is None:
+                    return response
 
-                for name, info in flask.g.timing_information.items():
+                timing_information_dash = timing_information.get("__dash_server", None)
+                if timing_information_dash is not None:
+                    dash_total = timing_information["__dash_server"]
```
alexcjohnson, line 1477:
```suggestion
                dash_total = timing_information.get("__dash_server", None)
                if dash_total is not None:
```

alexcjohnson (approved):
Thanks @remibaar - looks good, just a small simplifying suggestion. 💃 

You're right that the Percy change has nothing to do with this PR - strange, we must have inherited a new version of Werkzeug or something? Not a big deal.

remibaar:
Thanks @alexcjohnson  for the simplification! That was exactly how I planned to code it, but for some reason didn't haha. Thanks for catching it.

Do you also agree with the new integration test? I didn't exactly know where to put it, so I've putted in the `test_devtools_ui.py`

alexcjohnson:
The new test is excellent, thanks! I just added a changelog entry, will merge once tests pass.

Issue #1642, opened 2021-05-18
Missing classifiers for Python 3.8/3.9


samlishak-artemis:
Python 3.8 and 3.9 are missing from the classifiers in `setup.py`:

https://github.com/plotly/dash/blob/358c5089c929b2e99996f9d4ee6ec634f65437fe/setup.py#L55-L63

But there is no restriction to 3.7 or below in `python_requires`:
https://github.com/plotly/dash/blob/358c5089c929b2e99996f9d4ee6ec634f65437fe/setup.py#L29

Anecdotally, I've been using Dash on Python 3.9 with no issues. Reporting as an issue as I can't see any mention of Python 3.8 and 3.9 compatibility so far!

alexcjohnson:
Good call, the classifiers haven't been updated in a while. FYI in the upcoming Dash 2.0 we're planning to change `python_requires` to `">=3.6"` - dropping Py2 altogether as well as 3.5 and below.

samlishak-artemis:
I can submit a basic PR to add 3.8 and 3.9 if you like? Not a big job, but it's something 😄

samlishak-artemis:
Never mind, looks like you beat me to it - thanks very much!

PR #1643, opened 2021-05-18
Add py 3.8 and 3.9 to classifiers

alexcjohnson:
Simple classifier update for #1642


alexcjohnson:
FWIW we're already running tests against 3.9 😄

PR #1646, opened 2021-05-19
Dict callback decorator

summerswallow-whi:
The dictionary based callback decorator `@app.dict_callback` operates similarly to the normal `@app.callback` decorator. With two additional  options `strict` and `allow_missing`. The invocation of the decorator is virtually identical. 

## Basic Example with State 

This example is the same basic example given in the documentation
except using `dict_callback`.

```python
import dash

external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']

app = dash.Dash(__name__, external_stylesheets=external_stylesheets)

app.layout = html.Div([
    dcc.Input(id='dcinp-1-state', type='text', value='Montréal'),
    dcc.Input(id='dcinp-2-state', type='text', value='Canada'),
    html.Button(id='dc-submit-button', n_clicks=0, children='Submit'),
    html.Div(id='dc-output')
])


@app.dict_callback(Output('dc-output', 'children'),
                   Input('dc-submit-button', 'n_clicks'),
                   State('dcinp-1-state', 'value'),
                   State('dcinp-2-state', 'value'))
def update_output(inputs, states):
    return {'dc-output.children': u'''
        The Button has been pressed {} times,
        Input 1 is "{}",
        and Input 2 is "{}"
    '''.format(inputs['dc-submit-button.n_clicks'],
               states['dcinp-1-state.value'],
               states['dcinp-2-state.value'])}

```

## The `strict` and `allow_missing` options 

For performance reasons, the `dict_callback` decorated callback doesn't by default check to  make sure all keys in the output dictionary correspond to actual output ids and output  properties declared in the invocation. The `strict` option enforces that. If the output contains  a key not corresponding to a declared output `id` and `property` a `KeyError` is thrown. A good practice is to set `strict` to try while developing and debugging to catch possible errors. The `allow_missing` option is set to `True` by default. This options
allows for an incomplete output dictionary. All missing properties are unchanged. This is particularly useful for a complex callback where helper functions may be responsible for some outputs and not all outputs need to be updated.

In the following example, a `KeyError` is thrown since `output-2-children` is missing from the dictionary but if `allow_missing` were set to `True` it would be allowed. This pattern is very useful. Due to the constraints on outputs belonging to a single callback, you can easily be forced to conglomerate a bunch of callbacks into one. This pattern allows you to subdivide the callback into separate functions and one only needs to merge the dictionaries at the end.

```python
app = dash.Dash(__name__)
app.layout = html.Div([
    dcc.Input(id="input1", value="initial value"),
    dcc.Input(id="input2", value="state"),
    html.Div([html.Div(id="output-1"), html.Div(id="output-2")])
])

@app.dict_callback([Output("output-1", "children"), Output("output-2", "children")],
                   [Input("input1", "value")],
                   [State("input2", "value")], allow_missing=False)
def update_output(inputs, states):
    output = {"output-1.children": inputs['input1.value']
    }
    return output

```
In this example, `strict` is set. This will throw a `KeyError` as there is no output called `output-3.children`. While this is usually a good thing, it requires a separate check so could impact performance.

```python
app = dash.Dash(__name__)
app.layout = html.Div(
    [
        dcc.Input(id="input1", value="initial value"),
        dcc.Input(id="input2", value="state"),
        html.Div([html.Div(id="output-1"), html.Div(id="output-2")])
    ]
)

@app.dict_callback([Output("output-1", "children"), Output("output-2", "children")],
                   [Input("input1", "value")],
                   [State("input2", "value")], strict=True)
def update_output(inputs, states):
    output = {"output-1.children": inputs['input1.value'],
              "output-2.children": states['input2.value'],
              "output-3.children": "Another Value",
    }
    return output
```

Finally, the `dict_callback` decorator works with pattern matching as in this Pattern Matching Example (taken from the documentation). Note that `dict_callback` works with a `callback_dict` which is a subclass of `dict` that provides `pset` and `pget` that allows access to the dictionary using pattern matching keys without having to worry about the underlying hashing or the order.

```python
app = dash.Dash(__name__)

app.layout = html.Div([
    html.Button("Add Filter", id="dynamic-add-filter", n_clicks=0),
    html.Div(id='dynamic-dropdown-container', children=[]),
])

@app.dict_callback(
    Output('dynamic-dropdown-container', 'children'),
    Input('dynamic-add-filter', 'n_clicks'),
    State('dynamic-dropdown-container', 'children'))
def display_dropdowns(inputs, states):
    new_element = html.Div([
        dcc.Dropdown(
            id={
                'type': 'dynamic-dropdown',
                'index': inputs['dynamic-add-filter.n_clicks']
            },
            options=[{'label': i, 'value': i} for i in ['NYC', 'MTL', 'LA', 'TOKYO']]
        ),
        html.Div(
            id={
                'type': 'dynamic-output',
                'index': inputs['dynamic-add-filter.n_clicks']
            }
        )
    ])
    states['dynamic-dropdown-container.children'].append(new_element)
    return states


@app.dict_callback(
    Output({'type': 'dynamic-output', 'index': MATCH}, 'children'),
    Input({'type': 'dynamic-dropdown', 'index': MATCH}, 'value'),allow_missing=False
)
def display_output(inputs, states):
    id_, _ = inputs.pkeys()[0]
    output=dash.Dash.callback_dict()

    div = html.Div('Dropdown {} = {}'.format(id_['index'], inputs.pget(id_,'value')))
    output.pset(type='dynamic-output', index=id_['index'], property='children', value=div)
    return output
```


summerswallow-whi:
I added two tests that are skipped as I do not know how in this framework to assert an exception thrown by the callback wrapper.

alexcjohnson:
Thanks @summerswallow-whi! Some related ideas are actually already being explored in dash-labs, see https://github.com/plotly/dash-labs/blob/main/docs/02-CallbackEnhancements.md#tuple-and-dictionary-argument-grouping

Would you like to take a look at that section, and if you can see avenues to improve it make an issue over there to discuss?
cc @jonmmease

summerswallow-whi:
I'd be happy to join in the conversation. I'd certainly like anyone interest to try out my code. I feel it is very intuitive doesn't change the coding methodology or even callbacks that much. The only complication is when pattern matching was introduced. Originally, it was written as a decorator to decorate the callback decorator.

I jumped to the link. How do I join in?

alexcjohnson:
There are a few threads going on in the community forum that we'd welcome your input on - probably all linked off this one: https://community.plotly.com/t/introducing-dash-labs-dash-2-0-preview/52087

You can also make new issues in the dash-labs repo if you have specific suggestions how to combine those ideas with the ones in this PR. I'd prefer to build from what's already implemented there, since there's a decent amount of overlap and the ideas there have been through many iterations already.

summerswallow-whi:
I think the approaches are quite different and the dict callback as presented is philosophically quite different and not incompatible. But at the forums suggestion. This was developed into a dash plug in at https://github.com/WestHealth/dash-dict-callback. Also available on pypi `pip install dash-dict-callback`.

PR #1670, opened 2021-06-18
Global Dash Proxy Object for Plugins

chriddyp:
WIP / Proposal 

This PR adds a new import `dash.g.app` that can be used to register callbacks in other files.

**Usage**

`special_components.py`
```python
from dash.g import app

def SpecialComponent():
    return html.Div([
        dcc.Input(id='input'),

        html.Div(id='div-1'),
        html.Div(id='div-2'),
    ])

@app.callback(Output('div-1', 'children'), Input('input', 'value'))
def update_1(value):
    return f'Input 1 - {value}'

@app.callback(Output('div-2', 'children'), Input('input', 'value'))
def update_2(value):
    return f'Input 2 - {value}'
```

`app.py`
```python
import special_components

app = dash.Dash(__name__)

app.layout = html.Div([
    SpecialComponent(),
    html.Div(id='div-3')
])

@app.callback(Output('div-3', 'children'), Input('input', 'value'))
def update_3(value):
    return f'Input 3 - {value}'

app.run_server(debug=True)
```

This improves a couple of things:

1. Users don’t need to deal with circular `app` import issues. This is common with multi page apps (https://dash.plotly.com/urls)
2. It enables a more ergonomic API modular/coupled callback x component definitions in standalone files that could be distributed across multiple files or on pip. Current solutions involve passing in the app object (e.g. `my_component.register_callbacks(app)`) which is arduous and doesn’t feel as natural as a regular Dash component.

The implementation was a little tricky. I originally thought we could bind `app` to the `dash` module when the app is defined, like:
```python
class Dash():
    def __init__(self):
         dash.app = self
```

However, for modular callbacks users may use `g.app.callback` before `app = dash.Dash(__name__)` is defined.

So, I settled with an architecture that creates a sort of “proxy” global object that uses the `dash.callback` function (for validation, etc) and mutates its own `_callback_map` and `_callback_list` data structures. Then, before the first request is made to dash we merge the global mutable proxy data structures into the main dash app data structures.

This means that `dash.g.app` only has `@callback` defined. In other words, `dash.g.app` isn’t a true app object - it’s just a simple class that defines `def callback`. So, if users attempt to access other things in `dash.g.app` it’ll fail. I actually think that’s a good thing - It narrows the scope of what `dash.g.app` is meant to do.

Another idea would be to rename `@dash.g.app.callback` to something like `@dash.g.register_callback`. I think this is a bigger conceptual leap however and requires more code modification when transferring from a regular `@app.callback` to a `@dash.g.app.callback`


alexcjohnson:
On the one hand it's nice to be able to keep `@app.callback` in your modules, but it's potentially confusing as well, and feels like a lot of layers just to add delayed callback registration. What about just `@dash.callback`? Then you just need to change `app` to `dash`, but then `callback` isn't inside something called `app` so won't be mistaken for the original.

Another wrinkle is the multi-app pattern - not recommended but it's not forbidden and currently works, but would be very confusing with this new feature. So we should probably at least emit a warning if multiple apps are instantiated either before or after adding callbacks this way.

chriddyp:
Closing in favor of https://github.com/plotly/dash/pull/1710

PR #1677, opened 2021-06-29
Remove references to cleanModules job

HammadTheOne:
This PR provides a minor fix to the `dash-renderer` build process by removing the job to clean `node_modules` following the bundle generation. As pointed out by @alexcjohnson , this job is unnecessary and causes the following error:

```
> rm -rf lib && babel src --extensions=".ts,.tsx,.js,.jsx" --out-dir lib --copy-files
sh: babel: command not found
```


alexcjohnson (approved):
Thanks! 💃

PR #1679, opened 2021-06-30
Dash monorepo 

HammadTheOne:
This PR is the initial monorepo implementation. It integrates the `dash-core-components`, `dash-html-components`, and `dash-table` component libraries and their dependencies within the Dash namespace, and adds some tools for maintaining the monorepo.

- [x] Add `dcc`, `html`, and `dash_table` to the Dash namespace
- [x] Add dash dependencies to the Dash namespace 
- [x] Set up Lerna monorepo, and packages directory.
- [x] Added CLI script `dash-update-components` to access and update packages within Dash.
- [x] Update build process and Python package to include dcc, html, table dependencies.
- [x] Update linting to exclude `html`, `dcc` packages.
- [x] Integrate CI testing with dynamic configs with CircleCI (WIP)
- [x] Integrate dev branch git commits and histories with `lerna import` (must be done prior to merging the PR/updating components).
- [x] ~~Possibly merge `dash-table` to `dcc`.~~
- [x] Update CHANGELOG
- [x] Remove `test-app` before merging.
- [x] Fix remaining test conflicts
- [ ] Git rebase to clean up git history

A test app is present at `test-app/test-dash.py` to check that imports and dependencies are successfully imported through Dash.

## Contributor Checklist

### optionals

Closes plotly/dash-core#323


alexcjohnson:
@HammadTheOne this is looking good!

> adds some tools for maintaining the monorepo.

Can you describe these? This PR is big enough I don't really see a way to review it on GH, I'm just checking it out locally and playing with it.

> Integrate dev branch git commits and histories with `lerna import` (must be done prior to merging the PR/updating components).

Are you referring to currently open PRs? I was assuming we'd just do this manually but if there's an easier way I'm all 👂 

> Possibly merge `dash-table` to `dcc`.

Let's save that for a followup PR.

Two more things I notice now:
- As mentioned on slack: all the component `package-lock.json` files use hard tabs rather than two spaces. Seems like that's https://github.com/lerna/lerna/issues/2845. Let's just rerun `npm i` in each dir to get it back to spaces, and see if we can get away from `lerna bootstrap` so it doesn't come back.
- In the dev branch we probably want to `.gitignore` the component build artifacts (`dash/dcc`, `dash/html`, `dash/dash_table`)?

HammadTheOne:
> Can you describe these? This PR is big enough I don't really see a way to review it on GH, I'm just checking it out locally and playing with it.

This really just refers to the `dash-update-components` CLI command, and the `lerna` commands we discussed over Slack. Beyond that, it's mainly structural changes to the directory and `dash.py` to accomodate the dependencies. 

> Integrate dev branch git commits and histories with lerna import (must be done prior to merging the PR/updating components).

Unfortunately this doesn't include currently open PR's. There is a way to move the changes in open PRs from a different repo without having to manually copy over the work with by using `git cherry-pick` to move the range of commits from one repo to another. I haven't experimented with it just yet, but maybe after this PR, I can test it out and make a small guide in `CONTRIBUTING` which would outline the steps to do so. 

> In the dev branch we probably want to .gitignore the component build artifacts (dash/dcc, dash/html, dash/dash_table)

That makes sense - I'll add those to `.gitignore`.

alexcjohnson:
> Unfortunately this doesn't include currently open PR's.

Don't worry about this for now. We'll start out moving them manually - there aren't that many - but if that gets to be too cumbersome we'll revisit.

HammadTheOne:
@alexcjohnson The majority of tests are passing now. The last failing `dash-table` server tests involve a bug with `dash-table` where the `data_previous`prop isn't correctly updated when the contents of a cell/data are updated with copy + paste - This is unrelated to the monorepo changes, and occurs on the dev branch of `dash-table` as well afaict.

Does it make sense to skip these for now and address this in a separate issue?

alexcjohnson:
there's still `dash-table/tests/selenium/test_bootstrap.py`, failing with:
```
ModuleNotFoundError: No module named 'dash_html_components'
```
We should fix that one. But the others I'm OK with skipping if there's an issue to go back to them later. I know copy/paste can be finicky, might be browser-version-dependent or require that we enable the right permissions in the browser so copying is allowed.

alexcjohnson:
OIC it's actually `dash_bootstrap_components` importing html - OK, let's skip this too for now, but revisit it once DBC has a version compatible with dash 2 imports.

HammadTheOne:
> dash-table/tests/selenium/test_bootstrap.py

Ah right, sorry I skipped over explaining this one entirely once I saw it was an update needed for DBC itself. I'm tracking these down separately in a spreadsheet which I can transfer over to an issue.

alexcjohnson:
hmm one more issue: Percy just shows a single job here (as it should), but on their side there are actually 8 jobs generated, four of which have snapshots in them. We need these all to be combined into one. Not sure exactly what that'll entail, but at least collapsing to a single `percy-finalize` job that requires all test jobs that create snapshots to have completed first.
Screen Shot 2021-08-12 at 6 20 14 PM
https://percy.io/plotly/dash and select this branch in the upper right.

HammadTheOne:
@alexcjohnson  I added a `percy-tests` workflow with the jobs that include Percy snapshots, tied together to a single `percy-finalize` call. 

image

However, while the workflow seems correct on CircleCI, we still have 6 jobs generated still on Percy, 2 of which have snapshots (looks like `test-39` and `table-server-test`. The `PERCY_PARALLEL_TOTAL` env key we have set is correct for parallel builds, but it doesn't seem like they're being combined into a single build on Percy's end as they should according to [this](https://docs.percy.io/docs/parallel-test-suites). 

Could it be that we're using a somewhat older docker image for Percy which doesn't support this, rather than the recommended percy-agent CI orb? Or is it maybe the `parallelism` key within the jobs themselves?  Unfortunately the Percy docs don't cover this particular scenario in much  detail.

jonmmease:
The slider dcc tests were failing because of an orjson encoding issue that's getting fixed in https://github.com/plotly/plotly.py/pull/3351.

jonmmease:
`test_tdrp002_select_cell` in the table component tests is also failing under orjson. I'm looking into it, but you might want to remove orjson for the table tests as well.

HammadTheOne:
@alexcjohnson I split the dash-table ones which use percy storybook into a separate Percy build linked to the same repo, so they'll appear in the Github status, alongside the dash ones which would include dcc and html. The Dashboard Engine repo follows a similar pattern. 

If that looks fine, we can look at merging this PR tomorrow morning.

chriddyp (approved):

PR #1685, opened 2021-07-09
Move the renderer build into the renderer dir

alexcjohnson:
FYI @HammadTheOne - since the renderer is also an npm package, we want to keep its build output within that package for npm publishing to get all the right files. But I left all the other js deps in the deps folder where you put them in #1611.

@eff-kay please review - this is I think the last change we need before we can publish a new version of Dash.

### optionals

### DIFF
--- a/dash/development/build_process.py
+++ b/dash/development/build_process.py
@@ -21,19 +21,20 @@ class BuildProcess(object):
     def __init__(self, main, deps_info):
         self.logger = logger
         self.main = main
+        self.build_folder = self._concat(self.main, "build")
         self.deps_info = deps_info
         self.npm_modules = self._concat(self.main, "node_modules")
         self.package_lock = self._concat(self.main, "package-lock.json")
         self.package = self._concat(self.main, "package.json")
         self._parse_package(path=self.package)
-        self.asset_paths = (self.build_folder, self.npm_modules)
+        self.asset_paths = (self.deps_folder, self.npm_modules)
 
     def _parse_package(self, path):
         with open(path, "r") as fp:
             package = json.load(fp)
             self.version = package["version"]
             self.name = package["name"]
-            self.build_folder = self._concat(self.main, os.pardir, "deps")
+            self.deps_folder = self._concat(self.main, os.pardir, "deps")
             self.deps = package["dependencies"]
 
     @staticmethod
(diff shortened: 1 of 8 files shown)


alexcjohnson:
The test failure is very strange... `async-plotly.js` can't be found. It gets added to the `<head>` tag and executed - the app works just fine - but then somehow the script tag gets deleted before we can see it in the test. I was only able to see it by setting a breakpoint in `LazyLoader/plotly.js` and then stepping forward a little. I'll have to dig into this tomorrow... perhaps some new webpack behavior?

dash/development/build_process.py
```diff
@@ -21,19 +21,20 @@ class BuildProcess(object):
...
+        self.asset_paths = (self.deps_folder, self.npm_modules)
 
     def _parse_package(self, path):
         with open(path, "r") as fp:
             package = json.load(fp)
             self.version = package["version"]
             self.name = package["name"]
-            self.build_folder = self._concat(self.main, os.pardir, "deps")
+            self.deps_folder = self._concat(self.main, os.pardir, "deps")
             self.deps = package["dependencies"]
```
eff-kay, line 38: just so I understand this clearly. The new `self.build_folder` is located at `dash_renderer/build`, and the `dash_renderer/../deps `is now `self.deps_folder`??

dash/development/build_process.py
```diff
@@ -108,11 +111,11 @@ def digest(self):
 
     @job("copy and generate the bundles")
     def bundles(self, build=None):
-        if not os.path.exists(self.build_folder):
+        if not os.path.exists(self.deps_folder):
```
eff-kay, line 114: we can use the `exist_ok` flag in `os.makedirs` and do away with the if block altogether.

dash/development/build_process.py
```diff
@@ -138,20 +141,20 @@ def bundles(self, build=None):
...
                 self._concat(self.npm_modules, scope, name, subfolder, filename),
-                self._concat(self.build_folder, target),
+                self._concat(self.deps_folder, target),
             )
 
         _script = "build:dev" if build == "local" else "build:js"
         logger.info("run `npm run %s`", _script)
-        os.chdir(self._concat(self.build_folder, os.pardir, "dash-renderer"))
+        os.chdir(self.main)
         run_command_with_process("npm run {}".format(_script))
```
eff-kay, line 150: not exactly sure, but shouldn't we wait for this command to execute?

eff-kay:
LGTM

eff-kay:
> The test failure is very strange... `async-plotly.js` can't be found. It gets added to the `<head>` tag and executed - the app works just fine - but then somehow the script tag gets deleted before we can see it in the test. I was only able to see it by setting a breakpoint in `LazyLoader/plotly.js` and then stepping forward a little. I'll have to dig into this tomorrow... perhaps some new webpack behavior?

This looks like an async issue, something should wait for something to execute. The breakpoint takes a memory dump on each step, so makes sure all variables are resolved when you enter.

alexcjohnson:
It looks like webpack 5 indeed [removes the script after it executes](https://github.com/webpack/webpack/blob/main/lib/runtime/LoadScriptRuntimeModule.js#L140). AFAICT this was [silently introduced](https://github.com/webpack/webpack/pull/10992) with another feature and never documented. I'm just going to tweak the test to ensure the graph rendered and the sync module is present iff expected.

dash/development/build_process.py
alexcjohnson: exactly - `dash/deps` is all the 3rd-party packages we load on the page, and `dash/dash-renderer/build` is the bundles we build.

dash/development/build_process.py
alexcjohnson: `run_command_with_process` does [wait](https://github.com/plotly/dash/blob/ca4fcf5d0a209cd2865bd7ddba52952c83f2126e/dash/_utils.py#L221)

dash/development/build_process.py
alexcjohnson: good call -> https://github.com/plotly/dash/pull/1685/commits/f424c8b9769d104e28bcd5da5c9f868bc9e9696b

alexcjohnson: ah right, Py2 🙃 🔫

eff-kay: lol. k. I thought if its available for versions<=3.6 then it should be available for versions<2. But that is not the case, sorry for that.

PR #1691, opened 2021-07-23
Flexible callback support

jonmmease:
This PR includes several features of the `FlexibleCallback` plugin that was developed as part of Dash Labs.

In particular, this PR introduces support for:
 - Callback functions with named keyword arguments (https://github.com/plotly/dash-labs/blob/main/docs/02-CallbackEnhancements.md#positional-or-keyword-arguments)
 - Interchangeable `Input` and `State` (https://github.com/plotly/dash-labs/blob/main/docs/02-CallbackEnhancements.md#interchangeable-input-and-state)
 - Tuple and Dictionary grouping for both inputs and outputs (https://github.com/plotly/dash-labs/blob/main/docs/02-CallbackEnhancements.md#tuple-and-dictionary-argument-grouping)
 
For clarity, this PR does not make any changes to the `Input`/`State`/`Output` dependency objects, and so does not include the Dash Labs feature of allowing component objects to be passed in place of ids in dependency objects.

Also, this PR is not compatible with Python 2, as the plan is to deprecate Python 2 support for Dash 2.  Let me know if you'd like me to disable the Python 2 tests in this PR.


alexcjohnson:
> Also, this PR is not compatible with Python 2, as the plan is to deprecate Python 2 support for Dash 2. Let me know if you'd like me to disable the Python 2 tests in this PR.

Not just deprecate but in fact REMOVE Python 2 support in Dash 2. But I think it may be clearer if we make a separate PR that adds one Py2-incompatible change and does everything else associated: drops the tests and updates the classifiers, `python_requires`, `requires-*.txt`... then when that's merged we can update this PR.

jonmmease:
Sounds good. I'll open a PR to remove Python 2 support

alexcjohnson:
This PR looks great! A few questions though about how it plays with other features:

Presumably pattern-matching callbacks using `ALL` or `ALLSMALLER` still nest inside the inner level of whatever grouping is defined here?

`callback_context`: I don't think `inputs`, `states`, or `triggered` should be affected by this as they're intended to be flat prop_id->value dicts, but `outputs_list` should reflect the nested structure, and I'm guessing right now it's the flattened structure. Likewise, with inputs and state mingling, `inputs_list` and `states_list` don't even make sense anymore; there should just be one. In fact, `_list` doesn't make sense, perhaps these should be `output_structure` and `arg_structure`? Or `*_shape`? These are mostly useful with PMCs, so you know which IDs exist and where in the inputs and expected outputs - described in https://dash.plotly.com/pattern-matching-callbacks

Lastly clientside callbacks: It's OK I think to not implement it there for now, but if that's what we do, we should ensure that (a) you can't give the structured form to `app.clientside_callback` without an error, (b) we document this limitation clearly, (c) we have a sketch of a plan for how this will be added in the future.

jonmmease:
Thanks for taking a look @alexcjohnson.

> Presumably pattern-matching callbacks using ALL or ALLSMALLER still nest inside the inner level of whatever grouping is defined here?

Yes, that's right

> callback_context: I don't think inputs, states, or triggered should be affected by this as they're intended to be flat prop_id->value dicts, but outputs_list should reflect the nested structure, and I'm guessing right now it's the flattened structure. Likewise, with inputs and state mingling, inputs_list and states_list don't even make sense anymore; there should just be one. In fact, _list doesn't make sense, perhaps these should be output_structure and arg_structure? Or *_shape? These are mostly useful with PMCs, so you know which IDs exist and where in the inputs and expected outputs - described in https://dash.plotly.com/pattern-matching-callbacks

I've been using "grouping" to refer to these nested structures.  How about `outputs_grouping` and `args_grouping`?

> Lastly clientside callbacks: It's OK I think to not implement it there for now, but if that's what we do, we should ensure that (a) you can't give the structured form to app.clientside_callback without an error, (b) we document this limitation clearly, (c) we have a sketch of a plan for how this will be added in the future.

Sounds good. I'll add (a) to this PR and open an issue to discuss (b) and (c).

alexcjohnson:
> How about `outputs_grouping` and `args_grouping`

🏆

jonmmease:
I added `outputs_grouping` and `args_grouping` to callback context in 5f38d7b.  I didn't remove anything from callback context yet. We can talk about it, but I don't see a harm in leaving the `*_list` properties there for backward compatibility.

Do you have any suggestions for a good place to test these new callback context properties?

alexcjohnson:
> I don't see a harm in leaving the *_list properties there for backward compatibility

Sounds good, let's just officially deprecate them, both in the docs and via a warning when you use them.

> Do you have any suggestions for a good place to test these new callback context properties?

Mainly this is useful for PMCs, so I'd create an app similar to those in [test_wildcards](https://github.com/plotly/dash/blob/dev/tests/integration/callbacks/test_wildcards.py) but using the new semantics, spit out `args_grouping` and `outputs_grouping` onto the page somewhere and verify their values.

jonmmease:
I added a new test app for the callback context *_grouping props. And there is now a DeprecationWarning printed when accessing the old *_list properties iff there are any nested dependency groupings.

The legacy dependency validation logic is still being used for clientside callbacks, so that will catch scenarios where someone tries to used groupings in clientside callbacks.

jgff:
If the index_string argument in the Dash object is interacting with the component, how is this implemented?

.gitignore
```diff
@@ -80,3 +80,7 @@ VERSION.txt
 
 # vim
 *.swp
+
+# main-branch artifacts to explicitly include
+!dash/dash-renderer/build
+!dash/deps
```
alexcjohnson, line 86: What's this gitignore change about?

dash/_callback_context.py
```diff
@@ -96,5 +125,23 @@ def record_timing(name, duration=None, description=None):
 
         setattr(flask.g, "timing_information", timing_information)
 
+    @property
+    @has_context
+    def using_args_grouping(self):
+        """
+        Return True if this callback is using dictionary or nested groupings for
+        Input/State dependencies.
```
alexcjohnson, line 133: or interleaved inputs and state - does that trigger this flag?

.gitignore
jonmmease: Good catch. Not sure where that came from. reverted in 77c3c18

dash/_callback_context.py
jonmmease: Good point. It does now (9f41b86)

alexcjohnson:
@jgff I'm unclear what you're referring to, can you give an example?

dash/_grouping.py
```diff
@@ -0,0 +1,234 @@
...
+    :param default: Default scalar value to use if grouping scalar key is not present
+        in source
+    :return: grouping
+    """
+    return map_grouping(lambda s: source.get(s, default), schema)
+
+
+class SchemaValidationError(InvalidCallbackReturnValue):
+    def __init__(self, msg):
+        super().__init__(msg=msg)
```
alexcjohnson, line 154: Does this `__init__` change anything? Unless I'm missing something we'll still go through `DashException.__init__` with or without this, and that one dedents and strips the message - so the explicit 4-space indentation you give all the messages below will be removed, and we should be able to turn each collection of single-line-with-terminator f-strings into one big triple-quoted f-string, using whatever indentation is convenient since it'll be removed anyway.

Not a big deal, just a little cleaner IMO :)

dash/_grouping.py
```diff
@@ -0,0 +1,234 @@
...
+                f"    Expected type: {expected_type}\n"
+                f"    Received value of type {type(value)}:\n"
+                f"        {repr(value)}\n"
+            )
+        )
+
+    @classmethod
+    def check(cls, value, full_schema, path, expected_type):
+        if not isinstance(value, expected_type):
+            raise SchemaTypeValidationError(value, full_schema, path, expected_type)
```
alexcjohnson, line 172: Oh that's a nice pattern, I hadn't seen it done that way before.

dash/_grouping.py
jonmmease: Yeah, good idea. done in 449f16e. I also removed the `SchemaValidationError` base class because it's not actually doing anything

alexcjohnson (approved):
Looks great! I think the only thing we're still waiting for is a test that the new forms throw an error with clientside_callback https://github.com/plotly/dash/pull/1691#issuecomment-888516402 but then this is good to go!
💃

jonmmease:
Thanks!  clientside callback validation tests added in 5bb114a.  I'll merge on green

PR #1694, opened 2021-07-28
Drop python 2 support

jonmmease:
This PR removes the Python 2 circleci tests, updates the minimum Python version in setup.py to 3.6, and adds an f-string to dash.py for good measure.


setup.py
```diff
@@ -53,12 +53,6 @@ def read_req_file(req_type):
         "Intended Audience :: Science/Research",
         "License :: OSI Approved :: MIT License",
         "Programming Language :: Python",
-        "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.7",
-        "Programming Language :: Python :: 3",
```
alexcjohnson, line 58: Is it intentional to drop `Python :: 3`?

jonmmease: ahh, no. good catch

alexcjohnson:
The one other thing I was hoping this PR would do is drop the Py2 parts of `requires-dev.txt` and `requires-testing.txt`

alexcjohnson (approved):
Love it! 🌅  💃

PR #1700, opened 2021-08-06
Add support for adding "extra components" to an app's layout

jonmmease:
This PR is a foundational part of the coming `long_callback` PR, which will bring the `long_callback` decorator from Dash Labs to Dash.  I think it may be useful in other contexts as well, so I thought I'd split it off into a separate PR for discussion.

I added a new private `_extra_components` property to `dash.Dash` app instances that is initialized to an empty list. When components are added to this list, they are appended to the layout returned by `_layout_value`.

The motivation is to make it possible for `long_callback` to add `Interval` and `Store` components to the layout without returning them to the user and requiring the user to position them in the layout themselves.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -491,15 +491,13 @@ class Dash(object):
         return self._layout
 
     def _layout_value(self):
+        from dash_html_components import Div  # pylint: disable=import-outside-toplevel
+
         layout = self._layout() if self._layout_is_function else self._layout
 
-        # Add extra hidden components
+        # Add any extra components
         if self._extra_components:
-            if not hasattr(layout, "children"):
-                setattr(layout, "children", [])
-            for c in self._extra_components:
-                if c not in layout.children:
-                    layout.children.append(c)
+            layout = Div(children=[layout] + self._extra_components)
 
         return layout
 
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -400,6 +400,7 @@ class Dash(object):
         self._layout = None
         self._layout_is_function = False
         self.validation_layout = None
+        self._extra_components = []
 
         self._setup_dev_tools()
         self._hot_reload = AttributeDict(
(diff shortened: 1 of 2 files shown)


jonmmease:
Not sure what's going on with Percy :thinking:

jonmmease:
all green now. Ready for a look when you have a chance @alexcjohnson

dash/dash.py
```diff
@@ -490,7 +491,17 @@ def layout(self):
         return self._layout
 
     def _layout_value(self):
-        return self._layout() if self._layout_is_function else self._layout
+        layout = self._layout() if self._layout_is_function else self._layout
+
+        # Add extra hidden components
+        if self._extra_components:
+            if not hasattr(layout, "children"):
```
alexcjohnson, line 498: Some simple apps just use `app.layout = DataTable(...)` or a graph or something, that don't support children. Mostly this is just for testing purposes but maybe not always. In that case what do we do, throw an error? Wrap in `html.Div`? Actually there's something nice about wrapping in a div, then we don't need to keep testing if we've already added these components into the layout, and don't need to mutate it. I can't really think of how mutating the layout would cause a problem but it makes me uneasy.

jonmmease: Can you think of any cases where wrapping in a Div would break something?  Any concern if the top-level is a `ddk.App` or `dbc.App`?

alexcjohnson: I don't think this would cause a problem - we're already inside a div from the react entry point

https://github.com/plotly/dash/blob/a19c27426c7008769530d72420a88a68b9aed548/dash/dash.py#L86-L92

Could be nice at some point to allow the top level layout to be an array, but this would either also require wrapping in a div or would require changes to the renderer.

jonmmease: Ok, wrapping in a div seems like the best solution then. Thanks!

jonmmease: Done in b88d8a3

alexcjohnson (approved):
LGTM! 💃 after adding a changelog entry.

jonmmease:
> ...after adding a changelog entry.

This isn't a public feature (at least not yet), so I wasn't picturing adding a changelog for it by itself.

I'll go ahead and merge this one so that I can fix up the conflicts in https://github.com/plotly/dash/pull/1702, but I can add something more to the changelog over in that PR if you'd still like to mention this PR on its own.

Thanks!

alexcjohnson:
Ok, that makes sense

PR #1701, opened 2021-08-06
Preserve schema key order when flattening a dict grouping

jonmmease:
Small follow-on fix for flexible callback support added in https://github.com/plotly/dash/pull/1691.

For the case of a callback that returns a dictionary grouping, this fix removes the unintentional requirement that the dictionary key order of the returned value match the key order of the dependency specification.

For example, this case now works properly:

```python
@app.callback(dict(a=Output(...), b=Output(...)), ...)
def callback(...):
    return dict(b=0, a=1)
```

### DIFF
--- a/dash/_grouping.py
+++ b/dash/_grouping.py
@@ -30,6 +30,8 @@ def flatten_grouping(grouping, schema=None):
     """
     if schema is None:
         schema = grouping
+    else:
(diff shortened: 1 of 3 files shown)


alexcjohnson (approved):
Nice catch! And bonus points for reducing LOC (aside from tests) with the fix!
💃

PR #1707, opened 2021-08-12
Disable compression by default in Dash 2

jonmmease:
This PR changes the default value of the `compress` argument to the `dash.Dash` constructor from `True` to `False`.

Rational: Compression has a performance penalty, and is often handled at a higher level in the deployment stack (e.g. by NGINX or Dash Enterprise).  This PR makes Flask-level compression opt-in rather than opt-out for Dash 2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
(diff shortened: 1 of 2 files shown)


alexcjohnson (approved):
Also:

https://github.com/plotly/dash/blob/abfd9a81e18699355c79edaac81ecd73d2ff2806/dash/dash.py#L199-L200

Plus a changelog entry.

Then this is good to go 💃

PR #1710, opened 2021-08-12
`dash.callback`

chriddyp:
Provides `@dash.callback` as an alternative to `@app.callback`. Particularly useful when binding pattern matching callbacks to reusable components and multi-page apps.


chriddyp:
Alright, this is ready for review! The only thing missing right now is figuring out how to raise an error in the multiple-app scenario: 

```python
import dash


from flask import Flask
server = Flask(__name__)

app1 = dash.Dash(name='app1',
                server=server, url_base_pathname='/app1/')
app1.layout = html.Div([
    dcc.Input(id='input-app-1', value='app-1'),
    html.Div(id='output-app-1')
])
@dash.callback(Output('output-app-1', 'children'), Input('input-app-1', 'value'))
def update(value):
    return value
    
app1 = dash.Dash(name='app2',
                server=server, url_base_pathname='/app2/')
app1.layout = html.Div([
    dcc.Input(id='input-app-2', value='app-2'),
    html.Div(id='output-app-2')
])
@dash.callback(Output('output-app-2', 'children'), Input('input-app-2', 'value'))
def update(value):
    return value
```

PR #1718, opened 2021-08-20
`dash.callback`

chriddyp:
Provides `@dash.callback` as an alternative to `@app.callback`. Particularly useful when binding pattern matching callbacks to reusable components and multi-page apps.


jonmmease:
The PR looks good. I'm not sure why the long callback tests failed on CI. I kicked off the 3.9 tests again to see if it's repeatable.  Do you get any local failures on these tests?

jonmmease:
Trying something over in https://github.com/plotly/dash/pull/1719

jonmmease:
@chriddyp The test failures seem to have been a symptom of the new global state leaking across tests.  I made a change in #1719 that fixes the tests by moving the global values into the app's own values on use, then clearing the global values.  But I don't know if that fits the semantics you had in mind.

PR #1725, opened 2021-08-24
Add dependency on component stub packages

jonmmease:
This PR adds dependencies on the (not yet published) stub packages below:
 - dash_html_components==2.0.0rc1: https://github.com/plotly/dash-html-components/pull/198
 - dash_core_components==2.0.0rc1: https://github.com/plotly/dash-core-components/pull/985
 - dash_table==5.0.0rc1: https://github.com/plotly/dash-table/pull/937

It also adds a fix in 15ba716 that prevents raising a warning when using the built-in components (with the new imports) when the stub packages are installed.


HammadTheOne (approved):
:woman_dancing:

alexcjohnson (approved):
💃 🚀 🌟

PR #1726, opened 2021-08-24
Remove `future`, `__future__` and `py2/py3`

anders-kiaer:
As Python 2 support is dropped in `dev` :tada: here is a community PR on the transition of the code base.

`grep -rnw . -e 'substring'` now returns an empty list for all substrings `py2`, `py3`, `from future`, `import future`, `__future__`, `_strings`.

## Contributor Checklist

   -  [X] Checked for the substrings above and removed them + simplified code that followed from that.

### DIFF
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-from __future__ import unicode_literals
 import shlex
 import sys
 import uuid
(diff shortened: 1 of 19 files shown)


dash/_utils.py
```diff
@@ -109,7 +103,7 @@ def strip_relative_path(requests_pathname, path):
 
 # pylint: disable=no-member
 def patch_collections_abc(member):
-    return getattr(collections if utils.PY2 else collections.abc, member)
+    return getattr(collections.abc, member)
```
alexcjohnson, line 106: In fact we should be able to 🔪 `patch_collections_abc` entirely

dash/dash_table/__init__.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import print_function
```
alexcjohnson, line 1: Ah, there will be a little merge conflict with https://github.com/plotly/dash/pull/1734 where we remove all these build artifacts... (`dash/dash_table`, `dash/dcc`, `dash/html`)

alexcjohnson (approved):
This is fantastic @anders-kiaer - thanks for doing this! 💃

PR #1727, opened 2021-08-24
fixed bug causing 'dash_html_components.' to be imported instead of d…

yujin-wu:
…ash_html_components

Fixes bug in preparing resource fingerprints. The code accidentally imports "dash_html_components" as "dash_html_components." and this PR fixes that.

This issue causes problems with dash_embedded CI tests.

## Contributor Checklist

   -  [x] Fix and test.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -619,7 +619,7 @@ class Dash:
         # for cache busting
         def _relative_url_path(relative_package_path="", namespace=""):
             if any(
-                relative_package_path.startswith(x + '/')
+                relative_package_path.startswith(x + "/")
                 for x in ["dcc", "html", "dash_table"]
             ):
                 relative_package_path = relative_package_path.replace("dash.", "")
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -619,7 +619,7 @@ class Dash:
         # for cache busting
         def _relative_url_path(relative_package_path="", namespace=""):
             if any(
-                relative_package_path.startswith(x)
+                relative_package_path.startswith(x + '/')
                 for x in ["dcc", "html", "dash_table"]
             ):
                 relative_package_path = relative_package_path.replace("dash.", "")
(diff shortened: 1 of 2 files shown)


dash/dash.py
```diff
@@ -618,7 +618,10 @@ def _collect_and_register_resources(self, resources):
         # add the version number of the package as a query parameter
         # for cache busting
         def _relative_url_path(relative_package_path="", namespace=""):
-            if any(x in relative_package_path for x in ["dcc", "html", "dash_table"]):
+            if any(
+                relative_package_path.startswith(x)
```
alexcjohnson, line 622:
```suggestion
                relative_package_path.startswith(x + '/')
```
(Or just put the slashes in each list item)

The actual `relative_package_path` values we want to catch here are all like `html/dash_html_components.min.js` right? I'm worried about what would happen if someone made a new package called `html_helpers` or something.

alexcjohnson: oh right, black wants doublequotes 🙄

alexcjohnson (approved):
💃 Nice fix!

alexcjohnson:
OK so now we get a little further but fail at:

```
        with pytest.raises(TypeError) as e:
            html.Div(asdf=True)
>       assert str(e.value) == (
...
E         - The `Div` component received an unexpected keyword argument: `asdf`
E         + The `dash_html_components.Div` component (version 1.17.1) received an unexpected keyword argument: `asdf`
```

So somehow we're back now to sort of what we had prior to the monorepo PR, specifically [these test tweaks](https://github.com/plotly/dash/commit/d8147abd98d42a6453c468b3316d550bc4a889a1#diff-16dfc841434d6081a8e87125cac6a04eb38359005a6d6b214f7409d750bf7878L471)

It's not appropriate to call it `dash_html_components` anymore, nor is `version 1.17.1` correct (That's the `dcc` version, why are we getting that here? But really we shouldn't even get the `html` version, because that isn't a package anymore as far as users are concerned. We should get the `dash` version, and call the component either `html.Div` or `dash.html.Div`

yujin-wu:
@alexcjohnson It appears to be a side effect 😬 of the test I added (fixed). Now there is only percy/dash-table-test. It has nothing to do with this code, as other PRs have the identical UI diffs picked up by percy.

PR #1734, opened 2021-08-30
Remove build artifacts and update build process

HammadTheOne:
This PR removes build artifacts from the `dev` branch of `dash` and updates the build scripts. It also includes some cleanup for obsolete html components. 

## Contributor Checklist

- [x] TODO
   -  [x] Update package.json scripts to a single `npm build` script to build renderer and components.
   -  [x] Update CONTRIBUTING.md
   -  [x] Update dash-core release process (https://github.com/plotly/dash-core/pull/336)
   -  [x] Remove obsolete `html` components.
   -  [x] Fix table helpers import issue (if in scope)

- [x] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))

### optionals

- [x] I have added entry in the `CHANGELOG.md`

PR #1735, opened 2021-08-30
update plotly.js to v2.4.1

alexcjohnson:
FYI I specifically did NOT commit build artifacts because @HammadTheOne is working on removing these from the repo entirely (oh there is is, #1734). AFAICT the tests all rebuild all the components anyway.
- [x] Port over custom tooltip component https://github.com/plotly/dash-core-components/pull/982
- [x] Basic test of the `Tooltip` component


components/dash-core-components/src/components/Tooltip.react.js
```diff
@@ -0,0 +1,276 @@
...
+    return (
+        <>
+            <div className="dcc-tooltip-bounding-box">
+                <span
+                    data-dash-is-loading={is_loading || undefined}
+                    className={`hover hover-${props.direction}`}
+                >
+                    <span
+                        id={id}
+                        className={`hover-content ${props.className}`}
```
alexcjohnson, line 22: Discovered while writing the test, but anyway users will expect `props.id` to be in the DOM. I put it on the content `<span>`, because this is where `props.className` is applied, and that's mostly where you probably want your styles applied. Makes it a little awkward to style the container, but because of the pseudo-elements making the triangle this is discouraged anyway, and we have special props for bg and border colors there already.

jonmmease (approved):
This all looks reasonable to me, assuming others have tried out the tooltip component and are happy with the end user behavior.

PR #1745, opened 2021-09-04
Split dependencies

### LINKED ISSUE #1742: Split out Dash CI only requirements into `requires-ci.txt`?
Both `pip install dash[dev]` and `pip install dash[testing]` are extra public Dash features (respectively for building Dash components, and using the shipped Dash app test utilities).

Looking at [`requires-dev.txt`](https://github.com/plotly/dash/blob/dev/requires-dev.txt) and [`requires-testing.txt`](https://github.com/plotly/dash/blob/dev/requires-testing.txt), there appears to have leaked https://github.com/plotly/dash CI/test suite dependencies into both files (i.e. dependencies not used by the features in `dash[dev]` and `dash[testing]`). Latest one introduced is [`flaky`](https://github.com/plotly/dash/pull/1741).

Split out Dash CI only dependencies to a new file?

anders-kiaer:
Closes #1742.

## Contributor Checklist

   -  [X] I have checked locally which dependencies are needed in order to run `dash-generate-components`, and as far as I could see only `PyYAML`, `fire` and `coloredlogs` are needed in `dash[dev]`.
   -  [X] Moved dependencies that are used only by Dash CI/tests into `requires-ci.txt`.
   -  [X] Created two new "extra groups" for the long callback managers. This should make it easier for `dash` to potentially change which dependencies the callback managers require in the future, as well as communicate how to install what is needed.


anders-kiaer:
I'm a bit unsure if my `.circleci/config.yml` changes are optimal wrt. caching (not experienced with CircleCI, even though it has similarities with GitHub workflows, and the CircleCI setup was complex - probably because the monorepo transition is not complete, i.e. `dcc`, `html` etc. are still separate Python "packages" within the same repo). Also not sure why the Percy test fails. :laughing: 

Splitting out Dash CI dependencies will ensure projects using `dash[dev]` and `dash[testing]` pull downs only necessary requirements, which again will reduce installation time and improve security for downstream apps/projects.

setup.py
```diff
@@ -30,6 +30,8 @@ def read_req_file(req_type):
     extras_require={
         "dev": read_req_file("dev"),
         "testing": read_req_file("testing"),
+        "celery-manager": read_req_file("celery-manager"),
+        "diskcache-manager": read_req_file("diskcache-manager"),
```
alexcjohnson, line 34: This makes sense, thanks @anders-kiaer!

Let's include `ci` as an option here too. It's less important now that dcc, html, and table are combined into this repo, but we still like to use the same CI setup in other non-core component repos. And perhaps we can shorten the new ones by dropping `-manager`, just `celery` and `diskcache`?

dash/testing/plugin.py
```diff
@@ -189,10 +189,18 @@ def dashjl(request, dashjl_server, tmpdir):
...
+            DiskcacheLongCallbackManager,
+        )
+        import diskcache  # pylint: disable=import-outside-toplevel
+    except ImportError as missing_imports:
+        raise ImportError(
+            """\
+DiskcacheLongCallbackManager requires extra dependencies which can be installed doing
+
+$ pip install "dash[diskcache-manager]"\n"""
+        ) from missing_imports
```
alexcjohnson, line 203: 🌴  Perhaps make these imports a `@staticmethod` of `DiskcacheLongCallbackManager` so we don't need to repeat this, it's just something like:
```py
    from dash.long_callback import (  # pylint: disable=import-outside-toplevel
        DiskcacheLongCallbackManager,
    )
    diskcache = DiskcacheLongCallbackManager.imports()[0]
```

setup.py
alexcjohnson: Is there a convention for how to document available extras? We can of course add this to the docs site (not sure we want to go into all this detail in https://dash.plotly.com/installation but elsewhere on that site we can) but what about documentation within the package itself?

Perhaps a comment at the top of each `requires-*` file describing its purpose?

anders-kiaer: > And perhaps we can shorten the new ones by dropping `-manager`, just `celery` and `diskcache`?

Agree, I can shorten name the groups by dropping `-manager`. As well as make a `ci` option. 👍 

> Is there a convention for how to document available extras? We can of course add this to the docs site (not sure we want to go into all this detail in https://dash.plotly.com/installation but elsewhere on that site we can) but what about documentation within the package itself? Perhaps a comment at the top of each requires-* file describing its purpose?

The closest thing to a convention that I've seen is that they are documented along with the extra features depending on these extra dependencies (so that goes along your suggestion with not documenting it on https://dash.plotly.com/installation, but instead where the celery/diskcache managers are documented). A comment in each `requires-*` file feels natural, and these comments could easily be moved along into e.g. a [`pyproject.toml` optional section](https://www.python.org/dev/peps/pep-0621/#dependencies-optional-dependencies) or `setup.cfg` if doing that change at some point.

dash/testing/plugin.py
anders-kiaer: I solved the duplication in a slightly different way: If `cache` is not given to `DiskcacheLongCallbackManager` it defaults to initiate a `diskcache.Cache()` instance with default values (the same as the `pytest` fixture uses). This way the `pytest` fixture will naturally fail with the understandable error message provided centrally by `DiskcacheLongCallbackManager.__init__()` if dependencies are not installed.

Let me know @alexcjohnson if this way of doing it is in conflict some other existing/planned usage patterns with `DiskcacheLongCallbackManager`.

alexcjohnson: @jonmmease any reason not to do it this way? It seems like `diskcache` generally works fine without an explicit directory - which simplifies things for users, unless there's a common case where it fails.

jonmmease: Yeah, that sounds fine

alexcjohnson (approved):
💃  Thanks @anders-kiaer !

anders-kiaer:
Thanks @alexcjohnson for taking it over the 🏁 line. 🙂

Issue #1674, opened 2021-06-24
Display an error if propType value is a reference to a variable when building backend components wuth `dash-generate-components`


HammadTheOne:
As pointed out in https://github.com/plotly/dash-bio/pull/564, if we reference a variable within the `react.js` source code for a component `propType`, the components are generated for Python and R, but missing the values themselves when mapping  from the `PropTypes` js type object to the Python/R type. For Julia, the components fail to generate with a type-error. 

As @alexcjohnson suggested, it makes more sense to have the component generator throw an error when this happens so that component authors can fix this kind of problem up front rather than expecting them to notice a problem later and fix it.

gvwilson:
@HammadTheOne is this still a concern? thanks - @gvwilson

HammadTheOne:
No, we can close it for now.

PR #1747, opened 2021-09-06
Shorthand supports for dash-core-components

workaholicpanda:
_Replaced by #1763 and this can be closed, looks like too much commits_

This PR implements [this issue](https://github.com/plotly/dash-core-components/issues/960) to support optional shorthand supports for below components.
You can compare [old format](https://gist.github.com/jackparmer/864545d0c7f7596ed6b4c9b18a3b252c) and [new format](https://gist.github.com/workaholicpanda/e54d762fd8ae6e755eebfbef40c42bec).

## Dependency Update
 - [x] `Input` `Output` `State` can now accept component object as well as string value of the component key
    - [x] (Required) A separate PR will need to be merged before this

## Updated Components with change notes
 - Dropdown
   - [x] `options` props can now be passed without being a keyword argument
   - [x] `options` can now accept array of string as well as previous format (array of dictionary)

 - RangeSlider
   - [x] Changed props order to make it start with `min` `max` `step` 

 - Checklist
   - [x] `options` props becomes first argument, can be passed without keyword
   - [x] Added `inline` props to specify if `options` are array of string or dictionary (previous format)

- RadioItems
   - [x] `options` props becomes first argument, can be passed without keyword
   - [x] Added `inline` props to specify if `options` are array of string or dictionary (previous format)

- DataTable
  - [x] `data` and `columns` props becomes first 2 arguments, can be passed without keyword
  - [x] (Optional) Separate PR will be followed to extract `columns` automatically from given `data`


components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -33,68 +34,69 @@ export default class Checklist extends Component {
...
+                 */
+                value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+                    .isRequired,
+
+                /**
+                 * If true, this checkbox is disabled and can't be clicked on.
+                 */
+                disabled: PropTypes.bool,
+            }),
+            PropTypes.string
```
chriddyp: Let's add a comment here describing how this is used as label and value, and if the user wants a different thing for label and value then they should use a list instead.. These comments make their way into the Python docstrings in `dcc.Dropdown` via the component generator.

components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -187,6 +196,16 @@ Checklist.propTypes = {
...
+
+    /**
+     * Indicates whether given `options` are array or dictionary
+     * if True: 
+     *      `options` will receive: [`value1`, `value2`, `value3`, ...]
+     *      which equals to [{label: `value1`, value: `value1`}, {label: `value2, value: `value2`}, ...]
+     * if False: 
+     *      `options` will receive: [{label: `label1`, value: `value1`}, {label: `label2`, value: `value2`}, ...]
+     */
+    inline: PropTypes.bool,
```
chriddyp, line 215: 🔪 Let's remove this and "detect" the shape of `options` within the component itself.

components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -198,4 +217,5 @@ Checklist.defaultProps = {
     value: [],
     persisted_props: ['value'],
     persistence_type: 'local',
+    inline: false,
```
chriddyp, line 227: 🔪

components/dash-table/src/dash-table/dash/DataTable.js
```diff
@@ -110,17 +110,20 @@ export const defaultProps = {
...
-    active_cell: PropTypes.exact({
-        row: PropTypes.number,
-        column: PropTypes.number,
-        row_id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-        column_id: PropTypes.string
-    }),
+    data: PropTypes.arrayOf(PropTypes.object),
 
-    /**
+     /**
```
chriddyp, line 126: NBD, but I think the convention in the rest of the comments is for this to have a single tab and for the following `*` to have a tab plus one space.

components/dash-core-components/src/fragments/Dropdown.react.js
```diff
@@ -35,9 +35,15 @@ export default class Dropdown extends Component {
 
     UNSAFE_componentWillReceiveProps(newProps) {
         if (newProps.options !== this.props.options) {
+            const normalizedOptions = newProps.options.map(opt =>
```
chriddyp, line 38: Yeah - like this! Let's do something like this for the other components instead of `inline`

components/dash-core-components/src/components/Checklist.react.js
jackparmer: @chriddyp what do think of having a standard `inline=True` shortcut like dash bootstrap components, that substitutes and is the equivalent of the more verbose `style = {'display': 'inline'}`

components/dash-table/src/dash-table/dash/Sanitizer.ts
```diff
@@ -65,6 +66,11 @@ const getFixedRows = (
           (filter_action !== TableAction.None ? 1 : 0) +
           data2number(fixed.data);
 
+const populateColumnsFromData = (data: Data) =>
```
akksi, line 69: Not sure, but maybe we'd like to preserve the order of the columns we are passing to the object generator. It's not really possible using an array of objects, as object properties will always be ordered alphabetically.
So columns will always be displayed in an alphabetic order this way.

One possible solution for this could be instead of passing data from python as `DataFrame.to_dict('records')`, passing it as `DataFrame.to_dict('split')` that contains the headers as an array of strings that preserves column order.

components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -33,68 +33,87 @@ export default class Checklist extends Component {
...
-                                    newValue = append(option.value, value);
-                                }
-                                setProps({value: newValue});
-                            }}
-                        />
-                        {option.label}
-                    </label>
-                ))}
+                {options.map(option => {
+                    option =
```
akksi, line 37: I think it's not really necessary to check the row's data structure for each record one-by-one inside the map callback as we might not want to support mixed option types. If so, it's enough to check the first element and PropTypes takes care of the rest.

components/dash-table/src/dash-table/dash/DataTable.js
```diff
@@ -110,15 +110,18 @@ export const defaultProps = {
...
+     *      {'column-1': 8, 'column-2': 'boston', 'column-3': 'america'}
+     * ]
      */
-    active_cell: PropTypes.exact({
-        row: PropTypes.number,
-        column: PropTypes.number,
-        row_id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-        column_id: PropTypes.string
-    }),
+    data: PropTypes.arrayOf(PropTypes.object),
```
akksi, line 124: Maybe we could be a bit more descriptive/specific with the type, like defining it as something like `PropTypes.objectOf(PropTypes.oneOfType([PropTypes.string, PropTypes.number]))`.

components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -33,68 +33,87 @@ export default class Checklist extends Component {
...
+                        typeof option === 'string'
```
akksi, line 38: I noticed that other parts of the code use `type()` method from the `ramda` package for type checking, actually, I think it wouldn't add a lot, maybe just a bit more of consistency.

components/dash-core-components/src/components/RangeSlider.react.js
```diff
@@ -19,6 +19,21 @@ export default class RangeSlider extends Component {
 }
 
 RangeSlider.propTypes = {
+    /**
```
akksi, line 22: I think, calculating initial `ticks`, `step` and `value` properties are missing here if we want to use it as it's used int the format spec: https://gist.github.com/workaholicpanda/e54d762fd8ae6e755eebfbef40c42bec

It renders, but probably not the way we would expect, so it's not really usable if we omit all props other than `min` and `max`.

components/dash-table/src/dash-table/dash/DataTable.js
workaholicpanda: I think it's a thing to consider and good suggestion! But here are 2 things in my mind.
- Having object can give more flexibility to support formulas or reference to another cell like in Excel? Not even sure if this is a designed feature. :)
- I tried to not touch existing stuff that can potentially break anything. 😆 

@chriddyp What would you say?

components/dash-core-components/src/components/Checklist.react.js
workaholicpanda: I agree with you and better to have more consistency. 7497eb218b3545633b11fbeeba5a56bc7f49880b

components/dash-core-components/src/components/RangeSlider.react.js
workaholicpanda: I'm open to add that or copy things from your PR if it's okay. @akksi

akksi:
It's absolutely fine from my side.

components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -33,68 +33,87 @@ export default class Checklist extends Component {
...
              */
-            value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
-                .isRequired,
-
-            /**
-             * If true, this checkbox is disabled and can't be clicked on.
-             */
-            disabled: PropTypes.bool,
-        })
+            PropTypes.string,
```
AnnMarieW, line 115: Is it necessary to limit this to a string only?  Or could also include numbers?:
`PropTypes.oneOfType([PropTypes.string, PropTypes.number])`

chriddyp: Great point @AnnMarieW ! `bool` while we're at it too

workaholicpanda:
https://plotly.slack.com/archives/C01HT8G2RHQ/p1631290623354200?thread_ts=1631239702.344600&cid=C01HT8G2RHQ This still need to be done

components/dash-core-components/src/components/Checklist.react.js
workaholicpanda: @AnnMarieW Added `PropTypes.number` and use the number as a label of the option. 74c25af5
@chriddyp But question on `bool`: Is it okay to have empty `label` but with only `value` in an option?

chriddyp: For bool, I think we'd just convert the bools into strings, so `[{'label': 'True', 'value': True}, {'label': 'False', 'value': False}]`

components/dash-core-components/src/components/Dropdown.react.js
```diff
@@ -25,45 +25,46 @@ export default class Dropdown extends Component {
...
-            label: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
-                .isRequired,
-
+        PropTypes.oneOfType([
             /**
-             * The value of the dropdown. This value
-             * corresponds to the items specified in the
-             * `value` property.
+             * We now accept the single `value` as an option value,
+             * which equals to
```
chriddyp, line 36:
```suggestion
             * which is equal to
```

components/dash-core-components/src/components/RadioItems.react.js
```diff
@@ -63,39 +74,48 @@ export default class RadioItems extends Component {
 }
 
 RadioItems.propTypes = {
+    /**
+     * An array of options, or inline dictionary of options
+     */
+    options: PropTypes.oneOfType([
+        /**
+         * Simpler `options` representation in dictionary format
+         * {`label1`: `value1`, `label2`: `value2`, ... }
```
chriddyp, line 83: I believe that `value` needs to be unique but label doesn't need to be unique, so this should be inverted: `{'value1': 'label1', 'value2': 'label2'}`

components/dash-core-components/src/components/RadioItems.react.js
```diff
@@ -63,39 +74,48 @@ export default class RadioItems extends Component {
 }
 
 RadioItems.propTypes = {
+    /**
+     * An array of options, or inline dictionary of options
+     */
+    options: PropTypes.oneOfType([
```
chriddyp, line 80: Let's support the same 3 sets of option types across RadioItems, Dropdown, CheckList, and DataTable.columns for consistency:
- List of dictionaries (as before)
- Dictionary (value: label)
- List of numbers, strings, or bools

components/dash-core-components/src/components/RadioItems.react.js
```diff
@@ -63,39 +74,48 @@ export default class RadioItems extends Component {
...
+        /**
+         * Simpler `options` representation in dictionary format
+         * {`label1`: `value1`, `label2`: `value2`, ... }
+         * which equals to
+         * [{label: `label1`, value: `value1`}, {label: `label2`, value: `value2`}, ...]
+         */
+        PropTypes.object,
```
chriddyp, line 87: Maybe we can use `objectOf` here to get some more type checking. String, number, bool

components/dash-core-components/src/fragments/Dropdown.react.js
```diff
@@ -57,7 +65,7 @@ export default class Dropdown extends Component {
         } = this.props;
         const {filterOptions} = this.state;
         let selectedValue;
-        if (type(value) === 'array') {
+        if (type(value) === 'Array') {
```
chriddyp, line 68: 😱 wow wild that we didn't catch this before. Can we add a quick test to lock this down?

components/dash-core-components/src/fragments/RangeSlider.react.js
```diff
@@ -14,7 +14,12 @@ export default class RangeSlider extends Component {
             ? createSliderWithTooltip(Range)
             : Range;
         this._computeStyle = computeSliderStyle();
-        this.state = {value: props.value};
+
+        const {min, max, value, setProps} = props;
+        this.state = {value};
+        if (!value) {
+            setProps({value: calcValue(min, max, value)});
```
chriddyp, line 21: `setProps` is going to fire a Dash callback. I don't think that's what we want here.

components/dash-core-components/src/fragments/RangeSlider.react.js
```diff
@@ -101,8 +103,9 @@ export default class RangeSlider extends Component {
                         getTooltipContainer: node => node,
                     }}
                     style={{position: 'relative'}}
-                    value={value}
-                    marks={truncatedMarks}
+                    value={calcValue(min, max, value)}
+                    marks={calcMarks({min, max, marks, step})}
```
chriddyp, line 107: will want to skip this if `marks` is supplied by the user

components/dash-core-components/src/fragments/RangeSlider.react.js
```diff
@@ -101,8 +103,9 @@ export default class RangeSlider extends Component {
                         getTooltipContainer: node => node,
                     }}
                     style={{position: 'relative'}}
-                    value={value}
-                    marks={truncatedMarks}
+                    value={calcValue(min, max, value)}
```
chriddyp, line 106: skip if `value` is supplied by user

components/dash-table/src/dash-table/dash/DataTable.js
chriddyp: We don't support objects as values yet. So yeah string, number, bool would be fine as far as I know. And our tests should pick things up if not.

dash/development/base_component.py
```diff
@@ -163,6 +166,17 @@ def __init__(self, **kwargs):
 
             setattr(self, k, v)
 
+    def set_random_id(self):
+        if not hasattr(self, "id"):
+            rd = random.Random()
+            hexdigest = hashlib.md5(
+                bytes(str(self.to_plotly_json()), "utf8")
```
chriddyp, line 173: Hm, wouldn't components with identical properties end up having the same ID?

chriddyp: Let's use something like this instead:
```
import random
import uuid

r = random.Random(0)
r.seed(0)
print(uuid.UUID(int=r.randint(0, 2 ** 128)))
print(uuid.UUID(int=r.randint(0, 2 ** 128)))
print(uuid.UUID(int=r.randint(0, 2 ** 128)))
print(uuid.UUID(int=r.randint(0, 2 ** 128)))
```

chriddyp: Which will consistently generate the _same_ UUIDs like:
```
e3e70682-c209-4cac-629f-6fbed82c07cd
82e2e662-f728-b4fa-4248-5e3a0a5d2f34
d4713d60-c8a7-0639-eb11-67b367a9c378
23a7711a-8133-2876-37eb-dcd9e87a1613
```

components/dash-core-components/src/fragments/RangeSlider.react.js
workaholicpanda: `calcMarks` provides truncation only when it's provided, but we don't want any validation when supplied by user?

components/dash-core-components/src/components/RadioItems.react.js
workaholicpanda: RadioItems, Dropdown, CheckList updated in 63ced025

components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -84,36 +77,7 @@ Checklist.propTypes = {
...
-                    .isRequired,
-
-                /**
-                 * If true, this checkbox is disabled and can't be clicked on.
-                 */
-                disabled: PropTypes.bool,
-            }),
-        ])
-    ),
+    options: optionsType,
```
AnnMarieW, line 80: Although this is a very elegant solution, using variables in this way with PropTypes doesn't work with the Dash toolchain.  The docstrings fail to generate correctly. If you look at the docstring  in `Checklist.py`, you will see that none of the details from `optionsType` are included --  It only shows:

```
- options (optional):
        An array of options.
```
This may be related to #1674, except I didn't see type-errors with the Julia components. I'd love to know if there is a better workaround than just repeating the code. :hocho:  I ran into this in the `dash-bootstrap-components` library as well.

chriddyp: That's correct, and yeah there is no better workaround that I know of at this point

components/dash-core-components/src/components/RadioItems.react.js
```diff
@@ -38,9 +40,13 @@ export default class RadioItems extends Component {
...
                 style={style}
             >
-                {options.map(option => (
+                {sanitizeOptions(options).map(option => (
                     <label
-                        style={labelStyle}
+                        style={Object.assign(
+                            {},
+                            labelStyle,
+                            inline ? {display: 'inline'} : {}
```
chriddyp, line 48:
```suggestion
                            inline ? {display: 'inline-block'} : {}
```

components/dash-core-components/src/components/RadioItems.react.js
```diff
@@ -184,6 +169,13 @@ RadioItems.propTypes = {
      * session: window.sessionStorage, data is cleared once the browser quit.
      */
     persistence_type: PropTypes.oneOf(['local', 'session', 'memory']),
+
+    /**
+     * Indicates whether labelStyle should be inline or not
+     * True: Automatically set { 'display': 'inline' } to labelStyle
```
chriddyp, line 175:
```suggestion
     * True: Automatically set { 'display': 'inline-block' } to labelStyle
```

components/dash-core-components/src/components/RadioItems.react.js
```diff
@@ -184,6 +169,13 @@ RadioItems.propTypes = {
...
+     * False: No additional behavior to expect
```
chriddyp, line 176:
```suggestion
     * False: No additional styles are passed into labelStyle.
```

components/dash-core-components/src/utils/computeSliderMarkers.js
```diff
@@ -0,0 +1,135 @@
...
+                expectedSteps <= desiredCountMax + 1
+            );
+        }) || possibleValues[0];
+    return [
+        alignValue(min, finalStep) * stepValue,
+        alignValue(finalStep * stepValue, stepValue),
+    ];
+};
+
+export const autoGenerateMarks = (min, max, step = 1) => {
```
chriddyp, line 67: If user passes in `step`, then we should respect it. So I think you can remove ` = 1` here and add a `interval = step ? step : estimateBestSteps(min, max, step)`

components/dash-table/src/dash-table/dash/Sanitizer.ts
chriddyp: I think we're OK here - I believe `Object.keys` now preserves some sort of order. 
```
> Object.keys(JSON.parse('[{"sepal_length": 5.1, "sepal_width": 3.5, "petal_length": 1.4, "petal_width": 0.2, "species": "setosa", "species_id": 1}, {"sepal_length": 4.9, "sepal_width": 3.0, "petal_length": 1.4, "petal_width": 0.2, "species": "setosa", "species_id": 1}, {"sepal_length": 4.7, "sepal_width": 3.2, "petal_length": 1.3, "petal_width": 0.2, "species": "setosa", "species_id": 1}, {"sepal_length": 4.6, "sepal_width": 3.1, "petal_length": 1.5, "petal_width": 0.2, "species": "setosa", "species_id": 1}, {"sepal_length": 5.0, "sepal_width": 3.6, "petal_length": 1.4, "petal_width": 0.2, "species": "setosa", "species_id": 1}]')[0])

Array(6) [ "sepal_length", "sepal_width", "petal_length", "petal_width", "species", "species_id" ]
```

chriddyp: We can revisit later, but let's roll with this for now.

components/dash-core-components/src/components/RadioItems.react.js
chriddyp: Great. Let's skip DataTable for now. We can always do it later.

chriddyp:
Let's also add tests for each of the new component signatures. `dash_duo.percy_snapshot`-style tests are fine IMO. Just render a bunch of components with various ranges and take a screenshot.

The most involved test cases will be the sliders and making sure the marks function looks good.

`Slider` & `RangeSlider`
```
NUMBERS = [10**N for N in np.arange(-25, 25, 0.5)]

TEST_RANGES = []
LAYOUT = []

for n in NUMBERS:
    TEST_CASES.extend([
        [n, n * 1.5],
        [n * 0.9, n * 1.3],
        [n * 0.9 + 0.1, n * 1.3 * 0.4],
        [n * -1, n],
        [n * -1.5, n],
        [-1 * n, 0],
        [-1 * n, 0.3 * n],
        [0, n],
        [0, n * 0.43]
    ])

for t in TEST_CASES:
    layout.extend([
        html.Div(f'{t[0]} - {t[1]}'),
        dcc.Slider(t[0], t[1]),

        html.Div(f'{t[0]} - {t[1]}, {(t[1] - t[0]) / 5}'),
        dcc.Slider(t[0], t[1], (t[1] - t[0]) / 5),

        html.Div(f'{t[0]} - {t[1]}'),
        dcc.RangeSlider(t[0], t[1]),

        html.Div(f'{t[0]} - {t[1]}, {(t[1] - t[0]) / 5}'),
        dcc.RangeSlider(t[0], t[1], (t[1] - t[0]) / 5),
    ])

app.layout = html.Div(TEST_CASES)

# wait for something
# you'll need to fill this in...

dash_duo.percy_snapshot('SLIDER - TEST CASES') # look in the code to see if there is a naming convention we're using
```

Then for dropdowns etc just a simple rendering of the component without keyword arguments:

```python
app.layout = html.Div([
    dcc.Dropdown(['a', 'b', 'c']),
    dcc.Dropdown(['a', 'b', 'c'], 'b'),
    dcc.Dropdown(['a', 3, 'c']),
    dcc.Dropdown(['a', 3, 'c'], 3),
    dcc.Dropdown(['a', 3, 'c', true, false]),
    dcc.Dropdown(['a', 3, 'c', true, false], false),
```

dash/development/base_component.py
```diff
@@ -163,6 +165,14 @@ def __init__(self, **kwargs):
 
             setattr(self, k, v)
 
+    def set_random_id(self):
+        if not hasattr(self, "id"):
+            rd = random.Random(0)
```
chriddyp, line 170: I think this needs to be set in the global scope on import, otherwise this is going to produce the _same_ random ID each time it is called (because it's creating a new instance with the same seed)

dash/development/base_component.py
```diff
@@ -163,6 +165,14 @@ def __init__(self, **kwargs):
...
+            rd.seed(0)
+            v = str(uuid.UUID(int=rd.randint(0, 2 ** 128)))
+            setattr(self, "id", v)
+        return getattr(self, "id")
```
chriddyp, line 174: Let's add some tests for this as well with multiple components:
```
input1 = dcc.Input(value='Hello')
input2 = dcc.Input(value='Hello')
output1 = html.Div()
output2 = html.Div()
output3 = html.Div(id='output-3')

app.layout = html.Div([input1, input2, output1, output2, output3])

app.callback(Output(output1, 'children'), Input(input1, 'value'))
def update(v):
    return f'Input 1 {v}'

app.callback(Output(output2, 'children'), Input(input2, 'value'))
def update(v):
    return f'Input 2 {v}'

app.callback(Output('output-3', 'children'), Input(input1, 'value'), Input(input2, 'value'))
def update(v1, v2):
    return f'Output 3 - Input 1: {v1}, Input 2: {v2}'

# Verify the auto-generated IDs are stable
assert(output1.id, 'e3e70682-c209-4cac-629f-6fbed82c07cd')
assert(output2.id, '82e2e662-f728-b4fa-4248-5e3a0a5d2f34')
assert(output3.id, 'd4713d60-c8a7-0639-eb11-67b367a9c378')
assert(output4.id, '23a7711a-8133-2876-37eb-dcd9e87a1613')

# Do some assertions in the DOM to check that the values are getting returned OK
```

components/dash-core-components/src/components/RadioItems.react.js
workaholicpanda: f990b43c

components/dash-table/src/dash-table/dash/Sanitizer.ts
alexcjohnson: The JSON serializer we're using in Python sorts keys as of plotly.py 5.0 (whether or not you have `orjson` present), so yes I expect we are going to lose column order this way.

components/dash-core-components/src/components/RadioItems.react.js
workaholicpanda: @chriddyp @jackparmer @akksi
Two things here,
1. Unfortunately, I haven't found a way to iterate key as its own type, but all of Object methods (`entries`, `keys`, `for in`, etc) just gives string type key.
I think idea is good but we won't be dealing non-string `value` properly.

2. Also, I'm removing the `PropTypes.oneOfType([string, number, bool])` constraint because we'll be actually trying to validate the `label` type which is not what we really wanted.
https://plotly.slack.com/archives/C01HT8G2RHQ/p1631628303425800?thread_ts=1631542387.390300&cid=C01HT8G2RHQ

https://github.com/workaholicpanda/dash/pull/6

chriddyp: That's correct - `values` will only be able to be strings. That's OK - It's just a limitation of this data shape. Users won't necessarily see this in Python as Python will serialize the keys to strings.

components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -33,69 +34,107 @@ export default class Checklist extends Component {
...
-
-            /**
-             * If true, this checkbox is disabled and can't be clicked on.
-             */
-            disabled: PropTypes.bool,
-        })
-    ),
+    options: PropTypes.oneOfType([
+        /**
+         * Array of options - [string|number|bool]
```
chriddyp, line 82:
```suggestion
         * Array of options where the label and the value are the same thing - [string|number|bool]
```

components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -33,69 +34,107 @@ export default class Checklist extends Component {
...
+         */
+        PropTypes.arrayOf(
+            PropTypes.oneOfType([
+                PropTypes.string,
+                PropTypes.number,
+                PropTypes.bool,
+            ])
+        ),
+        /**
+         * Simpler `options` representation in dictionary format
```
chriddyp, line 92:
```suggestion
         * Simpler `options` representation in dictionary format. The order is not guaranteed. 
```

components/dash-core-components/tests/integration/misc/test_persistence.py
```diff
@@ -141,12 +141,12 @@ def make_output(*args):
 
     dash_dcc.find_element("#radioitems label:first-child input").click()  # red
 
-    range_slider = dash_dcc.find_element("#rangeslider")
-    dash_dcc.click_at_coord_fractions(range_slider, 0.5, 0.25)  # 5
-    dash_dcc.click_at_coord_fractions(range_slider, 0.8, 0.25)  # 8
+    # range_slider = dash_dcc.find_element("#rangeslider")
+    # dash_dcc.click_at_coord_fractions(range_slider, 0.5, 0.25)  # 5
+    # dash_dcc.click_at_coord_fractions(range_slider, 0.8, 0.25)  # 8
```
chriddyp, line 146: Is this no longer working?

tests/unit/development/test_base_component.py
```diff
@@ -473,3 +474,37 @@ def test_debc027_component_error_message():
...
+    def update(v1, v2):
+        return f"Output 3 - Input 1: {v1}, Input 2: {v2}"
+
+    # Verify the auto-generated IDs are stable
+    assert output1.id == "e3e70682-c209-4cac-629f-6fbed82c07cd"
+    assert input1.id == "82e2e662-f728-b4fa-4248-5e3a0a5d2f34"
+    assert output2.id == "d4713d60-c8a7-0639-eb11-67b367a9c378"
+    assert input2.id == "23a7711a-8133-2876-37eb-dcd9e87a1613"
+    # we make sure that the if the id is set explicitly, then it is not replaced by random id
+    assert output3.id == "output-3"
```
chriddyp, line 510: Could we do a quick assertion to verify that the callbacks are correctly fired? You can either assert that that the values are there by looking up the element in the DOM (see other tests) or you can take a screenshot.

workaholicpanda: 13e008f0

components/dash-core-components/tests/integration/misc/test_persistence.py
workaholicpanda: It is activated back and functional now.

components/dash-core-components/tests/integration/dropdown/test_dropdown_shorthands.py
```diff
@@ -0,0 +1,52 @@
...
+            ],
+            True,
+        ],
+    ]
+
+    layout = []
+    for definition in TEST_OPTIONS_N_VALUES:
+        (option, value) = definition if len(definition) > 1 else [definition[0], None]
+        layout.extend(
+            [html.Div(f"Options={option}, Value={value}"), dcc.Dropdown(option, value)]
```
chriddyp, line 45: Could we render `RadioItems` and `Checklist` here too? Just to lock down the behavior of all of the components that accept these new properties

workaholicpanda:
@chriddyp Can we close this and move on to #1763? Or do we need to keep all review comments & histories?

PR #1749, opened 2021-09-09
Add shorthand options to Core Components

akksi:
https://github.com/plotly/dash-core-components/issues/960

### DIFF
--- a/components/dash-core-components/src/components/Checklist.react.js
+++ b/components/dash-core-components/src/components/Checklist.react.js
@@ -19,6 +19,7 @@ export default class Checklist extends Component {
             inputStyle,
             labelClassName,
             labelStyle,
+            inline,
             options,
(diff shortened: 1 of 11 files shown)


akksi:
2 important notes:

1. If there is no value passed to RangeSlider, the React Component passes back the values derived from the Range values and the Python update function updates the values correctly, but until this happens, Python raises an error. This can be solved by adding default properties to `slider_range`. Not sure if it can be handled on React side.

2. For DataTable, the Python DataFrame object can't be passed directly to a React component so my current implementation accepts `df.to_dict('split')` as it contains both the columns and data in the right order (`df.to_dict('records')` doesn't preserve the order of columns).

alexcjohnson:
superseded by https://github.com/plotly/dash/pull/1763

PR #1750, opened 2021-09-12
Rename `run_server` `run` and add backwards compatible alias

chriddyp:
A little terser in the spirit of Dash 2.0!


chriddyp:
Ugh branched off of `main` instead of `dev` again

PR #1751, opened 2021-09-12
`app.run` instead of `run_server`

chriddyp:
In the spirit of Dash 2.0's terseness:

- Renames `app.run_server` `app.run`
- Preserves `app.run_server` for backwards compatibility 
- Points folks from `run_server` to `run` via docstring


alexcjohnson (approved):
💃

PR #1752, opened 2021-09-13
Set last shift-clicked element as active element

tmpacifitech:
Fixes https://github.com/plotly/dash-table/issues/837

### DIFF
--- a/components/dash-table/src/dash-table/handlers/cellEvents.ts
+++ b/components/dash-table/src/dash-table/handlers/cellEvents.ts
@@ -70,7 +70,8 @@ export const handleClick = (
 
     const newProps: Partial<ICellFactoryProps> = {
         is_focused: false,
-        end_cell: clickedCell
+        end_cell: clickedCell,
+        active_cell: clickedCell,
     };
 
     if (e.shiftKey && active_cell) {
@@ -84,9 +85,7 @@ export const handleClick = (
             visibleColumns,
             viewport
         );
-        newProps.active_cell = clickedCell;
     } else {
-        newProps.active_cell = clickedCell;
         newProps.start_cell = clickedCell;
         newProps.selected_cells = [clickedCell];
     }
--- a/components/dash-table/src/dash-table/handlers/cellEvents.ts
+++ b/components/dash-table/src/dash-table/handlers/cellEvents.ts
@@ -71,7 +71,7 @@ export const handleClick = (
     const newProps: Partial<ICellFactoryProps> = {
         is_focused: false,
         end_cell: clickedCell,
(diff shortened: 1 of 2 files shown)


components/dash-table/src/dash-table/handlers/cellEvents.ts
```diff
@@ -84,6 +84,7 @@ export const handleClick = (
             visibleColumns,
             viewport
         );
+        newProps.active_cell = clickedCell;
     } else {
         newProps.active_cell = clickedCell;
```
akksi, line 89: Looks like this value is not assigned conditionally anymore so I would just pull it out before the if statement instead of duplicating it in the else block.

tmpacifitech: right right, already pushed fix. thanks!

akksi:
It has fixed the issue on my end. :)

workaholicpanda:
pretty small thing, but looks like husky is not activated and you may want to run `eslint --fix`  :)

```
/home/circleci/dash/components/dash-table/src/dash-table/handlers/cellEvents.ts
  74:33  error  Unexpected trailing comma  comma-dangle

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

```

workaholicpanda:
You may want to move `test.py` to some where around `components/dash-table/tests/` according to what it actually does. :blush:

jackparmer:
@workaholicpanda @tmpacifitech is this one good to go?

chriddyp:
Let's get the Percy tests up and running:
image

chriddyp:
Could we add a test? I believe that there are already some copy & paste tests in the repo somewhere.

alexcjohnson:
> Could we add a test? I believe that there are already some copy & paste tests in the repo somewhere.

In fact there's even an example of using the shift key in those tests (just with the down arrow, not a click):

https://github.com/plotly/dash/blob/d0d8adbbc10f8c5cca16b0e6ab9b6c8e8d07994a/components/dash-table/tests/selenium/test_basic_copy_paste.py#L161-L162

@tmpacifitech why don't you 🔪  `test.py` but add a test case to `test_basic_copy_paste.py`.

And just a note for anyone else working in here: you can run and play with the app defined in any of these test cases using `--pause` - for example to run the app with that shift-down-arrow: `pytest -k tbcp004 --pause`. This also puts you in the Python debugger (within the selenium thread, not the app thread).

ndrezn:
This PR was superseded by: https://github.com/plotly/dash/pull/1789, which was released.

PR #1753, opened 2021-09-13
Fit column width to child elements

tmpacifitech:
Fixes https://github.com/plotly/dash-table/issues/867

### DIFF
--- a/components/dash-table/src/dash-table/components/Table/Table.less
+++ b/components/dash-table/src/dash-table/components/Table/Table.less
@@ -504,6 +504,10 @@
             th {
                 white-space: nowrap;
 
+                .column-header--wrapper {
+                    display: inline-flex;
+                }
+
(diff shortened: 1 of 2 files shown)


akksi (commented):
I think it's a nice approach to use flex instead of float here, the current solution has some alignment issues yet, a possible way to solve it:
1. Using `flex` instead of `inline-flex` as we should fill in the width of the column
2. adding `justtify: space-between` to keep the sorting icon aligned to the left and text to the right
3. removing 'float: left' from the wrapper that prevents it from taking the width of the column
(4. it would be nice to remove 'float: left' from the children too, as it becomes pointless with this solution)

jackparmer:
@akksi @tmpacifitech is this one good to go?

tmpacifitech:
@jackparmer Yes

PR #1758, opened 2021-09-15
Use invisible header row

tmpacifitech:
Fixes https://github.com/plotly/dash-table/issues/205 and https://github.com/plotly/dash-table/issues/432

### DIFF
--- a/components/dash-table/tests/visual/percy-storybook/DashTable.percy.tsx
+++ b/components/dash-table/tests/visual/percy-storybook/DashTable.percy.tsx
@@ -143,6 +143,20 @@ const dataA2J = (() => {
 
 const style_data_conditional = [{width: '100px'}];
 
+const columnsWithLongerTitle = ['Date sent to company', 'Company Response'].map(
+    id => ({id: id, name: id.toUpperCase()})
+);
+const dataWithLongerTitle = (() => {
+    const r = random(1);
+
+    return R.range(0, 100).map(() =>
+        ['Date sent to company', 'Company Response'].reduce((obj: any, key) => {
+            obj[key] = Math.floor(r() * 1000);
+            return obj;
+        }, {})
+    );
+})();
+
 storiesOf('DashTable/Fixed Rows & Columns', module)
     .add('with 1 fixed row, 2 fixed columns', () => (
         <DataTable
@@ -153,7 +167,7 @@ storiesOf('DashTable/Fixed Rows & Columns', module)
             fixed_columns={{headers: true}}
             fixed_rows={{headers: true}}
             row_deletable={true}
-            row_selectable={true}
+            row_selectable='single'
             style_data_conditional={style_data_conditional}
         />
     ))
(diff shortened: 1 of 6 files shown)


akksi (commented):
It kind of works, but it's quite flickery/laggy and sometimes unresponsive when scrolling vertically or resizing the window.

https://user-images.githubusercontent.com/14875243/133575525-90e658f5-5eae-4233-b2cc-75580ed8766a.mp4

I'm not sure how much work it would take, but maybe it would be worth restructuring the HTML table markup a bit (by joining the `header` and the `body` part to one `table`) to render a semantic table and format it with CSS instead of calculating the column sizes with js.

If not, maybe it's possible to optimize the calculations a bit more.

akksi:
A lower effort solution could be cloning the header row into the table body and make it visually hidden so it would essentially set the min. width of the columns - in theory.

tmpacifitech:
@akksi can you take a look at this PR again?

akksi:
> @akksi can you take a look at this PR again?

Looks good to me. :)

components/dash-table/tests/visual/percy-storybook/DashTable.percy.tsx
```diff
@@ -165,10 +179,21 @@ storiesOf('DashTable/Fixed Rows & Columns', module)
...
+    .add('with 1 fixed row, longer header width than content cell', () => (
+        <DataTable
+            setProps={setProps}
+            id='table'
+            data={dataWithLongerTitle}
+            columns={columnsWithLongerTitle}
+            fixed_rows={{headers: true}}
+            row_deletable={true}
+            row_selectable='single'
+        />
```
chriddyp, line 195: image

Could we add a test where the columns are as wide as `max(header, cell)`? In this test case, the table is very wide so we don't see the issue of the narrow headers.

chriddyp: Perhaps a test with several columns

chriddyp:
This percy diff is odd - Looks like we lost the header border for the row selection column?
image

chriddyp:
Another odd one - custom colors aren't being applied to that border
image

components/dash-table/tests/visual/percy-storybook/DashTable.percy.tsx
chriddyp: Ah well I guess this existing test covers it:
image

chriddyp:
Looks like this test is missing data now
`DashTable/Border (available space filled): with frozen rows and no frozen columns`
image

https://percy.io/plotly/dash-table-test/builds/12769932/changed/721646457?browser=chrome&browser_ids=19%2C20%2C21&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

chriddyp:
And this one isn't getting rendered

https://percy.io/plotly/dash-table-test/builds/12769932/changed/721646552?browser=firefox&browser_ids=19%2C20%2C21&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

image

chriddyp:
The widths of these columns changed considerably. Not sure what the "true" width should be for this case though.

image

chriddyp:
Same here
https://percy.io/plotly/dash-table-test/builds/12769932/changed/721646761?browser=edge&browser_ids=19%2C20%2C21&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

image

Again not sure what the width "should" be in this case.

chriddyp:
There is a double border below A now

https://percy.io/plotly/dash-table-test/builds/12769932/changed/721646637?browser=firefox&browser_ids=19%2C20%2C21&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

image

gvwilson:
@tmpacifitech we think most of the Dash DataTable has been superseded by AG Grid, so we're going to close this one as stale. Thank you again for having contributed - sorry it took us so long to get to this one.

PR #1768, opened 2021-09-20
Revert "Fit column width to child elements"

chriddyp:
Reverts plotly/dash#1753


chriddyp:
#1753 introduced a visual change in the column alignment.

PR #1776, opened 2021-09-23
Use flexbox for table header

tmpacifitech:


components/dash-table/src/dash-table/components/Table/Table.less
```diff
@@ -504,13 +516,22 @@
...
+                    &--hide,
+                    &--sort {
+                        .not-selectable();
+                        cursor: pointer;
+                    }
+
+                    > *:only-child {
+                        margin-left: auto;
+                    }
                 }
```
tmpacifitech, line 535: This is the only changes on this less file.. other changes are just prettier fix.

tmpacifitech:
Another fix for https://github.com/plotly/dash-table/issues/867

akksi (changes requested):
- Looks like one of the Percy tests is failing (https://app.circleci.com/pipelines/github/plotly/dash/2707/workflows/6936fe57-0c84-4d9f-b4b7-c47452fe9a6e/jobs/44310)
- If there are multiple buttons/icons in a column, they should be wrapped in a container so all of them would be aligned left OR, maybe it would work to always add `margin-left: auto` to the text, so it will put the spacing between the icons and the text (I like the second solution better).
- Do we need to keep the `float: left` properties?

alexcjohnson:
Looks like we had multiple parallel fixes for this... 🙈

Issue #1433, opened 2020-10-12
[BUG] react-dom: Each child in a list should have a unique "key" prop


tashrifbillah:
I experienced this issue in a bigger context but I found an easy way to reproduce it. Please run the `Quickstart` tutorial from https://dash.plotly.com/datatable

**Screenshots**
![image](https://user-images.githubusercontent.com/35086881/95768459-61258f80-0c84-11eb-9e0e-6b5b95347bba.png)

**Describe your context**
Windows 10 machine on both Edge and Chrome browsers

```
dash                   1.16.2
dash-core-components   1.12.1
dash-extensions        0.0.18
dash-html-components   1.1.1
dash-renderer          1.8.2
dash-table             4.10.1
```

**Expected behavior**

No warning.

**Additional info**
[This](https://github.com/plotly/dash/issues/891) issue might be related but does not appear to give a clear reproduction way which I did in the above.

themegaphoenix:
Hi,
I am also having the same issue.
This pull request might be what we need:
https://github.com/plotly/dash-table/pull/690

chriddyp:
These are just warnings and can be ignored unless you've found a way that this effects the behavior of the table itself.

themegaphoenix:
> These are just warnings and can be ignored unless you've found a way that this effects the behavior of the table itself.

I understand, but nothing makes me happier than a clean console :). It should hopefully be resolved soon

tashrifbillah:
I would love to test this PR. Two points--

* Is the PR from February? Seems a little old.
* How do I install just `dash-table`? I have always done `pip install dash` which had `dash-table` dependency. I can clone the `dash` master and try `pip install .` locally but how do I point to the PR for `dash-table` dependency?

antoinerg:
Closed by https://github.com/plotly/dash/pull/1778

michaelbabyn:
reopening since it looks like this issue still appears when the `presentation` column attribute is set to `markdown` in any of the columns.

E.g see this simple example from the forum:
```python
import dash_bootstrap_components as dbc
from dash import Dash, html, dash_table

import pandas as pd

df = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/solar.csv")

app = Dash(__name__)

app.layout = dash_table.DataTable(
    df.to_dict("records"),
    [{"name": i, "id": i, "presentation": "markdown"} for i in df.columns],
)

if __name__ == "__main__":
    app.run_server(debug=True, port=8092)
```

[...]

PR #1778, opened 2021-09-24
data-table: fixes warning about missing key prop

antoinerg:
Closes https://github.com/plotly/dash-table/issues/860 and potentially https://github.com/plotly/dash/issues/1433

As pointed out in https://github.com/plotly/dash-table/pull/690#issue-565576435, in order to see the warnings on the demo page, we need to use React's dev bundles (see https://github.com/plotly/dash/commit/3eac50a0e07dbad395c5d56149126a0e941a8cb0). I also changed the version of the React bundle to match the one used in `package-lock.json`. I don't know if this should stay in the PR but @chabb made a good argument for it to stay in.

The fix itself can be found in https://github.com/plotly/dash/commit/a25cca92254c34ff82cafe38d9e591703f465cda. It adds the one missing `key` prop. After this commit, the missing key warning disappears in the demo page!

cc @alexcjohnson 

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals

### DIFF
--- a/components/dash-table/CHANGELOG.md
+++ b/components/dash-table/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Fixed
+- [#1778](https://github.com/plotly/dash/pull/1778) Fix React warnings stating
+  that each child in a list should have a unique "key" prop
+
 ## [4.12.0] - 2021-07-09
 ### Fixed
 - [#907](https://github.com/plotly/dash-table/pull/907)
--- a/components/dash-table/demo/index.html
+++ b/components/dash-table/demo/index.html
@@ -5,8 +5,8 @@
     </head>
     <body>
         <div id='root'></div>
-        <script src='https://unpkg.com/react@17.0.1/umd/react.production.min.js'></script>
-        <script src='https://unpkg.com/react-dom@17.0.1/umd/react-dom.production.min.js'></script>
+        <script src='https://unpkg.com/react@16.14.0/umd/react.development.js'></script>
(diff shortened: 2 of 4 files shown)


alexcjohnson:
Nice! Is there a way for our tests to catch this warning so we can ensure it doesn't pop up again as we change things?

> I also changed the version of the React bundle to match the one used in package-lock.json. I don't know if this should stay in the PR but @chabb made a good argument for it to stay in.

Agreed - though at some point soon we will want to make react 17 an option you can enable for any given Dash app, at first just for our own testing purposes.

antoinerg:
> Nice! Is there a way for our tests to catch this warning so we can ensure it doesn't pop up again as we change things?

Via a Selenium test, it would probably be possible to check console logs and assert they're empty.

antoinerg:
> Via a Selenium test, it would probably be possible to check console logs and assert they're empty.

Actually, for this to work, we also need to use a development bundle of React in the tests. I'm not sure what's the best way to do that.

cc @alexcjohnson

alexcjohnson:
We serve the dev version of react etc when you're in debug mode. In tests we typically disable hot reloading when we turn on debug, like:

https://github.com/plotly/dash/blob/8c193bc174166513e1e043312303336b0ae169f7/tests/integration/devtools/test_devtools_ui.py#L72-L78

antoinerg:
> Nice! Is there a way for our tests to catch this warning so we can ensure it doesn't pop up again as we change things?

Done in https://github.com/plotly/dash/pull/1778/commits/50f8c0481a77ab5b9353c81d980d1ddabc36bd26

Here's proof that this test detects the warning: https://app.circleci.com/pipelines/github/plotly/dash/2751/workflows/f3be563b-0da9-44e2-a0e1-4f0ad7796b2b/jobs/45091

alexcjohnson (approved):
Love it, thanks for taking care of this! Please add a quick changelog entry, then 💃

PR #1782, opened 2021-09-27
Allow passing Python Objects to Dash components by serialization

akksi:


ndrezn:
Assigning myself as we are exploring this feature for Dash. It's not clear this is the implementation we will move forward with but this is a great proposed strategy for handling Python objects directly in Dash components!

PR #1784, opened 2021-09-28
Two way serialization

workaholicpanda:
## Idea
Provide non-lossy data exchange between server & renderer client in initial layout and callbacks, namely **_two-way serialization_** that allows either party sends any object then receive it back in its original type.

## CHANGELOG

Added a serialization support of `pandas.DataFrame` as the very first type, which will be followed by a few more types like `numpy.array`, `DateTime`, `Plotly figures` and more.

Quick preview of how serialization work
### - Layout
Before :
```python
xValues = numpy.linspace(0, 3, 4)
yValues = numpy.linspace(4, 7, 4)
df = pandas.DataFrame({'x': xValues, 'y': yValues})
table = DataTable(data=df.to_dict('records'), id='data_table_1')
```
After :
```python
xValues = numpy.linspace(0, 3, 4)
yValues = numpy.linspace(4, 7, 4)
df = pandas.DataFrame({'x': xValues, 'y': yValues})
table = DataTable(data=df, id='data_table_1')
```

### - Callback
Before : 
```python
@callback(Output('data_table_2', 'data'), Input('data_table_1', 'data'))
def update(data): # <-- You would receive array object
	return data
```
After :
```python
@callback(Output('data_table_2', 'data'), Input('data_table_1', 'data'))
def update(data): # <-- You would receive pandas.DataFrame object as it's what you've passed to data_table_1
	return data
```

## BREAKDOWNS OF TASKS 

_As described & outlined fully by_ @chriddyp
- [x] Milestone 1 - Strip out the object serialization stuff on either end of the request
- [x] Milestone 2 - Book-keeping: Pass {__type: 'JSON', __value: value} back
- [x] Milestone 3 - Python serializer
- [x] Milestone 4 - Integrate serializer into the `dash` library
- [x] Milestone 5 - Flesh out the serializers

### TESTINGS

Needs sophisticated tests to verify it's working with many various **edge cases**
- [ ] Sending `DataFrame` from server to client (layout)
- [ ] Sending `DataFrame` from server to client (callbacks)
- [ ] Sending `DataFrame` from client to server (callbacks)


dash/dash-renderer/src/serializers/utils.js
```diff
@@ -0,0 +1,40 @@
+import {type, prop} from 'ramda';
+
+const PROP_TYPE = '__type';
+const PROP_VALUE = '__value';
+export const DASH_BOOK_KEEPER = '__dash_serialized_props';
+
+export const createBookkeeper = layout => {
+    const markedLayout = {...layout, __dash_serialized_props: {}};
```
chriddyp, line 8:
```suggestion
    const markedLayout = {...layout, [DASH_BOOK_KEEPER]: {}};
```

dash/dash.py
```diff
@@ -1333,7 +1333,11 @@ def dispatch(self):
         except KeyError as missing_callback_function:
             msg = "Callback function not found for output '{}', perhaps you forgot to prepend the '@'?"
             raise KeyError(msg.format(output)) from missing_callback_function
-        response.set_data(func(*args, outputs_list=outputs_list))
+
+        result_data = DashSerializer.serialize_tree(
```
akksi, line 1337: I'm not sure we need this here, I think at this point, the data should already be serialized?

akksi (commented):
It looks and works well overall as far as I can tell, but seems like it doesn't convert callback inputs back to the original datatype, e.g. callbacks receive data as JSON in {"__type": "DataFrame", "__value": "xyz"} format instead of actual Pandas DataFrame objects.

dash/dash-renderer/src/serializers/utils.js
```diff
@@ -0,0 +1,40 @@
...
+    });
+    return markedLayout;
+};
+
+export const dashSerializeValue = (type, value) => {
+    if (!type) return value;
+    const serializedValue = {};
+    serializedValue[PROP_TYPE] = type;
+    serializedValue[PROP_VALUE] = value;
+    return JSON.stringify(serializedValue);
```
akksi, line 39: Not sure if we need to `stringify` here.

dash/dash-renderer/src/serializers/utils.js
```diff
@@ -0,0 +1,40 @@
...
+        props: {children}
+    } = layout;
+    if (type(children) === 'Array') {
+        markedLayout.props.children = children.map(child =>
+            createBookkeeper(child)
+        );
+    }
+
+    if (type(children) === 'Object') {
+        createBookkeeper(children);
```
akksi, line 20: `createBookkeeper()` returns a modified Object here, shouldn't we update markedLayout's value with the returned value?

anders-kiaer:
@workaholicpanda @akksi @chriddyp Sorry for bumping into this PR, but this looks interesting! 🙂 If I understand this PR correctly, you want to try to support serialization/deserialization of custom Python objects before/after data is sent/received on the server. If so this is great, and something we have seen the need of as well. 🎉 

Looking at the draft PR, I looks like the current sketch is to centralize all serialization in a `class DashSerializer`, using branching with `isinstance` to check for different types. I'm wondering if we risk the branch list could potentially grow quite a bit over time, with many `isinstance` branches/checks necessary on input/output (de)serialization, and still be inflexible for Dash components to e.g. let them define the optimal serialization in their case without them having to do some risky monkey patching of `DashSerializer` centrally? Would love to get your opinions on #1786 if that is relevant here, if not sorry for the noise.

anders-kiaer:
> Could you describe some use cases of both component and data types that you'd like to have custom serializers/deserializers for?

The overall pattern would be components that require a complex data set, which would be easier to modify/adjust/interact with/give as input on the Python side through an appropriate Python object compared with a JSON structure which typically could be structured to make most sense/be most performant in the frontend (React) component.

Here is an attempt on finding a wide distribution of concrete examples:

* I think quite a lot of the components in https://dash.plotly.com/dash-bio could get a simplified API if serialization/deserialization is supported on component level. In order to populate the components with data, the examples typically must fall back on either using "precompiled" JSON data structures or use read function utilities defined "outside" of the component. Having the serialization/deserialization closer to the component makes the usage of them easier, and reduce the risk of read utilities (converting domain specific Python object/data structure → JSON structure) or user made variations of the read utilities coming out of sync with the JSON structure expected by the component. Allowing higher level Python structures being given as input and serialization/deserialization within the component class also makes it easy for the components to internally change JSON structure they consume without it being a breaking change towards the component user.
* A Dash component representing a calculator interface where the user can build mathematical expressions in the frontend.
  * Expression validation happens in frontend (for instant validation and user feedback).
  * You don't want the component user to parse the expression in the backend as well (since the frontend already does it). Encapsulating the serialization/deserialization logic within the component class would make it easier to provide it as a directly consumable high level Python object which could be e.g. something from [`sympy`](https://docs.sympy.org/latest/tutorial/basic_operations.html) in this use case.
* The input data to the custom Dash component [here (choose PriorPosteriorDistribution from the top left dropdown)](https://equinor.github.io/webviz-subsurface-components/) is originally a `pd.DataFrame`, but it includes a lot of redundant data:
  - Each of e.g. 1000+ model realizations go through some iterative update process, where model parameters are updated (gradually from prior to posterior).
  - In the original `pd.DataFrame` there are five columns: Realization model label, update step label, parameter name, parameter value.
  - Quite a bit of the data is redundant since the number of rows in the dataframe is the product of all of the unique entries, `(#number of realizations)*(#number of update steps)*(#number of parameters)`. Utilizing the knowledge about the data structure of the input when converting the `pd.DataFrame` reduces the data sent quite a bit, and is also at the same time received in the frontend in a format more suitable for direct visualization with e.g. `d3` frontend.
  - The take away here is that `pd.DataFrame` is a convenient data container for a lot of use cases, but how it is best serialized (and which assumptions the serialization method can take) is case dependent.
* On the same demo page there are some map component prototypes where the data originally comes from an instance of a Python class defined in a standalone domain specific Python library (with Python C-code-extensions to make it performant). The classes could (through its C-code) efficiently convert detailed map data to associated `.png`-pictures + map coordinate bounds to the frontend component. To increase data image depth there are some tricks involved (e.g. utilize RGB channels not as colors but modulo 256 to increase data depth). All these serialization/deserialization steps would be nice to abstract away within the component itself (i.e. it could take in an instance of the map class and serialize to `base64` `.png` + map bounds, and deserialize back to instance of the map class if requested).

These are the ones I can think of at the moment, but I'm quite sure the community will discover more of them :slightly_smiling_face: 

> Yes the idea here is to support custom python objects but also to deserialize these objects in callbacks. So if you pass a dataframe into dcc.Store then you get a dataframe back from dcc.Store as Input or State.

Multiple valid input formats with serialization/deserialization on component level could perhaps take inspiration from this PR with the `__type` idea (if the below would work it would reduce the code changes necessary in the JavaScript world perhaps). Pseudo-code:
```python

 class Store(Component):    
     @classmethod
     def serialize(prop_name, value: typing.Union[pd.DataFrame, JSONStructure]):
         if isinstance(pd.DataFrame):
              return {"__type": "DataFrame", "data": df.to_json()}  # The frontend component itself now needs to handle the extra layer/key `__type`, but since this function is part of the component itself it should be easy to keep in sync
         return {"__type": "JSONStructure, "data": to_json(value}  # Not a df - serialize as today

     @classmethod
     def deserialize(prop_name, value) -> typing.Union[pd.DataFrame, JSONStructure]:
         if value["__type"] == "DataFrame":
             return pd.read_json(value["data"])
         return from_json(value["data"])
```
Or maybe the `__type` layer could be skipped altogether and let the user decide on callback level the requested de-serialized format useful in that particular callback. :thinking: I.e. a third optional argument to `Output/State(id, prop_name, format)` connected directly to a third optional argument to `deserialize`:
```python
    @classmethod
    def deserialize(prop_name, value, format: Literal["dataframe", "json"] = "json") -> typing.Union[pd.DataFrame, JSONStructure]:
         if format == "dataframe":
             return pd.read_json(value)
         return from_json(value)
```
This would also allow the type hint annotation to potentially be even more precise (i.e. through `typing.Overload` the component class can avoid `typing.Union` on the `deserialize` output and explicitly say what the output would be given a `format` input value).

Besides the flexibility component level serialization/deserialization gives, I think one of the biggest benefit long term having it on component level would probably be the type annotations (which could give the Dash framework automatic documentation building of props on components and their valid Python input formats not relying entirely on docstrings + live editor feedback to the dash app developer through standard Python static type check tools).

emmeb:
^^ Similar to what Anders says, component level serialization would simplify and make a lot of components in dash-bio faster. It would also reduce user error/frustration as files from different bio-software don't always come in the same format. Just under half of the dash-bio apps rely on JSON formats (Molecule3D, Needleplot, OncoPrint,  Circos, Ideogram, Pileup), with the rest being csvs, Fasta, or PDB. Fasta and PDB files are the only files with a set structure, but they are large files, and serialization/deserialization to cache relevant parts would make the interactivity faster. All other files need to be structured perfectly to return the desired visualization, so component-level serialization/deserialization would help a lot.

jackparmer:
Chatting with @chriddyp , we're going to keep the centralized direction that @workaholicpanda has started here. This won't prevent us from adding serializations for Dash Bio and other component suites, and could be a stepping stone to component-level serializations in the future.

PR #1795, opened 2021-10-07
Make long header columns' contents visible in fixed width tables

akksi:
fixes https://github.com/plotly/dash-table/issues/432


akksi:
> @akksi this is a weird one, the failures are repeatable but I don't see them locally. I see a bunch of other cases in that same test fail (all the `fixed_rows1` variants), but not the ones that fail in this PR.
> 
> I don't see any different behavior locally in this test between your PR and `dev` I'm afraid, and AFAICT the behaviors that test is trying to validate are all correct. But I wonder if the problem is related to horizontal scroll: when there are fixed rows and your mouse is over one of the fixed rows (including the header), you can't scroll horizontally. It would be great if we could figure out a fix for that, and a bonus if this makes the tests pass. If we can't fix that, perhaps just change the test to not look at the first row of data but the second?
> 
> FWIW if you want to run that test locally and interact with the exact state being tested, you can use the command:
> 
> ```
> pytest -k ttip001 --pause
> ```
> 
> `-k ttip001` picks just that test case (I don't know of a way to pick the variants you want, other than modifying the `@pytest.mark.parametrize` calls in the test itself), and `--pause` drops you into the python debugger with the test browser open. At that point `c` will go to the next variant, `q` will exit, `n` a couple of times will step you into the test case and then additional `n`'s will step through the test statements.

Thanks for looking into it and the suggestions, will try to move on that path.

akksi:
> DashTable/Fixed Rows & Columns: with 2 fixed rows, 4 fixed columns and merged cells
> DashTable/Fixed Rows & Columns: with 2 fixed rows, 3 fixed columns, hidden columns and merged cells

It looks like enabling `fixed_rows` makes columns narrower on the current `dev` branch; with this fix the column widths stay the same regardless of `fixed_rows`, so it seems like this fix makes the behavior more consistent.
Still feels like we have a width-calculation bug when `merge_duplicate_headers` and `style_data_conditional` are enabled, but I think it's unrelated to this issue (we should probably open a separate one for that).

I fixed the border regressions.

gvwilson:
hi @akksi - we think most of these problems have been addressed by Dash AG Grid, and this PR has gone stale because of delays at our end (for which I apologize), so I'm closing this one as part of our tidying-up effort. I'm sorry we didn't get to it sooner...

Issue #1606, opened 2021-04-27
The 'flask-compress' distribution was not found and is required by the application


europa502:
When using cx_freeze to build a dash app I get the following error- 

```
Traceback (most recent call last):
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\cx_Freeze\initscripts\__startup__.py", line 81, in run
    module.run()
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\cx_Freeze\initscripts\Console.py", line 36, in run
    exec(code, m.__dict__)
  File "strata.py", line 1, in <module>
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\dash_core_components\__init__.py", line 6, in <module>
    import dash as _dash
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\dash\__init__.py", line 5, in <module>
    from .dash import Dash, no_update  # noqa: F401,E402
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\dash\dash.py", line 53, in <module>
    _flask_compress_version = parse_version(get_distribution("flask-compress").version)
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\pkg_resources\__init__.py", line 480, in get_distribution
    dist = get_provider(dist)
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\pkg_resources\__init__.py", line 356, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\pkg_resources\__init__.py", line 899, in require
    needed = self.resolve(parse_requirements(requirements))
  File "C:\Users\menonAbhijit\AppData\Local\Programs\Python\Python39\lib\site-packages\pkg_resources\__init__.py", line 785, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'flask-compress' distribution was not found and is required by the application


```

Here are the versions of dash and flask modules I'm using-
```
dash==1.19.0
dash-bootstrap-components==0.12.0
dash-core-components==1.15.0
dash-html-components==1.1.2
dash-renderer==1.9.0
[...]

PR #1797, opened 2021-10-08
Fetch flask_compress version from __version__ module

europa502:
Fix for #1606

Added code to handle "pkg_resources.DistributionNotFound: The 'flask-compress' distribution was not found and is required by the application" while building app with cx_freeze.
Since __version__ is included only in and above cx_freeze 1.10.0, I've added the try/except condition.


ndrezn:
It looks like this issue was addressed independently in `cx_freeze` `v6.7`, so closing this PR. It doesn't seem like an unreasonable change but it also doesn't appear to be a `try`/`except` we _need_ to consider in Dash.

PR #1801, opened 2021-10-13
Fix Dash Table filter options position

akksi:
Fixes https://github.com/plotly/dash-table/issues/927

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -102,7 +102,6 @@ jobs:
           command: |
             python -m venv venv && . venv/bin/activate
             pip install --upgrade pip wheel
-            pip install webdriver-manager
             pip install -e .[ci,dev,testing] --no-cache-dir --progress-bar off
       - run:
           name: 🌸 Python & JS Lint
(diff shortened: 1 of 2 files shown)


alexcjohnson:
@akksi this looks great. I'm not sure what's going on with `test_browser_smoke[Firefox]` - all it says is
`Service geckodriver unexpectedly exited. Status code was: 64`
Which I don't see documented anywhere. This test runs just fine for me locally, as well as in the `lint-unit-36` job, but seems to be failing repeatably in `lint-unit-39`.

Only thing I can think of is perhaps we're getting a new version of geckodriver that's incompatible with the version of Firefox on the test image `circleci/python:3.9.2-buster-node-browsers` - these images are deprecated in fact, and we really should update to the [`cimg` series](https://circleci.com/developer/images/image/cimg/python) - perhaps [cimg/python:3.9-browsers](https://github.com/plotly/dashboard-engine/blob/cdd1a04c87cc0dec914b79cec7ff16ceda7d29ff/.circleci/config.yml#L316)?

PR #1838, opened 2021-11-17
Two way serialization

workaholicAB:
_Replaces [previous PR](https://github.com/plotly/dash/pull/1784) with squashed commits_ 

## Idea
Provide non-lossy data exchange between server & renderer client in initial layout and callbacks, namely **_two-way serialization_** that allows either party sends any object then receive it back in its original type.

## CHANGELOG

Added a serialization support of `pandas.DataFrame` as the very first type, which will be followed by a few more types like `numpy.array`, `DateTime`, `Plotly figures` and more.

Quick preview of how serialization work
### - Layout
Before :
```python
xValues = numpy.linspace(0, 3, 4)
yValues = numpy.linspace(4, 7, 4)
df = pandas.DataFrame({'x': xValues, 'y': yValues})
table = DataTable(data=df.to_dict('records'), id='data_table_1')
```
After :
```python
xValues = numpy.linspace(0, 3, 4)
yValues = numpy.linspace(4, 7, 4)
df = pandas.DataFrame({'x': xValues, 'y': yValues})
table = DataTable(data=df, id='data_table_1')
```

### - Callback
Before : 
```python
@callback(Output('data_table_2', 'data'), Input('data_table_1', 'data'))
def update(data): # <-- You would receive array object
	return data
```
After :
```python
@callback(Output('data_table_2', 'data'), Input('data_table_1', 'data'))
def update(data): # <-- You would receive pandas.DataFrame object as it's what you've passed to data_table_1
	return data
```


ndrezn:
Assigning myself as we are exploring this feature for Dash. It's not clear this is the implementation we will move forward with but this is a great proposed strategy for handling Python objects directly in Dash components!

robertclaus:
Closing this since it's a very old PR and is likely too stale to matter now. The more recent work with websockets and streaming callbacks likely supercedes this by quite a bit.

Issue #1195, opened 2020-04-14
Ability to access the wrapped function for testing or to import into other programs


chriddyp:
i.e.
```
@app.callback(...)
def update_graph(value):
    ...
```

it would be helpful to be able to access or call `update_graph` _unwrapped_. Currently, if you call `update_graph`, you'll call the _wrapped_ version.

Three use cases:
1. Ability to unit test `update_graph`
2. Ability to call `update_graph` in e.g. Jupyter
3. Ability to call `update_graph` in other programs

I believe this violates how decorators work, so we might have to expose some method on the function like `.inner` so the caller would call e.g. `update_graph.inner(value)`

alexcjohnson:
Actually we don't do anything with the return value of the decorator right now - internal to the decorator we add the wrapped function to the callback map:

https://github.com/plotly/dash/blob/3ec55b5a97fdd34ed8384085df884d86730b1a9c/dash/dash.py#L941

But then we return the wrapped function:

https://github.com/plotly/dash/blob/3ec55b5a97fdd34ed8384085df884d86730b1a9c/dash/dash.py#L943

I suspect if instead we just `return func` everyone will be happy 🤞 (aside from any of our own tests that use the wrapped function - those will have to reference `app.callback_map` instead)

alexcjohnson:
Closed in #1839

shea-parkes:
Just dropping a note a couple years later for others that might land here:

We were actually expecting that we'd have to do `myfunc.__wrapped__` in our unit tests type work.  But since the current `dash.callback` code returns the original func (instead of the wrapped version it created), you can just work directly with the function as-if it wasn't decorated with `dash.callback()`.

We basically overthought things.

PR #1839, opened 2021-11-18
Calling callback functions directly (eg. for tests)

marekpiotrowski:
As per https://github.com/plotly/dash/issues/1195, sometimes there's a need to unit-test some of the callback functions. What we're doing in the team right now is adding a wrapper so that the callback calls a wrapper and we're able to call the function outside of the dash framework.

Initially I thought that what has been suggested in the issue description would be more appropriate (so, having `.inner` attribute on the returned value), but eventually the other suggested solution is more "intuitive" (in the former case, we'd have to update the documentation as well :( )

**The solution is not ideal, obviously, cause if somebody uses the callback context within the callbacks which he/she intends to call in isolation, it's gonna crash (unless it's somehow mocked in the calling context)**

## Contributor Checklist

   -  [X] Implement as suggested in the issue discussion
   -  [X] Adjust unit tests
   -  [X] Add integration test

### optionals


alexcjohnson (approved):
Thanks @marekpiotrowski! And apologies for taking so long to look at this. Great idea to make this change, I just simplified the test and added a changelog entry. 💃

marekpiotrowski:
No problem at all, thanks for checking! Cheers. 

The idea wasn't mine at all - it's all in the linked issue (https://github.com/plotly/dash/issues/1195). Not sure about your processes but would probably make sense to update its status?

By the way, the framework is awesome, keep up the great work you're doing!

alexcjohnson:
Haha ok well thanks for resurrecting the idea and reminding me what I proposed 2 years ago!

PR #1857, opened 2021-12-06
Fix slider marks and steps bugs

### LINKED ISSUE #1843: [BUG] - Docs example 'Non-Linear Slider and Updatemode' doesn't render correctly in Dash 2.1
This Slider example: https://dash.plotly.com/dash-core-components/slider#non-linear-slider-and-updatemode has a functioning slider on the docs page. 
image

Using Dash 2.1, and without any update to the example code, the slider no longer works.

image

I can't see any changes in the Slider's functioning that should cause this. `step` and `marks` are both provided.

### LINKED ISSUE #1851: [BUG] - Dash 2.1 - Slider/RangeSlider - can't restrict selection to defined marks

**Describe your context**
```
dash                 2.1
```

**Describe the bug**
In versions of Dash prior to 2.1, it was possible to create a Slider or RangeSlider that only allowed you to select values defined with `marks`. 

From the docs:

"If slider marks are defined and `step` is set to `None` then the slider will only be able to select values that have been predefined by the `marks`. Note that the default is step=1, so you must explicitly specify None to get this behavior."
[With Slider and RangeSlider i](https://dash.plotly.com/dash-core-components/rangeslider#marks-and-steps) 

This is not possible with Dash 2.1

**Expected behavior**

That it would be possible to create a Slider or RangeSlider that only allowed you to select values defined with `marks`.

HammadTheOne:
This PR addresses the Slider bugs brought up in #1851 and #1843. 

Changes in this PR:

- Fixed a regression where steps were not being set to marks if `None` was passed as the prop argument. 
- Added a check to set the `min` and `max` based on the range of `marks` if they are not explicitly defined.

TODO:

- [x] Add additional tests for slider marks and steps.

- [x] Add CHANGELOG entry.

Closes #1851 
Closes #1843


components/dash-core-components/src/utils/computeSliderMarkers.js
```diff
@@ -144,15 +166,18 @@ export const sanitizeMarks = ({min, max, marks, step}) => {
         return undefined;
     }
 
+    const [min_mark, max_mark] = Object.values(setUndefined(min, max, marks));
```
alexcjohnson, line 169:
```suggestion
    const {min_mark, max_mark} = setUndefined(min, max, marks);
```
Does the same thing right? Except the previous way is dependent on `Object.values` always putting `min_mark` first, which is implementation-dependent (implementation of JS itself and our implementation of `setUndefined`)

alexcjohnson (approved):
💃 Love it - just one small comment. Great tests!

PR #1866, opened 2021-12-13
Adding prop reorder exceptions

### LINKED ISSUE #1840: [BUG] Dash 2.1 - `RadioItems` doesn't take `value` as 2nd positional argument
Dash 2.1 - `RadioItems` doesn't take `value` as 2nd positional argument.

```
dash                 2.1
```
**Describe the bug**

Running the following app where `value` is provided as a `keyword` argument loads with `Montreal` selected:

```
import dash

app = dash.Dash(__name__)

app.layout = dcc.RadioItems(
    options=[
        {'label': 'New York City', 'value': 'NYC'},
        {'label': 'Montréal', 'value': 'MTL'},
        {'label': 'San Francisco', 'value': 'SF'}
    ],
    value='MTL'
)

if __name__ == "__main__":
    app.run_server(debug=True)

```

Running it with positional arguments for `options` and `value`, no `value` is selected on load:

```
import dash

app = dash.Dash(__name__)

app.layout = dcc.RadioItems(
    [
        {'label': 'New York City', 'value': 'NYC'},
        {'label': 'Montréal', 'value': 'MTL'},
        {'label': 'San Francisco', 'value': 'SF'}
    ],
    'MTL'
)

if __name__ == "__main__":
    app.run_server(debug=True)
```

**Expected behavior**

Montreal should show on load as the value is set to 'MTL' with the second positional argument

### LINKED ISSUE #1844: [BUG] Dash 2.1 - DataTable not accepting correct positional arguments
```
dash                 2.1.0
```
**Describe the Bug**
The changelog says 
`Rearranged Keyword Arguments - data and columns the first two keyword arguments which means they can be supplied as positional arguments without the keyword.`

But the following code, based on the example in the changelog under Datatable, throws an error.

```from dash import Dash, dash_table
import pandas as pd

df = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/solar.csv')

app = Dash(__name__)

app.layout = dash_table.DataTable(df.to_dict('records'),id='table')

if __name__ == '__main__':
    app.run_server(debug=True)

```
Error:

```
Invalid argument `active_cell` passed into DataTable with ID "table".
Expected `object`.
Was supplied type `array`.
Value provided: 
[
  {
    "Average MW Per Plant": 15.3,
    "Generation (GWh)": 10826,
    "Installed Capacity (MW)": 4395,
    "Number of Solar Plants": 289,
    "State": "California"
  },
  {
    "Average MW Per Plant": 22.5,
    "Generation (GWh)": 2550,
    "Installed Capacity (MW)": 1078,
    "Number of Solar Plants": 48,
    "State": "Arizona"
  },.....
```
**Expected behavior**
That based on the position of the argument, this would be taken as the data and columns would be generated from the first row.

### LINKED ISSUE #1845: [BUG] - Dash 2.1 list of properties for components are not updated
**Describe your context**

```
dash         2.1.0
```

**Describe the bug**
Generating the docs for a component with `dash.dcc.Dropdown`, for example, shows the properties lists are not updated. 
Here for `Dropdown`, `id` is listed first, which doesn't suggest that `options` and `value` can be provided as first positional arguments. 

**Expected behavior**

Updated list with arguments in new order.

HammadTheOne:
This PR adds an exception conditional for the prop reordering that occurs in `py_components_generation.py`. Previously, the prop order for certain `dcc` components and `dash-table` was changed in the component JS source files, but these changes aren't reflected in the Python backend as props are reordered alphabetically during Python component generation. 

With this PR, there is an additional flag for `dash-generate-components` passed with `-k (keep-prop-order)`. The argument is a comma-separated string of component names which will be exempt from alphabetical prop reordering. The keyword "ALL" can be passed to exempt all components from a particular package. 

Closes #1844
Closes #1845
Closes #1840 

### optionals


HammadTheOne:
@alexcjohnson This PR should be ready for review now :slightly_smiling_face:

LiamConnors:
@HammadTheOne, should this address https://github.com/plotly/dash/issues/1840 too? I see it's not mentioned above, so I guess not. I built off this branch and see the following for:

```help(dash.dcc.RadioItems)```

Output:

```RadioItems(options=undefined, id=undefined, value=undefined...x```

File: components/dash-table/package.json [hunk 1]
```diff
@@ -17,7 +17,7 @@
     "private::build:js": "run-s \"private::build -- --mode production\"",
     "private::build:js-test": "run-s \"private::build -- --mode development --config webpack.test.config.js\"",
     "private::build:js-test-watch": "run-s \"private::build -- --mode development --config webpack.test.config.js --watch\"",
-    "private::build:backends": "dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json && cp dash_table_base/** dash_table/ && dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json --r-prefix 'dash' --r-suggests 'dash' --jl-prefix 'dash' && black dash_table",
+    "private::build:backends": "dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json && cp dash_table_base/** dash_table/ && dash-generate-components src/dash-table/dash/DataTable.js dash_table -p package-info.json -k DataTable --r-prefix 'dash' --r-suggests 'dash' --jl-prefix 'dash' && black dash_table",
```
alexcjohnson, line 20: If we make prop order explicit for table, I think we should be a little more careful about what order we end up with, especially since this will be reflected in the docs. The first two are right: `data` and `columns`. And it makes sense to keep `derived_*` and the dash standards (`loading_state`, `persistence`) at the end. Other than that it's currently a bit random, will be a bit hard for users to find what they want. @chriddyp do you want to give this a shot?

components/dash-core-components/package.json
```diff
@@ -25,7 +25,7 @@
     "test:pyimport": "python -m unittest tests/test_dash_import.py",
     "prebuild:js": "cp node_modules/plotly.js/dist/plotly.min.js dash_core_components_base/plotly.min.js",
     "build:js": "webpack --mode production",
-    "build:backends": "dash-generate-components ./src/components dash_core_components -p package-info.json && cp dash_core_components_base/** dash_core_components/ && dash-generate-components ./src/components dash_core_components -p package-info.json --r-prefix 'dcc' --r-suggests 'dash,dashHtmlComponents,jsonlite,plotly' --jl-prefix 'dcc' && black dash_core_components",
+    "build:backends": "dash-generate-components ./src/components dash_core_components -p package-info.json  && cp dash_core_components_base/** dash_core_components/ && dash-generate-components ./src/components dash_core_components -p package-info.json -k RangeSlider,Slider,DataTable,Dropdown,RadioItems,Checklist --r-prefix 'dcc' --r-suggests 'dash,dashHtmlComponents,jsonlite,plotly' --jl-prefix 'dcc' && black dash_core_components",
```
alexcjohnson, line 28: Likewise here, there are some funny orders that I think we should be more careful about. In `Checklist` and `RadioItems`, what's `inline` doing way at the end, past the dash standards and far from any of the other style props? Some of these are alphabetized, other than the ones we explicitly want upfront, but some aren't - can we be consistent about this?

(BTW `DataTable` isn't needed in this list 😁)

components/dash-core-components/package.json
```diff
@@ -25,7 +25,7 @@
     "test:pyimport": "python -m unittest tests/test_dash_import.py",
     "prebuild:js": "cp node_modules/plotly.js/dist/plotly.min.js dash_core_components_base/plotly.min.js",
     "build:js": "webpack --mode production",
-    "build:backends": "dash-generate-components ./src/components dash_core_components -p package-info.json && cp dash_core_components_base/** dash_core_components/ && dash-generate-components ./src/components dash_core_components -p package-info.json --r-prefix 'dcc' --r-suggests 'dash,dashHtmlComponents,jsonlite,plotly' --jl-prefix 'dcc' && black dash_core_components",
+    "build:backends": "dash-generate-components ./src/components dash_core_components -p package-info.json  && cp dash_core_components_base/** dash_core_components/ && dash-generate-components ./src/components dash_core_components -p package-info.json -k RangeSlider,Slider,Dropdown,RadioItems,Checklist --r-prefix 'dcc' --r-suggests 'dash,dashHtmlComponents,jsonlite,plotly' --jl-prefix 'dcc' && black dash_core_components",
```
alexcjohnson, line 28: Checklist doesn't have mods below - mostly it looks correct already, but perhaps `id` should move down below the style & className props?

components/dash-table/src/dash-table/dash/DataTable.js
```diff
@@ -253,27 +280,40 @@ export const propTypes = {
...
+             * Defaults to 'any' if undefined.
+             *
              *
              */
-            selectable: PropTypes.oneOfType([
-                PropTypes.oneOf(['first', 'last']),
-                PropTypes.bool,
-                PropTypes.arrayOf(PropTypes.bool)
-            ]),
+            type: PropTypes.oneOf(['any', 'numeric', 'text', 'datetime']),
```
alexcjohnson, line 316: Do nested props also get their order preserved by this PR? If not don't worry about it for now, but assuming at some point they will, let's put `type` and `presentation` up top, right below `id` and `name`, as these are very important and commonly used.

alexcjohnson:
We're getting there, looking good!

In the table, let's move [`row_deletable`](https://github.com/plotly/dash/blob/2f779ad71437a4ba178451dafa1f89dcf82c5c49/components/dash-table/src/dash-table/dash/DataTable.js#L551) down a little after [`row_selectable`](https://github.com/plotly/dash/blob/2f779ad71437a4ba178451dafa1f89dcf82c5c49/components/dash-table/src/dash-table/dash/DataTable.js#L569). Also I think [`active_cell`](https://github.com/plotly/dash/blob/2f779ad71437a4ba178451dafa1f89dcf82c5c49/components/dash-table/src/dash-table/dash/DataTable.js#L446) should go next to the other read-only interaction state props, so perhaps right before [`selected_cells`](https://github.com/plotly/dash/blob/2f779ad71437a4ba178451dafa1f89dcf82c5c49/components/dash-table/src/dash-table/dash/DataTable.js#L578)?

alexcjohnson:
Oh actually, re: `CheckList` - see also my comment above about `inline` https://github.com/plotly/dash/pull/1866#discussion_r771787388

components/dash-table/src/dash-table/dash/DataTable.js
HammadTheOne: Updated in 15599ae.

HammadTheOne: Unfortunately nested props aren't preserved. For the moment though, I've made the change in d9e386b.

alexcjohnson (approved):
Excellent! 💃

PR #1869, opened 2021-12-17
plotlyjs 2.8.1

alexcjohnson:
And other dep updates across the board while I'm here.

FYI I encountered a new (to me) flavor of dependency hell here. Both `eslint` and `webpack` depend on the package `eslint-scope`
```
├─┬ eslint@8.4.1
│ └── eslint-scope@7.1.0
└─┬ webpack@5.65.0
  └── eslint-scope@5.1.1
```
One of these will be added to the root of `node_modules`, but because they're different versions we don't know which one. As long as everyone is referencing it directly it's all fine, the correct one will be chosen. However, another package, `babel-eslint`, does some funny resolving that seems to always pick the root version:
```
> cat node_modules/babel-eslint/lib/require-from-eslint.js
"use strict";

const resolve = require("resolve");
const eslintBase = require.resolve("eslint");

module.exports = function requireFromESLint(id) {
  const path = resolve.sync(id, { basedir: eslintBase });
  return require(path);
};
```

I first upgraded deps in `dash-table`, including `eslint 7.30.0 -> 8.4.1`, and everything was fine. I then did the same in `dash-core-components` and the build failed. That's because in `dash-table` we ended up with `eslint-scope@5.1.1` in the root of `node_modules`, which is what `webpack` needed, but in `dash-core-components` we ended up with `eslint-scope@7.1.0` in the root.

The only solution I could find was to downgrade `eslint` back to `^7.30.0` - which to be safe I did in both places.


dash/dash-renderer/package.json
```diff
@@ -35,51 +35,51 @@
...
+    "@typescript-eslint/eslint-plugin": "^5.7.0",
+    "@typescript-eslint/parser": "^5.7.0",
     "babel-loader": "^8.2.3",
     "chai": "^4.3.4",
-    "css-loader": "^5.2.6",
-    "es-check": "^5.2.4",
-    "eslint": "^7.29.0",
+    "css-loader": "^6.5.1",
+    "es-check": "^6.1.1",
+    "eslint": "^8.4.1",
```
archmoj, line 68: So this `eslint` v8 is fine?

alexcjohnson: oh thanks - yeah this one is also subject to the same danger. I'll revert it to v7.

PR #1872, opened 2021-12-23
Fix #1871

### LINKED ISSUE #1871: dash won't install on Python 3.6.4
Good day,

We have large clustered Data environment with no access to the internet.  I'm attempting to install "dash" from our local pip repository, but it doesn't want to install on Python 3.6.4.  I've tried version 2.0.0 and 1.21.1. 

When attempting to install dash 2.0.0, I get the following error:

ERROR: Package 'dash' requires a different Python: 3.6.4 not in '>=3.6'

When attempting to install dash 1.21.1, I get the following error:

ERROR: Package 'dash' requires a different Python: 3.6.4 not in '!=3.0.*,!=3.1.*,!=3.2.*,>=2.7'

Neither of these are appear to actually state 3.6.4 is not supported, however I can't for the life of me get it to install.  Is there a version of dash specific to Python 3.6.4, or is it simply not supported?  It's not an easy environment to update the Python version so this will likely not permit us to use dash in the near future if it is simply not supported.

Apologies if this is not the correct place for this question, I'm just getting a little desparate.

Thanks,

Greg

alexcjohnson:
For some reason Py3.6.4 fails to install dash without explicitly pinning importlib-metadata. Hopefully this will help.

Thanks @gfrair for finding this issue!

### DIFF
--- a/requires-install.txt
+++ b/requires-install.txt
@@ -4,3 +4,4 @@ plotly>=5.0.0
 dash_html_components==2.0.0
 dash_core_components==2.0.0
 dash_table==5.0.0
+importlib-metadata==4.8.3;python_version<"3.7"


archmoj:
Thanks for the PR.
On another note : Could you possibly upgrade plotly.js from v2.8.1 to v2.8.3 please?

PR #1876, opened 2022-01-02
Delayed App Config

bormanjo:
Adds the ability to set/update `Dash.config` attributes via `**kwargs` passed to `Dash.init_app()`.

## Contributor Checklist

   -  [X] Make AttributeDict additive
   -  [X] Update `Dash.__init__()` and `Dash.init_app()`

## Alternative Multi-Page App Example

The delayed app config allows app routes to be decoupled from `Dash.__init__()` when `server=False` 

```python
from dash import Dash, html
from flask import Flask

server = Flask(__name__)

app1 = Dash(name='app1', server=False)
app1.layout = html.Div('This is App1')

app2 = Dash(name='app2', server=False)
app2.layout = html.Div('This is App2')

routes = {
    '/dash-app1/': app1,
    '/dash-app2/': app2,
}

for route, app in routes.items():
    app.init_app(server, routes_pathname_prefix=route, requests_pathname_prefix=route)


@server.route('/')
def index():
    return "<p>Hello, World!</p>"

server.run()

```

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -446,7 +446,6 @@ class Dash:
             config.routes_pathname_prefix.replace("/", "_"), "dash_assets"
         )
 
-        print(assets_blueprint_name)
         self.server.register_blueprint(
             flask.Blueprint(
                 assets_blueprint_name,
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -361,6 +361,10 @@ class Dash:
             ],
             "Read-only: can only be set in the Dash constructor",
         )
+        self.config.finalize(
+            "Invalid config key. Some settings are only available "
+            "via the Dash constructor"
+        )
 
         # keep title as a class property for backwards compatibility
         self.title = title
@@ -433,10 +437,6 @@ class Dash:
             "Read-only: can only be set in the Dash constructor or during init_app()",
         )
 
-        self.config.finalize(
-            "Invalid config key. Some settings are only available "
-            "via the Dash constructor"
-        )
         config = self.config
 
         if app is not None:
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -356,17 +356,10 @@ class Dash:
                 "assets_folder",
                 "assets_url_path",
                 "eager_loading",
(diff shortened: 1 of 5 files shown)


jborman-exos:
@alexcjohnson Thoughts on this? The two percy render tests that failed appear to be issues over the change in month/year (e.g expected December 2021, got January 2022)

dash/dash.py
```diff
@@ -427,8 +420,23 @@ def __init__(
...
+                "routes_pathname_prefix",
+                "requests_pathname_prefix",
+            ],
+            "Read-only: can only be set in the Dash constructor or during init_app()",
+        )
+
+        self.config.finalize(
+            "Invalid config key. Some settings are only available "
+            "via the Dash constructor"
+        )
```
alexcjohnson, line 439: Can we leave `finalize` where it was? If it's here, users could add random items to config before `init_app` is called, but I believe all the valid keys have already been initialized in the `AttributeDict` constructor.

Alternatively we could get rid of `finalize` altogether and make a custom constructor for `AttributeDict` that bakes this in.

jborman-exos: I agree with your first point - I previously misunderstood finalize as "no further changes can be made to the `AttributeDict`". Having re-read the `__setitem__()` implementation this should be moved back where it was.

alexcjohnson: OK great! So let's move it back, then add a changelog entry for this and it'll be ready to go - thanks!

bormanjo: Updated `dash.py`, added a changelog entry and merged with the latest `dev` branch as of c50cbf4

jborman-exos:
@alexcjohnson Does this look good to approve + merge?

alexcjohnson (approved):
Looks great, thanks very much @jborman-exos! 💃

t89:
This set of changes seems to cause an incompatibility with the `jupyter-dash 0.4.0`: https://github.com/plotly/jupyter-dash/issues/75#issue-1118382598

PR #1886, opened 2022-01-11
Encourage cloning with SSH not HTTPS

archmoj:
@alexcjohnson


alexcjohnson (approved):
💃 Sure - though it's interesting reading about GH's own evolving stance on this issue over the years... they've strongly recommended each at different times, now they seem to extremely carefully make no recommendation at all.

PR #1891, opened 2022-01-15
Depend on `plotly.js-dist-min` package instead of `plotly.js`

archmoj:
Use [`plotly.js-dist-min`](https://www.npmjs.com/package/plotly.js-dist-min) package instead of [`plotly.js`](https://www.npmjs.com/package/plotly.js).

`dist/plotly.min.js` appears to be the only file used from `plotly.js` package.

This PR can help reduce dependency conflicts and speed up builds & testing.
See simplified `components/dash-core-components/package-lock.json` diff.

TODO:
- [ ] add entry in the `CHANGELOG.md`

@alexcjohnson


components/dash-core-components/package.json
```diff
@@ -42,12 +42,13 @@
     "@fortawesome/react-fontawesome": "^0.1.16",
     "base64-js": "^1.5.1",
     "color": "^4.1.0",
+    "d3-format": "^1.4.5",
```
archmoj, line 45: For now I used the same version as loaded by `plotly.js`. 
Using newer versions of this module could also be considered in future.

alexcjohnson (approved):
💃 Good call @archmoj - Not that it was done for this purpose, but the small code duplication due to `d3-format` really isn't worth the extra complexity (and locking the `d3-format` version as you point out), much cleaner to switch to the dist version.

alexcjohnson:
I think this one is sufficiently internal-facing that it doesn't need a changelog entry, I'll merge as is.

PR #1894, opened 2022-01-19
error when autogenerated IDs are used with persistence or snapshots

alexcjohnson:
also give set_random_id a leading underscore so it doesn't
need to become a reserved word (disallowed prop name)

- [x] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR (didn't try to mock snapshots appearing in `sys.modules` though, that kind of scares me...)

cc @chriddyp @emmeb seem like reasonable error messages? The traceback will point to the callback definition, and users CAN just set the component ID right there though probably better is to follow the variable reference back to where the component is first created and set the ID at that point.


dash/development/base_component.py
```diff
@@ -167,7 +167,29 @@ def __init__(self, **kwargs):
...
+            )
+        if "dash_snapshots" in sys.modules:
+            raise RuntimeError(
+                """
+                Attempting to use an auto-generated ID in an app with `dash_snapshots`.
+                This is prohibited because snapshots saves the whole app layout,
+                including component IDs, and auto-generated IDs can easily change.
+                Callbacks referencing the new IDs will not work old snapshots.
+
+                Please assign an explicit ID to this component.
```
chriddyp, line 189: Is there any way for us to print out "which" component it is? Or any other identifier (besides the ID) like the component type? If a user has a ton of components or callbacks, it can be difficult for them to find out which component the error message is referring too.

chriddyp: See the error string prefix in `__init__` as an example - might be able to reuse that. the id itself might not be helpful, but at least seeing the component name and type

chriddyp (approved):
Besides my comment about the error message, 💃🏼

dash/development/base_component.py
alexcjohnson: good call, the callback definition will point to multiple components, including the type will help the user narrow it down.

dash/development/base_component.py
```diff
@@ -167,11 +167,38 @@ def __init__(self, **kwargs):
...
 
-    def set_random_id(self):
-        if not hasattr(self, "id"):
-            v = str(uuid.UUID(int=rd.randint(0, 2 ** 128)))
-            setattr(self, "id", v)
-        return getattr(self, "id")
+    def _set_random_id(self):
+
+        if hasattr(self, "id"):
+            return getattr(self, "id")
```
alexcjohnson, line 173: 😅 I had this logic wrong at first - the whole point of this PR is to force folks to add this ID explicitly, so if it's already there we should skip the checks below! [Now that case is tested](https://github.com/plotly/dash/pull/1894/files#diff-16dfc841434d6081a8e87125cac6a04eb38359005a6d6b214f7409d750bf7878R533)

dash/development/component_generator.py
```diff
@@ -23,7 +23,6 @@
     "UNDEFINED",
     "REQUIRED",
     "to_plotly_json",
-    "set_random_id",
```
alexcjohnson, line 26: We really want to limit new additions to `reserved_words`, as these become ineligible as prop names for all dash components anyone ever makes. So I prepended an underscore to the method name.

tests/unit/development/test_base_component.py
```diff
@@ -508,3 +508,29 @@ def update(v1, v2):
     assert input2.id == "23a7711a-8133-2876-37eb-dcd9e87a1613"
     # we make sure that the if the id is set explicitly, then it is not replaced by random id
     assert output3.id == "output-3"
+
+
+def test_debc029_random_id_errors():
+    app = Dash(__name__)
+
+    input1 = dcc.Input(value="Hello Input 1", persistence=True)
```
eff-kay, line 516: I am pretty sure its probably late, and not sure if its valid here. But we can force the user to always Instantiate `Input` with a defined id by making it a required keyword-arg. That way it will simply be a compile error. 

```
class Input():
        def __init__(self, value=None, style=None, *, id):
                ...
                
        
# The following line will throw an error 
# The error states that __init__ is missing a required keyword-arg 'id'

input1 = Input(value="something", style={'color': 'orange'})

              
```

eff-kay: cc @alexcjohnson @chriddyp

eff-kay: Basically anything after the `*` becomes a required keyword arg. and its part of python 3.6.

alexcjohnson: Dash components do have the ability to specify required args - and this gets handled in each component's `__init__` via the component generator. The challenge here is that it's coupled, though you're right that in principle we could handle that in `__init__`, but it would have to be explicit code. This would prevent creating the component and then later setting its ID, but not sure if that's really a pattern we want to promote.

PR #1895, opened 2022-01-20
Support debug=True if native namespace-packages are present

lschmelting:
debug=True fails if namespace-packages are present.

Traceback:
```
  File "tmp.py", line 67, in <module>
    app.run_server(dev_tools_silence_routes_logging=True, debug=True)
  File "/home/lschmelting/.pyenv/versions/py38/lib/python3.8/site-packages/dash/dash.py", line 1958, in run_server
    debug = self.enable_dev_tools(
  File "/home/lschmelting/.pyenv/versions/py38/lib/python3.8/site-packages/dash/dash.py", line 1721, in enable_dev_tools
    component_packages_dist = [
  File "/home/lschmelting/.pyenv/versions/py38/lib/python3.8/site-packages/dash/dash.py", line 1726, in <listcomp>
    else package.filename
AttributeError: '_NamespaceLoader' object has no attribute 'filename'
```

This PR solves the problem for me, though i don't know if there is a more elegant way.


lschmelting:
Linting failed because i'm accessing _path:
W0212: Access to a protected member _path of a client class (protected-access)

but as far as i know, there is no public access to path for namespace-packages.

I'm looking forward to your suggestions on how to handle this.

alexcjohnson:
Super, thanks for the PR @lschmelting!

You can override this linting rule on the relevant line(s):

https://github.com/plotly/dash/blob/0f1b299dce356dbec6c669731663ba7ce6ef057d/dash/dash.py#L535

Is there an easy way we could add a test for this? ie include a namespace package in a test like the [hot reloading test](https://github.com/plotly/dash/blob/dev/tests/integration/devtools/test_hot_reload.py) - we don't need to actually trigger hot reloading from the namespace package IMO, but we need to hit the hot reload code while there's a namespace package loaded. Perhaps turn something like [dash-test-components](https://github.com/plotly/dash/tree/dev/%40plotly/dash-test-components) into a namespace package? Or perhaps there's a way to mock this in the test by directly manipulating `ComponentRegistry.registry` in a way that would fail before this patch and succeed with it?

Then last thing is to add a changelog entry.

lschmelting:
The problem only seems to affect **native** namespace packages. Turning dash-test-components into a native namespace package would require a different folder structure, since implicit namespace packages don't have an __init__.py in the package directory (https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages). 

I took a closer look at the code surrounding `ComponentRegistry.registry`, but i don't see how we could simply mock the existence of a namespace-package yet.

alexcjohnson:
Alright, let's not worry about a test - would be an awfully big lift compared to the code changes here. So the linting fix and a changelog entry and we'll be ready to go :)

alexcjohnson (approved):
💃 Thank you very much @lschmelting!

Issue #1596, opened 2021-04-15
Datepicker start, and end_date not persistent when set from callback. 


vegovs:
I have a callback that sets the datepickers start- and end_date to today. When I refresh the site the dates change to the previous manually selected dates. I.e the dates set by the callback are not persistent. 

```
dash                      1.19.0
dash-bootstrap-components 0.11.3
dash-core-components      1.15.0
dash-extensions           0.0.46
dash-html-components      1.1.2
dash-renderer             1.9.0
dash-table                4.11.2
```
-  if frontend related, tell us your Browser, Version and OS

    - OS: [Ubuntu 20.0]
    - Browser [chrome]
    - Version [22]

**Describe the bug**

As described above. Dates set manually by a user are persistent, but dates set by callbacks are not.

Datepicker:
```
datepicker = dcc.DatePickerRange(
    id='datepicker',
    start_date=date.today(),
    end_date=date.today(),
    start_date_placeholder_text='Fra',
    end_date_placeholder_text='Til',
    initial_visible_month=datetime.now(),
    first_day_of_week=1,
    minimum_nights=0,
    persistence=True,
    persistence_type='session',
    persisted_props=['start_date', 'end_date'],
    with_portal=True,
    show_outside_days=True,
),
```

Callback:
```
@app.callback(
    [Output('datepicker', 'start_date'),
     Output('datepicker', 'end_date')],
    Input('today_button', 'n_clicks'),
)
def set_datepicker_today(n):
    return [date.today(), date.today()] if n else dash.no_update
```

gvwilson:
Hi - we are tidying up stale issues and PRs in Plotly's public repositories so that we can focus on things that are most important to our community. If this issue is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. (Please note that we will give priority to reports that include a short reproducible example.) If you'd like to submit a PR, we'd be happy to prioritize a review, and if it's a request for tech support, please post in our community forum. Thank you - @gvwilson

ndrezn:
Reopening as there is an open PR related to this issue.

arc12:
Also seems to affect the single date picker. Is there a workaround?
Which is the open PR referred to in the last comment

AnnMarieW:
This feature is now available as of v3.1.0  
Fixed by #3279

PR #1900, opened 2022-01-25
Store props from callback

nickmelnikov82:
As described in issue #1596 props passed by callback are not saved in storage. This PR makes it possible to save props.

## Contributor Checklist

   -  [ ] Implement saving persisted props passed by callback
   -  [ ] Add integration test

### optionals


dash/dash-renderer/src/persistence.js
```diff
@@ -306,6 +306,55 @@ const getProps = layout => {
     };
 };
 
+export function setPersistance(layout, newProps, dispatch) {
```
alexcjohnson, line 309: @nickmelnikov82 my apologies for not reviewing this quicker!

Can you explain the differences between this new function and `recordUiEdit`? Ideally we'd 🌴 combine them into one (which we could call `recordEdit` I guess) with just a flag to distinguish the two situations, then any differences would be crystal clear.

The differences I see:
- Here you needed to create `finalPersistence` - makes sense because the UI can't edit `persistence` - but in fact I don't see a harm in including this in `recordUiEdit`, it's not a big overhead. Which brings up that in the callback case it's also possible to alter `persistence_type`, so presumably we should also create `finalPersistenceType`.
- Why the short-circuit on `finalPersistence !== persistence`? I guess maybe it's more complicated in this case, at least if you change from one truthy `persistence` value to another, but it seems to me like we'd still want the new value stored.
- Can you explain the `storage.removeItem` case? Is that situation accessible from the UI as well (ie a case we were always missing)?
- `recordUiEdit` has `const vals = originalVal === undefined ? [newVal] : [newVal, originalVal];` - ie we don't store `undefined`. I'm not sure the practical consequences of this, but presumably we want to keep that behavior?

nickmelnikov82: 1. About finalPersistenceType it makes sense
2. Yes, it's related to case when one truthy persistence value to another. When persistence is changed, the current value is immediately written to the storage as the original value.
3. This is for the case described in the [documentation](https://dash.plotly.com/persistence#explicitly-clearing-saved-data)
4. Without these lines, more than half of the tests fail. I'm not quite sure how storage without the original value is used.

AnnMarieW:
A related issue is the persistence doesn't work when a prop is set using query strings or path variables in a multi-page app.  

Here's a great example (thanks @nopria ) : https://github.com/nopria/dash-persistence-test

Will this PR address this case as well?

gvwilson:
@T4rk1n can you please let us know if this one is still current or has been superseded or gone stale? thanks - @gvwilson

robertclaus:
Closing since this feature was implemented in a different PR and already available in Dash 3.1

Issue #1474, opened 2020-12-01
[BUG] Animated Graph Conflicts with Animated Figure


ghost:
Thank you so much for helping improve the quality of Dash!

We do our best to catch bugs during the release process, but we rely on your help to find the ones that slip through.

**Describe your context**
My Dash Server is hosted on WSL2, and browse the App from my browser.

- result of `pip list | grep dash`:
```
dash                 1.17.0
dash-core-components 1.13.0
dash-html-components 1.1.1
dash-renderer        1.8.3
dash-table           4.11.0
```
-  Browser, Version and OS, reproduced on both Firefox and Edge:

    - OS: Windows 10 64 bits
    - Browser: Firefox, Edge
    - Version: Firefox 83.0, Edge 87.0

**Describe the bug**

When both `dcc.Graph` and `go.Figure` have animation, then, when the second animation in `Figure` is executed, the `Frames` from the first animation are played instead of the second one. There is also a change in the dataset to display.

My findings showed that all the data is correctly sent by the server, and is shown in the callback debug UI. I could not pinpoint exactly where the bug occurs in the JavaScript code as my knowledge is limited. See screenshots for more information.

**Expected behavior**

The animation from the second (and following) request should be played instead of the animation of the first requested data.

**Screenshots**

Sorry for the quality of the GIFs!

Expected result:

![Expected Result](https://user-images.githubusercontent.com/75329862/100779549-29a6aa00-33ce-11eb-958e-881492d8d406.gif)

Observed result:

![Observed Result](https://user-images.githubusercontent.com/75329862/100779588-388d5c80-33ce-11eb-8ace-f3223a75b5dc.gif)

**Reproducible App**

Use the following snippet to reproduce easily the issue:

```python
import dash
import dash_core_components as dcc
import dash_html_components as html
import plotly.graph_objects as go
from dash.dependencies import Input, Output


app = dash.Dash(__name__)

fig = go.Figure(
    data=[],
    layout=go.Layout(
        title='Select a dataset to play',
        yaxis=dict(range=[-1, 5]),
        xaxis=dict(range=[-3, 3]),
        updatemenus=[dict(type="buttons",
                          buttons=[dict(label="Play",
                                        method="animate",
                                        args=[None, {"frame": {"duration": 100, "redraw": True},
                                                     "fromcurrent": False,
                                                     "transition": {"duration": 500,
                                                                    "easing": "quadratic-in-out"}}])])]
    ),
    frames=[],
)

app.layout = html.Div([
[...]

PR #1915, opened 2022-02-10
Fix graph animate frames

nickmelnikov82:
As described in issue #1474, when both dcc.Graph and go.Figure have animation, then, when the second animation in Figure is executed, the Frames from the first animation are played instead of the second one. 

## Contributor Checklist

   -  [x] Fix graph frame update
   -  [x] Make test

### optionals


components/dash-core-components/src/fragments/Graph.react.js
```diff
@@ -155,7 +154,19 @@ class PlotlyGraph extends Component {
...
+            if (figure.frames) {
+                result = Plotly.deleteFrames(gd).then(() => {
+                    return Plotly.addFrames(gd, figure.frames).then(() => {
+                        return Plotly.animate(gd, figure, animation_options);
+                    });
+                });
+            } else {
+                result = Plotly.animate(gd, figure, animation_options);
+            }
+            return result;
```
alexcjohnson, line 169:
```suggestion
            if (figure.frames) {
                return Plotly.deleteFrames(gd)
                    .then(() => Plotly.addFrames(gd, figure.frames))
                    .then(() => Plotly.animate(gd, figure, animation_options));
            }
            return Plotly.animate(gd, figure, animation_options);
```
Nice find! Just simplifying the code a little...

I wonder though, it looks inconsistent that `layoutClone` happens after this. To cover the responsive case wouldn't we want the `Plotly.animate` calls to take `{data: figure.data, layout: layoutClone}` instead of `figure`?

nickmelnikov82: Сalling the react and animate functions does not happen at the same time, but I think it makes sense to pass the same data.

alexcjohnson: OK - I'm pretty sure only `data` and `layout` get used by `Plotly.animate` but at least `frames` was passed in before and hasn't seemed to cause any problems, `config` should be the same.

alexcjohnson:
Looks good! Please merge the latest `dev` and then add a changelog entry for this, then it'll be ready to go!

alexcjohnson (approved):
💃 Thanks!

Issue #1722, opened 2021-08-21
[Feature Request]Add parameters to dash.Dash() to let user ignore "detected a Component for a prop other than `children`" error


CNFeffery:
When I was developing the Dash component, I wanted to pass some Component objects to the component as other parameters, such as a defined icon prefix, so I thought it would be better to add parameters that allow the user to control whether or not to perform the checks shown in the figure below😀

![image](https://user-images.githubusercontent.com/49147660/130328379-d492ecd7-78bd-4dc2-8cd0-df65e52a5e9f.png)

gvwilson:
Hi - we are tidying up stale issues and PRs in Plotly's public repositories so that we can focus on things that are most important to our community. If this issue is still a concern, please add a comment letting us know what recent version of our software you've checked it with so that I can reopen it and add it to our backlog. (Please note that we will give priority to reports that include a short reproducible example.) If you'd like to submit a PR, we'd be happy to prioritize a review, and if it's a request for tech support, please post in our community forum. Thank you - @gvwilson

PR #1916, opened 2022-02-10
Skip check on whether a component is passed as a non-children property.

emilhe:
Officially, Dash only supports passing other Dash components via the `children` prop. And for that reason (I guess), a check is being performed in the Python layer (in `base_component.py`) to catch if people try to do it,

```
if k != "children" and isinstance(v, Component):
    raise TypeError(...)
```

However, it _is_ possible to render components passed via non-children properties - in fact I have developed some JS code (available in [dash-extensions-js](https://github.com/thedirtyfew/dash-extensions-js/))  that makes it possible to do it in [a single line of code](https://github.com/snehilvj/dash-mantine-components/blob/9f8c67a369ffc22bf1a79dc8eaf1319f807e2dc6/src/lib/components/Button.react.js#L25). This option dramatiacally improves the development experience and the flexibility of the resulting Dash components allowing stuff like,

```
import dash
from dash_iconify import DashIconify as Iconify

app = dash.Dash()
app.layout = dmc.Button(
    "Send Mail", leftIcon=[Iconify(icon="fluent:folder-mail-16-filled")]  # dash component passed as non-children property
)

if __name__ == "__main__":
    app.run_server()
```

image

This code already works with the current version of dash, but **only** because the list wrapping of the `Iconify` component cheats the check listed above. Hence the following code will **not** work with the current version of Dash,

```
import dash
from dash_iconify import DashIconify as Iconify

app = dash.Dash()
app.layout = dmc.Button(
    "Send Mail", leftIcon=Iconify(icon="fluent:folder-mail-16-filled")  # dash component passed as non-children property
)

if __name__ == "__main__":
    app.run_server()
```

since the check will be hit. In this PR, I am proposing to remove the check so that the above code will work too.

The main limitation of this approach is that components rendered this way are now known by the dash renderer, so they _cannot be targeted by callbacks_.

### DIFF
--- a/dash/development/base_component.py
+++ b/dash/development/base_component.py
@@ -139,14 +139,6 @@ class Component(metaclass=ComponentMeta):
                     )
                 )
 
-            if k != "children" and isinstance(v, Component):
-                raise TypeError(
-                    error_string_prefix
-                    + " detected a Component for a prop other than `children`\n"
-                    + "Did you forget to wrap multiple `children` in an array?\n"
-                    + "Prop {} has value {}\n".format(k, repr(v))
-                )
-
             if k == "id":
                 if isinstance(v, dict):
                     for id_key, id_val in v.items():


alexcjohnson:
@emilhe really interesting solution with `dash-extensions-js`! This issue has been on our radar for a long time - the only mention I can find of it in a quick search is #1722 but we certainly discussed it internally long before that.

A big caveat with this method is that the dash renderer doesn't know that these components exist, so they can't be used in callbacks as either inputs or outputs - we get that info by crawling the layout looking for IDs, but we only descend into `children`, no other props. That's fine for your use case here, but app devs could easily get tripped up if they wanted to for example change the icon, they'd have to change the whole `leftIcon` prop, not just the `icon` prop inside the `Iconify` component.

It's also a little worrisome that `dash-extensions-js` duplicates a bunch of the logic from `dash-renderer` - this part of the renderer has been pretty stable, but if it ever did change in the future that could break components using this.

On the other hand, `renderDashComponents` being imperative has a nice side effect that you can call it selectively. One use case for components inside arbitrary props would be richer content in `dataTable`, but if we _always_ needed to crawl all the way into the `data` prop looking for components to render, that would have a huge performance cost. The way you've implemented this it could be enabled only when you're making a small table where the cost is acceptable, or only for the cells that actually get rendered.

So I'd love to see if we can find a way to accomplish this, but keep the cool ability to selectively render but do it off the dash-renderer codebase, and maintain the ability to link these components to callbacks.

emilhe:
@alexjohnson yes, I have looked through (some of) the discussions. The performance considerations mentioned there was one of the main reasons I decided to work on an solution via external library rather than pursuing a PR.

I am (painfully) aware of the callback limitation. I actually have a lot of use cases where this limitation is a show stopper, but I figured that with an external library this was probably 'the best that could be done' (I guess that to enable callbacks, a solution implemented inside the dash renderer would be needed). And even without callbacks, I think this solution is a significant step forward as compared to not being able to pass components at all.

I actually started writing my own very simple render code, which worked in terms of rendering, but it emitted some weird React warnings. After trying to debug it without success, I decided to mimic the behaviour of the dash renderer more closely. As a result, the warnings disappered, but the code is now vastly more complex and contains lot's of code duplicated from the dash renderer. I guess the proper solution here would be to seperate out the render functionality in a seperate npm package, which can then be used both by the dash renderer and dash-extensions-js (or the custom component authors directly).

I think that making the rendering optional for the developer at the prop level is the key to avoid performance issues. We could event provide two options; render a prop with callbacks enabled, or render it without callbacks enabled for improved performance. I believe that would yield the best of both worlds.

snehilvj:
following

PR #1923, opened 2022-02-14
`get_relative_path, strip_relative_path, get_asset_url`

AnnMarieW:
Closes https://github.com/plotly/dash-labs/issues/65

This PR creates the following functions:
- `dash.get_relative_path`
- `dash.strip_relative_path`
- `dash.get_asset_url`

This is similar to `dash.callback` where you don't need the `app` object. It makes it possible to use these functions in the `pages` folder of a multi-page app without running into the circular `app` imports issue.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1473,7 +1473,7 @@ class Dash:
         ]
 
     def get_asset_url(self, path):
-        return _get_paths.real_get_assets_url(self.config, path)
+        return _get_paths.real_get_asset_url(self.config, path)
 
     def get_relative_path(self, path):
         """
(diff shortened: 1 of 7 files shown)


dash/dash.py
```diff
@@ -424,6 +424,9 @@ def __init__(
 
         self.logger.setLevel(logging.INFO)
 
+    def _get_config(self):
+        return self.config
```
alexcjohnson, line 428: is this needed?

AnnMarieW: Nope - that was an early attempt at a solution.  I'll delete it

AnnMarieW: done

alexcjohnson (approved):
💃 very nice!

Issue #1157, opened 2020-03-20
[BUG] Javascript error when using dcc.Loading around 3+ Graph objects that use modBarButtons


jordanweaver:
**Describe your context**
Please provide us your environment so we can easily reproduce the issue.

-  replace the result of `pip list | grep dash` below
```
dash                 1.3.1
dash-core-components 1.2.1
dash-html-components 1.2.1
dash-renderer        1.1.0
dash-table           4.3.0
```
-  don't think this is frontend related, but here are my details:

    - OS: macOS
    - Browser Firefox (also seen on Chrome)
    - Version 76.0a1 (2020-03-19)

**Describe the bug**

The app will throw a Javascript error if it uses a `dcc.Loading` component to wrap a collection of `Graph` objects, specifically ones that specify a `config` containing `modBarButtons`. There doesn't seem to be a problem when we only have 2 Graphs, but once we add a third (duplicated) Graph the error appears.

```
must provide button 'click' function in button config
```

Also posted on the Plotly forums [here](https://community.plot.ly/t/click-function-in-button-config-not-defined/36292)

I have also included a reasonably small test app that illustrates the problem. It's probably a bit longer than it needs to be but we weren't sure how many aspects of our setup were complicating the issue. I can try to simplify further if needed.

[click_bug.py.txt](https://github.com/plotly/dash/files/4360394/click_bug.py.txt)

**Expected behavior**

Given that the app works as long as we only duplicate our graph once (for a total of two graph objs), we'd expect this to continue to work when adding additional objects.

**Screenshots**

![2020-03-20 10 18 51](https://user-images.githubusercontent.com/423534/77177998-75a9fc80-6a94-11ea-9c84-925913764494.gif)

cc @ZachPope531

alexcjohnson:
As I mentioned in #1924 - which I closed because we should fix it in plotly.js:

> The problem is [here](https://github.com/plotly/plotly.js/blob/b8905568ee672a60f1749a73dd8703835849087e/src/components/modebar/manage.js#L342) where we mutate the input - which is the user-provided array, something we should not alter. Instead we should create a copy with the object inserted in place of the string. Would be a one-line change if we used `ramda` in plotly.js, but we don't...

PR #1924, opened 2022-02-14
Fix modeBarButtons issue

nickmelnikov82:
As described in issue #1157, when dcc.Graph with the specified "modBarButtons" in the configuration is duplicated, an error occurs. The problem is that dash cannot pass a javascript function to the configuration.

## Contributor Checklist

   -  [x] Fix issue
   -  [x] Make test

### optionals


components/dash-core-components/src/fragments/Graph.react.js
```diff
@@ -227,7 +228,39 @@ class PlotlyGraph extends Component {
...
+        }
+
+        let configClone = mergeDeepRight(
+            config,
+            this.getConfigOverride(responsive)
+        );
+
+        if (this.configModeBarButtons) {
+            configClone = mergeDeepRight(configClone, {
+                modeBarButtons: this.configModeBarButtons,
```
alexcjohnson, line 243: Is there a reason for the `this.configModeBarButtons` indirection? Looks like the idea might have been that from one update to the next if you removed `modeBarButtons` from the `config` (or removed `config` entirely?) the previous set of buttons would remain? I don't see why we'd want that, but also this wouldn't actually happen because `this.getModeBarButtons` will always return a truthy value - unlike in Python, in JS `[]` is truthy. Likewise we're always going to enter the `if (this.configModeBarButtons)` branch here.

So I'd suggest to remove `this.configModeBarButtons` entirely, and have this branch be `if (modeBarButtons.length)`

nickmelnikov82: This was done because the configuration is overwritten inside the component. If we have an empty parameter `modeBarButtons` then it remains empty and component get default config. But if there is a string configuration there, then after initialization  the string is replaced with a JS object and the click function is not passed to the python code. So I made local storage for the string parameter so that it can be reused.

alexcjohnson: > the configuration is overwritten inside the component

That's why we're cloning the config and copying in the original `modeBarButtons` spec, but I still don't see the purpose of `this.configModeBarButtons`. Can you show a situation (even better if it's a test!) where we need this bit of internal state to get the right behavior? I'm just wary of it holding on to a state we no longer want and thus introducing a new more subtle bug.

nickmelnikov82: You can check example code from issue #1157. The problem manifests itself when trying to copy or move a component. It turns out that such parameters `modeBarButtons=[['resetScale2d']]` are transferred during initilization, and such `'modeBarButtons': [[{'name': 'resetScale2d', '_cat': 'resetscale', 'attr': 'zoom', 'val': 'reset', 'icon': {'width': 928.6, 'height': 1000, 'path': 'm786 296v-267q0-15-11-26t-25-10h-214v214h-143v-214h-214q-15 0-25 10t-11 26v267q0 1 0 2t0 2l321 264 321-264q1-1 1-4z m124 39l-34-41q-5-5-12-6h-2q-7 0-12 3l-386 322-386-322q-7-4-13-4-7 2-12 7l-35 41q-4 5-3 13t6 12l401 334q18 15 42 15t43-15l136-114v109q0 8 5 13t13 5h107q8 0 13-5t5-13v-227l122-102q5-5 6-12t-4-13z', 'transform': 'matrix(1 0 0 -1 0 850)'}}]]},` are returned when copying the configuration.

alexcjohnson:
OK - I think I understand what you're doing here now. But I strongly suspect there's a manifestation of this bug in plain plotly.js as well, so I'd much rather we address it there.

The problem is [here](https://github.com/plotly/plotly.js/blob/b8905568ee672a60f1749a73dd8703835849087e/src/components/modebar/manage.js#L342) where we mutate the input - which is the user-provided array, something we should not alter. Instead we should create a copy with the object inserted in place of the string. Would be a one-line change if we used `ramda` in plotly.js, but we don't...

@nickmelnikov82 I'm going to close this PR, if you want to solve it on the plotly.js side please go ahead! Otherwise we'll get it onto @archmoj's plate (or perhaps @emilykl's?) sometime soon.

PR #1926, opened 2022-02-15
update packages diff noticed during build

archmoj:
`npm run build` on current `dev` branch gives diffs for various `package-lock` files.
I thought we could/should commit these changes?
@alexcjohnson

### DIFF
--- a/components/dash-core-components/package-lock.json
+++ b/components/dash-core-components/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dash-core-components",
-  "version": "2.0.0",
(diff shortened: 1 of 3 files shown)


alexcjohnson (approved):
Ugh annoying, good catch. This is currently correct on master but not on dev... will have to take a look at the procedure next time I do a release so we get this into dev too, but you're right that incoming PRs will be nicer if we do this now.
💃

archmoj:
Thank you @alexcjohnson. 
Let's merge this please.

PR #1911, opened 2022-02-04
bump `plotly.js-dist-min` v2.9.0

archmoj:
@alexcjohnson

PR #1929, opened 2022-02-15
use plotly.js-dist-min in webpack externals

archmoj:
Follow up of #1911.
Now that `plotly.js-dist-min` is the name of the package used instead of `plotly.js`.
@alexcjohnson


alexcjohnson:
Can we just delete those? html and table shouldn't use plotly.js at all, right?

archmoj:
> Can we just delete those? html and table shouldn't use plotly.js at all, right?

Good call. Looks like we could. Removed in https://github.com/plotly/dash/pull/1929/commits/0a8cf57ab3cd5c7426747eadb1d029b541f5eac4.

PR #1931, opened 2022-02-16
Fix split by timings for pytest parallelization

archmoj:
Currently on `dev` branch split by timing of parallelized tests is broken:
https://app.circleci.com/pipelines/github/plotly/dash/3084/workflows/f0ae2fb1-1ad0-4d19-8313-e709c10ee72c/jobs/51326

timing-split

As a result tests e.g. `table-server-test` take too much time to complete.
This PR addresses this problem thanks to https://discuss.circleci.com/t/pytest-parallelization-timings/42716 and https://stackoverflow.com/questions/64796476/pytest-junit-xml-report-doesnt-have-file-field

@alexcjohnson

### DIFF
--- a/components/dash-core-components/pytest.ini
+++ b/components/dash-core-components/pytest.ini
@@ -1,4 +1,5 @@
 [pytest]
+junit_family = xunit1
 testpaths = tests/
 addopts = -rsxX -vv
 log_format = %(asctime)s | %(levelname)s | %(name)s:%(lineno)d | %(message)s
--- a/components/dash-core-components/pytest.ini
+++ b/components/dash-core-components/pytest.ini
@@ -1,5 +1,6 @@
 [pytest]
 junit_family = xunit1
+
 testpaths = tests/
 addopts = -rsxX -vv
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Doesn't look like that solved the issue. I did play with this a bunch prior to the dash 2.1 release (unsuccessfully, of course) and I think the issue is with finding the file paths properly, given that we're running these tests from a subdirectory. Let's not spend any more time on this right now - we can make it a cooldown project if you like.

archmoj:
It looks like the timing would be adjusted by circleci after the first successful run now having the benchmarks.

alexcjohnson:
Pretty sure I tried some flavor of `junit_family` already, but if you want to trigger one more run on this branch (with a noise commit) to see if it's now fixed go ahead. If that doesn't do it lets leave this alone until cooldown.

archmoj:
Before: 
before1

After:
after1

Before:
before2

After:
after2

alexcjohnson (approved):
OK great! 💃

PR #1932, opened 2022-02-17
Fix IE11

alexcjohnson:
- Fixes #1925 - Main purpose of this PR is to re-establish Dash compatibility with IE11 (https://github.com/plotly/dash/commit/5b3745d0c3c2490e2d908db9a7a3acf90800b389). Unfortunately as part of this I had to pin `fontawesome` versions back a bit, as the newest ones use a feature (regexp `u` flag) that IE does not support. Some of this will only be caught in manual tests with IE, but I did improve our es-check coverage by changing the sourcemap method away from `eval` which puts all the dev bundle code in strings - since the dash-renderer dev bundle is much bigger than its prod bundle (it contains all the devtools) that was a pretty big blind spot.

While I was at it I tackled two other small bugs:

- Fixes #1914 - text alignment is preserved in dash_table headers. This bug was introduced by #1788 as part of fixing https://github.com/plotly/dash-table/issues/867, the updated css AFAICT covers all of these cases as desired (https://github.com/plotly/dash/commit/f5a9a0c3bfeb7a3ca57baf00710a1976af88c6be)

- Fixes #1919 - `webdriver-manager` was added in #1801, but I believe this was just a misunderstanding of the new CircleCI `cimg` images. I removed it, looks like the tests are running just fine.

- [x] I have added entry in the `CHANGELOG.md`

PR #1938, opened 2022-02-18
Abort build process when encountering npm errors

archmoj:
@alexcjohnson


archmoj:
> Wait this does not work. Let me fix it.

Now catching and displaying errors as expected after https://github.com/plotly/dash/pull/1938/commits/de5e2d6965af5b984ef4d63904b7a85182c248a5.

alexcjohnson (approved):
💃 I think this may no longer be necessary after #1932 tweaked the build command internals - but could still be nice, thanks!

PR #1939, opened 2022-02-18
npm clean-install

archmoj:
https://docs.npmjs.com/cli/v8/commands/npm-ci

We may add a flag to switch back to `npm i` when rebuilding; but for now I think most of these could be converted to `ci`.   

@alexcjohnson


dash/development/update_components.py
```diff
@@ -31,7 +31,7 @@ def booststrap_components(components_source):
     )
 
     cmd = shlex.split(
-        "npx lerna exec --scope *@({})* -- npm i".format(source_glob),
```
alexcjohnson, line 34: During the dev and release processes I think this one should still be `npm i`, but on CI it'd be nice to use `ci` - may need a flag to switch between them?

components/dash-core-components/README.md
```diff
@@ -18,7 +18,7 @@ $ python -m venv venv && . venv/bin/activate
 $ pip install -e .[dev,testing]  # in some shells you need \ to escape []
 
 # run the build process - this will build all of dash, including dcc
-$ npm i && npm run build
+$ npm ci && npm run build
```
alexcjohnson, line 21: I'm not sure the component readmes (and CONTRIBUTING.md files) are still relevant - may have some useful info in them, but a lot of it is probably obsolete.

archmoj: It is mentioned [here](https://github.com/plotly/dash/blob/dev/CONTRIBUTING.md#development-of-dash-core-components-dash-html-components-and-dash_table) that contributors should look into these.
But yeah it would be nice to merge/update those.

dash/development/update_components.py
archmoj: Addressed in https://github.com/plotly/dash/pull/1939/commits/96e547f9c077c135b105b89cd44c570d8265ed65.

archmoj:
Resolved various merge conflicts in https://github.com/plotly/dash/pull/1939/commits/96e547f9c077c135b105b89cd44c570d8265ed65.

alexcjohnson (approved):
💃 Thanks!

PR #1945, opened 2022-02-23
clear unused resolve alias for `plotly.js` in webpack config

archmoj:
Looks unused and rather confusing.
@alexcjohnson


alexcjohnson (approved):
True, no longer needed 💃

Issue #76, opened 2017-07-08
Cannot import dash


nmiculinic:
I've installed dash via pip3 on Ubuntu 16.04LTS, however upon importing dash I have huge error popping out:

```
Type "help", "copyright", "credits" or "license" for more information.
>>> import dash
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/dash/__init__.py", line 1, in <module>
    from .dash import Dash
  File "/usr/local/lib/python3.5/dist-packages/dash/dash.py", line 3, in <module>
    import plotly
  File "/usr/local/lib/python3.5/dist-packages/plotly/__init__.py", line 31, in <module>
    from plotly import (plotly, dashboard_objs, graph_objs, grid_objs, tools,
  File "/usr/local/lib/python3.5/dist-packages/plotly/plotly/__init__.py", line 10, in <module>
    from . plotly import (
  File "/usr/local/lib/python3.5/dist-packages/plotly/plotly/plotly.py", line 29, in <module>
    from plotly import exceptions, files, session, tools, utils
  File "/usr/local/lib/python3.5/dist-packages/plotly/tools.py", line 16, in <module>
    from plotly import exceptions, optional_imports, session, utils
  File "/usr/local/lib/python3.5/dist-packages/plotly/utils.py", line 27, in <module>
    pandas = get_module('pandas')
  File "/usr/local/lib/python3.5/dist-packages/plotly/optional_imports.py", line 23, in get_module
    return import_module(name)
  File "/usr/lib/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/usr/lib/python3/dist-packages/pandas/__init__.py", line 42, in <module>
    import pandas.core.config_init
  File "/usr/lib/python3/dist-packages/pandas/core/config_init.py", line 17, in <module>
    from pandas.core.format import detect_console_encoding
  File "/usr/lib/python3/dist-packages/pandas/core/format.py", line 10, in <module>
    from pandas.core.index import Index, MultiIndex, _ensure_index
  File "/usr/lib/python3/dist-packages/pandas/core/index.py", line 31, in <module>
    from pandas.io.common import PerformanceWarning
  File "/usr/lib/python3/dist-packages/pandas/io/common.py", line 68, in <module>
    from boto.s3 import key
  File "/usr/lib/python3/dist-packages/boto/__init__.py", line 1216, in <module>
[...]

PR #1947, opened 2022-02-24
Add `pages` for multi-page app

AnnMarieW:
This PR moves the `pages` plug-in from dash-labs to dash.
Until the tests are written, the  app in the `usage` folder can be used to try the new features.

To Do
- [x] Find a better way to build `dash.register_page`.   See [dash-labs #76](https://github.com/plotly/dash-labs/issues/76).  Status:  Deferred for now, see comments in dash-labs issue.
- [x] How to call `pages` function only when those features are being used. 
- [x] Delete the `usage` folder
- [x] Add entry in the `CHANGELOG.md`

Test coverage
- test navigation
    - [x] by url
    - [x] by dcc.Link
- [x] test for path variables.
- [x] test for query string
- [x] test for adding arbitrary page information to dash.page_registry 
- [x] test for redirects 
- [x] test for the title
- [x] test for the extra metadata different by page.
- [x] test for layout as function.
- test invalid cases that should be handled.
    - [x] no layout in page.
    - [ ] fail to validate the layout page container.
- [x] test load pages from directory.
- [x] test not found 404 page.
- [x] test for specifying page order

Find a live demo of the `pages` features at https://dashlabs.pythonanywhere.com/


AnnMarieW:
@alexcjohnson 

How much information would you like in the CHANGELOG.md?   Should it just be a pointer to the [forum post?]( https://community.plotly.com/t/introducing-dash-pages-a-dash-2-x-feature-preview/57775)


I'm having an issue with when to call the `pages` function.   The problem is that `dash.register_page` can be called after the app starts, so it can't be used to trigger whether pages features are enabled.    What do you think about having the following Dash args:


> `pages`:  Default `False`.  Set to `True` to enable the multi-page apps features


> `pages_folder`  Default "pages".  A path relative to the current working directory for files used with the multi-page app `pages` feature.



And then for convenience, we can set  `pages=True` if  a `pages_folder` name is supplied, so you don't have to set both props.

dash/dash.py
```diff
@@ -141,6 +166,14 @@ class Dash:
         requested.
     :type assets_folder: string
 
+    :param pages_folder: a path, relative to the current working directory,
+        for pages of a multi-page app. Default ``'pages``.
+    :type pages_folder: string
+
+    : param pages:  Dafault False.  When True, the `pages` feature for multi-page
```
eiriklid, line 173: ``
```suggestion
    : param pages:  Default False.  When True, the `pages` feature for multi-page
```

AnnMarieW (commented):
This [commit](https://github.com/plotly/dash/pull/1947/commits/4764c9cdf46ae11c302c2fc2d33e1bbd121b23e0) is a fix for the bug reported in dash-labs https://github.com/plotly/dash-labs/issues/85

T4rk1n (changes requested):
This is a very nice feature, I like it.

I think now that it's in dash directly instead of dash-labs, we can have better integration for the index rendering.

I am also wondering if there would be a way to do without a `dcc.Location` and container to serve the proper page layout directly in `_dash-layout` instead of an additional callback.

dash/_pages.py
```diff
@@ -0,0 +1,273 @@
...
+def _filename_to_name(filename):
+    return filename.split(".")[-1].replace("_", " ").capitalize()
+
+
+def _infer_path(filename, template):
+    if template is None:
+        path = filename.replace("_", "-").replace(".", "/").lower().split("pages")[-1]
+        path = "/" + path if not path.startswith("/") else path
+        return path
+    else:
```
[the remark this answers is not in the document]
AnnMarieW: Of course :blush:  -- fixed

dash/dash.py
```diff
@@ -13,10 +13,13 @@
 import base64
 from urllib.parse import urlparse
 
+from textwrap import dedent
+import warnings
```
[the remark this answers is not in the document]
AnnMarieW: done

dash/dash.py
```diff
@@ -265,6 +298,8 @@ def __init__(
         name=None,
         server=True,
         assets_folder="assets",
+        pages_folder="pages",
+        pages=False,
```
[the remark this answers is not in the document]
AnnMarieW: Good suggestion.  I changed it to `use_pages` for now, but I think @alexcjohnson was looking for a way to do this without using another attribute.  (but I'm not sure how to do that, so I'll leave this open)

dash/dash.py
```diff
@@ -419,11 +461,17 @@ def __init__(
...
+        # amw
+        self.page_registry = collections.OrderedDict()
+
         if self.server is not None:
             self.init_app()
 
         self.logger.setLevel(logging.INFO)
 
+        if pages:
+            self.enable_pages()
```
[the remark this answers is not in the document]
AnnMarieW: Done

dash/dash.py
```diff
@@ -141,6 +166,14 @@ class Dash:
         requested.
     :type assets_folder: string
 
+    :param pages_folder: a path, relative to the current working directory,
+        for pages of a multi-page app. Default ``'pages``.
+    :type pages_folder: string
+
+    : param pages:  Dafault False.  When True, the `pages` feature for multi-page
+     apps is enabled.
```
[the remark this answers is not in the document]
AnnMarieW: Done

dash/dash.py
```diff
@@ -2041,3 +2092,211 @@ def verify_url_part(served_part, url_part, part_name):
...
+        walk_dir = self.config.pages_folder
+        for (root, _, files) in os.walk(walk_dir):
+            for file in files:
+                if file.endswith(".py") and not file.startswith("_"):
+                    with open(os.path.join(root, file)) as f:
+                        content = f.read()
+                        if "register_page" not in content:
+                            continue
+                if file.startswith("_") or not file.endswith(".py"):
+                    continue
```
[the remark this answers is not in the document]
AnnMarieW: Done

dash/dash.py
```diff
@@ -2041,3 +2092,211 @@ def verify_url_part(served_part, url_part, part_name):
...
+        return {}, None
+
+    def enable_pages(self):
+        if not self.pages:
+            return
+        if self.pages_folder and not os.path.exists(self.config.pages_folder):
+            warnings.warn(
+                f"A folder called {self.pages_folder} does not exist.", stacklevel=2
+            )
+        if os.path.exists(self.config.pages_folder) and self.pages:
```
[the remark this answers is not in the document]
AnnMarieW: I removed the extra `self.pages` check.  I'm not sure what you mean by the other two. I'd like to suppress the warning if `pages_folder` is set to `None` because it is possible to use the `pages` features without having a pages folder.

dash/_pages.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    # COERCE
+    # - Set the order
+    # - Inferred paths
+    page = dict(
+        module=module,
+        supplied_path=path,
+        path_template=None
+        if path_template is None
+        else _validate.validate_template(path_template),
```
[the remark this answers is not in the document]
AnnMarieW: I fixed this by moving the check for `if path_template is None` to the `validate_template` function.  Do you think this is OK?

AnnMarieW:
@T4rk1n Thanks so much for your review! 

Can you say more about how we might do this:
>I am also wondering if there would be a way to do without a dcc.Location and container to serve the proper page layout directly in _dash-layout instead of an additional callback.

dash/dash.py
```diff
@@ -2041,3 +2092,211 @@ def verify_url_part(served_part, url_part, part_name):
...
+                    image=image,
+                    favicon=kwargs["favicon"],
+                    css=kwargs["css"],
+                    app_entry=kwargs["app_entry"],
+                    config=kwargs["config"],
+                    scripts=kwargs["scripts"],
+                    renderer=kwargs["renderer"],
+                )
+
+            self.interpolate_index = interpolate_index
```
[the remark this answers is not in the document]
AnnMarieW: Good point - I'll work on a solution.

dash/dash.py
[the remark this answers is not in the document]
T4rk1n: It could just be check for `os.path.exists(pages_folder)`, but I think for now it's better to have this flag for backward compatibility as users may have a `pages` directory with another multi pages pattern. It could be removed later on in a major version.

dash/dash.py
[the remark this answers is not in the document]
T4rk1n: I didn't think of using the system without the pages directory, in that case I think it's good like that.

T4rk1n:
In `Dash.serve_layout`, you can use `flask.request.referrer` to find the path of the page that sent the request. Could also wrap the layout in the container there so you still have the `dcc.Store(id='_page_title', value=title)` to update the title with the clientside callback.

dash/_pages.py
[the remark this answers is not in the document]
T4rk1n: :+1:

AnnMarieW:
@T4rk1n 
> In `Dash.serve_layout`, you can use `flask.request.referrer` to find the path of the page that sent the request. Could also wrap the layout in the container there so you still have the `dcc.Store(id='_page_title', value=title)` to update the title with the clientside callback.

I see how this could work, but it would also be necessary to include `dash.page_container` [here](https://github.com/plotly/dash/blob/1f38e4f87933c69774b21cd51a24e1ab22759d0b/dash/dash.py#L556) as well, correct?  If so, then it wouldn't be possible for the user to determine where the `dash.page_container`  is placed in the `app.py` layout.

dash/dash.py
```diff
@@ -2041,3 +2092,211 @@ def verify_url_part(served_part, url_part, part_name):
...
+                page_filename = os.path.join(root, file).replace("\\", "/")
+                _, _, page_filename = page_filename.partition("pages/")
+                page_filename = page_filename.replace(".py", "").replace("/", ".")
+                page_module = importlib.import_module(f"pages.{page_filename}")
+
+                self.page_registry = _pages.PAGE_REGISTRY.copy()
+                if f"pages.{page_filename}" in self.page_registry:
+                    self.page_registry[f"pages.{page_filename}"]["layout"] = getattr(
+                        page_module, "layout"
+                    )
```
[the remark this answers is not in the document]
AnnMarieW: Actually, the default error message in the Traceback is pretty good.   Here's an example:

```
  File "/home/amward/dash/usage/app.py", line 5, in <module>
    app = Dash(__name__, use_pages=True)
  File "/home/amward/dash/dash/dash.py", line 472, in __init__
    self.enable_pages()
  File "/home/amward/dash/dash/dash.py", line 2137, in enable_pages
    self._import_layouts_from_pages()
  File "/home/amward/dash/dash/dash.py", line 2113, in _import_layouts_from_pages
    self.page_registry[f"pages.{page_filename}"]["layout"] = getattr(
AttributeError: module 'pages.redirect' has no attribute 'layout'
```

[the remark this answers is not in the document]
AnnMarieW: Would you like to see a different error message, or can I mark this as resolved?

dash/dash.py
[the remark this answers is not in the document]
AnnMarieW: Fixed  in [this commit.](https://github.com/plotly/dash/pull/1947/commits/fdf72b401b51eb231b3cd221244f24e85943ce37)  I think its  a big improvement - thanks for the suggestion!

dash/dash.py
```diff
@@ -405,8 +402,9 @@ def __init__(
 
         _get_paths.CONFIG = self.config
         _pages.CONFIG = self.config
+
         self.pages_folder = pages_folder
-        self.use_pages = use_pages
+        self.use_pages = True if pages_folder != "pages" else use_pages
```
T4rk1n, line 407: Nice :+1:

dash/dash.py
```diff
@@ -875,6 +910,10 @@ def index(self, *args, **kwargs):  # pylint: disable=unused-argument
 
         # use self.title instead of app.config.title for backwards compatibility
         title = self.title
+        pages_metas = ""
+        if self.use_pages:
+            pages_metas = self._pages_meta_tags()
+            title = ""
```
T4rk1n, line 913: `self._pages_meta_tags` could return a tuple  `pages_metas, title = self._pages_meta_tags()`

dash/dash.py
```diff
@@ -827,6 +827,41 @@ def _generate_meta_html(self):
...
+
+        description = start_page.get("description", "")
+        if callable(description):
+            description = (
+                description(**path_variables) if path_variables else description()
+            )
+
+        return dedent(
+            f"""           
+            <meta name="viewport" content="width=device-width, initial-scale=1">
```
T4rk1n, line 845: Would the regular index benefit from this meta too?

AnnMarieW: absolutely!  I'll add it there instead.

dash/dash.py
AnnMarieW: fixed - no need for tuple

dash/__init__.py
```diff
@@ -26,3 +25,14 @@
...
+
+from ._pages import register_page, PAGE_REGISTRY as page_registry  # noqa: F401,E402
+from .dash import (  # noqa: F401,E402
+    Dash,
+    no_update,
+    page_container,
+    _ID_DUMMY,
+    _ID_LOCATION,
+    _ID_STORE,
+    _ID_CONTENT,
```
T4rk1n, line 37: Not sure we need to export those ?

T4rk1n:
There is an issue on circle where it dash locally, but the components are not built at that time so the import to build components because dcc is imported for the layout container. We need to delay the import of dcc for the container, add a function like `get_page_container()` and import the components inside, then you can call that function where you need the container.

AnnMarieW:
In an early version, I used  `get_page_container()`  but @alexcjohnson suggested doing it this way so it didn't need to be a function.  Is there any other workaround?

T4rk1n:
With a check `hasattr(html, 'Div')` you could set the page_container to None if it's not there. It's just for the initial building of components.

dash/dash.py
```diff
@@ -117,14 +117,17 @@
...
 
-page_container = html.Div(
-    children=[
-        dcc.Location(id=_ID_LOCATION),
-        html.Div(id=_ID_CONTENT),
-        dcc.Store(id=_ID_STORE),
-        html.Div(id=_ID_DUMMY),
-    ]
-)
+if not hasattr(html, "Div"):
```
T4rk1n, line 122: It also needs to check `dcc` because that is also used.

dash/_pages.py
```diff
@@ -0,0 +1,271 @@
...
+
+def _infer_image(module):
+    """
+    Return:
+    - A page specific image: `assets/<title>.<extension>` is used, e.g. `assets/weekly_analytics.png`
+    - A generic app image at `assets/app.<extension>`
+    - A logo at `assets/logo.<extension>`
+    """
+    assets_folder = CONFIG.assets_folder
+    valid_extensions = ["apng", "avif", "gif", "jpeg", "png", "webp"]
```
alexcjohnson, line 22: `jpg` is still pretty common too

dash/_pages.py
```diff
@@ -0,0 +1,271 @@
...
+    if template is None:
+        path = (
+            filename.replace("_", "-").replace(".", "/").lower().split(pages_folder)[-1]
+        )
+        path = "/" + path if not path.startswith("/") else path
+        return path
+    # replace the variables in the template with "none" to create a default path if no path is supplied
+    path_segments = template.split("/")
+    default_template_path = ["none" if s.startswith("<") else s for s in path_segments]
+    return "/".join(default_template_path)
```
alexcjohnson, line 68: In the `if template is None` block you ensure the final path starts with `/`, but here it look like it might not - should it?

dash/_pages.py
```diff
@@ -0,0 +1,271 @@
...
+       (string or function) The <meta type="description"></meta>.
+       If not supplied, then nothing is supplied.
+
+    - `image`:
+       The meta description image used by social media platforms.
+       If not supplied, then it looks for the following images in `assets/`:
+        - A page specific image: `assets/<title>.<extension>` is used, e.g. `assets/weekly_analytics.png`
+        - A generic app image at `assets/app.<extension>`
+        - A logo at `assets/logo.<extension>`
+        When inferring the image file, it will look for the following extensions: APNG, AVIF, GIF, JPEG, PNG, SVG, WebP.
```
alexcjohnson, line 171: SVG isn't in the list above

dash/dash.py
```diff
@@ -179,6 +207,14 @@ class Dash:
         requested.
     :type assets_folder: string
 
+    :param pages_folder: a path, relative to the current working directory,
+        for pages of a multi-page app. Default ``'pages``.
+    :type pages_folder: string
+
+    : param pages:  Default False.  When True, the `pages` feature for multi-page
+        apps is enabled.
```
alexcjohnson, line 215:
```suggestion
    :param use_pages:  Default False, or True if you set a non-default ``pages_folder``.
        When True, the ``pages`` feature for multi-page apps is enabled.
```

dash/dash.py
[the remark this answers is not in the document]
alexcjohnson: Yeah I think the way you have it now is about as minimal as we can get: either `use_pages=True` or `pages_folder=<something_nondefault>` to enable the feature. So let's keep it this way.

dash/dash.py
```diff
@@ -2076,3 +2170,137 @@ def verify_url_part(served_part, url_part, part_name):
...
+                    if "register_page" not in content:
+                        continue
+
+                page_filename = os.path.join(root, file).replace("\\", "/")
+                _, _, page_filename = page_filename.partition(pages + "/")
+                page_filename = page_filename.replace(".py", "").replace("/", ".")
+                module_name = ".".join([pages, page_filename])
+                page_module = importlib.import_module(module_name)
+
+                self.page_registry = _pages.PAGE_REGISTRY.copy()
```
alexcjohnson, line 2193: Can you say more about why we're copying `PAGE_REGISTRY` into the app instance instead of just continuing to use the one in `_pages`? Note that `.copy()` is shallow, so the assignment you're doing below is modifying the original anyway.

If we want `app.page_registry` to be available for users we could do that (`self.page_registry = _pages.PAGE_REGISTRY` once in the constructor) but even that I don't really see the purpose, since we also have `dash.page_registry`. And keeping the copy if we don't need it is a risk of bugs if there's ever a way the two copies could get out of sync.

dash/dash.py
```diff
@@ -2076,3 +2170,137 @@ def verify_url_part(served_part, url_part, part_name):
...
+                    self.strip_relative_path(pathname)
+                )
+
+                # get layout
+                if page == {}:
+                    if "pages.not_found_404" in self.page_registry:
+                        layout = self.page_registry["pages.not_found_404"]["layout"]
+                        title = self.page_registry["pages.not_found_404"]["title"]
+                    else:
+                        layout = html.H1("404")
```
alexcjohnson, line 2251:
```suggestion
                        layout = html.H1("404 - Page not found")
```
A little friendlier to non-techy viewers :)

dash/dash.py
```diff
@@ -2076,3 +2170,137 @@ def verify_url_part(served_part, url_part, part_name):
...
+                query_parameters = _parse_query_string(search)
+                page, path_variables = self._path_to_page(
+                    self.strip_relative_path(pathname)
+                )
+
+                # get layout
+                if page == {}:
+                    if "pages.not_found_404" in self.page_registry:
+                        layout = self.page_registry["pages.not_found_404"]["layout"]
+                        title = self.page_registry["pages.not_found_404"]["title"]
```
alexcjohnson, line 2249: non-blocking, but this could be made a little more DRY by starting with `not_found_404 = self.page_registry.get("pages.not_found_404")`

tests/integration/multi_page/test_pages_layout.py
```diff
@@ -0,0 +1,136 @@
...
+    for page in dash.page_registry.values():
+        dash_duo.find_element("#" + page["id"]).click()
+        dash_duo.wait_for_text_to_equal("#text_" + page["id"], "text for " + page["id"])
+        assert dash_duo.driver.title == page["title"], "check that page title updates"
+
+    # test redirects
+    dash_duo.wait_for_page(url="http://localhost:8050/v2")
+    dash_duo.wait_for_text_to_equal("#text_redirect", "text for redirect")
+    dash_duo.wait_for_page(url="http://localhost:8050/old-home-page")
+    dash_duo.wait_for_text_to_equal("#text_redirect", "text for redirect")
```
alexcjohnson, line 62: Should the url have changed to `/redirect` in these cases? Can we test that?

dash/dash.py
AnnMarieW: I changed it to a try/except AttributeError.  If it's better to check for all the attributes with hasattr, please let me know and I'll change it back

dash/_pages.py
AnnMarieW: Good catch.  I updated to include it.

dash/dash.py
AnnMarieW: Great suggestion.  Plus I just noticed that I was still using "pages" rather than self.pages_folder in the module name and updated that as well.

dash/dash.py
```diff
@@ -2076,3 +2170,137 @@ def verify_url_part(served_part, url_part, part_name):
...
+                + [self.layout]
+            )
+            if _ID_CONTENT not in self.validation_layout:
+                raise Exception("`dash.page_container` not found in the layout")
+
+            # Update the page title on page navigation
+            self.clientside_callback(
+                """
+                function(data) {{
+                    document.title = data.title || 'Dash'
```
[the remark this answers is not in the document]
AnnMarieW: I just removed the `|| 'Dash'` because data.title is updated [here](https://github.com/plotly/dash/blob/2780bb6d1a66535a8f3799d0c800ab7852d3efd6/dash/dash.py#L2269)

tests/integration/multi_page/test_pages_layout.py
AnnMarieW: Added the test

dash/dash.py
AnnMarieW: I thought a `self.page_registry` was required, but it's unnecessary.  It's much better without it.

dash/dash.py
alexcjohnson: LGTM - I might just suggest adding a comment about _why_ we need this, so we don't accidentally remove it later.

dash/dash.py
```diff
@@ -2076,3 +2166,136 @@ def verify_url_part(served_part, url_part, part_name):
...
+                        layout = not_found_404["layout"]
+                        title = not_found_404["title"]
+                    else:
+                        layout = html.H1("404 - Page not found")
+                        title = self.title
+                else:
+                    layout = page["layout"]
+                    title = page["title"]
+
+                if isinstance(layout, patch_collections_abc("Callable")):
```
alexcjohnson, line 2252: Despite the fact that we have this pattern elsewhere in this repo, `callable(layout)` like you [have below](https://github.com/plotly/dash/pull/1947/files#diff-187d1883eb8273e47c7fae609bb5667d5a9488f9406dfc6a9ca5f19d8d7cc205R2268) is a [better pattern](https://stackoverflow.com/questions/624926/how-do-i-detect-whether-a-python-variable-is-a-function)

(and [6 lines below](https://github.com/plotly/dash/pull/1947/files#diff-187d1883eb8273e47c7fae609bb5667d5a9488f9406dfc6a9ca5f19d8d7cc205R2258) as well)

dash/_pages.py
```diff
@@ -0,0 +1,294 @@
...
+       e.g. `pages.weekly_analytics` to `Weekly analytics`
+
+    - `description`:
+       (string or function) The <meta type="description"></meta>.
+       If not supplied, then nothing is supplied.
+
+    - `image`:
+       The meta description image used by social media platforms.
+       If not supplied, then it looks for the following images in `assets/`:
+        - A page specific image: `assets/<title>.<extension>` is used, e.g. `assets/weekly_analytics.png`
```
LiamConnors, line 185: Would <module>.<extension> be more accurate here?

AnnMarieW: Yes, good catch - thanks

dash/dash.py
```diff
@@ -443,6 +443,7 @@ def __init__(  # pylint disable=too-many-statements
 
         _get_paths.CONFIG = self.config
         _pages.CONFIG = self.config
+        _get_paths.SERVER = self.server
```
alexcjohnson, line 446: We’ll need to cover the case where you create a server later and add it via `app.init_app(server)` - so I think this line should move down after this one:

https://github.com/plotly/dash/blob/693b62cc68701628b4836fc80f6306af5ae50a9b/dash/dash.py#L481

You asked (on slack) about tests - these can be pretty simple, and should go in `tests/unit` somewhere. Just create an app - once the normal way, once passing in a server later - but in each case finish with:

`assert dash.get_server() == app.server`

AnnMarieW: Great - thanks.
I put this in the `_get_paths` folder with the other functions like `dash.get_asset_url` since it's kinda similar.  Should it be somewhere else, or rename this folder to something more generic in case there are other functions that would would like to add here as well?

T4rk1n (approved):

PR #1949, opened 2022-02-25
Implement SVG rendering of equations in `dcc.Graph` and `dcc.Markdown` using `MathJax` v3

archmoj:
Also bumping `plotly.js` to `v2.10.1`.
See additions and fixes in [v2.10.0](https://github.com/plotly/plotly.js/releases/tag/v2.10.0) and [v2.10.1](https://github.com/plotly/plotly.js/releases/tag/v2.10.1).

Example app:

```py
import dash
from dash import dcc
from dash import html

app = dash.Dash(__name__)

false = False
true = True

fig = {
    "data":[{
        "x": [0, 1],
        "y": [0, 1.414],
        "name": "$E^2=m^2c^4+p^2c^2$"
    }, {
        "x": [0, 1],
        "y": [1.4, 0.1],
        "type": "bar",
        "name": "$x=\\frac{-b \\pm \\sqrt{b^2-4ac}}{2a}$"
    }, {
        "type": "pie",
        "values": [1, 9],
        "labels": ["$\\frac{1}{10}=10\\%$", "$\\frac{9}{10}=90\\%$"],
        "domain": {"x": [0.3, 0.75], "y": [0.55, 1]}
    }, {
        "type": "heatmap",
        "z": [[1,2],[3,4]],
        "xaxis": "x2",
        "yaxis": "y2",
        "colorbar": {"y": 0.225, "len": 0.45}
    }],
    "layout": {
        "yaxis":{"domain": [0, 0.45], "title": {"text": "$y=\\sin{2 \\theta}$"}},
        "xaxis":{
            "domain": [0, 0.45],
            "title": {"text": "$x=\\int_0^a a^2+1$"},
            "tickvals": [0, 1],
            "ticktext": ["$\\frac{0}{100}$", "$\\frac{100}{100}$"]
        },
        "xaxis2": {"domain": [0.85, 1], "anchor": "y2"},
        "yaxis2": {
            "domain": [0, 0.45],
            "anchor": "x2",
            "title": {"text": "$(||01\\rangle+|10\\rangle)/\\sqrt2$"}
        },
        "height":500,
        "width":800,
        "margin": {"r": 250},
        "title": {"text": "$i\\hbar\\frac{d\\Psi}{dt}=-[V-\\frac{-\\hbar^2}{2m}\\nabla^2]\\Psi$"},
        "annotations":[
            {
                "text":"$(top,left)$","showarrow":false,"xref":"paper","yref":"paper",
                "xanchor":"left","yanchor":"top","x":0,"y":1,"textangle":10,
                "bordercolor":"#0c0","borderpad":3,"bgcolor":"#dfd"
            },
            {
                "text":"$(right,bottom)$","xref":"paper","yref":"paper",
                "xanchor":"right","yanchor":"bottom","x":0.2,"y":0.7,"ax":-20,"ay":-20,
                "textangle":-30,"bordercolor":"#0c0","borderpad":3,"bgcolor":"#dfd",
                "opacity":0.5
            },
            {"text":"$not-visible$", "visible": false},
            {
                "text":"$^{29}Si$","x":0.7,"y":0.7,"showarrow":false,
                "xanchor":"right","yanchor":"top"
            },
            {
                "text":"$^{17}O$","x":0.7,"y":0.7,"ax":15,"ay":-15,
                "xanchor":"left","yanchor":"bottom"
            }
        ]
    }
}

    dcc.Markdown('# h1 tag with inline MathJax: $E=mc^2$', mathjax=True),

    dcc.Markdown('''
        ## h2 tag with MathJax block:
        $$
        \\frac{1}{(\\sqrt{\\phi \\sqrt{5}}-\\phi) e^{\\frac25 \\pi}} =
        1+\\frac{e^{-2\\pi}} {1+\\frac{e^{-4\\pi}} {1+\\frac{e^{-6\\pi}}
        {1+\\frac{e^{-8\\pi}} {1+\\ldots} } } }
        $$
        ## Next line.
    ''', mathjax=True),

    dcc.Graph(
        mathjax=True,
        id='graph-with-math',
        figure=fig
    ),

    dcc.Markdown('### No MathJax: Apple: $2, Orange: $3'),

    dcc.Graph(
        id='graph-without-math',
        figure=fig
    )
])

if __name__ == '__main__':
    app.run_server(debug=True)
```


components/dash-core-components/package.json
```diff
@@ -46,9 +46,10 @@
     "fast-isnumeric": "^1.1.4",
     "file-saver": "^2.0.5",
     "highlight.js": "^11.4.0",
+    "mathjax": "3.2.0",
     "moment": "^2.29.1",
     "node-polyfill-webpack-plugin": "^1.1.4",
-    "plotly.js-dist-min": "2.9.0",
+    "plotly.js-dist-min": "file:packages/plotly.js-dist-min-2.9.0.tgz",
```
archmoj, line 52: https://github.com/plotly/plotly.js/pull/6073

components/dash-core-components/package.json
```diff
@@ -46,9 +46,10 @@
     "fast-isnumeric": "^1.1.4",
     "file-saver": "^2.0.5",
     "highlight.js": "^11.4.0",
+    "mathjax": "3.2.0",
```
archmoj, line 49: https://www.npmjs.com/package/mathjax/v/3.2.0
Perhaps we could change & relax this to "^3.0.0"?

components/dash-core-components/package.json
```diff
@@ -60,6 +61,7 @@
     "react-resize-detector": "^6.7.6",
     "react-select-fast-filter-options": "^0.2.3",
     "react-virtualized-select": "^3.1.3",
+    "remark-math": "^3.0.1",
```
archmoj, line 64: We should decide which version to use.
I was able to use v1 and v2 as well; but not v4 and v5.
https://www.npmjs.com/package/remark-math?activeTab=versions

components/dash-core-components/src/fragments/Markdown.react.js
```diff
@@ -131,7 +134,16 @@ export default class DashMarkdown extends Component {
...
+                    plugins={mathjax ? [RemarkMath] : []}
                     renderers={{
+                        math: props => (
+                            <Math tex={props.value} inline={false} />
+                        ),
+
+                        inlineMath: props => (
+                            <Math tex={props.value} inline={true} />
+                        ),
+
```
archmoj, line 146: Thanks to @anders-kiaer for the contribution in :trophy: :trophy: :trophy: https://github.com/plotly/dash-core-components/pull/634 :trophy: :trophy: :trophy:

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -93,3 +93,56 @@ def test_mkdw002_dcclink(dash_dcc):
     dash_dcc.percy_snapshot("mkdw002 - markdowns display")
 
     assert dash_dcc.get_logs() == []
+
+
+def test_mkdw003_without_mathjax(dash_dcc):
+    app = Dash(__name__, eager_loading=False, assets_folder="../../assets")
```
alexcjohnson, line 99:
```suggestion
    app = Dash(__name__)
```
Best to leave these extras out unless you're explicitly testing them 😄

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -93,3 +93,56 @@ def test_mkdw002_dcclink(dash_dcc):
...
+
+    app.layout = html.Div(
+        [
+            dcc.Markdown("# No MathJax: Apple: $2, Orange: $3"),
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    dash_dcc.percy_snapshot("mkdw003 - markdown without mathjax")
```
alexcjohnson, line 108:
```suggestion
    dash_dcc.wait_for_text_to_equal("h1", "No MathJax: Apple: $2, Orange: $3")
    assert not dash_dcc.driver_execute_script("return !!window.MathJax")
```

We aren't testing anything particularly visual here - so I'd rather we not add a snapshot, instead just make DOM & JS assertions

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -93,3 +93,56 @@ def test_mkdw002_dcclink(dash_dcc):
...
+    app = Dash(__name__, eager_loading=False, assets_folder="../../assets")
+
+    app.layout = html.Div(
+        [
+            dcc.Markdown("# h1 tag with inline MathJax: $E=mc^2$", mathjax=True),
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    dash_dcc.percy_snapshot("mkdw004 - markdown inline mathjax")
```
alexcjohnson, line 122: This part is visual, so a snapshot makes sense, but I'd probably combine this and the next test into one that covers both cases, just so there are fewer snapshots to look at.

components/dash-core-components/tests/integration/graph/test_graph_varia.py
```diff
@@ -847,3 +859,77 @@ def graph_dims():
...
+    dash_dcc.start_server(
+        app,
+        debug=True,
+        use_reloader=False,
+        use_debugger=True,
+        dev_tools_hot_reload=False,
+    )
+
+    # Give time for the async dependency to be requested (if any)
+    time.sleep(2)
```
alexcjohnson, line 889:
```suggestion
    dash_dcc.wait_for_element("button#btn")
```
I hate to add explicit waits... can't we just wait for the content on the page to be complete?

Same below, we should have the graph invoke MathJax, then just test for the presence of a MathJax element in the graph.

components/dash-core-components/tests/integration/graph/test_graph_varia.py
```diff
@@ -847,3 +859,77 @@ def graph_dims():
...
+    assert dash_dcc.get_logs() == []
+
+
+def test_grva011_without_mathjax(dash_dcc):
+    app = Dash(__name__, eager_loading=False, assets_folder="../../assets")
+
+    app.layout = html.Div([dcc.Graph(id="output", figure={"data": [{"y": [3, 1, 2]}]})])
+
+    dash_dcc.start_server(app)
+    dash_dcc.percy_snapshot("grva011 - graph without mathjax")
```
alexcjohnson, line 913: Here too, no need for a snapshot (we test these visuals on the plotly.js side), but let's put something in the graph that would render as math and test that it shows up as plain text, as well as testing that `window.MathJax` doesn't exist.

alexcjohnson: Also maybe we should parametrize `eager_loading` like `grva001` does, and assert `find(Sync|Async)MathJax` gives the expected results? I guess in the eager case `window.MathJax` will be defined, but the math should still not render.

We should parametrize `eager_loading` in grva012 too - and again, I'd probably drop the snapshot and just make some DOM assertions to show that MathJax renders... but if you keep it after parametrizing note you'd need to make the names unique.

components/dash-core-components/tests/integration/markdown/test_markdown.py
alexcjohnson: I guess this one wants both `eager_loading` flavors as well?

components/dash-core-components/src/fragments/Math.react.js
```diff
@@ -0,0 +1,51 @@
...
+            window.MathJax.typeset([this.span_element.current]);
+        }
+    }
+
+    render() {
+        return (
+            <span ref={this.span_element}>
+                {this.props.inline ? String.raw`\(` : String.raw`\[`}
+                {this.props.tex}
+                {this.props.inline ? String.raw`\)` : String.raw`\]`}
```
alexcjohnson, line 34:
```suggestion
                {this.props.inline ? '\\(' : '\\['}
                {this.props.tex}
                {this.props.inline ? '\\)' : '\\]'}
```
I had never seen `String.raw` before - thanks @anders-kiaer 🧑‍🏫  - but doesn't it seem unnecessary here with only one escaped character?

components/dash-core-components/src/fragments/Math.react.js
```diff
@@ -0,0 +1,51 @@
...
+        if (
+            prevProps.tex !== this.props.tex ||
+            prevProps.inline !== this.props.inline
+        ) {
+            this.renderMath();
+        }
+    }
+
+    renderMath() {
+        if (window.MathJax && window.MathJax.typeset) {
```
alexcjohnson, line 24:
```suggestion
        if (window.MathJax?.typeset) {
```
Living in the modern JS world over here, not like plotly.js 😅

components/dash-core-components/src/fragments/Graph.react.js
```diff
@@ -159,8 +159,10 @@ class PlotlyGraph extends Component {
             return Plotly.animate(gd, figure, animation_options);
         }
 
-        const configClone = this.getConfig(config, responsive);
         const layoutClone = this.getLayout(figure.layout, responsive);
+        const configClone = this.getConfig(config, responsive);
+        // add typesetMath | not exposed to the dash API
+        configClone.typesetMath = this.props.mathjax;
```
alexcjohnson, line 165: Nonblocking, but ideally this would be baked into `this.getConfig(config, responsive, mathjax)`

components/dash-core-components/src/fragments/Graph.react.js
```diff
@@ -170,6 +172,9 @@ class PlotlyGraph extends Component {
             frames: figure.frames,
             config: configClone,
         }).then(() => {
+            // clear typesetMath
+            configClone.typesetMath = undefined;
```
alexcjohnson, line 176: Is it necessary to clear this? It's already a clone of the input config.

components/dash-core-components/package.json
alexcjohnson: Probably because we're on such an old version of `react-markdown`. Let's keep it at 3.0.1, it's going to be a substantial project (likely mostly on the QA side) to upgrade these.

components/dash-core-components/package.json
alexcjohnson: We don't really package DCC for anyone else to use at the JS level - so we've generally gone with `^latest` - ie `^3.2.0` in this case.

archmoj: Done in 8ba1dcf8b22e251d7f77f020a6b9204246e8744d.

archmoj:
@alexcjohnson This PR is now ready for review. Thank you!

components/dash-core-components/src/fragments/Graph.react.js
archmoj: Addressed in https://github.com/plotly/dash/pull/1949/commits/0ee422d5aae317c609b511eec8a6625b8e5e59ea.

archmoj:
@alexcjohnson please view my recent comments regarding `eager` and `percy` tests.
If needed, please feel free to push your suggestions to the branch, while I would be OOO.
Thank you!

components/dash-core-components/tests/integration/graph/test_graph_varia.py
```diff
@@ -847,3 +859,77 @@ def graph_dims():
...
+
+
+@pytest.mark.parametrize("is_eager", [True, False])
+def test_grva011_without_mathjax(dash_dcc, is_eager):
+    app = Dash(__name__, eager_loading=is_eager, assets_folder="../../assets")
+
+    app.layout = html.Div([dcc.Graph(id="output", figure={"data": [{"y": [3, 1, 2]}]})])
+
+    dash_dcc.start_server(app)
+    assert dash_dcc.get_logs() == []
```
alexcjohnson, line 914: Here we should insert something into the graph that WOULD be rendered as math if that were enabled, and verify that it is left as normal text. So if using the title I suggested for the previous test, the assertion could be like:
```py
assert dash_dcc.wait_for_element(".gtitle").text == "$E=mc^2$"
```
If it's rendered as math, that text is `""` because it's just a bunch of SVG paths.

Also we should assert that `find(Sync|Async)MathJax` is as expected - I guess `sync==is_eager`, `async==False`?

components/dash-core-components/tests/integration/graph/test_graph_varia.py
```diff
@@ -847,3 +859,77 @@ def graph_dims():
...
+                figure={
+                    "data": [{"y": [3, 1, 2]}],
+                    "layout": {"title": {"text": "Equation: $E=mc^2$"}},
+                },
+            )
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    assert dash_dcc.get_logs() == []
```
alexcjohnson, line 935: Same here: check that we really see rendered math (`dash_dcc.wait_for_element(".gtitle-math")`) and `find(Sync|Async)MathJax` is as expected

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
     )
 
     dash_dcc.start_server(app)
-    dash_dcc.percy_snapshot("mkdw002 - markdowns display")
+    assert dash_dcc.get_logs() == []
+
+
+@pytest.mark.parametrize("is_eager", [True, False])
+def test_mkdw003_without_mathjax(dash_dcc, is_eager):
+    app = Dash(__name__)
```
alexcjohnson, line 99:
```suggestion
    app = Dash(__name__, eager_loading=is_eager)
```
Which is presumably going to require changing the assertion below to:
```py
assert dash_dcc.driver.execute_script("return !!window.MathJax") == is_eager
```

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+
+    dash_dcc.start_server(app)
+    dash_dcc.wait_for_text_to_equal("h1", "No MathJax: Apple: $2, Orange: $3")
+    assert not dash_dcc.driver.execute_script("return !!window.MathJax")
+    assert dash_dcc.get_logs() == []
+
+
+@pytest.mark.parametrize("is_eager", [True, False])
+def test_mkdw004_inline_mathjax(dash_dcc, is_eager):
+    app = Dash(__name__, eager_loading=False, assets_folder="../../assets")
```
alexcjohnson, line 115:
```suggestion
    app = Dash(__name__, eager_loading=is_eager, assets_folder="../../assets")
```

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    assert dash_dcc.get_logs() == []
+
+
+@pytest.mark.parametrize("is_eager", [True, False])
+def test_mkdw005_block_mathjax(dash_dcc, is_eager):
+    app = Dash(__name__, eager_loading=False, assets_folder="../../assets")
```
alexcjohnson, line 129:
```suggestion
    app = Dash(__name__, eager_loading=is_eager, assets_folder="../../assets")
```

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+                    $$
+                    ## Next line.
+                """,
+                mathjax=True,
+            ),
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    assert dash_dcc.get_logs() == []
```
alexcjohnson, line 149: Here it looks like the math isn't in the `<h2>` but we can still verify that it gets rendered, maybe give the markdown an ID like `"md"` then
```py
dash_dcc.wait_for_element("#md svg")
```
(just to make sure we aren't catching some other weird SVG on the page...)

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+
+    # Second click: Back to initial state except that MathJax library is still loaded
+    dash_dcc.wait_for_contains_text("#md", gravity)
+    dash_dcc.wait_for_no_elements("#md svg")
+    assert dash_dcc.driver.execute_script("return !!window.MathJax")
+
+
+@pytest.mark.parametrize("is_eager", [True, False])
+def test_mkdw007_load_mathjax(dash_dcc, is_eager):
+    app = Dash(__name__)
```
alexcjohnson, line 202:
```suggestion
    app = Dash(__name__, eager_loading=is_eager)
```
And again this will change the initial state of `!!window.MathJax`

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    assert dash_dcc.get_logs() == []
+
+
+@pytest.mark.parametrize("is_eager", [True, False])
+def test_mkdw006_toggle_mathjax(dash_dcc, is_eager):
+    app = Dash(__name__)
```
alexcjohnson, line 154:
```suggestion
    app = Dash(__name__, eager_loading=is_eager)
```

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+                mathjax=True,
+            ),
+            dcc.Graph(mathjax=True, id="graph-with-math", figure=fig),
+            dcc.Markdown("### No MathJax: Apple: $2, Orange: $3"),
+            dcc.Graph(id="graph-without-math", figure=fig),
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    dash_dcc.percy_snapshot("mkdw008 - markdown and graph with/without mathjax")
```
alexcjohnson, line 366: This looks like the one case where we have both graph and markdown with math. Can we do something here to verify that only one copy of MathJax is requested? count the number of `<script>` elements pointing to MathJax?

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+
+    # One click: MathJax is loaded and rendered on the second, unformatted text is gone
+
+    dash_dcc.wait_for_element("#md2 svg")
+    assert gravity not in dash_dcc._get_element("#md2").text
+    assert dash_dcc.driver.execute_script("return !!window.MathJax")
+
+
+def test_mkdw008_mathjax_visual(dash_dcc):
+    app = Dash(__name__, eager_loading=True, assets_folder="../../assets")
```
alexcjohnson, line 243:
```suggestion
    app = Dash(__name__, assets_folder="../../assets")
```
Eager loading False is the default (and what 99% of users do) - so let's use that most of the time. Or even better parametrize, but then you have to give each percy snapshot a unique name.

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+    dash_dcc.wait_for_element("#md svg")
+    assert gravity not in dash_dcc._get_element("#md").text
+    assert dash_dcc.driver.execute_script("return !!window.MathJax")
+
+    btn.click()
+
+    # Second click: Back to initial state except that MathJax library is still loaded
+    dash_dcc.wait_for_contains_text("#md", gravity)
+    dash_dcc.wait_for_no_elements("#md svg")
+    assert dash_dcc.driver.execute_script("return !!window.MathJax")
```
alexcjohnson, line 197: Can we make a version of `mkdw006` and `mkdw007` for graphs as well?

components/dash-core-components/tests/integration/graph/test_graph_varia.py
```diff
@@ -847,3 +859,77 @@ def graph_dims():
...
+    dash_dcc.wait_for_element("button#btn")
+
+    scripts = dash_dcc.driver.find_elements(By.CSS_SELECTOR, "script")
+    assert findSyncMathJax(scripts) is None
+    assert findAsyncMathJax(scripts) is None
+
+    dash_dcc.find_element("#btn").click()
+
+    # Give time for the async dependency to be requested (if any)
+    time.sleep(2)
```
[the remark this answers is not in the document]
archmoj: Addressed in 7c12efac4f58731f55b3295ae33d8474d800f2de.

components/dash-core-components/tests/integration/graph/test_graph_varia.py
archmoj: Addressed in 6623ba2ddf93efbc8e43a89d06c02b67f0e77c2c.

components/dash-core-components/tests/integration/graph/test_graph_varia.py
archmoj: Addressed in fa482689296fd82a3d363f087bc4b2713939e2a2.

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+    app = Dash(__name__, eager_loading=False, assets_folder="../../assets")
+
+    app.layout = html.Div(
+        [
+            dcc.Markdown("# h1 tag with inline MathJax: $E=mc^2$", mathjax=True),
+        ]
+    )
+
+    dash_dcc.start_server(app)
+    assert dash_dcc.get_logs() == []
```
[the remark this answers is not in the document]
archmoj: Addressed in a58ac546e64b4b8d6fc5d9e7fd9877cfaff0af4a.

components/dash-core-components/tests/integration/markdown/test_markdown.py
archmoj: Addressed in c61b771f2e31b5f7165b0c866695a0f8ee586e57.

components/dash-core-components/tests/integration/markdown/test_markdown.py
archmoj: Good call. Fixed react bug and added tests in f86985a854443b0af3a22b45ba9281f82a0a8e72 and eca3d5251272131ea15fc89a6dc9633c1711bbab.

components/dash-core-components/tests/integration/markdown/test_markdown.py
```diff
@@ -90,6 +91,278 @@ def test_mkdw002_dcclink(dash_dcc):
...
+
+    dash_dcc.wait_for_element("#md2 svg")
+    assert gravity not in dash_dcc._get_element("#md2").text
+    assert dash_dcc.driver.execute_script("return !!window.MathJax")
+
+
+def test_mkdw008_mathjax_visual(dash_dcc):
+    app = Dash(__name__, eager_loading=True, assets_folder="../../assets")
+
+    false = False
```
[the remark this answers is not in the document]
archmoj: Simply wanted to keep the figure portable for copying to `plotly.js` newPlot.

components/dash-core-components/tests/integration/graph/test_graph_varia.py
archmoj: In respect to `find(Sync|Async)MathJax`, I added assertions in 980961f99a2211a4c72bc390d45a34863a3eb170.
But please verify the cases are actually correct?

alexcjohnson: Ah those assertions are a bit off, but also maybe not really useful: In the `eager_loading=True` case I see:
```
<script src="/_dash-component-suites/dash/dcc/async-mathjax.v2_2_0m1646920173.js"></script>
```
It's still using `async-` because unlike `plotly.js` we use the same bundle for the sync and async versions. TBH I don't recall why we did that for plotly.js, here it should be fine to use the same. But then in the default `eager_loading=False` case I see no MathJax script. I had forgotten about this, but it's because Webpack for some reason recently started deleting its async script tags after they finish loading. https://github.com/plotly/dash/pull/1685#issuecomment-877199466

So actually let's forget about the script tags entirely, and just make assertions about `!!window.MathJax`. That's what we ultimately care about anyway.

components/dash-core-components/src/utils/LazyLoader/mathjax.js
```diff
@@ -0,0 +1,9 @@
+export default (mathjax) => Promise.resolve(window.MathJax || (
+    mathjax === false ?
+        undefined :
+        import(/* webpackChunkName: "mathjax" */ 'mathjax/es5/tex-svg.js').then(() => {
+            window.MathJax.config.startup.typeset = false;
```
alexcjohnson, line 5: Ah shoot, this `typeset = false` isn't included in the chunk, so in eager mode we may end up going ahead with the initial typesetting. Most of the time this shouldn't be a problem because the page won't have any content on it at that point, but in dash-embedded it could, and that's now perhaps the major reason people would use `eager_loading=True`.

I think what we may need to do is make a separate file that's just:
```js
import 'mathjax/es5/tex-svg.js';

window.MathJax.config.startup.typeset = false;
```
And then in this file we `import()` THAT file as the `mathjax` chunk.

components/dash-core-components/tests/integration/markdown/test_markdown.py
alexcjohnson: Ignore this comment, per the weird webpack behavior https://github.com/plotly/dash/pull/1685#issuecomment-877199466 as I [mentioned above](https://github.com/plotly/dash/pull/1949#discussion_r824278502)

components/dash-core-components/tests/integration/graph/test_graph_varia.py
alexcjohnson: https://github.com/plotly/dash/pull/1949/commits/70af21e8cf96bab4627246ea1327d7534baa09e1

archmoj:
@alexcjohnson How the percy image test for MathJax look like?

alexcjohnson:
[Looks good!](https://percy.io/plotly/dash/builds/16458149/changed/926379986?browser=firefox&browser_ids=19&subcategories=approved&viewLayout=overlay&viewMode=new&width=1280&widths=375%2C1280) Some parts of the graphs are grossly misaligned, but that's unfortunately normal for percy, we're missing some CSS from plotly.js
Screen Shot 2022-03-11 at 19 17 57

alexcjohnson (approved):
💃 Fantastic work @archmoj!

Issue #1818, opened 2021-10-25
[Feature Request] Better Callback Context


AnnMarieW:
As suggested by @chriddyp in [this forum post](https://community.plotly.com/t/duplicate-callback-outputs-solution-api-discussion/55909/8?u=annmariew): creating a new `triggered_dict` would remove some boilerplate and make it possible to have more readable and intuitive code when using the `callback_context` 

>provide a more ergonomic:
```python 
triggered_id = callback_context.triggered[0]['prop_id']
```
>For example, we could make a new triggered_dict that would have the same content but be a dictionary instead of a list. The logic would become:

```python
def update_graph(b1, b2):
    if 'reset.n_clicks' in dash.callback_context.triggered_map:
        return reset_graph()
    elif 'draw.n_clicks'  in dash.callback_context.triggered_map:
        return draw_graph()
```
------

It would also be great to address PMC as well.  See this [this forum post](https://community.plotly.com/t/parsing-callback-context/54258):

>Is there any good practice for parsing `callback_context` ? I recently started working more with pattern-matching callbacks and extracting certain values from the `callback_context` became essential. Until today, I have been doing things such as

`idx = callback_context.states['packaging_concepts-div-waste_items.children'][-2]['props']['id']['index']`

:grimacing: :grimacing:

>There must be a better way! Right?!

alexcjohnson:
Closed by @AnnMarieW in #1952 🎉

Issue #1810, opened 2021-10-20
Support Dash prop callback connection closer to function signature


anders-kiaer:
**Is your feature request related to a problem? Please describe.**
When creating complex Dash dashbords/callbacks, there can be a long list of input/state/output. We in some cases have seen things of the size of:
```python
@app.callback(
    Output("someid", "someprop"),
    Output("someid", "someprop"),
    Output("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    Input("someid", "someprop"),
    State("someid", "someprop"),
    State("someid", "someprop"),
    State("someid", "someprop"),
    State("someid", "someprop"),
    State("someid", "someprop"),
    State("someid", "someprop")
)
def _some_call_back(
    arg1,
    arg2,
    arg3,    
    arg4,    
    arg5,    
    arg6,    
    arg7,    
    arg8,    
    arg9,    
    arg10,    
    arg11,
    arg12,    
    arg13,    
    arg14,    
    arg15,    
    arg16,    
):
    ...
```
😱 Both during development, but also code review/maintenance, there is mental overhead with respect to seeing which `(id, prop)` pair belongs to which named argument in the function. This has gotten a bit better with https://dash.plotly.com/flexible-callback-signatures, but the underlying problem with `(id, prop)` and named argument in the function being defined in two  different lines remains.

**Describe the solution you'd like**
Utilize the Python 3+ typing syntax to annotate the `(id, prop)` closer to the argument. By using that we at the same time can avoid explicitly stating `Output` as that is already explicit by being defined as the returned type(s) of the function. The only thing to differentiate between is `Input`/`State` which could be done by some extra annotation tag. E.g. something like
```python
from typing import Annotated as A

@app.callback
def _some_call_back(
    arg1: A[List[str], "someid", "someprop"],
    arg2: A[List[str], "someid", "someprop"],
    arg3: A[List[str], "someid", "someprop"],
    arg4: A[List[str], "someid", "someprop"],
    arg5: A[List[str], "someid", "someprop", "state"],
) -> Tuple[
        A[List[str], "someid", "someprop"],
        A[List[str], "someid", "someprop"],
        A[List[str], "someid", "someprop"],
    ]:
    ...
```
[...]

Issue #1054, opened 2019-12-19
[Feature Request] simplified syntax when several Inputs can trigger a callback 


emmanuelle:
With Dash, it is not possible to have multiple callbacks modifying a given `Output` (see #153). The reason is that if this were possible, one could imagine having several callback sharing as Input the same property and modifying the same `Output`, and one could not decide which one to fire.

Therefore, the solution is to group all the inputs in a big callback, and since a different behaviour is often expected depending on which `Input` fired the callback, one ends up with a quite verbose syntax such as 
```
ctx = dash.callback_context
if not ctx.triggered:
    button_id = 'No clicks yet'
else:
    button_id = ctx.triggered[0]['prop_id'].split('.')[0]

if button_id == 'button1':
...
elif ....
``` 

@alexcjohnson suggested that we could introduce a `Trigger`  object which would be a variant of `Input`. It would be called as  `Trigger("button1", "n_clicks")` and would feed the callback True (if that was the trigger) or False. Therefore we could write callbacks like
```
@app.callback(...
)
def update(arg1_changed, arg2_changed)
    if arg1_changed:
          return update_output_arg1(...)
    elif arg2_changed:
          return update_output1_arg2(...)
```

As a side note it would be great to update our documentation to encourage more breaking callbacks into smaller functions like above, so that a multiple-input callback is more a router. 

A variant of the `Trigger` would be to have a tuple `(triggered_bool, prop_value)` so that one would write `arg1_changed[0]` for the bool and `arg1_changed[1]` for the value. (Therefore we would not need to use `callback_context` at all).

emilhe:
I was thinking a about making an attempt to implement a `Trigger` component. However, in terms of syntax, i was thinking about simply removing the argument from the callback, i.e. the syntax would be,

```
@app.callback(Output("output_id", "output_prop"), Trigger("button", "n_clicks"))
  def func():  # note that "n_clicks" is not included as an argument 
```

What do you think about this kind of syntax? I made a demo implementation `dash-extensions`, so you can try it out if you like :) 

https://pypi.org/project/dash-extensions/

alexcjohnson:
> removing the argument from the callback

That works if there's only one input - though in that case the existing `Input` would work just as well, wouldn't it? But it's very common to have multiple possible triggers, as in the pattern @emmanuelle showed at the beginning of this issue, and to support that case I think it's important to include the boolean as an argument to the function.

emilhe:
[...]

PR #1952, opened 2022-02-28
Improved `callback_context`

AnnMarieW:
closes # https://github.com/plotly/dash/issues/1818
https://github.com/plotly/dash/issues/1054

See also: [Community Forum Discussion](https://community.plotly.com/t/dash-2-x-feature-preview-simplified-callbacks/61738/7)

#### This PR adds features to `dash.callback_context`  to make it easier to use:
-  simplifies the syntax for determining which input triggered a callback. 
- adds `ctx` as an alias for `dash.callback_context`
- adds ability access the ctx dictionary with the dot notation
- simplifies the callback function signature when there are many inputs. 
- makes the dict id easier to access in callbacks using pattern matching. 

#### Determining which input triggered a callback
 The new features will replace boilerplate  such as:

```python
button_id = dash.callback_context.triggered[0]['prop_id'].split('.')[0]
```

Currently, when you have multiple Inputs in a callback, you  might do something like:

```python
@callback(
    Output("graph", "figure"), Input("btn-1", "n_clicks), Input("btn-2", "n_clicks)
)
def display(btn1, btn2):
    ctx = dash.callback_context
    button_id = ctx.triggered[0]['prop_id'].split('.')[0]

    if button_id = "btn-1":
         return update_graph1()
    if button_id = "btn-2":
         return update_graph2()
```

Here one new way introduced by this PR

```python
@callback(
    Output("graph", "figure"), Input("btn-1", "n_clicks), Input("btn-2", "n_clicks)
)
def display(btn1, btn2):
   if  "btn-1.n_clicks" in ctx.triggered_ids:
       return update_graph1()
   if  "btn-2.n_clicks" in ctx.triggered_ids:
      return update_graph2()
```

#### :point_right:  New:  `dash.ctx`  

ctx is a commonly used alias  for  `dash.callback_context`.  The shorter name makes it convenient to add with the import statements, for example:
```
from dash import Dash, html, Input, Output, ctx
```

#### :point_right: New:    `ctx.triggered_ids` 
`triggered_ids` is a `CallbackContext` method that returns a dict of ids that triggered the callback.    When using pattern matching callbacks, it also has the dictionary form of the ids rather than just the stringified version that's  in `callback_context.triggered`.

- keys  (str) : the triggered "prop_id",  
 - values (str or dict):  the id of the component that triggered the callback.          

For example, with regular callback, the ctx.triggered_ids dict would look something like:
```python
{"btn-1.n_clicks": "btn-1"}
```

With  pattern matching ids:
```python
{'{"index":0,"type":"filter-dropdown"}.value': {"index":0,"type":"filter-dropdown"}}
```

----------------------
---------------------

## Update 

Here is a summary of the discussion with  @alexcjohnson and @chriddyp on the items shown above:

`ctx.triggered_ids` is an improvement over the current syntax, but will still be hard to use with long or complicated ids, such as pattern matching, or auto-generated ids.

A better way would be something like:

```python
if variable_name.triggered:
     do_something()
```

We discussed having a type of input called `Triggered`, but that adds complexity to the callback and is not beginner friendly.

I discovered that by using [flexible callback signatures](https://dash.plotly.com/flexible-callback-signatures),  it's not necessary to make any changes to the callback dependencies.

### :tada: Version II:

```python
@app.callback(
    inputs=dict(btn1=Input("btn-1", "n_clicks"), btn2=Input("btn-2", "n_clicks")),
)
def display(btn1, btn2):
    c = ctx.args_grouping
    if c.btn1.triggered:
        return f"Button 1 clicked {btn1} times"
    elif c.btn2.triggered:
        return f"Button 2 clicked {btn2} times"
    else:
        return "No clicks yet"
```

#### :point_right: New Improved:  `ctx.args_grouping`

`args_grouping` is a `CallbackContext` method that returns a dict of the inputs when using flexible callback signatures.

- "id": (string or dict) the component id.  If it's a pattern matching id, it will be a dict.
- "id_str": (str) for pattern matching ids, it's the strigified dict id with no white spaces.
- "property": (str) The component property
- "value": the value of the component property at the time the callback was fired
- "triggered":  (bool)Whether this input triggered the callback

So in the above example, if you click the Button 1, then `ctx.args_grouping` looks like:

```
{
    'btn1': {'id': 'btn-1', 'property': 'n_clicks', 'triggered': True, 'value': 1},
    'btn2': {'id': 'btn-2',   'property': 'n_clicks', 'triggered': False,  'value': None}
}
```

#### :point_right: New: Access the `ctx.args_grouping` dict with a dot notation:

```python
c = ctx.args_grouping

# you can access values in the dict like this:
c["btn1"]["value"]

# Or you can use the dot notation:
c.btn1.value

# See if it triggered the callback like this:
if c.btn1.triggerd:
   do_something()
```

----------------

#### Bonus 1:  Simplified callback function signature
See issue #1810 [ Feature Request: Support Dash callback closer to function signature](https://github.com/plotly/dash/issues/1810)

Here's the other great thing about the new `ctx.args_grouping`:  You don't need to repeat all the variable names in the callback function.  This is super convenient  when you have a large callback with many inputs.

##### :cake: In the example below, note that  `all_inputs` is the only arg in the callback function:

```python

@app.callback(
    inputs={
        "all_inputs": {
            "btn1": Input("btn-1", "n_clicks"),
            "btn2": Input("btn-2", "n_clicks"),
            "dropdown": Input("dropdown", "value"),
            "store": Store("store", "data"),
        }
    },
)
def display(all_inputs):
    c = ctx.args_grouping.all_inputs  

    if c.btn1.triggered:
        return f"Button 1 clicked {c.btn1.value} times"
    elif c.btn2.triggered:
        return f"Button 2 clicked {c.btn2.value} times"
   #  ... etc
    else:
        "no update"


```

#### Bonus 2:  Easier to get the dict id when using pattern matching callbacks.

It's often necessary to get the component's dictionary id in a callback.  However, the current `dash.callback_context.triggered`. only has the  `"prop_id"` which is a stringified dictionary with no white space. It's necessary to parse the string and use json loads to get the dictionary version.

Here is an example of the current `dash.callback_context.triggered`.
```
[
  {
    'prop_id': '{"index":0,"type":"filter-dropdown"}.value',
    'value': 'NYC'
  }
]
```

__Here is an example app using the new `ctx.triggered_ids` to get the index of the button that triggered the callback__
```python
import numpy as np
import plotly.express as px
from dash import Dash, Input, Output, ctx, dcc, html, ALL

N=5

app = Dash(__name__)


def make_figure(n):
    x = np.linspace(0, 8, 81)
    y = x ** int(n)
    return px.scatter(x=x, y=y)


app.layout = html.Div(
    [
        html.Div(
            [html.Button(f"x^{i}", id={"index":i}) for i in range(N)]
        ),
        dcc.Graph(id="graph"),
    ]
)

@app.callback(
    Output("graph", "figure"), Input({"index":ALL}, "n_clicks")
)
def update_graph(btns):
    if ctx.triggered:
        # The ctx.triggered_ids is a dict with the ids of the input that triggered the callback
        # This gets the key, which is the stringified dict id: "  
        prop_id=ctx.triggered_ids.first()
        
        # This gets the value which is the component's dictionary `id`:
        dict_id= ctx.triggered_ids[prop_id]
        
        # This gets the index number of the button the triggered the callback.  Note you can 
        # use the dot notation or dict_id["index"]
        n= dict_id.index
        
        return make_figure(n)
    return {}


if __name__ == "__main__":
    app.run_server(debug=True)

```

---
__Here is the same app, but instead it uses the new `ctx.args_grouping` to get the index of the button that triggered the callback:__

```python
import numpy as np
import plotly.express as px
from dash import Dash, Input, Output, ctx, dcc, html, ALL

N=5

app = Dash(__name__)


def make_figure(n):
    x = np.linspace(0, 8, 81)
    y = x ** int(n)
    return px.scatter(x=x, y=y)


app.layout = html.Div(
    [
        html.Div(
            [html.Button(f"x^{i}", id={"index":i}) for i in range(N)]
        ),
        dcc.Graph(id="graph"),
    ]
)

@app.callback(
    Output("graph", "figure"), dict(btns=Input({"index":ALL}, "n_clicks"))
)
def update_graph(btns):
    # Since we are using {"index":ALL}, then
    # ctx.args_grouping.btns is a list of dicts - one dict for each button.
    for btn in ctx.args_grouping.btns:
        if btn.triggered:
            n = btn.id.index
            return make_figure(n)
    return {}


if __name__ == "__main__":
    app.run_server(debug=True)

```

### :tada: Version III:

Based on community feedback, 
- renamed `ctx.triggered_ids` to `ctx.triggered_prop_ids`
- added `ctx.triggered_id`, which is the component id of the the Input that triggered the callback.

So now you can use either:
```
if "btn-1" == ctx.triggered_id
   do_something()
```
or
```
if "table-1.selected_rows" in ctx.triggered_prop_ids
    do_something()
if "table-1.selected_columns" in ctx.triggered_prop_ids
    do_something_else()
```

And now, the pattern matching example above can be simplified to:
```
btn = ctx.triggered_id
n = btn.id.index
return make_figure(n)
```


dash/__init__.py
```diff
@@ -19,7 +19,7 @@
 from . import html  # noqa: F401,E402
 from . import dash_table  # noqa: F401,E402
 from .version import __version__  # noqa: F401,E402
-from ._callback_context import callback_context  # noqa: F401,E402
+from ._callback_context import callback_context, ctx  # noqa: F401,E402
```
alexcjohnson, line 22: Does this need to be a separate instance? Can we just say `ctx = callback_context` after all the imports, and remove the [extra instantiation](https://github.com/plotly/dash/pull/1952/files#diff-f19c52717a102d5ef8b91b157d234b9075d07bbcc2cd1d0a0cd8060bc9c7f608R290)?

alexcjohnson:
As just discussed on slack, this looks great! Aside from my comment above, this just needs:
- [ ] a changelog entry
- [ ] resolving conflicts
- [ ] flattening the docstring indentation [here](https://github.com/plotly/dash/pull/1952/files#diff-f19c52717a102d5ef8b91b157d234b9075d07bbcc2cd1d0a0cd8060bc9c7f608R54-R57) and [here](https://github.com/plotly/dash/pull/1952/files#diff-f19c52717a102d5ef8b91b157d234b9075d07bbcc2cd1d0a0cd8060bc9c7f608R75-R78)

AnnMarieW:
@alexcjohnson I completed the following:  

> As just discussed on slack, this looks great! Aside from my comment above, this just needs:
> 
> * [x]  a changelog entry
> * [x]  resolving conflicts
> * [x]  flattening the docstring indentation [here](https://github.com/plotly/dash/pull/1952/files#diff-f19c52717a102d5ef8b91b157d234b9075d07bbcc2cd1d0a0cd8060bc9c7f608R54-R57) and [here](https://github.com/plotly/dash/pull/1952/files#diff-f19c52717a102d5ef8b91b157d234b9075d07bbcc2cd1d0a0cd8060bc9c7f608R75-R78)

Also, I removed the extra `callback_context` instance per your comment.
One test  in `test-39` fails, but it's unrelated to any changes in this PR --  I'm not sure what's wrong.

alexcjohnson (approved):
Fantastic! 💃

MattesR:
IDK whether this is the right spot. I'd just want to mention that there's a typo in this example:
example app using the new ctx.triggered_ids to get the index of the button that triggered the callback:

it says `prop_id=ctx.triggered_ids.first()` but it should be  `prop_id=ctx.triggered_id.first()`. Found out when trying to update my code with the new ctx.

`ctx` has the attributes `'triggered_id'`, and  `'triggered_prop_ids'` according to `dir(ctx)`

AnnMarieW:
@MattesR Thanks for your comment.  I think the examples in the initial write-up for this PR were confusing because the names changed during review.  For example  `triggered_ids` was renamed to `triggered_prop_ids`.   Then a new attribute was added: `ctx.triggered_id`, which is the component id of the the Input that triggered the callback.

I just updated the comment to remove all the examples here and point people to the ones in the announcement on the community forum.

PR #1953, opened 2022-02-28
fix unrelenting alert when hot reload fails

### LINKED ISSUE #1783: Less aggressive Reloader alert
Currently if an app fails, the reloader alert message is pretty aggressive. If you close it, it will re-appear almost immediately.

We should only show the `alert` once per run.

alexcjohnson:
Fixes #1783 

This really only gets triggered in workspaces - must be because the app is behind a proxy there. When you're pointed directly at the server a failure manifests as a refused connection, which the devtools detect as "the server stopped." But in workspaces the server appears present, just responding with a 500 error. To test this locally I temporarily threw an exception [here](https://github.com/plotly/dash/blob/97470a0a8aabd93c480f5dff64009baa88906cc6/dash/dash.py#L604).

Not intending to add a test for this, it would be awfully cumbersome and unclear that it would actually be representative of the real condition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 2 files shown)


chriddyp (approved):
💃🏼

PR #1958, opened 2022-03-04
fixed dropdown bug #1908

### LINKED ISSUE #1908: [BUG] Dropdown:  Selected options not showing when the `value` contains a comma
As [reported on the forum by @ marcus](https://community.plotly.com/t/dcc-dropdown-bug-suspected-please-confirm-or-correct-me/60585):

Problem is in this spot: 'value': "50 , 10"
Specifically comma in `value` string makes that selection is not being displayed. When coma is removed or replaced (dot creates no issue) it works fine, but when coma is within the string, the selections are not displayed in the input box and are not removed from the `options`

 This only occurs in Multi-Value dropdowns.  This is new in Dash V2.1.0  It works as expected in V2.0.0

```

app = Dash(__name__)

app.layout = html.Div(
    [
        dcc.Dropdown(
            id="dropdown",
            className="inputbox-long",
            options=[
                {"label": "New York City", "value": "50, 10"},
                {"label": "Montreal", "value": "MTL" },
                {"label": "San Francisco", "value": "SF"},
            ],
            placeholder="Select one or more",
            multi=True,
        ),
        html.Div(id="output"),
    ]
)


@app.callback(
    Output("output", "children"), Input("dropdown", "value"),
)
def update(value):    
    return value


if __name__ == "__main__":
    app.run_server(debug=True)
```
dropdown_bug

AnnMarieW:
Fixes #1908 Dropdown: Selected options not showing when the value contains a comma

As [reported on the forum](https://community.plotly.com/t/dcc-dropdown-bug-suspected-please-confirm-or-correct-me/60585) by Marcus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Fixed
 - [#1953](https://github.com/plotly/dash/pull/1953) Fix bug [#1783](https://github.com/plotly/dash/issues/1783) in which a failed hot reloader blocks the UI with alerts.
 
+
+## Dash Core Components
+
+### Fixed
+- [#1958](https://github.com/plotly/dash/pull/1958) Fix dropdown bug [#1908](https://github.com/plotly/dash/issues/1908) where the value didn't display in the input box if it contained a comma.
+
+
 ## [2.2.0] - 2022-02-18
 
 ### Added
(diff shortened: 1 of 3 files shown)


alexcjohnson:
@AnnMarieW the test you deleted makes sense to delete, it doesn't look like behavior we want. But can you add a test that shows the behavior that this PR fixes?

AnnMarieW:
Closing this because it's now included in #1970

PR #1965, opened 2022-03-11
Component as props

T4rk1n:
Add component as prop support.

## Contributor Checklist

### optionals


components/dash-core-components/src/components/Checklist.react.js
```diff
@@ -108,6 +108,7 @@ Checklist.propTypes = {
                     PropTypes.string,
                     PropTypes.number,
                     PropTypes.bool,
+                    PropTypes.node,
```
alexcjohnson, line 111: Should these switch to just `label: PropTypes.node`? Strings and numbers are valid as nodes - as are booleans, though they aren't rendered. And this is already reflected in the docstrings: `node` currently creates the docstring "a list of or a singular dash component, string or number" (ignoring bool, which I think is appropriate), so as is this is shown as:
```
- label (string | number | boolean | a list of or a singular dash component, string or number; required):
    The option's label.
```

It's good to know that `oneOfType` works with `node` though!

T4rk1n: > a list of or a singular dash component, string or number

In this case, component in a list of dictionary, it cannot be a list of components but a single component. Maybe we can put PropTypes.element instead ?

@plotly/dash-test-components/src/components/ComponentAsProp.js
```diff
@@ -0,0 +1,37 @@
...
+            {list_of_shapes && <ul>{list_of_shapes.map(e => <li key={e.value}>{e.label}</li>)}</ul> }
+        </div>
+    )
+}
+
+ComponentAsProp.propTypes = {
+    id: PropTypes.string,
+    element: PropTypes.node,
+
+    elements: PropTypes.arrayOf(PropTypes.node),
```
alexcjohnson, line 22: `children` props - which just have type `PropTypes.node` - allow a list of components, and this is reflected in the docstring we use for node, "a list of or a singular dash component, string or number"

And as your example `{elements || element}` shows, in practical usage a node or an array of nodes generally (always?) work the same.

So if there were a way we could make it work the same way for other node props that would be the best. Otherwise we'll need to carve out a special case for `children` in the generated docs, vs other node props.

dash/development/base_component.py
```diff
@@ -132,13 +134,26 @@ def __init__(self, **kwargs):
...
+                            self._children_props.append(k)
+                            break
+                        if isinstance(item, dict):
+                            found = False
+                            for key, value in item.items():
+                                if isinstance(value, Component):
+                                    self._children_props.append("[]" + k + "." + key)
+                                    found = True
+                            if found:
+                                break
```
alexcjohnson, line 156: I don't know if calculating `_children_props` at component instantiation is the right way - I was envisioning we'd do this either while building the components, flag all the props and prop parts that are allowed to be nodes... or even somehow have the renderer extract this info from the component `propTypes` at run time though the `PropTypes` package really doesn't seem to want you to do that...

Anyway the two problems I see with this:
- if you instantiate a component with only strings in these props, then a callback later provides components directly to the props, those props won't be in `_children_props`
- this requires iterating through potentially enormous props, like table data, even if we have no intention of allowing components there.
- and one due to your specific implementation (that we could fix, but as I said I think a bigger change is needed): if a list of dicts allows components in two different keys, but the first item in the list to have a component only has it on one of those keys, the other key won't be in `_children_props`

I'd love if we could handle this entirely in the renderer, so the only back-end update would be removing the . I suppose a real simple way would be to require manually enumerating component props in the js source code - like `Checklist.childrenProps = ['[]options.label']`. Not very 🌴 , but we could probably write something into the component generator to at least throw an error if this isn't done correctly?

dash/development/base_component.py
```diff
@@ -132,13 +134,26 @@ def __init__(self, **kwargs):
                     f"\nAllowed arguments: {allowed_args}"
                 )
 
-            if k != "children" and isinstance(v, Component):
-                raise TypeError(
-                    error_string_prefix
-                    + " detected a Component for a prop other than `children`\n"
-                    + "Did you forget to wrap multiple `children` in an array?\n"
-                    + f"Prop {k} has value {v!r}\n"
```
alexcjohnson, line 140: hmph and right after @chriddyp just upgraded this error message #2027 - I suppose an advantage of building the `_children_props` list into the Python component generator is we could keep this error, just qualify it further to
```
if k != "children" and k not in self._children_props and isinstance(v, Component):
```

dash/dash-renderer/src/actions/utils.js
```diff
@@ -41,6 +54,30 @@ export const crawlLayout = (object, func, currentPath = []) => {
             const newPath = concat(currentPath, propsChildren);
             crawlLayout(children, func, newPath);
         }
+        const childrenProps = pathOr([], ['childrenProps'], object);
```
alexcjohnson, line 57: [`propOr`](https://ramdajs.com/docs/#propOr) would suffice :)

dash/development/base_component.py
T4rk1n: I added the node paths from the metadata into a class attribute, the renderer now wrap the props all the time, works fine but there is a little extra overhead for the union types. I was thinking maybe add a symbol in the key like `?` so it only performs a check in those cases?

T4rk1n: It also adds the children_props to all the components json, the additional data could be pretty significant for repetitive info. I think we should supply that info somewhere else.

alexcjohnson: Right, the ideal would still be to provide these once per component class - which is why I suggested putting it directly into the JS source, but I still don't see a way to do that without manual duplication. But what about appending it to the entry point bundle when it's served, in [`serve_component_suites`](https://github.com/plotly/dash/blob/07be9e2d0c060b929540aee88457bba717123a66/dash/dash.py#L810)? Could be as simple as constructing a string like:
```
{namespace}.{component_name}.childrenProps = {repr(children_props)};
{namespace}.{component_name2}.childrenProps = {repr(children_props2)};
...
```
and appending it to the served file - assuming we can figure out which one is the entry point bundle. A little hacky perhaps... but maybe workable?

T4rk1n: I opted to use the component meta & registry to gather the children props, they are then served in the configs json script tag on the index. The TreeContainer has a prop `_dashprivate_config` so it's easy to get from there and extracted the configs from the store to use in `crawlLayout`. I think it works good now.

@plotly/dash-test-components/src/components/ComponentAsProp.js
T4rk1n: Changed, now nodes have the same behavior in all cases.

alexcjohnson: Can we then get rid of `elements` as a prop here, and instead in the tests let `element` and some of the nested props be lists of components?

components/dash-core-components/src/components/Checklist.react.js
alexcjohnson: This is still open: rather than adding to the `oneOfType` I think this should be just `label: PropTypes.node`. Note that today `boolean` doesn't work anyway (both `True` and `False` show no label at all) so `node` that already allows `string` and `number` (and based on your other work lists of all of these are now allowed?) should be exactly what we want, right?

alexcjohnson: And same for dropdown and radioitems

dash/development/base_component.py
```diff
@@ -392,6 +402,14 @@ def __repr__(self):
             props_string = repr(getattr(self, "children", None))
         return f"{self._type}({props_string})"
 
+    @classmethod
+    def _get_base_nodes(cls):
+        if not cls._base_nodes:
+            cls._base_nodes = [
+                n.replace("[]", "").split(".")[0] for n in cls._children_props
```
alexcjohnson, line 409: Could this result in duplicate entries in this list, like if `a.b` and `a.c` both support components? Does that matter?

tests/unit/development/test_collect_nodes.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_dcn001_collect_nodes():
+    nodes = collect_nodes(metadata)
+
+    assert nodes == [
+        "shape.node",
+        "list_of_nodes",
+        "[]list_of_union.b",
+        "[]list_of_union",
+        "[]list_of_shapes.label",
```
alexcjohnson, line 59: Putting the `[]` at the front may be limiting eventually, it obscures the difference between `a.b[0].c` and `a[0].b.c`

```py
>>> metadata = {"a": {"type": {"name": "shape", "value": {"b": {"type": {"name": "arrayOf", "value": {"name": "shape", "value": {"c": {"name": "node"}}}}}}}}}
>>> collect_nodes(metadata)[0]
'[]a.b.c'
>>> md2 = {"a": {"type": {"name": "arrayOf", "value": {"name": "shape", "value": {"b": {"type": {"name": "shape", "value": {"c": {"name": "node"}}}}}}}}}
>>> collect_nodes(md2)[0]
'[]a.b.c'
```
whereas if the brackets go after the attribute, these would look like `a.b[].c` and `a[].b.c`. In principle you could also imagine `a[].b[].c` vs `a[][].b.c` and so on.

Maybe not an issue in the short term, but I feel like with just that change of notation we could reach complete generality of anywhere a component might allow nodes.

dash/development/base_component.py
T4rk1n: It's only used for a check, better change it to a set.

alexcjohnson: actually this is used just in the [type check](https://github.com/plotly/dash/pull/1965/files#diff-ac4d356fec919ee468c3b73bbd242cddf1deb95b985a6b860e44a79176117211R143), right? That isn't digging into nested prop vals, so I'd think we only want the props that support components as their top level - ie children props that don't have `[]` or `.` in them - otherwise you could provide a component at the top level for that prop, bypass this error and only get the error on the front end.

(and a small optimization: might as well add `'children'` to the list here rather than in the type check, and in fact only include it if the component has `'children'` in its `_prop_names` - or `available_properties`, IDK why we have both of these, maybe we should drop `_prop_names` and just use `available_properties`, with a fallback in the constructor to move `_prop_names` over to `available_properties` if some old component only has `_prop_names`?)

T4rk1n: Good point, should only check the flat props on the backend.

alexcjohnson: > `available_properties` is generated in the init, I think they should be class members like the children props as it's static data.

Good point. But if we change that we'll need a test to ensure old generated components still work as expected. I don't imagine there would be a problem but the old components will be in use for a long time.

tests/unit/development/test_collect_nodes.py
T4rk1n: Changed the notation of the brackets to the back.

dash/development/base_component.py
```diff
@@ -392,6 +402,14 @@ def __repr__(self):
             props_string = repr(getattr(self, "children", None))
         return f"{self._type}({props_string})"
 
+    @classmethod
+    def _get_base_nodes(cls):
+        if not cls._base_nodes:
```
alexcjohnson, line 407: if a component has lots of props but none of them accept components, will we run this iteration every time one of them is instantiated? If so, can we avoid that? Moving `+ ["children"]` in here would be one way perhaps?

alexcjohnson: oh sorry, it's only iterating over `_children_props` - so the problem only comes if there are lots of nested props and no base props. Still would be best to avoid.

T4rk1n: Let's move that to the generator.

T4rk1n: Added a legacy test from the old standard component output.

PR #1969, opened 2022-03-14
Speedup parallel tests of table server

archmoj:
Before: 17:30 minutes 
After: 10:30 minutes
@alexcjohnson

### DIFF
--- a/components/dash-table/tests/selenium/test_sizing_x.py
+++ b/components/dash-table/tests/selenium/test_sizing_x.py
@@ -9,9 +9,5 @@ from utils import (
 
 
 @pytest.mark.parametrize("props", basic_modes)
-@pytest.mark.parametrize(
-    "data_fn",
-    [generate_mock_data],
-)
-def test_szng004_on_focus(test, props, data_fn):
-    on_focus(test, props, data_fn)
+def test_szng004_on_focus(test, props):
+    on_focus(test, props, generate_mock_data)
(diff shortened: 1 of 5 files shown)


components/dash-table/tests/selenium/test_sizing_x.py
```diff
@@ -0,0 +1,17 @@
...
+    generate_mock_data,
+)
+
+
+@pytest.mark.parametrize("props", basic_modes)
+@pytest.mark.parametrize(
+    "data_fn",
+    [generate_mock_data],
+)
+def test_szng004_on_focus(test, props, data_fn):
```
T4rk1n, line 16: The `data_fn` parameter can be removed as there is now only one.

archmoj: I am not super familiar with these `pytest`. 
Also I'd prefer to keep the changes in each PR as minimal as possible. 
So if that's non-blocking, I'd suggest we merge and then possibly address extra changes in a separate PR?

alexcjohnson: @archmoj I’d say this is a decent opportunity to learn a bit about `pytest`. @T4rk1n is right, the obvious thing to do here is just remove `data_fn` and it’s associated `@pytest.mark.parametrize` decorator, and move `generate_mock_data` into its place in the function body. This is sufficiently related to what you’re doing in this PR that it should be done here, not in a separate PR.

alexcjohnson (approved):
Looks great, thanks! 💃 This will be a big improvement to the dev experience!

PR #1970, opened 2022-03-14
Fix Dropdown multi option removed update value.

### LINKED ISSUE #1868: [BUG] dcc.Dropdown `value` does not update when selected option removed from `options` -- but UI does update
**Describe your context**

```
dash               2.0.0
dash-renderer      1.9.1
```

Bug is not browser-specific but these are my stats:

    - OS: Mac OS 11.2.2 (Big Sur)
    - Browser: Safari
    - Version: 14.0.3

**Describe the bug**

When the `options` of a `dcc.Dropdown` are updated to remove a currently selected option, the UI updates to remove that value, but the `value` parameter does not update.

This results in unexpected app behaviour because the `value` of the dropdown is different from what's displayed in the UI.

This behaviour happens with both single-select and multi-select dropdowns.

**Expected behavior**

Removing an option from `options` also removes that option from `value`. Value in UI and code remain in sync. 

**Screenshots**

https://user-images.githubusercontent.com/4672118/146413770-350543f7-31e9-4399-b931-0e88bbc12d3b.mov

**Minimal example**

```
import dash
from dash import dcc, html

app = dash.Dash(__name__)
server = app.server  # expose server variable for Procfile

# Set display title
app_title = "Dropdown bug"
app.title = app_title

seats = ["11A", "11B", "11C", "12A", "12B", "15E", "15F"]


def layout():
    return html.Div(
        [
            html.Div(
                [
                    html.P("Available seats: "),
                    dcc.Dropdown(
                        id="available-seats",
                        multi=True,
                        options=[{"label": i, "value": i} for i in seats],
                        value=seats,
                    ),
                ],
            ),
            html.Div(
                [
                    html.P("Chosen seats: "),
                    dcc.Dropdown(id="chosen-seats", multi=True,),
                ],
            ),
            html.Button(id="submit-btn", children="Submit seat choice"),
            html.Div(id="output"),
        ],
    )


app.layout = layout()


@app.callback(
    Output("chosen-seats", "options"), Input("available-seats", "value"),
)
def update_options(available_seats):
    if available_seats is None:
        return []
    else:
        return [{"label": i, "value": i} for i in available_seats]

@app.callback(
    Output("output", "children"),
    Input("submit-btn", "n_clicks"),
    State("chosen-seats", "options"),
    State("chosen-seats", "value"),
    prevent_initial_call=True,
)
def print_value(n_clicks, options, value):
    return [
        html.P(f"Chosen seats are: {value}"),
        html.P(f"Available seats are: {[i['value'] for i in options]}."),
    ]


if __name__ == "__main__":
    app.run_server(debug=True)
```

Adding this callback results in what I think is the expected behaviour:

```
@app.callback(
    Output("chosen-seats", "value"),
    Input("chosen-seats", "options"),
    State("chosen-seats", "value"),
)
def update_value_when_options_change(options, cur_value):
    if cur_value is None:
        return None
    else:
        option_values = {i["value"] for i in options}
        return [i for i in cur_value if i in option_values]
```

### LINKED ISSUE #1908: [BUG] Dropdown:  Selected options not showing when the `value` contains a comma

T4rk1n:
- Fix #1868
- Fix #1908 
- Add tests for option removal.
- Refactor Dropdown to functional component

## Contributor Checklist

### DIFF
--- a/components/dash-core-components/src/fragments/Dropdown.react.js
+++ b/components/dash-core-components/src/fragments/Dropdown.react.js
@@ -40,7 +40,7 @@ const Dropdown = props => {
         return [
             sanitized,
             createFilterOptions({
-                options: options,
+                options: sanitized,
                 tokenizer: TOKENIZER,
             }),
         ];
@@ -82,7 +82,7 @@ const Dropdown = props => {
     useEffect(() => {
         if (optionsCheck !== sanitizedOptions && !isNil(value)) {
             const values = sanitizedOptions.map(option => option.value);
-            if (multi) {
+            if (multi && Array.isArray(value)) {
                 const invalids = value.filter(v => !values.includes(v));
                 if (invalids.length) {
                     setProps({value: without(invalids, value)});
--- a/components/dash-core-components/src/fragments/Dropdown.react.js
+++ b/components/dash-core-components/src/fragments/Dropdown.react.js
@@ -1,4 +1,4 @@
-import {isNil, pluck, omit, without} from 'ramda';
+import {isNil, pluck, without, pick} from 'ramda';
 import React, {useState, useCallback, useEffect, useMemo} from 'react';
 import ReactDropdown from 'react-virtualized-select';
 import createFilterOptions from 'react-select-fast-filter-options';
@@ -20,6 +20,18 @@ const TOKENIZER = {
         );
     },
 };
+
+const RDProps = [
+    'multi',
+    'clearable',
+    'searchable',
+    'search_value',
+    'placeholder',
+    'disabled',
+    'optionHeight',
+    'style',
+    'className',
+];
 
 const Dropdown = props => {
     const {
@@ -107,7 +119,7 @@ const Dropdown = props => {
                 backspaceRemoves={clearable}
                 deleteRemoves={clearable}
                 inputProps={{autoComplete: 'off'}}
-                {...omit(['setProps', 'value', 'options'], props)}
+                {...pick(RDProps, props)}
             />
         </div>
     );
--- a/components/dash-core-components/src/fragments/Dropdown.react.js
+++ b/components/dash-core-components/src/fragments/Dropdown.react.js
@@ -113,7 +113,7 @@ const Dropdown = props => {
                 onInputChange={onInputChange}
                 backspaceRemoves={clearable}
                 deleteRemoves={clearable}
-                inputProps={{autoComplete: 'of'}}
+                inputProps={{autoComplete: 'off'}}
                 {...omit(['setProps', 'value', 'options'], props)}
             />
         </div>
--- a/components/dash-core-components/src/fragments/Dropdown.react.js
+++ b/components/dash-core-components/src/fragments/Dropdown.react.js
@@ -1,4 +1,4 @@
-import {isNil, pluck, omit, type, without} from 'ramda';
+import {isNil, pluck, omit, without} from 'ramda';
 import React, {useState, useCallback, useEffect, useMemo} from 'react';
 import ReactDropdown from 'react-virtualized-select';
 import createFilterOptions from 'react-select-fast-filter-options';
@@ -20,8 +20,6 @@ const TOKENIZER = {
         );
     },
 };
-
-const DELIMITER = ',';
 
 const Dropdown = props => {
     const {
@@ -45,11 +43,6 @@ const Dropdown = props => {
             }),
         ];
     }, [options]);
-
-    const selectedValue = useMemo(
-        () => (type(value) === 'Array' ? value.join(DELIMITER) : value),
-        [value]
-    );
 
     const onChange = useCallback(
         selectedOption => {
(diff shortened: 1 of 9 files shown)


AnnMarieW:
@T4rk1n I really like the refactoring you did here  --  Would you like to incorporate [#1958 ](https://github.com/plotly/dash/pull/1958)  too?

T4rk1n:
@AnnMarieW 

Yes, I thought that test and the array join was kinda weird.

components/dash-html-components/dash_html_components_base/__init__.py
```diff
@@ -33,27 +33,27 @@
 
 _js_dist = [
     {
-        "relative_package_path": 'html/{}.min.js'.format(_this_module),
+        "relative_package_path": "html/{}.min.js".format(_this_module),
```
alexcjohnson, line 36: ha, we should include this file in our linting / formatting runs. If we did, the new linter would ask for this to become an f-string ;)

components/dash-html-components/setup.py
```diff
@@ -2,22 +2,22 @@
 import json
 from setuptools import setup
 
-with open('package.json') as f:
+with open("package.json") as f:
```
alexcjohnson, line 5: I think we can delete `components/*/setup.py` now, right? Also `MANIFEST.in`, maybe some others too... same in `dash-renderer`.

Anyway I won't make you do this now, but I'm reminded of it after seeing this file being edited.

package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "license": "UNLICENSED",
   "scripts": {
-    "private::format.black": "black dash tests --exclude metadata_test.py",
+    "private::format.black": "black dash tests components --exclude metadata_test.py",
```
alexcjohnson, line 5: That's a bit over-broad, and anyway both dcc and table already have their own linters and formatters. But since html doesn't, if you want to include just html here go ahead - just please also add it to the lint commands for black, flake8, and pylint.

components/dash-core-components/src/fragments/Dropdown.react.js
```diff
@@ -21,90 +21,99 @@ const TOKENIZER = {
...
+            <ReactDropdown
+                filterOptions={filterOptions}
+                options={sanitizeOptions(options)}
+                value={value}
+                onChange={onChange}
+                onInputChange={onInputChange}
+                backspaceRemoves={clearable}
+                deleteRemoves={clearable}
+                inputProps={{autoComplete: 'off'}}
+                {...omit(['setProps', 'value', 'options'], props)}
```
alexcjohnson, line 110: Thanks for converting this component to functional - and of course getting rid of one `UNSAFE_` method 🎉  That will become very important as soon as someone tries to use Dash with React 18 (which is officially out as of a few weeks ago)

FWIW not necessary to include in this PR, but just because I see it carried over in your refactoring, I'd like us to eventually get rid of all the `ramda.omit` calls fowarding props to underlying components https://github.com/plotly/dash-core-components/issues/703

alexcjohnson (approved):
Nice job! And excellent tests 🏆 

💃 after considering my comments - I think the only blocking one is about the [`format.black` command](https://github.com/plotly/dash/pull/1970/files#r854652676), the others I'll leave to your discretion.

package.json
T4rk1n: I think I added that for convenience because it's linted somewhere in ci but running `npm run format` from root didn't format the tests in dcc.

alexcjohnson: I see - then the right way to handle this is to make new commands for each component package that mirror the renderer, ie they call out to the format commands defined in those packages:

https://github.com/plotly/dash/blob/4b03e5ac9e6406a4fd35f8b7b2fde33663959885/package.json#L6

And if these new commands fit the `private::format.*` pattern they'll be picked up by the root `npm run format` command

JohnMulligan:
Hello, sorry to comment on a closed issue, but has this fix been included in a release or is it just in the dev branch for now? I'm seeing this behavior in 2.3.1 and just wanted to check.

ned2:
You can see the fix for this in the Unreleased section of the `CHANGELOG.md`, so I think that means we should see it in the next Dash release.

    Fix bug https://github.com/plotly/dash/issues/1908 Selected options not showing when the value contains a comma.

Issue #1962, opened 2022-03-10
[BUG] DatePickerSingle is extremely slow when using large number of disabled_days


Bougeant:
**Describe your context**
Please provide us your environment, so we can easily reproduce the issue.

```
dash                      2.0.0
dash-core-components      2.0.0
dash-extensions           0.0.67
dash-html-components      2.0.0
dash-table                5.0.0
jupyter-dash              0.4.0

OS: macOS
Browser: Chrome
Version: 96
```

**Describe the bug**

`DatePickerSingle` and `DatePickerRange` are extremely slow when provided a long list of `disabled_days`. 

For example, if one wants to disabled every day between 2010 and 2022, except the first day of each month (4000+ disabled days), then the `DatePickerSingle` takes about 30 seconds to open and 30 additional seconds each time we change the month. This issue had already been reported here: https://github.com/plotly/dash-core-components/issues/986

Here's a minimal reproducible example:

```
from dash import Dash, dcc
import pandas as pd


min_date = pd.to_datetime("2010-01-01")
max_date = pd.to_datetime("2022-01-01")
disabled_days = [x for x in pd.date_range(min_date, max_date, freq="D") if x.day != 1]

app = Dash()

app.layout = dcc.DatePickerSingle(
    min_date_allowed=min_date,
    max_date_allowed=max_date,
    disabled_days=disabled_days
)

app.run_server(port=8051)
```
**Expected behavior**

The `DatePickerSingle` should appear and be interactable instantly.

T4rk1n:
Fixed by #1976

PR #1976, opened 2022-03-17
Removed unnecessary recalculation of dates

nickmelnikov82:
As described in issue #1962 DatePickerSingle and DatePickerRange are extremely slow when provided a long list of disabled_days.

## Contributor Checklist

   -  [x] fix is_outside_range function
   -  [x] make test

### optionals


components/dash-core-components/src/fragments/DatePickerSingle.react.js
```diff
@@ -10,23 +10,60 @@ import convertToMoment from '../utils/convertToMoment';
...
+        if (Object.keys(state).length) {
+            this.setState(state);
+        }
+    }
 
+    UNSAFE_componentWillReceiveProps(newProps) {
+        this.propsToState(newProps);
+    }
+
+    UNSAFE_componentWillMount() {
```
T4rk1n, line 55: Don't have to do this now as it's bigger than just this bug, but instead of using the `UNSAFE` methods it could be worthwhile to refactor the component to functional and `useMemo` for the values that needs mapping.

T4rk1n (commented):
I ran the test before rebuilding with the fix and it froze for over 15 seconds before I closed the browser. With the fix it passed almost instantly.

:woman_dancing: After merging dev and adding a changelog entry.

components/dash-core-components/src/fragments/DatePickerSingle.react.js
nickmelnikov82: So I need to rewrite the component into a functional one? Or leave it as is for now?

T4rk1n: You can leave it as is.

alexcjohnson (approved):

PR #1995, opened 2022-03-29
Fix werkzeug 2.1.0 import & dev tools error html rendering.

### LINKED ISSUE #1992: ImportError: cannot import name 'get_current_traceback' from 'werkzeug.debug.tbtools'
I am working on a Dash app and it was working great till yesterday. When I tried today, I am seeing the following:

  File "/app/app-v2.py", line 12, in <module>
    import dash
  File "/usr/local/lib/python3.9/site-packages/dash/__init__.py", line 5, in <module>
    from .dash import Dash, no_update  # noqa: F401,E402
  File "/usr/local/lib/python3.9/site-packages/dash/dash.py", line 18, in <module>
    from werkzeug.debug.tbtools import get_current_traceback
ImportError: cannot import name 'get_current_traceback' from 'werkzeug.debug.tbtools' (/usr/local/lib/python3.9/site-packages/werkzeug/debug/tbtools.py)

I am running the app on macOS Catalina. Happy to provide more details.

T4rk1n:
- Fix #1992
- Fix html rendering of debugger errors in the devtools with new version werkzeug.

### optionals

### DIFF
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -8,6 +8,8 @@ import subprocess
 import logging
 import io
 import json
+import secrets
+import string
 from functools import wraps
 
 logger = logging.getLogger()
@@ -206,3 +208,9 @@ def job(msg=""):
         return _wrapper
 
     return wrapper
+
+
+def gen_salt(chars):
+    return "".join(
+        secrets.choice(string.ascii_letters + string.digits) for _ in range(chars)
+    )
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -11,13 +11,11 @@ import time
 import mimetypes
 import hashlib
 import base64
+import traceback
 from urllib.parse import urlparse
 
 import flask
 from flask_compress import Compress
-
-from werkzeug.debug import tbtools
-from werkzeug.security import gen_salt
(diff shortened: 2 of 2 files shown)


alexcjohnson:
This looks good, and I've verified it works with both previous and new `werkzeug` versions. Given that this became necessary though, I'm a little uneasy depending on ANYTHING inside `werkzeug` submodules, because I don't see the things we care about [documented](https://werkzeug.palletsprojects.com/en/2.0.x/search/?q=tbtools) so they clearly would feel free to change these whenever they see fit.

So I wonder if we could minimize the chance of future problems... first, `gen_salt` could be inlined easily:
```py
def gen_salt(chars):
    return "".join(secrets.choice(string.ascii_letters + string.digits) for _ in range(chars))
```
Second, as nice as `tbtools` HTML output is, can we have a second fallback in case we don't find either of these (or perhaps in case we don't even find the `tbtools` module), that just uses `traceback.format_exc()` or something?

alexcjohnson (approved):
💃 Excellent. This should prevent breakage from any sort of future changes they might make.

PR #2006, opened 2022-04-07
Fix ci failures

T4rk1n:
- dash_duo ThreadRunner: Kill all new threads created during the tests.
- Temporary fixes for a  bug introduced in Werkzeug 2.1.0 update with the 204 responses that we return for the `no_update` callbacks.
  - Add `pytest-rerunfailures` to automatically reruns flaky tests.
  - Skip & xfail for the tests that are more prone to fails.


components/dash-core-components/tests/integration/calendar/test_date_picker_single.py
```diff
@@ -71,6 +73,11 @@ def test_dtps010_local_and_session_persistence(dash_dcc):
     assert dash_dcc.get_logs() == []
 
 
+@pytest.mark.xfail(
```
alexcjohnson, line 76: Is there a pattern to which tests you marked `xfail` and which you marked `skipif`?

tests/integration/callbacks/test_prevent_update.py
```diff
@@ -94,7 +95,10 @@ def show_clicks(n):
 
     dash_duo.start_server(app)
 
-    dash_duo.multiple_click("#btn", 10)
+    btn = dash_duo.wait_for_element("#btn")
+    for i in range(10):
+        btn.click()
+        time.sleep(0.2)
```
alexcjohnson, line 101: Can we add this as maybe a `delay` kwarg to [`multiple_click`](https://github.com/plotly/dash/blob/f6abdfb823cdbe445eb7e660b29f1444e636a02a/dash/testing/browser.py#L527)? Seems like this would be useful more than once.

dash/testing/browser.py
```diff
@@ -88,15 +78,8 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, traceback):
         try:
             self.driver.quit()
-            if self._percy_run and self._percy_finalize:
```
alexcjohnson, line 91: shall we remove the `percy_run` and `percy_finalize` args, and the `--nopercyfinalize` pytest option?

dash/testing/browser.py
```diff
@@ -465,6 +448,10 @@ def _get_chrome(self):
                 "safebrowsing.disable_download_protection": True,
             },
         )
+        options.add_argument("--disable-dev-shm-usage")
+        options.add_argument("--no-sandbox")
+        options.add_argument("--disable-gpu")
+        options.add_argument("--remote-debugging-port=9222")
```
alexcjohnson, line 454: Why are we adding these options?

T4rk1n: I can't run chromedriver without those arguments on ubuntu, I thought they would help.

components/dash-core-components/tests/integration/calendar/test_date_picker_single.py
T4rk1n: I skipped those that were guaranteed to fail, xfail for those that can pass, but I added the xfail after reruns failed on dcc but it wasn't actually set, so I think now they will xpass most of the time.

tests/integration/callbacks/test_prevent_update.py
T4rk1n: Good idea.

alexcjohnson:
Any idea how adding one package to dash-table caused its package-lock to get 6000 lines SMALLER?

dash/testing/browser.py
alexcjohnson: OK - seems to still work for me this way on mac, we can keep them.

alexcjohnson:
Ah now that the "deprecated SDK" banner is gone, percy has a new banner:
Screen Shot 2022-04-07 at 21 51 58

That said some of the snapshots look oddly empty - may be worth doing a bit of manual comparison of the old vs new and/or a comparison to what these look like when run locally, to see if there's something odd going on with the new snapshot framework.

dash/testing/plugin.py
```diff
@@ -41,19 +41,6 @@ def pytest_addoption(parser):
         "--headless", action="store_true", help="set this flag to run in headless mode"
     )
 
-    dash.addoption(
-        "--percy-assets",
-        action="store",
-        default="tests/assets",
-        help="configure how Percy will discover your app's assets",
-    )
```
alexcjohnson, line 49: Does the new percy runner have a different way of collecting assets so it doesn't need this?

T4rk1n: I think that was specific to the old percy client, can't find something similar for the new one. Do you think it would be better to make the migrate in a separate PR with a proper percy baseline build to compare ?

alexcjohnson: If they can make do without that it's way better... let's just verify that assets like images *are* in fact making it into the snapshots, and keep going.

alexcjohnson (approved):
💃 what a slog. Thanks for all the iterations @T4rk1n - let's see what Percy support says, whether there's a way for us to bring back their new SDK or not... for now at least this gets us onto the new rendered Firefox.

PR #2009, opened 2022-04-11
Clientside Callback Promise Handling

### LINKED ISSUE #1364: ClientSide Callbacks Promise Handling
It would be a nice funcionallity to be able to return promises in the clientside callbacks.
Right now it is impossible to make web requests in the clientside.
This would help to take some load from the server side, which would greatly improve the dash apps performance

Example of what I mean:
```javascript
async function MakeWebRequest(type) {
    var _url = "https://test.api.com/entities/?type="+type
    var headers = {
        "Accept":"application/json"
    }
    var config = {
        url: _url,
        headers: headers,
        method: 'get'
    }
    return await axios(config)
        .then(data => {
            return data
        })
        .catch(error => {
            console.log(error);
            return error;
        });
}

async function updateDropdown(name, path, previous_value) {

    var disabled = false
      ...
    if (courseID !== '') {
        var options = await MakeWebRequest("course");
    
        return [disabled,courseID, options]
    }
}

window.dash_clientside = Object.assign({}, window.dash_clientside, {
    newQuestionnaire: {
    updateDropdown: async function (name, path, previous_value) {
                return await updateDropdown(name, path, previous_value);
            }
}


```
This raises this type of error in the Web Page:
image

I think this should be a feature to really consider in future releases. 
I really like how dash lets you abstract from javascript, But it is still lacking the the ability to offload demanding jobs to the clientside...

MrTeale:
This PR aims to be the implementation of Client-side Callback's being able to handle Promises. This feature has been long discussed and requested as seen in #1364 and in this [Community Forum Discussion](https://community.plotly.com/t/client-side-data-from-external-source/41118). 

The current solution converts the error message for having a promise in a client-side callback into an `await` that gets the output from the promise instead. This does mean the client-side callback handler is now an `async function`.

Will close #1364 

## Contributor Checklist

   -  [x] Implement Async Promise Handling for Client-side Callbacks
   -  [x] Implement/Modify tests to handle new feature!
   -  [x] Seek discussions/opinions of key members (@alexcjohnson, @chriddyp)


MrTeale:
I feel like this feature was way to easy to implement. This is most likely due to a large amount of functionality and edge cases I'm unaware of. Would love your input/guidance on what else you'd need to see for this to be complete @chriddyp and @alexcjohnson.

alexcjohnson:
I'm not surprised this is pretty easy at this point - there have been a bunch of architectural changes bringing the async server-side callback execution and sync client-side execution closer to parallel. We just hadn't gotten around to trying it - and adding tests to ensure it works as intended. But I wonder if this wouldn't already work just by removing the test for a promise? I see later in the process we do a different test for promises (coming I presume from server-side callbacks) and we await these there:

https://github.com/plotly/dash/blob/f6abdfb823cdbe445eb7e660b29f1444e636a02a/dash/dash-renderer/src/observers/executingCallbacks.ts#L21-L24

Did you try this and see it fail? My main concern with blanket converting the synchronous function to async is whether we'll cause multiple page reflows/repaints in the case of multiple concurrent clientside callbacks, which can slow things down or make them look unnecessarily jittery. This is why for example plotly.js makes its own [`syncOrAsync` function](https://github.com/plotly/plotly.js/blob/12a49c0f225dea785c98bd36f6820091d98d2d85/src/lib/index.js#L446-L463) to ensure we keep execution as synchronous as possible.

Anyway as I alluded to above the main additional thing this will need is some tests. Perhaps:
- One adding a clientside callback that returns a promise that's known to resolve fairly quickly (don't add any external network requests, those become unreliable in CI tests, just make a small timeout or something) where another callback (clientside or serverside, doesn't matter) uses the output of the first one as an input.
- A pair of tests with a clientside and serverside callback both triggered by the same input, where we engineer it so in one case the serverside callback returns first and in the other case the clientside returns first, and ensure that the page updates first with the faster value before the slow has returned, followed by the slow value. Don't do this with time delays, as those are flaky and slow. Instead have the test code tell the callbacks when they're allowed to return. Server-side this can be done with [`multiprocessing.Lock`](https://github.com/plotly/dash/blob/f6abdfb823cdbe445eb7e660b29f1444e636a02a/tests/integration/callbacks/test_basic_callback.py#L42). Client-side you should be able to do something like:
```js
function cb(inputVal) {
    return new Promise(function(resolve) {
        window.callbackDone = function(deferredVal) {
            resolve(inputVal + ' - ' + deferredVal);
        }
    })
}
```
then in the test code, wait for the serverside output to be visible then call:
```python
dash_duo.driver.execute_script("window.callbackDone('the wait is over!')")
```

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [2.3.2]
```
alexcjohnson, line 5:
```suggestion
## [Unreleased]
```
We'll decide on the version number later, but this will be at least 2.4 since this is a new feature :)

alexcjohnson:
@MrTeale thanks very much for kicking this off! Also, don't worry about the existing test failures just yet - we're working on that right now in #2006, hopefully will have that done shortly and then we can merge that in here.

CHANGELOG.md
MrTeale: Ahhh true! Will fix this in the next commit

MrTeale:
Hey @alexcjohnson! Thanks for your review, was really detailed and helped a tonne to know what direction you wanted to go as the project owner. Quickly to begin, I did try removing the test for a promise, no dice unfortunately.

I've added the tests you suggested and they're passing well. If you (or anyone else) glance across them and sanity check to ensure they do what they're supposed to do. New tests added include:

- [`clsd015_clientside_chained_callbacks_returning_promise`](https://github.com/plotly/dash/blob/4d6f9cadc65180d28c9ffad2acf9b5dbe10fae53/tests/integration/clientside/test_clientside.py#L723-L747) - Tests using an output from a client-side callback (returning a promise) as the input to a server-side component. 
- [`clsd016_serverside_clientside_shared_input_with_promise`](https://github.com/plotly/dash/blob/4d6f9cadc65180d28c9ffad2acf9b5dbe10fae53/tests/integration/clientside/test_clientside.py#L750-L775) - Tests a shared input on both client-side (returning a promise) and server-side where the server-side callback finishes first.
- [`clsd017_clientside_serverside_shared_input_with_promise`](https://github.com/plotly/dash/blob/4d6f9cadc65180d28c9ffad2acf9b5dbe10fae53/tests/integration/clientside/test_clientside.py#L778-L807) - Tests a shared input on both client-side (returning a promise) and server-side where the client-side callback finishes first.

Note: I have xfailed `clsd005_clientside_fails_when_returning_a_promise` as this test will no longer be needed with this addition. I know some projects prefer to remove them, others xfail for posterity.

On the concern of converting a synchronous function to async and the desire to keep it as synchronous as possible, I completely agree. Perhaps we could determine whether the return value is a function or not ([currently done here](https://github.com/plotly/dash/blob/4d6f9cadc65180d28c9ffad2acf9b5dbe10fae53/dash/dash-renderer/src/actions/callbacks.ts#L249-L253)) outside of `handleClientside` and direct callbacks returning promises to a seperate async version of `handleClientside`?

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -274,8 +272,6 @@ function handleClientside(
             throw e;
         }
     } finally {
-        delete dc.callback_context;
```
alexcjohnson, line 277: Oh actually, we might want to keep the delete here too - so that if there's an error in a sync callback we still delete the context.

MrTeale: What would happen if we kept the deletion here and let it be stale within the async part of the callbacks? The guidelines is to not use it in the async portions of your callbacks anyway.

alexcjohnson: My worry is that it would mostly work, so users would come to rely on it, but then there would be intermittent problems, for example if there's a callback linked to a `dcc.Interval` or if the user triggers two actions in quick succession, that would be very hard to debug. Better to make it never work than fail sporadically.

MrTeale: Ah yes, that would definitely be the case. Let's have it both places for now and it can be removed in future if a way to have a stable context is found.

MrTeale:
> I presume it also works if these are declared as async functions, rather than regular functions that return promises? Including using the inline syntax

Yes it does. I have just added a couple more tests to cover this scenario. These two cover both `async function() { ... }` and `new Promise( function (resolve) { ... }` styles of async calls.

> Feels to me like the safest way to handle this would be to move that deletion right after the sync function call

Definitely agree that,  for now, the safest option is to read the `callback_context` in the synchronous portion. I've added an additional `delete dc.callback_context` to between the synchronous call to the callback and the check for a promise.

alexcjohnson (approved):
Beautiful! Thanks so much for your contribution @MrTeale 💃

PR #2027, opened 2022-04-27
provide simple example of the error

chriddyp:
I run into this error all of the time. Thought a simple example in the error message might help out users who are seeing it for the first time.


alexcjohnson (approved):
Nice touch 💃

PR #2029, opened 2022-04-28
Limit generated components to 250 explicit args by default

alexcjohnson:
configurable to any other number

Keep this to 253 or below for Py3.6 support. Props beyond those explicitly listed can still be used, they must just be invoked as kwargs. Also I believe that in Py3.6 it will still not be possible to use ALL props of a component if there are >253 - it's not just the function signature that's limited but also any individual function call. That's not a big deal, app developers want access to all of these props but they're very unlikely to actually use them all simultaneously. But it's possible that we'll want to set the limit lower than 250, in case the limit is (total # of named props) + (total unnamed props used simultaneously) - this I haven't tested.

In general, if you get up to hundreds of props in your component, it's a sign that you might want to refactor... group some of them together to make them easier to learn and work with, for example. But there are cases you want more - this PR is prompted by `dash-ag-grid` where we're wrapping another component (ag-grid) that exposes 265 props, and we're just forwarding them on to dash.

### DIFF
--- a/dash/development/_py_components_generation.py
+++ b/dash/development/_py_components_generation.py
@@ -111,9 +111,10 @@ def generate_class_string(
         if len(default_arglist) > final_max_props:
             default_arglist = default_arglist[:final_max_props]
             docstring += (
-                "\n\nNote: due to the large number of props for this component,"
-                "\nnot all of them appear in the constructor signature, but"
-                "\nthey may still be used as keyword arguments."
+                "\n\n"
+                "Note: due to the large number of props for this component,\n"
+                "not all of them appear in the constructor signature, but\n"
+                "they may still be used as keyword arguments."
             )
 
     default_argtext += ", ".join(default_arglist + ["**kwargs"])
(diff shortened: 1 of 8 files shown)


dash/development/_py_components_generation.py
```diff
@@ -91,18 +96,27 @@ def __init__(self, {default_argtext}):
...
+    ]
+
+    if max_props:
+        final_max_props = max_props - (1 if "children" in props else 0)
+        if len(default_arglist) > final_max_props:
+            default_arglist = default_arglist[:final_max_props]
+            docstring += (
+                "\n\nNote: due to the large number of props for this component,"
+                "\nnot all of them appear in the constructor signature, but"
+                "\nthey may still be used as keyword arguments."
```
T4rk1n, line 116: The newline at the beginning is kinda weird, it would be more readable to put them at the end of the line.

T4rk1n (approved):
:woman_dancing: Looks good.

dash/development/_py_components_generation.py
alexcjohnson: Good point - especially `\nnot` looks weird. I did this because I needed some newlines at the start of the string, but better to put those on their own and the others at the end of their lines -> https://github.com/plotly/dash/pull/2029/commits/b43044df12a4b4a10494e226f5145880c2be70be

Issue #1827, opened 2021-11-08
Add link_target option to dcc.Markdown


nopria:
Since `dash_table.DataTable` already has a `markdown_options` that allows to easily change the link target behavior of Markdown inside DataTable, I think  that even more so `dcc.Markdown` should have such a useful option.

The default link target behavior in HTML is `_self`, which opens the link in the same window. But most often you want the user to stay in your page, opening the link in a new window through the `_blank` HTML option.

Currently, to change the target behavior of a Markdown link you have to enable HTML inside Markdown, a cumbersome solution that poses **security issues**, as evident even from the name of the option which is `dangerously_allow_html`.

aGitForEveryone:
I'd like to support this feature. Would be very useful.

alexcjohnson:
Closed by @adrianoesch in #2034 🎉

PR #2034, opened 2022-05-03
New dcc.Markdown prop 'link_target' 

adrianoesch:
This PR is adding the prop `link_target` to dcc.Markdown component, analogous to the behaviour in datatable and thereby closing #1827. 

## Contributor Checklist

   -  [x] Figure out [react-markdown](https://github.com/remarkjs/react-markdown) integration
   -  [x] Add necessary props and documentation string
- [x] I have run the tests locally and they passed (though only dash-core-components markdown test).
- [x] I have added a test.

### optionals


alexcjohnson:
Beautiful, with a test and everything 😍 
Just add a changelog entry and we'll merge.

adrianoesch:
Great! Just added the changelog.

alexcjohnson:
The Percy tests here are showing a problem (Python errors showing up in the browser as a raw string rather than rendered HTML) that I suspect is due to the latest Werkzeug release. It certainly isn't anything you did in this PR @adrianoesch, but I don't want to merge until we fix it so we don't lose the correct Percy baselines. @T4rk1n can you take a look at addressing this?

T4rk1n:
@alexcjohnson The doctype changed to all lowercase in werkzeug 2.1.2, I modified the check to accommodate all case in https://github.com/plotly/dash/pull/2035

alexcjohnson (approved):
💃

PR #2041, opened 2022-05-09
Fix missing newline for split html debugger.

T4rk1n:
Think I missed something when I tested the fix for the html debugger with newer version of werkzeug.


alexcjohnson (approved):
💃

Issue #2003, opened 2022-03-30
`dcc.Markdown(mathjax=True, dangerously_allow_html=True)` combination


anders-kiaer:
Thanks for the MathJax support in the latest Dash version 🎉 I've been trying it out today, and there is small detail that (if possible to solve) would make it easier to integrate in our application.

We currently use `dcc.Markdown(content, dangerously_allow_html=True)` where user content has been through [`bleach`](https://pypi.org/project/bleach/) and includes some allowed content pre-processed by us (e.g. `<img>` with text caption / centering support etc, `<details>/<summary>` parts).

It looks like `dangerously_allow_html=True` + `mathjax=True` works in some cases, and in some cases not. Given
```python
from dash import Dash, dcc

app = Dash(__name__)

CONTENT = ...

app.layout = dcc.Markdown(CONTENT, dangerously_allow_html=True, mathjax=True)

app.run_server()
```
these are some experiments:
<table>
<tr>
<th><pre>CONTENT</pre></th>
<th>Math rendered</th>
</tr>
<tr>
<td><pre>&lt;details&gt;
    &lt;summary&gt;Topic&lt;/summary&gt;
    Some details
&lt;/details&gt;<br/>
$E = mc^2$</pre></td>
<td>✔️</td>
</tr>
<tr>
<td><pre>&lt;p&gt;Some paragraph&lt;/p&gt;<br/>
$E = mc^2$</pre></td>
<td>✔️</td>
</tr>
<tr>
<td><pre>&lt;p&gt;Some paragraph&lt;/p&gt;
$E = mc^2$</pre></td>
<td>❌</td>
</tr>
<tr>
<td><pre>&lt;p&gt;Some paragraph&lt;/p&gt; $E = mc^2$</pre></td>
<td>❌</td>
</tr>
<tr>
<td><pre>&lt;p&gt;Some paragraph with $E = mc^2$ inline math&lt;/p&gt;
<td>❌</td>
</tr>
</table>

Easy fix to support the given combination of arguments to `dcc.Markdown`? If not we could reconsider our use case of supporting allowed html tags 🙂

PR #2043, opened 2022-05-11
Fix markdown mathjax in html

nickmelnikov82:
As described in issue #2003, `dangerously_allow_html=True` + `mathjax=True` works in some cases, and in some cases not.
The problem is that the [markdown-parse](https://github.com/remarkjs/remark/tree/main/packages/remark-parse) recognizes whole string as the html object and stops further parsing. This PR fixes the processing of the mathjax in html.

## Contributor Checklist

   -  [x] fix issue
   -  [x] add test

### optionals

### DIFF
--- a/components/dash-core-components/src/fragments/Markdown.react.js
+++ b/components/dash-core-components/src/fragments/Markdown.react.js
@@ -108,6 +108,20 @@ export default class DashMarkdown extends Component {
                     )}
                 />
             ),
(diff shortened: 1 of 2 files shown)


alexcjohnson:
@nickmelnikov82 this looks great! Can you include the other two failing cases from #2003 in the tests, and add a changelog entry? The test failure looks spurious so with any luck it'll simply pass on the next try.

PR #2047, opened 2022-05-12
Fixed bug that prevented env DASH_DEBUG to work

### LINKED ISSUE #1979: DASH_DEBUG variable has no effect
https://dash.plotly.com/reference

Definition of external environment variable `export DASH_DEBUG=True` does not trigger debugging mode. It appears that it must be set within:

```
f __name__=='__main__':
    app.run_server(host='localhost', debug=True)
```

I have the latest dash. You can try to reproduce the issue by removing `debug=True` from code of your favorite app and defining `DASH_DEBUG` via `export`.

PGrawe:
Hey,
setting the debug value with the environment variable `DASH_DEBUG=true`, didn't work.
The default, if no debug value is given, is still false. Everything else works as intended.
This also fixes #1979 .

## Contributor Checklist

   -  [x] Changing the default debug value for `app.run` to `None`
   -   [x] Changing the default debug value to `False` in `enable_dev_tools`

### DIFF
--- a/tests/unit/test_configs.py
+++ b/tests/unit/test_configs.py
@@ -480,6 +480,5 @@ def test_debug_mode_enable_dev_tools(empty_environ, debug_env, debug, expected):
     if debug_env:
         os.environ['DASH_DEBUG'] = debug_env
     app = Dash()
-    # with pytest.raises(AssertionError):
     app.enable_dev_tools(debug=debug)
     assert app._dev_tools.ui == expected
--- a/tests/unit/test_configs.py
+++ b/tests/unit/test_configs.py
@@ -437,3 +437,49 @@ def test_app_invalid_delayed_config():
     app = Dash(server=False)
     with pytest.raises(AttributeError):
         app.init_app(app=Flask("test"), name="too late 2 update")
+
+
+@pytest.mark.parametrize(
+    "debug_env, debug, expected",
+    [
+        (None, None, False),
+        (None, True, True),
+        (None, False, False),
+        ('True', None, True),
+        ('True', True, True),
+        ('True', False, False),
+        ('False', None, False),
+        ('False', True, True),
+        ('False', False, False),
+    ],
+)
+def test_debug_mode_run(empty_environ, debug_env, debug, expected):
+    if debug_env:
+        os.environ['DASH_DEBUG'] = debug_env
+    app = Dash()
+    with pytest.raises(AssertionError):
+        app.run(debug=debug, port=-1)
+    assert app._dev_tools.ui == expected
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Thanks @PGrawe! There's a little more going on though - I suspect this PR breaks the documented behavior of `app.enable_dev_tools`:

https://github.com/plotly/dash/blob/b84fad8d4d2a9eb6b4ac18d3cd2e23bc705d594f/dash/dash.py#L1653-L1657

So maybe we need two calls to `get_combined_config` - put the one inside `enable_dev_tools` back to the way it was, but add another in `run` that defaults to `False`?

Now, how can we test this behavior? [`test_configs.py`](https://github.com/plotly/dash/blob/dev/tests/unit/test_configs.py) seems like the right place and has a lot of similar examples, and ideally we should test ten cases: (`app.enable_dev_tools` or `app.run`) x (`DASH_DEBUG=true | false`, or no env var with `debug=True | False | None` in the method call). The `debug` value isn't explicitly stored, but you can look at for example `app._dev_tools.ui` which should have the same value as `debug` if you haven't done anything else special.

Testing `app.run` in a unit test is tricky because it kicks off the blocking server run at the end. But I bet we could hack around this by getting it to throw an error after its internal `enable_dev_tools` call - for example `app.run(debug=debug, port=-1)` would throw [here](https://github.com/plotly/dash/blob/cebad0734215cc12a300a0173859cf1b4ea43bb0/dash/dash.py#L2004-L2009) so you could do something like:
```py
with pytest.raises(AssertionError):
    app.run(debug=debug, port=-1)
assert app._dev_tools.ui == expected_debug
```

PGrawe:
Hey Alex,
thank you so much for your help! I really appreciate it.  🙏 
I changed everything that you suggested and implemented the tests and they passed.
Let me know what you think.

tests/unit/test_configs.py
```diff
@@ -437,3 +437,49 @@ def test_app_invalid_delayed_config():
...
+        ('False', None, False),
+        ('False', True, True),
+        ('False', False, False),
+    ],
+)
+def test_debug_mode_enable_dev_tools(empty_environ, debug_env, debug, expected):
+    if debug_env:
+        os.environ['DASH_DEBUG'] = debug_env
+    app = Dash()
+    # with pytest.raises(AssertionError):
```
alexcjohnson, line 483:
```suggestion

```

alexcjohnson (approved):
💃 Excellent! Can you add a changelog entry? Then we'll merge 🎉

PGrawe:
Amazing! 💯 Just updated the changelog. Thanks for all your help :smiley:

PR #2049, opened 2022-05-13
Wait for class testing utilities

MrTeale:
As more and more CSS libraries and frameworks are using utility classes, there is a growing need for people to check an elements class attribute contains or equals a value. I am currently in this circumstance building a set of custom AIO components utilising TailwindCSS and FontAwesome icons.

My reasoning for making this addition was to check a callback had successfully changed the icon within a div. FontAwesome uses the `::before` pseudo-element with a `content` css property to contain the unicode character of the icon. The process to access psuedo-elements within Selenium is not the most developer friendly method ([example here](https://stackoverflow.com/a/59706267)) but if they're being created by classes, why not just check and wait for those? 

The testing documentation ([located here](https://dash.plotly.com/testing)) would need to be updated to include these new options under Browser APIs.

## Contributor Checklist

   -  [X] Add wait_for_class_to_equal
   -  [X] Add wait_for_contains_class

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 [#2003](https://github.com/plotly/dash/issues/2003) in which 
 `dangerously_allow_html=True` + `mathjax=True` works in some cases, and in some cases not.
 
-- [#2049](https://github.com/plotly/dash/pull/2043) Added `wait_for_class_to_equal` and `wait_for_contains_class` methods to `dash.testing`
+### Added
 
+- [#2049](https://github.com/plotly/dash/pull/2043) Added `wait_for_class_to_equal` and `wait_for_contains_class` methods to `dash.testing`
 
 ## [2.4.1] - 2022-05-11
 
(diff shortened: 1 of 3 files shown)


dash/testing/wait.py
```diff
@@ -68,6 +68,23 @@ def __call__(self, driver):
...
+        self.classname = classname
+
+    def __call__(self, driver):
+        try:
+            elem = driver.find_element(By.CSS_SELECTOR, self.selector)
+            classname = elem.get_attribute("class")
+            logger.debug(
+                "contains class {%s} => expected %s", classname, self.classname
+            )
+            return self.classname in str(classname)
```
alexcjohnson, line 83:
```suggestion
            return self.classname in str(classname).split(" ")
```
So we don't match part of a longer class, only an exact match of one of the classes.

alexcjohnson (approved):
Very nice addition @MrTeale! 💃 after my two minor comments are addressed.

dash/testing/wait.py
MrTeale: Ahhh yes! Great idea

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 [#2003](https://github.com/plotly/dash/issues/2003) in which 
 `dangerously_allow_html=True` + `mathjax=True` works in some cases, and in some cases not.
 
+- [#2049](https://github.com/plotly/dash/pull/2043) Added `wait_for_class_to_equal` and `wait_for_contains_class` methods to `dash.testing`
```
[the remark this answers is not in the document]
MrTeale: Moved with the latest commit

MrTeale:
Thanks @alexcjohnson! I've made those changes and should be 💃  now

PR #2053, opened 2022-05-16
Fix namespace props not resolving in ts components.

T4rk1n:
- Fix tsconfig.json compilerOptions not passed to ts compiler.
- Fix namespace props imports, `esModuleInterop` needs to be true. Resolve plotly/dash-typescript-component-template#4

### DIFF
--- a/dash/extract-meta.js
+++ b/dash/extract-meta.js
@@ -37,7 +37,7 @@ if (!src.length) {
 }
 
 if (fs.existsSync('tsconfig.json')) {
-    tsconfig = JSON.parse(fs.readFileSync('tsconfig.json'));
+    tsconfig = JSON.parse(fs.readFileSync('tsconfig.json')).compilerOptions;
 }
 
 let failedBuild = false;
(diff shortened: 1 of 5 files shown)


dash/extract-meta.js
```diff
@@ -37,7 +37,7 @@ if (!src.length) {
 }
 
 if (fs.existsSync('tsconfig.json')) {
-    tsconfig = JSON.parse(fs.readFileSync('tsconfig.json'));
+    tsconfig = JSON.parse(fs.readFileSync('tsconfig.json')).compilerOptions;
```
Jordan-Hall, line 40: Just remembered some 3rd party tools that would wrap around this (I'm thinking like NX which i use alot) actually uses Jsonc. Can we change out the JSON.parse to support this? Or add another ticket

dash/extract-meta.js
```diff
@@ -180,7 +180,7 @@ function gatherComponents(sources, components = {}) {
         return components;
     }
 
-    const program = ts.createProgram(filepaths, tsconfig);
+    const program = ts.createProgram(filepaths, {...tsconfig, esModuleInterop: true, jsx: "react"});
```
Jordan-Hall, line 183: What about if you use jsxImportSource and jsxFactory then JSX cant use JSX

alexcjohnson (approved):
💃

Issue #1769, opened 2021-09-21
[BUG] long_callback short interval makes job run in loop


sam1902:
I'm really excited to use `long_callback`s because it helps a lot for my use-cases, but [while reading the doc](https://dash.plotly.com/long-callbacks) (Example 5) I found that there exists an undocumented parameter `interval` which is set to `1000` in this example.

I tried setting it to `1000` too, but the progress updates were too slow, so I progressively lowered it. It behaved as I wanted when setting it to `100` (ms), but when running my job with `interval=100`, when the job finishes, it is immediatly executed again, even though the launch button's event is not triggered.

## Minimal replication code

```python3
#!/usr/bin/env python3
import time
import dash
import dash_html_components as html
import dash_core_components as dcc
from dash.long_callback import DiskcacheLongCallbackManager
from dash.dependencies import Input, Output
import plotly.graph_objects as go

## Diskcache
import diskcache
cache = diskcache.Cache("./cache")
long_callback_manager = DiskcacheLongCallbackManager(cache)

def make_progress_graph(progress, total):
    progress_graph = (
        go.Figure(data=[go.Bar(x=[progress])])
        .update_xaxes(range=[0, total])
        .update_yaxes(
            showticklabels=False,
        )
        .update_layout(height=100, margin=dict(t=20, b=40))
    )

    return progress_graph


app = dash.Dash(__name__, long_callback_manager=long_callback_manager)

app.layout = html.Div(
    [
        html.Div(
            [
                html.P(id="paragraph_id", children=["Button not clicked"]),
                dcc.Graph(id="progress_bar_graph", figure=make_progress_graph(0, 10)),
            ]
        ),
        html.Button(id="button_id", children="Run Job!"),
        html.Button(id="cancel_button_id", children="Cancel Running Job!"),
        html.Span(id="heavy-data-span", children="", style={"display": "none"}),
        html.Span(id="output-span", children="", style={"display": "none"}),
    ]
)

@app.long_callback(
    output=[Output("paragraph_id", "children"),Output("output-span", "children"),],
    inputs=Input("button_id", "n_clicks"),
    running=[
        (Output("button_id", "disabled"), True, False),
        (Output("cancel_button_id", "disabled"), False, True),
        (
            Output("paragraph_id", "style"),
            {"visibility": "hidden"},
            {"visibility": "visible"},
        ),
        (
            Output("progress_bar_graph", "style"),
            {"visibility": "visible"},
            {"visibility": "hidden"},
        ),
[...]

PR #2058, opened 2022-05-18
Fix long_callback for slow responses and small polling intervals

sebbegg:
This fixes (https://github.com/plotly/dash/issues/1769) 

This avoids the race-condition between the long_callback interval triggers and the long callback returning slow/large payloads.
As mentioned in #1769 this PR basically does the following:

- Instead of returning the result of a long callback directly from the interval callback, the interval callback only signals that the result is ready by setting the result key and job id on an newly added "result" helper store
- When the result is ready, the interval is disabled directly (not delayed to the next interval trigger) and the user store data is cleared
- A newly added callback on the "result" store fetches the result

The code snippets for reseting the user store seem a little redundant between the `should_cancel` branch and the new parts of code. I'd be happy to refactor that somewhat, but wanted to keep the diff as small as possible for a start.

## Contributor Checklist

- [x] I have run ~~the tests~~ `pytest -k long_callback` locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))

### optionals

<sub>Sebastian Eckweiler (sebastian.eckweiler@mercedes-benz.com), Mercedes-Benz AG on behalf of MBition GmbH ([Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md))</sub>

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1161,7 +1161,7 @@ class Dash:
         self._long_callback_count += 1
         long_callback_id = self._long_callback_count
 
-        # Create Interval and Store for long callback and add them to the app's
+        # Create Interval and Stores for long callback and add them to the app's
         # _extra_components list
         interval_id = f"_long_callback_interval_{long_callback_id}"
         interval_component = dcc.Interval(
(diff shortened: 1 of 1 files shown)


T4rk1n:
This fix is only partial, works ok for larger response. But when I tested this with the first app example from #1769, I still get infinite loops and the callback doesn't resolve and I can't cancel.

I think as long as the long callbacks requests are made via an interval component, there is bound to have issues with the polling. There is also issues with the extra components not playing well with the coming up multi page feature.

As such, we have decided to improve the long callbacks and bring them on par with regular callbacks. I am refactoring the long callback to be handled in the renderer with a request polling chain to replace the interval and extra components added by the current long callback solution. There will only be one request at time, when the fetch completes, it will either have the progress updates and restart the fetch after a timeout, or complete the initial callback outputs.

Thank you for your contribution, really helped me have a better grasp of the timing issues. I will close this PR as it's not completely fixed and would require more work while I am currently working on an overhaul of the long callbacks.

PR #2062, opened 2022-05-23
Fixes AIO Callback testing with Dash-Duo

### LINKED ISSUE #1933: dash_duo does not work with All-in-One components
```
dash                      2.1.0
dash-bootstrap-components 1.0.3
pytest                    7.0.1
```

**Describe the bug**

It seems that dash testing utils don't work property with All-in-One components. For instance, given the following test scenario:

```python
import uuid
from typing import Optional

import dash
from dash import MATCH, Input, Output, callback, html


class SomeButton(html.Button):
    class ids:
        button = lambda aio_id: {"component": "SomeButton", "aio_id": aio_id}

    def __init__(self, aio_id: Optional[str] = None):
        if aio_id is None:
            aio_id = str(uuid.uuid4())

        super().__init__(
            "Loading...",
            id=self.ids.button(aio_id),
            disabled=True,
        )

    @callback(
        Input(ids.button(MATCH), "children"),
        output=dict(
            label=Output(ids.button(MATCH), "children"),
            disabled=Output(ids.button(MATCH), "disabled"),
        ),
    )
    def update_button(_):
        return dict(label="Loaded", disabled=False)


def test_some_button_first_time(dash_duo: dash.testing.composite.DashComposite):
    app = dash.Dash("foo123")
    app.layout = html.Div(SomeButton(), id="wrapper")

    dash_duo.start_server(app)
    dash_duo.wait_for_contains_text("#wrapper", "Loaded")


def test_some_button_second_time(dash_duo: dash.testing.composite.DashComposite):
    app = dash.Dash("bar123")
    app.layout = html.Div(SomeButton(), id="wrapper")

    dash_duo.start_server(app)
    dash_duo.wait_for_contains_text("#wrapper", "Loaded")
```

The first run `test_some_button_first_time` is perfectly fine. But `test_some_button_second_time` results with `selenium.common.exceptions.TimeoutException: Message: text -> Loaded not found inside element within 10s` error.
I ran it with `--pause` option and in the DevTools I noticed that for the second test the update callback `_dash-update-component` is not being triggered (see the attached screenshots).

Moreover, if I run `test_some_button_second_time` separately or if I comment out the first test everything is fine.
For some reason callbacks are fired only for the first test.

**Expected behavior**

It seems that `dash_duo` does not work with `@callback` for all-in-one components.

**Screenshots**

CleanShot 2022-02-17 at 17 25 05

CleanShot 2022-02-17 at 17 26 26

MrTeale:
This PR fixes #1933. The AIO components are unable to be tested using Pytest and the Dash-Duo object. The callbacks set via `dash.callback` (rather than `app.callback`) are added to a global map that is cleared after the first instance of `dash.Dash._setup_server`.

To get around this, I'm currently exploring whether it is possible to leave the `GLOBAL_CALLBACK_MAP` and `GLOBAL_CALLBACK_LIST` populated rather than clearing them on setup.

## Contributor Checklist

   -  [ ] Explore whether it is as simple as allowing the MAP and LIST to remain populated
   -  [ ] If not, explore whether we can have another MAP as suggested by @T4rk1n in #1933 

### optionals

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1400,10 +1400,9 @@ class Dash:
                     "assigned with `app.callback`."
                 )
 
-            self.callback_map[k] = _callback.GLOBAL_CALLBACK_MAP.pop(k)
+            self.callback_map[k] = _callback.GLOBAL_CALLBACK_MAP[k]
 
         self._callback_list.extend(_callback.GLOBAL_CALLBACK_LIST)
-        _callback.GLOBAL_CALLBACK_LIST.clear()
 
     def _add_assets_resource(self, url_path, file_path):
         res = {"asset_path": url_path, "filepath": file_path}


MrTeale:
Turns out it is possible to not remove them from the `GLOBAL_CALLBACK_MAP` and clear the `GLOBAL_CALLBACK_LIST` and everything work fine.

Was there any particular reasoning as to removing them the way they were?

gvwilson:
@T4rk1n can you please have a look at this at some point in cycle 17? thanks - @gvwilson

Issue #1985, opened 2022-03-24
[BUG] Long callback function executed without being triggered


keurfonluu:
**Describe your context**
Please provide us your environment, so we can easily reproduce the issue.

```
dash                      2.3.0
dash-bootstrap-components 1.0.3
dash-core-components      2.0.0
dash-html-components      2.0.0
dash-table                5.0.0
dash-vtk                  0.0.9
```

**Describe the bug**

I have a rather complex app where the layout content is generated by a callback (single page app). For one specific layout content, I need a long callback because one callback function takes some time to execute. The problem is that, somehow, right after the layout is generated, the long callback is triggered automatically (even when using prevent_initial_call). No problem if I use a standard callback, though.

Please find below a sample script to reproduce this issue based on the long callback example in the documentation. I also defined a `validation_layout` as recommended by the documentation.

```python
import dash
from dash import html
from dash.long_callback import DiskcacheLongCallbackManager
from dash.dependencies import Input, Output

import diskcache
cache = diskcache.Cache("./cache")
long_callback_manager = DiskcacheLongCallbackManager(cache)

app = dash.Dash(
    __name__,
    suppress_callback_exceptions=True,
    long_callback_manager=long_callback_manager,
)

layout1 = html.Div(
    [
        html.Button("Generate layout", id="button_1"),
        html.Div(id="main"),
    ],
)
layout2 = html.Div(
    [
        html.Div([html.P(id="paragraph_id", children=["Button not clicked"])]),
        html.Button(id="button_id", children="Run Job!"),
    ]
)

app.layout = layout1
app.validation_layout = html.Div([layout1, layout2])

@app.callback(
    Output("main", "children"),
    Input("button_1", "n_clicks"),
    prevent_initial_call=True,
)
def generate_layout(n_clicks):
    return layout2

@app.long_callback(
    Output("paragraph_id", "children"),
    Input("button_id", "n_clicks"),
    prevent_initial_call=True,
)
def callback(n_clicks):
    return [f"Clicked {n_clicks} times"]

if __name__ == "__main__":
    app.run_server(debug=True)
```

Execute the app, click on "Generate layout", wait a few seconds, `"paragraph_id"` children will automatically change to "Clicked None times".

**Expected behavior**

The long callback not to be triggered after the new layout is generated. For the example above, the `children` property of the component `"paragraph_id"` to be updated only when the button is clicked.

celia-plotly:
I'm posting this here since it provides more information about why it might be happening:
[...]

Issue #1346, opened 2020-07-27
[BUG] dcc.Location not honoring refresh=False for dynamic layouts starting with dash v 1.13


giftculture:
Thank you so much for helping improve the quality of Dash!

We do our best to catch bugs during the release process, but we rely on your help to find the ones that slip through.

**Describe your context**
Please provide us your environment so we can easily reproduce the issue.

```
dash                         1.13.0
dash-auth                    1.3.2
dash-bootstrap-components    0.10.1a0
dash-core-components         1.10.0
dash-daq                     0.1.7
dash-html-components         1.0.3
dash-renderer                1.5.0
dash-table                   4.8.0
dash-table-experiments       0.6.0
dash.ly                      0.17.3
```

**Describe the bug**
Starting with dash v 1.13, dcc.Location isn't respecting refresh=False for dynamic layouts, it is getting triggered over and over again

index.py
```
@app.callback(Output(‘page-content’, ‘children’),
[Input(‘url’, ‘pathname’)])
def display_page(pathname):
     if pathname == ‘/apps/cumulative_pnl’:
          return cumulative_pnl.return_layout()
```
cumulative_pnl.py looks like:
```
from dash.dependencies import Input, Output, State
import dash_html_components as html
import dash_core_components as dcc
import datetime

from . import render_graphs

from app import app



def return_layout():
    return html.Div(
        [
            datetime.datetime.now(),
            dcc.Location(id='url', refresh=False),
        ],
    )
```

In Dash 1.13, it seems to get stuck in an infinite loop, triggering over and over, with the datetime continually updating.  If I comment out dcc.Location, I get the expected behavior, so there's something going on with dcc.Location - it appears to not be respecting refresh=False
**Expected behavior**

Layout renders one time, with the current datetime (works under dash 1.12)
Each time I go to /apps/cumulative_pnl, the return_layout function is invoked and my layout renders once, showing the datetime as of page load

chriddyp:
@giftculture - Thanks for reporting! It is possible for you to create a very simple, standalone example that reproduces this issue? We haven't seen this issue come up yet and our tests didn't fail.

giftculture:
Hello!  

A simple example is included here, let me know if you have any issues reading it
[simple_reload_test.tar.zip](https://github.com/plotly/dash/files/4995821/simple_reload_test.tar.zip)

giftculture:
@chriddyp  Not sure if you got notification for my post above, but I attached a zipped tar file that includes a simple example

[...]

Issue #1312, opened 2020-06-25
[BUG] No callback to dcc.Location when two or more dcc.Location are placed in layout


papalagichen:
Thank you so much for helping improve the quality of Dash!

We do our best to catch bugs during the release process, but we rely on your help to find the ones that slip through.

**Describe your context**
Please provide us your environment so we can easily reproduce the issue.

-  replace the result of `pip list | grep dash` below
```
dash                      1.13.3
dash-auth                 1.3.2
dash-bootstrap-components 0.10.2
dash-core-components      1.10.1
dash-html-components      1.0.3
dash-renderer             1.5.0
dash-table                4.8.1
```
-  if frontend related, tell us your Browser, Version and OS

    - OS: [e.g. macOS]
    - Browser [e.g. chrome, safari, firefox]
    - Version [e.g. 83.0.4103.116, 13.1.1, 77.0.1]

**Describe the bug**

In a multi-page app, browser's back and forward buttons do not work when multiple dcc.Location components are placed in layout. No callback is triggered nor request is sent when browser's back or forward button is clicked. As a result, I see the URL is changed while the page is not rendered accordingly.

Two dcc.Location are placed in the layout like following code snippet.

```python
url_bar_and_content_div = html.Div([
    dcc.Location(id='url', refresh=False),
    dcc.Location(id='url2', refresh=False),
    html.Div(id='page-content')
])
```

**Expected behavior**

When URL is changed, the corresponding callback should be called and page should be rendered accordingly.

**Screenshots**

wmachadoa:
@papalagichen Did you find a solution to this? I'm experiencing the same issue

jowlo:
I encountered the same problem while using pages. The bug is triggered when one of the `dcc.Location`s is removed from the layout, which unregisters the `window.onpopstate` event handler that triggers the callback. 

Having a `Location` in one of the pages is enough to trigger it, which will also break the browser's back button for the pages. Adapting the example by @AnnMarieW above:
```python
import dash
from dash import Dash, html, dcc

app = Dash(__name__, use_pages=True, pages_folder="")


dash.register_page("stocks", layout=[
    html.Div("stocks page"),
    dcc.Location(id="url", refresh=False),
])
dash.register_page("home", path="/", layout=html.Div("home page"))

app.layout = html.Div(
    [
        dcc.Location(id="url2", refresh=False),

        dcc.Link('Stocks', href='/stocks'),
        html.Br(),
        dcc.Link('Home', href='/'),
[...]

PR #2068, opened 2022-05-27
dcc location callback nav

AnnMarieW:
## Improve page navigation when `dcc.Location` is updated in a callback.

Currently when dcc.Location is updated in a callback, the browser url is updated, but it does not navigate to the new page unless the browser is refreshed.  This PR makes it unnecessary to refresh the browser -- giving the same fast navigation as when  user clicks on a `dcc.Link`. 

As requested on the [Dash Community Forum](https://community.plotly.com/t/introducing-dash-pages-a-dash-2-x-feature-preview/57775/123?u=annmariew)

### Limitations
The following issues occur when there are multiple `dcc.Location` components in a multi-page app.  These are all existing issues and are not new based on changes in this PR.  It they aren't fixed as part of the PR, then it would be helpful to document the workarounds.  

- There are potential conflicts when callbacks update multiple `dcc.Location` location causing the browser to crash.
- When using in a multi-page app with multiple `dcc.Location` components, they must all be on the same page - typically the `app.py` file.  (Or the `index.py` if using the multi-page structure where `app` is defined in it's own file)

- When navigating to a page with the callback to update the `dcc.Location` the callback is fired even when `prevent_initial_call=True`  This has been reported in [issue 1346](https://github.com/plotly/dash/issues/1346) and [issue 1985](https://github.com/plotly/dash/issues/1985)

- Browser forward/back doesn't work when there are multiple `dcc.Location` components.  
    -   For Pages apps, this is working now :tada:  This was solved by adding `refresh="callback-nav"` to the `dcc.Location` in the `dash.page_container`

## Contributor Checklist

### optionals

### Sample usage
``` python
import dash
from dash import Dash, html, dcc, Input, Output, State

app = Dash(__name__, use_pages=True, pages_folder="")


def layout_stocks(stock=None):
    return html.Div(f"Stock analysis for {stock}")


dash.register_page("stocks", layout=layout_stocks, path_template="/stocks/<stock>")
dash.register_page("home", path="/", layout=html.Div("home page"))

app.layout = html.Div(
    [
        dcc.Location(id="url", refresh="callback-nav"),
        html.Div("Enter Stock Symbol"),
        dcc.Input(id="stock-input"),
        html.Button("submit", id="submit", n_clicks=0),
        html.Hr(),
        dash.page_container,
    ]
)


@app.callback(
    Output("url", "href"), Input("submit", "n_clicks"), State("stock-input", "value")
)
def go(n, stock):
    if n > 0:
        return f"/stocks/{stock}"
    return dash.no_update


if __name__ == "__main__":
    app.run(debug=True)
```


T4rk1n:
The PR look good.

Is there an issue for the limitation:
> There are potential conflicts when callbacks update multiple dcc.Location location causing the browser to crash.

alexcjohnson:
This looks good to me too. As @T4rk1n says we should make sure there's an issue open regarding the browser crash scenario, and any of the other limitations we aren't addressing here. In particular:

> When using in a multi-page app with multiple `dcc.Location` components, they must all be on the same page

Does that include the automatic one added by the pages skeleton? If so is the statement actually "none of the individual pages can have a `dcc.Location`, only the main `app.layout`? Why would you want to do this anyway, and does this apply even if those other `dcc.Location` components are only used as inputs?

> Browser forward/back doesn't work when there are multiple `dcc.Location` components

Only with `refresh=True` now, I gather, since as you say it's fixed for `'callback-nav'`. Is there a concise statement we can make in the docs about when to use each of the three `refresh` options? Something like:
- Use `True` to navigate outside this Dash app
- Use `False` if the same callback that updates the `Location` component is also updating the page content
- Use `'callback-nav'` if a different callback will respond to the new `Location` with updated content.

Do I have that right or is there more nuance (or things I got wrong)?

Anyway none of this needs to go _in_ this PR, but before merging we should have answers/issues for these points. And resolve the merge conflicts 😏

AnnMarieW:
Merge conflicts are fixed 


Here is clarification on the use-case and limitations:


- Use `True` to navigate outside the Dash app or to manually refresh a page.
- Use `False` if the same callback that updates the Location component is also updating the page content - typically used in multi-page apps that do not use Pages.
- Use `'callback-nav'` if a different callback will respond to the new Location with updated content. This is typical with multi-page apps that use Pages.  This will allow for navigating to a new page without refreshing the page

------------------------
Limitations

> There are potential conflicts when callbacks update multiple `dcc.Location` location causing the browser to crash.

I haven't yet been able to create a minimal example to duplicate this issue.  It may have been caused by multiple callbacks updating the URL with multiple `dcc.Location` components.  I'll continue to work on making a MRE, but is it necessary to document that issue before this PR is merged?  It might be an edge case. (Or an old bug in one of my apps)

> To make components available for all pages of a multi-page app the component must be in the app.py file.  

This was my workaround for the issue above.  If I used one `dcc.Location` in `app.py` shared between all pages of a multi-page app, then there were no conflicts with different callbacks updating the URL.  Since I can't reproduce the issue, this shouldn't be considered a limitation.


> Browser forward/back doesn't work when there are multiple dcc.Location components

I added a minimal example in https://github.com/plotly/dash/issues/1312

alexcjohnson (approved):
💃 Thanks @AnnMarieW for adding this mode, and for adding examples for the remaining buggy behavior 🎉 

@LiamConnors would you look over the description @AnnMarieW posted for when to use each of these three `refresh` modes https://github.com/plotly/dash/pull/2068#issuecomment-1445161251 and ensure they're reflected in the docs?

AnnMarieW:
@alexcjohnson once @LiamConnors reviews the descriptions, should I update the docstrings for those props as well?

alexcjohnson:
Sure, that could be useful 🌟 
Probably doesn't need to be coordinated - you can update the docstring now if you like!

PR #2069, opened 2022-05-28
Slim percy

alexcjohnson:
Making more efficient use of our percy snapshots.
- Table: uses storybook, and made many many very similar images that have been very stable over time. I didn't remove any, but combined like tables into larger images, and made some of the individual tables smaller. I think this will actually be more useful than the individual images anyway, as it's easier to compare the cases and see what's different between them.
  - Reduced from 358 to 31 images
  - Compare [before](https://percy.io/plotly/dash-table-test/builds/18461151/search/preview/1035387820?browser=firefox&searchParam=dashtable%2Fmultiline&viewLayout=overlay&viewMode=new&width=1280) to [after](https://percy.io/plotly/dash-table-test/builds/18466624/changed/1035744881?browser=firefox&browser_ids=22&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1280&widths=1280)
- Others: lots of snapshots were being used as very coarse stand-ins for what's actually a simple assertion about the DOM - and in fact the DOM assertion is more useful in these cases because it targets exactly the behavior we want to lock in. The snapshots that really are about visuals I left unchanged.
  - Reduced from 151 to 92 images
  - Compare [before](https://percy.io/plotly/dash/builds/18460940/unchanged/1035375176?browser=firefox&viewLayout=overlay&viewMode=new&width=1280) to [after](https://percy.io/plotly/dash/builds/18466631/changed/1035745082?browser=firefox&browser_ids=22&subcategories=approved&viewLayout=overlay&viewMode=new&width=1280&widths=1280)

In total this drops us a factor of 4, from 509 to 123 snapshots per build.


components/dash-core-components/tests/integration/graph/test_graph_varia.py
```diff
@@ -178,11 +175,19 @@ def render_content(click, prev_graph):
...
+    )
+
+    dash_dcc.wait_for_element("#click").click()
     time.sleep(2)  # Wait for graph to re-render
-    dash_dcc.percy_snapshot(
-        "render-empty-graph ({})".format("eager" if is_eager else "lazy")
+    assert dash_dcc.driver.execute_script(
+        """
+        return (document.querySelector('.js-plotly-plot').data || []).length === 0;
+    """
```
T4rk1n, line 190: This could be refactored in a function to avoid repetition.

components/dash-core-components/tests/integration/input/test_input_basics.py
```diff
@@ -27,6 +27,12 @@ def input_id(type_):
             for _ in ALLOWED_TYPES
         ]
         + [html.Div(id="output")]
+        + [
+            dcc.Input(
+                id=input_id(_) + "2", type=_, placeholder="input type {}".format(_)
+            )
```
T4rk1n, line 33: f-string

components/dash-core-components/tests/integration/input/test_input_basics.py
```diff
@@ -65,6 +69,12 @@ def test_inbs002_user_class(dash_dcc):
     dash_dcc.start_server(app)
 
     dash_dcc.find_element(".test-input-css")
-    dash_dcc.percy_snapshot("dcc styled input - width: 100%, border-color: hotpink")
+
+    def get_style(attr):
+        sel = "document.querySelector('.test-input-css input')"
+        return dash_dcc.driver.execute_script(f"return getComputedStyle({sel}).{attr}")
+
```
T4rk1n, line 76: There is `wait_for_style` with the browser or is that because we don't want to wait?

T4rk1n (commented):
Nice, I like this, will be much easier to review the percy diff.

components/dash-core-components/tests/integration/input/test_input_basics.py
alexcjohnson: Good call, I had forgotten about that one -> https://github.com/plotly/dash/pull/2069/commits/4d869c2cc693f88ff22c5e2e013bd9d83397ad3f

components/dash-core-components/tests/integration/graph/test_graph_varia.py
alexcjohnson: definitely. https://github.com/plotly/dash/pull/2069/commits/4d869c2cc693f88ff22c5e2e013bd9d83397ad3f

components/dash-core-components/tests/integration/input/test_input_basics.py
T4rk1n: The style test fails, I think the colors may need the alpha value.

PR #2084, opened 2022-06-08
Fixed viewport meta-tags

AnnMarieW:
In dash 2.5.0, a default viewport meta tag was added as recommended for mobile-optimized sites by [mdn.](https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag)
This PR makes it so that this feature can be disabled by providing an empty viewport meta tag. 
 e.g. `app = Dash(meta_tags=[{"name": "viewport"}])`

## Contributor Checklist

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -133,6 +133,29 @@ def test_inin007_meta_tags(dash_duo):
         assert meta_tag.get_attribute("content") == meta_info["content"]
 
 
+def test_inin007b_change_viewport_meta_tag(dash_duo):
+    """
+    As of dash 2.5 the default viewport meta tag is:
+        [{"name": "viewport", "content": "width=device-width, initial-scale=1"}]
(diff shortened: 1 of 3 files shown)


tests/integration/test_integration.py
```diff
@@ -133,6 +133,29 @@ def test_inin007_meta_tags(dash_duo):
         assert meta_tag.get_attribute("content") == meta_info["content"]
 
 
+def test_inin007b_change_viewport_meta_tag(dash_duo):
```
alexcjohnson, line 136: I'd rather we not squeeze test numbers in between with `b` or such - makes it harder to run just one test with eg `pytest -k inin007`. Just shift some of the other test numbers so this one can have its own. Maybe the ones up front since this file starts at `inin004`?

alexcjohnson (approved):
💃 Excellent!

Issue #2086, opened 2022-06-09
[BUG] Use of "id"  in Pattern Matching Callbacks breaks args_grouping


MrTeale:
**Describe your context**
 Currently running on Python 3.9.10 on MacOS Monterey (12.3.1) with the below dash libraries installed
```
dash                 2.5.0
dash-core-components 2.0.0
dash-cytoscape       0.3.0
dash-daq             0.5.1
dash-html-components 2.0.0
dash-table           5.0.0
```

**Describe the bug**

Using `id` as a key within a component's id breaks the new callback context's `args_grouping` function. An example of a component within my use cases looks like:
```python
dcc.Store(id={"id": "A-store", "type": "scenario"}, storage_type="local"),
dcc.Store(id={"id": "B-store", "type": "scenario"}, storage_type="local"),
dcc.Store(id={"id": "C-store", "type": "scenario"}, storage_type="local"),
```
and the callback Input looks like:

```python
Input({"type": "scenario", "id": ALL}, "data"),
```

Output Traceback from Error (apologies, I've had to hide some personal file paths):

```bash
Traceback (most recent call last):
  File ".....", line 172, in update_graph
    pprint(ctx.args_grouping)
  File "/Users/...../.venv_core/lib/python3.9/site-packages/dash/_callback_context.py", line 18, in assert_context
    return func(*args, **kwargs)
  File "/Users/...../.venv/lib/python3.9/site-packages/dash/_callback_context.py", line 182, in args_grouping
    recursive_update(grouping)
  File "/Users/...../.venv/lib/python3.9/site-packages/dash/_callback_context.py", line 180, in recursive_update
    recursive_update(i)
  File "/Users/...../.venv/lib/python3.9/site-packages/dash/_callback_context.py", line 179, in recursive_update
    update_args_grouping(i)
  File "/Users/...../.venv/lib/python3.9/site-packages/dash/_callback_context.py", line 157, in update_args_grouping
    prop_id = f"{str_id}.{g['property']}"
KeyError: 'property'
```

**Expected behavior**

The ability to use any key within a pattern matching ID and still have args grouping working or this limitation documented within the Dash Docs. Current documentation recommends using `type`, `index` or `id` as keys for readability, so this would need to change.

MrTeale:
cc @AnnMarieW - Awesome work on this feature! Makes pattern matching a tonne easier. Figured I'd CC you since you'd know this section of Dash best.

AnnMarieW:
[...]

PR #2087, opened 2022-06-09
Fix args_grouping to handle "id" in the `id`

AnnMarieW:
fixes [#2086](https://github.com/plotly/dash/issues/2086) where using id as a key within a component's id breaks the new callback context's args_grouping function. 
## Contributor Checklist

### DIFF
--- a/dash/_callback_context.py
+++ b/dash/_callback_context.py
@@ -152,7 +152,7 @@ class CallbackContext:
         grouping = getattr(flask.g, "args_grouping", {})
 
         def update_args_grouping(g):
-            if isinstance(g, dict) and "id" in g:
+            if isinstance(g, dict) and isinstance(g.get("id"), dict):
                 str_id = stringify_id(g["id"])
                 prop_id = f"{str_id}.{g['property']}"
 
--- a/dash/_callback_context.py
+++ b/dash/_callback_context.py
@@ -170,7 +170,8 @@ class CallbackContext:
             if isinstance(g, (tuple, list)):
                 for i in g:
                     update_args_grouping(i)
-                    recursive_update(i)
+                    if not isinstance(i.get("id"), dict):
+                        recursive_update(i)
             if isinstance(g, dict):
                 for i in g.values():
                     update_args_grouping(i)
--- a/dash/_callback_context.py
+++ b/dash/_callback_context.py
@@ -1,11 +1,10 @@
 import functools
 import warnings
 import json
-from copy import deepcopy
 import flask
 
 from . import exceptions
-from ._utils import stringify_id, AttributeDict
+from ._utils import AttributeDict
 
 
 def has_context(func):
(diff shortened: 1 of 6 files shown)


dash/_callback_context.py
```diff
@@ -152,7 +152,7 @@ def display(btn1, btn2):
         grouping = getattr(flask.g, "args_grouping", {})
 
         def update_args_grouping(g):
-            if isinstance(g, dict) and "id" in g:
+            if isinstance(g, dict) and "id" in g and "property" in g:
```
alexcjohnson, line 155: What if someone makes a PMC ID with both `id` AND `property` as keys? Is there some other way we can tell the difference between a dict we should recurse into and a dict ID?

AnnMarieW: D'oh!  
ok, I think it's correct in the 3rd update :slightly_smiling_face:

alexcjohnson: I'm still worried that there could be ways to fool us - now perhaps we wouldn't recurse when we should. Like:
```py
@callback(
    inputs=dict(
        a=dict(id=Input(...), property=Input(...))
    )
)
```

Feels to me like the more robust way to handle this would be when we have a known flat list [`inputs_state`](https://github.com/plotly/dash/blob/c12c957d5fff8d0d65ca06276af7f9e90b424971/dash/dash.py#L1431-L1432) - add the extra info to THAT before we map that into its grouped form.

Separately, would it be worthwhile augmenting `outputs_grouping` as well? That doesn't have `triggered` or `value` but it could benefit from `str_id` and having all its parts be `AttributeDict`s

AnnMarieW: Your above example works fine, so do all of the following examples. So far, I haven't found anything that _didn't_ work.

As far as `outputs_grouping`, I think it's a good idea, and I agree that it should be a separate issue.  



--------------
```python
from dash import Dash, html, Input, Output,ctx
app = Dash(__name__, suppress_callback_exceptions=True)

app.layout = html.Div([
    html.Button("btn1",id="btn1", n_clicks=0),
    html.Button("btn2", id="btn2", n_clicks=0),
    html.Div(id='container', children=[]),
])


@app.callback(
    Output('container', 'children'),
    inputs=dict(
        a=dict(id=Input("btn1", "n_clicks"), property=Input("btn2", "n_clicks")),
    )
)
def update1(a):
    g=ctx.args_grouping
    return  f"btn1 clicked {g.a.id.value} times, btn2 clicked {g.a.property.value} times"


if __name__ == '__main__':
    app.run_server(debug=True)

```

This example has component ids as "id" and "property" as well

-----------------
```python

from dash import Dash, html, Input, Output, ctx
app = Dash(__name__, suppress_callback_exceptions=True)

app.layout = html.Div([
    html.Button("id",id="id", n_clicks=0),
    html.Button("property", id="property", n_clicks=0),
    html.Div(id='container', children=[]),
])


@app.callback(
    Output('container', 'children'),
    inputs=dict(
        a=dict(id=Input("id", "n_clicks"), property=Input("property", "n_clicks")),
    )
)
def update1(a):
    g=ctx.args_grouping
    return  f"id btn clicked {g.a.id.value} times, property btn clicked {g.a.property.value} times"


if __name__ == '__main__':
    app.run_server(debug=True)


```

------

Pattern matching dict id's are fine too:

```python
from dash import Dash, html, Input, Output, ctx
app = Dash(__name__, suppress_callback_exceptions=True)

app.layout = html.Div([
    html.Button("id",id="id", n_clicks=0),
    html.Button("property", id="property", n_clicks=0),
    html.Div(id='container', children=[]),
])


@app.callback(
    Output('container', 'children'),
    inputs=dict(
        a=dict(id=Input("id", "n_clicks"), property=Input("property", "n_clicks")),
    )
)
def update1(a):
    g=ctx.args_grouping
    return  f"id btn clicked {g.a.id.value} times, property btn clicked {g.a.property.value} times"


if __name__ == '__main__':
    app.run_server(debug=True)
```

alexcjohnson (approved):
Love it! 💃 

FYI re: `args_grouping_values` - normally not a good idea to remove a feature outside of a major version, but (a) it was never documented, (b) this was partially broken, (c) it was only added in 2.4 so hasn't been available for long, and (d) Google is unaware of any mentions of it outside of the source code in which it was defined. So I think this is OK.

Issue #2066, opened 2022-05-26
[BUG] typescript complier RangeError: Maximum call stack size exceeded


Jordan-Hall:
Repo: https://github.com/communitiesuk/gov-uk-dash-react-components/tree/move-to-ts

**Output:**

```
c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:52545
                        return type.flags & 1048576 /* TypeFlags.Union */ ? ts.factory.createUnionTypeNode(typeNodes) : ts.factory.createIntersectionTypeNode(typeNodes);
                                                                                       ^

RangeError: Maximum call stack size exceeded
    at typeToTypeNodeHelper (c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:52545:88)
    at serializeTypeForDeclaration (c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:53892:30)
    at symbolToParameterDeclaration (c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:53321:41)
    at c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:53231:274
    at Array.map (<anonymous>)
    at signatureToSignatureDeclarationHelper (c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:53231:240)
    at createTypeNodesFromResolvedType (c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:52990:43)
    at createTypeNodeFromObjectType (c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:52837:35)
    at visitAndTransformType (c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:52764:34)
    at createAnonymousTypeNode (c:\dev\clients\communitiesuk\uk_gov_dash_components\node_modules\typescript\lib\typescript.js:52702:36)

Error generating metadata in uk_gov_dash_components (status=1)
````

Jordan-Hall:
@T4rk1n  I can take a look soon.  I assume its you that will be dealing with TS complier at plotly

T4rk1n:
[...]

PR #2088, opened 2022-06-10
WIP: feat: add better handing for typescript complier

Jordan-Hall:
fix: soft file there is no props defined

Check if export is either a class or  function that returns JSX element
This allows for none exported JSX functions in the file.

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals

### DIFF
--- a/dash/extract-meta.js
+++ b/dash/extract-meta.js
@@ -27,7 +27,7 @@ function help() {
     console.error('usage: ');
     console.error(
         'extract-meta ^fileIgnorePattern ^forbidden$|^props$|^patterns$' +
-            ' path/to/component(s) [path/to/more/component(s) ...] > metadata.json'
+        ' path/to/component(s) [path/to/more/component(s) ...] > metadata.json'
     );
 }
 
@@ -93,7 +93,7 @@ function isReservedPropName(propName) {
         if (reservedPattern.test(propName)) {
             process.stderr.write(
                 `\nERROR: "${propName}" matches reserved word ` +
-                    `pattern: ${reservedPattern.toString()}\n`
+                `pattern: ${reservedPattern.toString()}\n`
             );
             failedBuild = true;
         }
@@ -113,9 +113,14 @@ function checkDocstring(name, value) {
 function docstringWarning(doc) {
     checkDocstring(doc.displayName, doc.description);
 
-    Object.entries(doc.props).forEach(([name, p]) =>
-        checkDocstring(`${doc.displayName}.${name}`, p.description)
-    );
+    if (doc && doc.props) {
+        Object.entries(doc.props).forEach(([name, p]) =>
+            checkDocstring(`${doc.displayName}.${name}`, p.description)
+        );
+    } else {
+        // Soft fail
+        console.warn(`${doc.displayName} does not have any props`)
+    }
 }
 
 function zipArrays(...arrays) {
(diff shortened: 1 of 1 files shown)


Jordan-Hall:
@T4rk1n before i carry on and add Unit test for these. This is what I was thinking to solve https://github.com/plotly/dash/issues/2066#issuecomment-1150916950. The error comes from not checking the type is related to JSX before trying to work out the props. 

````
export const defaultProps = {
   address: "abc...."
}
````

would cause the complier to break. So while i was at it, i thought its best to check the return type is an element. If this is ok, I'll take a look at how to add support for multiple components in a single file. Currently you always override the docs each time which isn't ideal

File: dash/extract-meta.js [hunk 1]
```diff
@@ -738,12 +760,14 @@ function gatherComponents(sources, components = {}) {
                     .join('');
             }
             const doc = {
-                displayName: name,
+                displayName: `${name} - ${rootExp.name}`,
```
Jordan-Hall, line 763: Added the export name to the display name - Reasoning for this is for better debugging handling by end user. Especially if they exporting multiple things here. It was quiet hard to know what was failing before. Plus  I believe it will be needed to support multiple export in same file

Jordan-Hall:
The failed unit test relates to JSX just returning null. In real world I can't ever expect it not to be Element. But happy to remove that extra check for JSX

T4rk1n:
> @T4rk1n before i carry on and add Unit test for these. This is what I was thinking to solve [#2066 (comment)](https://github.com/plotly/dash/issues/2066#issuecomment-1150916950). The error comes from not checking the type is related to JSX before trying to work out the props.
> 
> ```
> export const defaultProps = {
>    address: "abc...."
> }
> ```
> 
> would cause the complier to break. So while i was at it, i thought its best to check the return type is an element. If this is ok, I'll take a look at how to add support for multiple components in a single file. Currently you always override the docs each time which isn't ideal

> The failed unit test relates to JSX just returning null. In real world I can't ever expect it not to be Element. But happy to remove that extra check for JSX

Seems ok, just make sure that `null` is also a valid return type, some components don't need to render anything (eg: dcc.Store) and thus return `null`.

gvwilson:
@T4rk1n can you please take another look at this one in cycle 17? thanks - @gvwilson

Jordan-Hall:
I'm happy to take a look and make any changes

@plotly/dash-generator-test-component-typescript/src/index.ts
```diff
@@ -3,11 +3,12 @@ import TypeScriptClassComponent from './components/TypeScriptClassComponent';
 import MemoTypeScriptComponent from './components/MemoTypeScriptComponent';
 import StandardComponent from './components/StandardComponent.react';
 import WrappedHTML from './components/WrappedHTML';
-
+import ExportWithObject from './components/ExportWithObject';
```
T4rk1n, line 6: Just need a space between the import and exports

dash/extract-meta.js
```diff
@@ -702,6 +708,23 @@ function gatherComponents(sources, components = {}) {
                 }
             }
 
+            const isArrowFunction = typeSymbol.declarations && typeSymbol.declarations[0].kind === ts.SyntaxKind.ArrowFunction
+            const isClass = ts.isClassDeclaration(declaration);
+
+            if (!isArrowFunction && !isClass) {
```
T4rk1n, line 714: I think it could be a regular `function(props: Props) {return <div></div>;` not only an arrow function.

Jordan-Hall: Good spot I'll make the changes tonight

PR #2100, opened 2022-06-19
Fix Pages not_found_404 parsing

jacobswe:
When initializing the Dash object using `use_pages=True` and `pages_folder=""` you cannot pass a `dash.register_page("not_found_404", layout=CUSTOM_404_LAYOUT)`. Line 2245 joins on a pages_folder and 404 path making the searched module name ".not_found_404". This small edit fixes the error and allows for passing a 404 page using module name "not_found_404" in the case of pages_folder="". Currently, you can (outside documentation) pass ".not_found_404" with a custom layout.

## Contributor Checklist

   -  [x] Fix filepath parsing in pages 404

### optionals


jacobswe:
I have been migrating my app to 3.5 and, in using the new pages plugin, I manually registered pages after doing some set up (hence the page_folder=""). I noticed couldn't pass the not_found_404 and get it to register correctly, but drilling into the source code I found this would fix the issue, and that using ".not_found_404" as the module name was a working alternative. Opening this to offer a solution (if it's not purposeful behavior) or alternatively that this would be good to document (if it is purposeful behavior).

alexcjohnson:
Nice catch @jacobswe! Would you be able to add this case to a test? Perhaps [`pala001`](https://github.com/plotly/dash/blob/ac7b37d0a40ffd7e48780e3c84f645c0554eb26f/tests/integration/multi_page/test_pages_layout.py#L51) via [`get_app`](https://github.com/plotly/dash/blob/ac7b37d0a40ffd7e48780e3c84f645c0554eb26f/tests/integration/multi_page/test_pages_layout.py#L7)?

jacobswe:
Yep! Will write this and update PR over the weekend.

alexcjohnson (approved):
💃 Looks great, thanks @jacobswe and @AnnMarieW!

jacobswe:
Thanks for this, I was having dev environment issues and have been traveling for work.

PR #2102, opened 2022-06-21
fixed pages to exclude files starting with `.`

AnnMarieW:
- Fix bug as reported in [dash-labs 113](https://github.com/plotly/dash-labs/issues/113) where files starting with `.` were not excluded when building `dash.page_registry`.
- Removed `usage-pages` file 

### optionals


tests/integration/multi_page/pages/.page1.py
```diff
@@ -0,0 +1,21 @@
+# This page should not be imported since it start with `.`
+
+import dash
+from dash import html, Input, Output, callback
+
+
+dash.register_page(__name__, id="page1")
```
alexcjohnson, line 7: So the way we know this file isn't imported is that `page1` is already taken, and duplicating it would cause an error? I suppose that works, though it would be more obvious if this file directly threw an error, like:
```py
raise Exception("files starting with . should not be imported")
```

alexcjohnson (approved):
Lovely! 💃

charmeem:
Like to know in which release this fix will be included.

charmeem:
I have upgraded to 2.6.0 but this issues does not seems to be resolved.

PR #2110, opened 2022-06-29
Dropdown component label search

### LINKED ISSUE #2108: [BUG] Styling Components as Option Labels Destroys Search
When using the example from in [documentation](https://dash.plotly.com/dash-core-components/dropdown#styling-components-as-option-labels). The search functions does not work.

```
dash                      2.5.1
dash-bootstrap-components 1.1.0
dash-extensions           0.1.3
```

**Describe the bug**

When using the example from in [documentation](https://dash.plotly.com/dash-core-components/dropdown#styling-components-as-option-labels). The search functions does not work.

**Expected behavior**

I expect that search should work even though the label is wrapped in a div.

T4rk1n:
Add `search` prop for the options of `dcc.Dropdown`. Fix #2108 

- Use if the `label` is a component or provide a custom search value different from the label.
- If no search value and the label is a component, the `value` will be used for search.

### DIFF
--- a/components/dash-core-components/src/fragments/Dropdown.react.js
+++ b/components/dash-core-components/src/fragments/Dropdown.react.js
@@ -1,4 +1,4 @@
-import {isNil, pluck, without, pick} from 'ramda';
+import {isNil, pluck, without, pick, head} from 'ramda';
 import React, {useState, useCallback, useEffect, useMemo} from 'react';
 import ReactDropdown from 'react-virtualized-select';
 import createFilterOptions from 'react-select-fast-filter-options';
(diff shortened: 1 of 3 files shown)


components/dash-core-components/src/fragments/Dropdown.react.js
```diff
@@ -47,12 +47,26 @@ const Dropdown = props => {
     } = props;
     const [optionsCheck, setOptionsCheck] = useState(null);
     const [sanitizedOptions, filterOptions] = useMemo(() => {
-        const sanitized = sanitizeOptions(options);
+        let sanitized = sanitizeOptions(options);
+        const firstOption = sanitized ? head(sanitized) : null;
+        let labelKey = 'label';
+        if (firstOption && firstOption.search) {
+            labelKey = 'search';
+        } else if (firstOption && React.isValidElement(firstOption.label)) {
```
alexcjohnson, line 55: I can see people using strings for some options and components for others - like if they want a few options to have custom font color but others just use the default behavior. So just looking at the first one may not be robust. Also wondering if there's a way to pull the text out of a component for the same use case, particularly when the value is encoded in a way that isn't useful for searching (matching names to ID numbers or something). But it could still be nice to search on value too, especially if the search would match EITHER label or value. Is that possible?

T4rk1n: > But it could still be nice to search on value too, especially if the search would match EITHER label or value. Is that possible?

It's possible to add field indexes to the search, maybe we should add a new prop `search_indexes`, then it could match any field in a dictionary of options ?

alexcjohnson (approved):
💃

Issue #1978, opened 2022-03-18
[BUG]


ned2:
**Describe the bug**

When a `DataTable` is initialised with `cell_selectable=False`, text cannot be copied to the clipboard. This is independent of the issue in #1975 where cells in columns with `presentation` value of `input` do not appear to be visually selected. The text can actually be selected, however when attempting to copy the selection, nothing is added to the clipboard.

**Expected behavior**
When `cell_selectable` is false, text selected inside DataTable cells should be able to be copied to the clipboard.

**Minimal Example**
Text in any cell (regardless of `presentation`) cannot be copied to the clipboard.

```python
import dash
import pandas as pd
from dash import html
from dash.dash_table import DataTable

url = "https://github.com/plotly/datasets/raw/master/26k-consumer-complaints.csv"
df = pd.read_csv(url, nrows=100)
columns = [{"name": col, "id": col, "selectable": True} for col in df.columns]
columns[2]["presentation"] = "markdown"

app = dash.Dash(__name__)
app.layout = html.Div(
    [
        DataTable(
            columns=columns,
            data=df.to_dict("records"),
            cell_selectable=False,
        )
    ]
)

if __name__ == "__main__":
    app.run_server(debug=True, port=8050)
```  

**Environment**
```
dash                          2.3.0
dash-bootstrap-components     1.0.3
dash-core-components          2.0.0
dash-html-components          2.0.0
dash-table                    5.0.0

```

    - OS: Ubuntu 21.10
    - Browser Chrome
    - Version 99.0.4844.51

ned2:
I did a bit of digging and it looks like the issue is that the `onCopy` event handler in https://github.com/plotly/dash/blob/464a743881a002bb7ae9d57214c1ff6abb970905/components/dash-table/src/dash-table/components/ControlledTable/index.tsx makes the assumption that the DataTable is in `cell_selectable=True` mode, as it always defers to `TableClipboardHelper.toClipboard`, which takes data from its `selectedCells` param, however this will always be empty when `cell_selectable=False`.

I'm wondering if what's needed is for the copy event handling logic to opt-out of using the DataTable's copying machinery and defer to the browser's default handling?

@alexcjohnson is this something we could address with a relatively small change? If so I'd be up for trying to put together a PR. Or is this opening a broader question that needs untangling first?

alexcjohnson:
[...]

PR #2114, opened 2022-07-01
Enable copying text to clipboard for tables with `cell_selectable=False`

benedikt-budig:
This PR makes sure that for DataTables with `cell_selectable=False`, the _text content_ of a cell can be selected and copied to the clipboard as is normally handled by the browser.

It fixes issue https://github.com/plotly/dash/issues/1978.

## Contributor Checklist

### optionals

### DIFF
--- a/components/dash-table/src/dash-table/components/ControlledTable/index.tsx
+++ b/components/dash-table/src/dash-table/components/ControlledTable/index.tsx
@@ -794,14 +794,17 @@ export default class ControlledTable extends PureComponent<ControlledTableProps>
             include_headers_on_copy_paste
         } = this.props;
 
-        TableClipboardHelper.toClipboard(
-            e,
-            selected_cells,
-            columns,
-            visibleColumns,
-            viewport.data,
-            include_headers_on_copy_paste
-        );
+        // if no cells are selected, fall back to the browser's default copy event handling
+        if (selected_cells.length) {
+            TableClipboardHelper.toClipboard(
+                e,
+                selected_cells,
+                columns,
+                visibleColumns,
+                viewport.data,
+                include_headers_on_copy_paste
+            );
+        }
         this.$el.focus();
     };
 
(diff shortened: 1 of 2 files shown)


benedikt-budig:
the `percy/dash` action seems to be failing because the snapshot still has a datepicker that shows June (and not July). Are the baseline screenshots updated automatically after some time?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [UNRELEASED]
 
 ### Fixed
+- [#2114](https://github.com/plotly/dash/pull/2114) Fix bug [#2095](https://github.com/plotly/dash/issues/2095) where text could not be copied from cells in tables with `cell_selectable=False`.
```
alexcjohnson, line 8:
```suggestion
- [#2114](https://github.com/plotly/dash/pull/2114) Fix bug [#1978](https://github.com/plotly/dash/issues/1978) where text could not be copied from cells in tables with `cell_selectable=False`.
```

alexcjohnson:
@benedikt-budig thanks for the PR! Looks great, and don't worry about Percy, I'll approve those when we're ready.

The one thing we should do is add a test that shows that the bug is fixed. This should be a good starting point (and this test I believe verifies that the normal `cell_selectable=True` copy still works correctly with your fix?):

https://github.com/plotly/dash/blob/565de531e77e99b339bab7e0838ac51a4a02511a/components/dash-table/tests/selenium/test_basic_copy_paste.py#L154-L174

benedikt-budig:
@alexcjohnson thanks a lot for your feedback! In the meantime, I have:
- added a unit test covering the fixed issue, and
- slightly extended `Contents` to add a new CSS class to table cells indicating whether they are selectable or not.

However, there are now some `long_callback` tests failing. It seems to me that these tests are operating on the edge of the 10 seconds timeout and I'm not sure if my changes have actually something to do with them failing. Could you help me once again?

alexcjohnson (approved):
💃 Beautiful - great call adding the `selectable` class so we can keep `background-color: transparent;` in that case. And nice job on the test!

Will merge once tests pass, let's hope merging dev will fix them 🤞

benedikt-budig:
@alexcjohnson @T4rk1n  awesome, thanks for your help and for merging this! 🎉

PR #2116, opened 2022-07-04
Rename long callbacks to background.

T4rk1n:
- Rename `long` argument to `background`
- Remove `long` prefixes.
- Deprecate
  - `dash.long_callback.managers.CeleryLongCallbackManager` -> `dash.CeleryManager`
  - `dash.long_callback.managers.DiskcacheLongCallbackManager` -> `dash.DiskcacheManager`
  - `long_callback_manager` dash constructor argument -> `background_callback_manager`

### DIFF
--- a/dash/long_callback/managers/celery_manager.py
+++ b/dash/long_callback/managers/celery_manager.py
@@ -12,8 +12,8 @@ from dash.exceptions import PreventUpdate
 from dash.long_callback.managers import BaseLongCallbackManager
 
 
-class CeleryLongCallbackManager(BaseLongCallbackManager):
-    """Deprecated: use `import CeleryManager from dash` instead."""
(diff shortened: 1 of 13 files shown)


dash/long_callback/managers/celery_manager.py
```diff
@@ -13,6 +13,8 @@
 
 
 class CeleryLongCallbackManager(BaseLongCallbackManager):
+    """Deprecated: use `import CeleryManager from dash` instead."""
```
alexcjohnson, line 16: Can we give the primary class here the name we want (`CeleryManager`) and make `CeleryLongCallbackManager` into the subclass, defined right in this file? That would make it easier for people to read the code going forward, by removing a layer of indirection from the recommended usage.

Also the deprecation note should say `from dash import ...` not `import ... from dash` right?

PR #2146, opened 2022-07-18
No console log

T4rk1n:
- Remove leftover debug console log
- Add eslint rule to error on console.log for the renderer.


alexcjohnson (approved):
💃

PR #2152, opened 2022-07-20
Fix component as prop with multiple components in list of objects

### LINKED ISSUE #2128: [BUG] No rendering if two component as props inside a dictionary.
When defining props on component, if you have two props that can be components inside a list of dictionaries, the components are not rendered and a react error indicate objects are invalid.

Example props:
```tsx
type Props = {
   fields: {
      label?: JSX.Element;
      input_component?: JSX.Element;
   }[];
}
```

T4rk1n:
- Fix #2128


tests/integration/renderer/test_component_as_prop.py
```diff
@@ -90,6 +90,16 @@ def test_rdcap001_component_as_prop(dash_duo):
...
                 },
             ),
+            ComponentAsProp(
+                id="multi-component",
+                multi_components=[
+                    {
+                        "id": "multi",
+                        "first": Span("first"),
+                        "second": Span("second"),
+                    }
```
alexcjohnson, line 100: For completeness perhaps give this array a second element too?

alexcjohnson (approved):
💃

PR #2167, opened 2022-07-25
Improve background callback test stability

T4rk1n:
- Fix failure to start the threaded server.
- Update dependencies

There is still a timeout error due to zombie processes created by diskcache background callbacks.


alexcjohnson (approved):
💃

PR #2182, opened 2022-08-04
fixed validation_layout in pages

### LINKED ISSUE #2172: [Feature Request] Make app validation optional when using `pages`
When using `pages` to make a multi page app, one of the nice features is that it automatically validates the layout.  However as pointed out in [this forum post](https://community.plotly.com/t/examples-of-multi-page-apps-with-dash-pages/66489/4?u=annmariew) it can make the initial load of an app slow if some of the layout functions are slow.

A possible solution is to make the validation optional, then [this section](https://github.com/plotly/dash/blob/dev/dash/dash.py#L2078:L2092) would be skipped.  (I tried it and it seems to work)

Validation could be turned off like this:

```
app = das.Dash(__name__, use_pages=True, set_validation_layout=False,  suppress_callback_exceptions=True)
```

or maybe if `set_validation_layout=False`, it could automatically set `suppress_callback_exceptions=True`

Any downside to doing this?

AnnMarieW:
Closes #2172 

Makes it so that when using `pages`,  if `suppress_callback_exceptions=True` the `validation_layout` is not set.


alexcjohnson (approved):
💃 Looks great - and yeah, cleaner to do this without the before_(first_)request part.

Please add a changelog note, then we'll merge 🎉

CHANGELOG.md
```diff
@@ -14,7 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-- [#2175](https://github.com/plotly/dash/pull/2175) Fix [#2173](https://github.com/plotly/dash/issues/2173) callback output of ndarray and no_update check.
+- [#2175](https://github.com/plotly/dash/pull/2175) Fix [#2172](https://github.com/plotly/dash/issues/2172) callback output of ndarray and no_update check.
```
alexcjohnson, line 18: Unintentional change?

PR #2188, opened 2022-08-09
Fix prepend/extends graph tests

### LINKED ISSUE #2181: [BUG] Tests failing due to Chrome 104
Something funny is going on with async chunks in the latest version of Chrome. It's causing tests `grva004` and `grva005` (about extending/prepending data to traces) to fail. This is a fairly unusual edge case so the problems may be very narrow, but needs to be investigated ASAP to confirm how big an issue this is.

@T4rk1n suspects [Feature: Subresource loading with Web Bundles](https://chromestatus.com/feature/5710618575241216)

T4rk1n:
Replace graph prepend/extend trace test assertions with DOM nodes graph inspection. Resolve #2181


components/dash-core-components/tests/integration/graph/test_graph_varia.py
```diff
@@ -193,6 +194,37 @@ def num_traces():
...
+    )
+
+    traces = browser.find_elements(f"{graph_id} .points")
+
+    for trace in traces:
+        elements = trace.find_elements(By.CSS_SELECTOR, "path")
+        for element in elements:
+            ActionChains(browser.driver).move_to_element(element).perform()
+            hover = browser.find_element(f"{graph_id} .hoverlayer")
+            content = hover.text
```
alexcjohnson, line 210: That'll work, though perhaps an easier solution (and more robust against changes in plotly.js implementation details) would be to write a little JS to look at `graphDiv._fullData[i].(x|y)`, where `graphDiv` is `document.querySelector("#{graph_id} .js-plotly-plot")`

T4rk1n:
@alexcjohnson 
The thread killing doesn't work properly with flask 2.2.x, replaced the default runner with a multiprocess based runner. Not sure if we should replace `ProcessRunner` with the new one?

dash/testing/application_runners.py
```diff
@@ -206,6 +207,59 @@ def stop(self):
...
+                app.run(threaded=True, **options)
+            except SystemExit:
+                logger.info("Server stopped")
+                raise
+            except Exception as error:
+                logger.exception(error)
+                raise error
+
+        # pylint: disable=import-outside-toplevel,no-member
+        import multiprocess as mp
```
alexcjohnson, line 235: Why does this import need to be inside here?

T4rk1n: Moved :ok_hand:

alexcjohnson (approved):
💃

components/dash-core-components/tests/integration/graph/test_graph_varia.py
alexcjohnson: As discussed elsewhere, we can leave this for now, I may come back to it later myself.

PR #2194, opened 2022-08-15
Fix unordered background callback inputs & state.

### LINKED ISSUE #2190: [BUG] Order of Inputs, Outputs and States in long_callbacks causing problems
An issue I had was [described as a potential bug on community.plotly](https://community.plotly.com/t/issues-switching-from-callback-to-long-callback/66882/4) and it was recommended I share here.

The following code runs properly:

```
import webbrowser
import diskcache

from threading import Timer
from dash import Dash, dcc, html, Input, Output, State
from dash.long_callback import DiskcacheLongCallbackManager

cache=diskcache.Cache("./cache")
long_callback_manager=DiskcacheLongCallbackManager(cache)

list_of_tables=[['Generic Table',['Table Columns']]]

app = Dash(__name__)

app.layout = html.Div([
    dcc.Checklist([table[0] for table in list_of_tables],id='table checklist'),
    html.H3(children='Run simulation: '),
    html.Button(id='run simulation button',n_clicks=0,children='Run'),
    html.Div(id='my-output'),
])

@app.callback(
    Output('my-output', 'children'),
    State('table checklist','value'),
    Input('run simulation button', 'n_clicks'),
    prevent_initial_call=True
    )
def test(tables_to_compute,n_clicks):

    if 'Generic Table' in tables_to_compute:
        string='Generic table is checked'
    else:
        string='Generic table is not checked'

    return string

if __name__ == '__main__':
    Timer(1, webbrowser.open('http://127.0.0.1:8054/')).start();
    app.run_server(debug=True,use_reloader=False,port=8054)
```
but when I change the callback decorator to
```
@app.long_callback(
    Output('my-output', 'children'),
    State('table checklist','value'),
    Input('run simulation button', 'n_clicks'),
    manager=long_callback_manager,
    prevent_initial_call=True
    )
```
an error occurs, telling me "TypeError: argument of type ‘int’ is not iterable". Here is the reply from community.plotly:

"Thanks for the reproducible example!

My suspicion before running was that it was mixing the order between Input and State and assigning the click count to tables_to_compute (therefore your error). I can confirm that this is exactly what happened and the error disappears if you switch the order in the decorator and in the function signature.

I think you can consider this a bug, although it is a good practice to have the Input before State in callbacks… I would recommend you to open an issue on GH to report it just in case."

T4rk1n:
Fix #2190

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -362,7 +362,7 @@ def register_callback(  # pylint: disable=R0914
                     job = callback_manager.call_job_fn(
                         cache_key,
                         job_fn,
-                        args,
+                        func_args if func_args else func_kwargs,
                         AttributeDict(
                             args_grouping=callback_ctx.args_grouping,
                             using_args_grouping=callback_ctx.using_args_grouping,
--- a/tests/integration/long_callback/app_unordered.py
+++ b/tests/integration/long_callback/app_unordered.py
@@ -1,0 +1,31 @@
+from dash import Dash, Input, Output, dcc, State, html, callback
+
+from tests.integration.long_callback.utils import get_long_callback_manager
+
+long_callback_manager = get_long_callback_manager()
+handle = long_callback_manager.handle
+
+app = Dash(__name__, long_callback_manager=long_callback_manager)
+
+app.layout = html.Div(
+    [
+        html.Div(id="output"),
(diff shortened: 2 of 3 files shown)


alexcjohnson (approved):
💃

Issue #2185, opened 2022-08-08
[BUG]: Dash-Table unexpected behavior when copying and pasting double quotes


amy-morrill:
Hi Dash team! I'm hoping to get some help with a bug I found in dash-table. 

**The Bug:** In my application, users copy and paste data into a dash-table, and some of this data includes two consecutive double quotes. However, when copied and pasted into the table, the consecutive double quotes are automatically changed to be one double quote (e.g.: a""b becomes a"b). This is especially hard to fix because some data also contains one double quote (e.g.: a"b could be a valid input). 

**The Expected Behavior:** Strings copied and pasted into dash-table cells should not be changed i.e. any consecutive double quotes in a string should not be converted to one double quote.

**My Context:**
```
dash                 2.0.0
dash-core-components 2.0.0
dash-html-components 2.0.0
dash-table           5.0.0
```
 - OS: macOS Monterey; version 12.4
- Browser: Google Chrome; version 103.0.5060.134

Thank you for your help!! I also would be happy to try to contribute to solving this, although I'm a bit unsure of where to start.

alexcjohnson:
Thanks for the report @amy-morrill - this is strange behavior indeed!

It looks like it's coming from [sheetclip](https://www.npmjs.com/package/sheetclip) - specifically `SheetClip.prototype.parse` here:

https://github.com/plotly/dash/blob/e746bdbbac5a3581a6cac7ba0431ef03bf7859b5/components/dash-table/src/dash-table/utils/TableClipboardHelper.ts#L101

[...]

PR #2202, opened 2022-08-22
Fixed double quotes copy and paste bug

amy-morrill:
Addresses #2185
I created a new parse function based on the previously used [SheetClip.parse function](https://github.com/warpech/sheetclip/blob/master/sheetclip.js#L21), removing the parts which replace consecutive double quotes with one double quote. Moreover I added a test in components/dash-table/tests/selenium/test_basic_copy_paste.py which verifies that consecutive double quotes are un-modified when copied and pasted into a dash-table. 

## Contributor Checklist

   -  [X] Replace SheetClip.parse with inline function fixing bug
   -  [X] Add a test to verify that bug is fixed

### optionals


components/dash-table/src/dash-table/utils/TableClipboardHelper.ts
```diff
@@ -111,4 +111,13 @@ export default class TableClipboardHelper {
             overflowRows
         );
     }
+
+    private static parse(str: string) {
+        const temprows = str.split('\n');
+        if (temprows.length > 1 && temprows[temprows.length - 1] === '') {
+            temprows.pop();
+        }
+        const rows = temprows.map(row => row.split('\t'));
```
alexcjohnson, line 120: Thanks for tracking down the original [`SheetClip.parse`](https://github.com/warpech/sheetclip/blob/f6fc218347e37ad6e630e01c757a83e2c4055445/sheetclip.js#L21). I'm a little worried that this simple replacement - while fixing the problem you encountered with multiple quotes within a cell - will cause other problems due to extra cases that were covered by `SheetClip` and missed here. I'm thinking particularly about cells that have tab or newline characters in them, which presumably are the whole reason for all that monkey business with quotes, so if a cell starts and ends with quote characters it can have tabs and newlines inside it and these don't cause the parser to move to the next column or row.

Then of course we could ask what happens if there are ALSO quote characters inside the cell text. I'd have guessed these would be escaped (with a backslash before the quote) but I see no code in the original `SheetClip.parse` to handle that so perhaps the quote character is doubled? That would explain their "replace two quotes with one" step, it's just applied too broadly.

Anyway I think the goal of `SheetClip` as expressed in its [head comment](https://github.com/warpech/sheetclip/blob/f6fc218347e37ad6e630e01c757a83e2c4055445/sheetclip.js#L5) is our goal as well:
> This tiny library transforms JavaScript arrays to strings that are pasteable by LibreOffice, OpenOffice, Google Docs and Microsoft Excel.

Meaning we want to be able to copy and paste rectangular sections of dash table to and from common spreadsheet programs, as well as to other dash tables, regardless of what text is in those cells. So I think this requires a little more investigation about how such cells are encoded by spreadsheets - and it may mean we have more work to do on the copy side as well as on paste.

amy-morrill: That makes a lot of sense -- I was wondering if this solution was too simplistic. 

One thought I have (which may also be too simplistic) is that when parsing pasted cells we could first "escape" quotes by adding a backslash before each quote and then keep the SheetClip.parse function the same except for changing `.replace(/""/g, '"')` steps to something like `.replace('\"', '"')`. Does that general idea sound okay to you or am I oversimplifying?

Regardless, I'm happy to keep looking at this. Thank you again for your help!

components/dash-table/tests/selenium/test_basic_copy_paste.py
```diff
@@ -322,15 +348,75 @@ def test_tbcp010_copy_from_unselectable_cells_table(test):
     source.cell(2, 2).double_click()
     assert source.cell(2, 2).get_text() == test.get_selected_text()
 
-    # copy the source text to clipboard using CTRL+C
+    # copy the source text to clipboard using CTRL+C or COMMAND+C
     test.copy()
+    # Uncomment the following two lines if using a Mac computer
+    # with test.hold(Keys.COMMAND):
+    #     ActionChains(test.driver).send_keys("c").perform()
```
alexcjohnson, line 355: Oh interesting, I thought I had run these tests before on my Mac but I guess not because you're right, they fail. The better way to do this would be to modify `test.paste` etc so they use `Keys.COMMAND` on Mac and `Keys.CONTROL` on others.

In fact we do exactly that in one place (which has an older pattern in other ways):

https://github.com/plotly/dash/blob/29a389245c8c44802d55ba847b6c2e54a28fe2eb/components/dash-table/tests/selenium/conftest.py#L271

We just need to do the same in copy and paste:

https://github.com/plotly/dash/blob/29a389245c8c44802d55ba847b6c2e54a28fe2eb/components/dash-table/tests/selenium/conftest.py#L520-L526

The cleanest way to do this would be to pull the `CMD =` line up to the top of the module and then use `CMD` in copy & paste.

components/dash-table/src/dash-table/utils/TableClipboardHelper.ts
```diff
@@ -85,6 +85,7 @@ export default class TableClipboardHelper {
         includeHeaders: boolean
     ): {data: Data; columns: Columns} | void {
         const text = Clipboard.get(ev);
+        console.log(text);
```
alexcjohnson, line 88: 🔪 Let's take this debugging statement out before merging

alexcjohnson (approved):
💃 Nicely done! The quote-doubling behavior is funny, must be some extremely old artifact from Excel... nice job tracking that down!

Ready to go after my last two comments.

components/dash-table/src/dash-table/utils/TableClipboardHelper.ts
amy-morrill: Fixed!

alexcjohnson:
@amy-morrill a number of bug fixes have landed so it's about time for us to make a patch release. There are a couple of other loose ends we're trying to tie up but likely end of this week or early next we'll make a new release.

PR #2133, opened 2022-07-14
Experimental: use microbundled react-cytoscapejs

akx:
As requested by @alexcjohnson in https://github.com/plotly/react-cytoscapejs/pull/99#issuecomment-1182411086 , this pulls in the https://github.com/plotly/react-cytoscapejs/pull/99 version of react-cytoscapejs and adds cytoscape as a direct dependency (as it is a peer dependency of react-cytoscapejs going forward) for dash-renderer's dev tools.

* 👉  This PR should _not_ be merged as-is, though, since the package.json specifier for react-cytoscapejs is not a published version

## Contributor Checklist

- [x] I have run **npm run test in `dash/dash-renderer` locally and it passed with flying colors.
- [x] I have run manual tests with this version of dash linked in to a downstream project:

I tried things out in my `dash-cytoscape` https://github.com/plotly/dash-cytoscape/pull/180 branch with

```
pip install -e ../dash
pip install -e ../dash/components/dash-table/
pip install -e ../dash/components/dash-html-components/
pip install -e ../dash/components/dash-core-components/
```
and e.g. the `python usage-responsive-graph.py` demo shows the callback graph just fine.

PR #2217, opened 2022-09-02
react-cytoscapejs v2 and lint updates

alexcjohnson:
Use new `react-cytoscapejs@2.0.0` - from https://github.com/plotly/react-cytoscapejs/pull/99 (cc @akx)
Supersedes #2133

@T4rk1n unclear why I need the lint updates here, yet CI was passing?

(probably doesn't need a changelog entry, as this has no user-facing consequences)


T4rk1n:
> @T4rk1n unclear why I need the lint updates here, yet CI was passing?

The linting for black seems to be broken:

```
> private::lint.black
> if [[ ${PYVERSION:-python39} != python36 ]]; then black dash tests --exclude metadata_test.py --check; fi

sh: 1: [[: not found
```

T4rk1n (approved):
:woman_dancing:

PR #2223, opened 2022-09-09
exclude dirs in the pages folder that start with . or _

AnnMarieW:
Fixes bug as [reported on the forum](https://community.plotly.com/t/dash-pages-ignore-hidden-files-and-folders/67819) where hidden directories in the `pages` folder were being imported.

### DIFF
--- a/tests/integration/multi_page/pages/_no_import.py
+++ b/tests/integration/multi_page/pages/_no_import.py
@@ -1,3 +1,5 @@
 from dash import register_page
 
+register_page(__name__)
+
 raise Exception("files starting with _ should not be imported")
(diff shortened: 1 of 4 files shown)


tests/integration/multi_page/pages/_no_import.py
```diff
@@ -1 +1,5 @@
+from dash import register_page
+
+register_page(__name__)
```
alexcjohnson, line 3: Oh good catch, it wouldn't execute this file without `register_page`, would it!

alexcjohnson (approved):
💃 Nicely done 🎉

PR #2226, opened 2022-09-12
Fix background callbacks & register page

### LINKED ISSUE #2193: [BUG] Import from a page without application (For testing purpose)
When `register_page` is called during import, if there is no app defined before it will throw a bare `Exception` with a message indicating there was no app  before the import. https://github.com/plotly/dash/blob/20b6467bf61024aba75fcdac60ee9952363355c4/dash/_validate.py#L465

This makes it hard to ignore that particular exception in the case of unit testing callbacks defined within a page file as [reported on the forums](https://community.plotly.com/t/dash-2-6-released-background-callbacks-unit-testing-persistent-selections-dropdown-features/66322/3) and should be changed to a custom exception.

### LINKED ISSUE #2196: [Feature Request] Handle import exceptions in pages apps
In a large pages app (several dozen unrelated pages coded by different individuals) it would be helpful to not let an import exception in one page prevent the entire site from loading.  We have hit cases where one developer has made assumptions that work in their local dev/test env but don't in production.  It seems like `_import_layouts_from_pages()` could catch exceptions on import, print/log a message indicating which page is failing, and continue loading other pages.  

A tangiential use case would be intentional prevention of page loading by simply throwing an import error, something like:

```
if app_enabled(__name__):
    dash.register_page(__name__, "/page1")
else:
    raise ImportError(f"{__name__} is currently not enabled")
``` 

Just guarding register_page is enough to preclude access but still loads unneeded resources from the file which an exception would prevent.  My developers really only want to load the page(s) they're working on and ignore dozens of others for rapid test cycles.  I'm sure there are many other solutions for conditional page import, this just seemed like one path requiring very little dash support aside from handling import expections (that doesn't involve moving/renaming files).  

I can help with a PR if there is agreement on a path forward.  Thanks.

### LINKED ISSUE #2219: [BUG] Background callbacks not working with pages
**Describe your context**
Dash Callback Example 4: Progress Bar from https://dash.plotly.com/background-callbacks

```
dash                      2.6.1                                                                                         
dash-bootstrap-components 1.2.1                                                                                        
dash-daq                  0.5.0                                                                                         
```
    - OS: Windows
    - Browser Chrome
    - Version 104.0.5112.81

**Describe the bug**

using background callbacks does not work with pages.

not using pages the code runs correctly

app.py
```

import os
import dash
from dash import DiskcacheManager, CeleryManager, html, page_container

# code to test without pages
# import pages.page as page

if 'REDIS_URL' in os.environ:
    # Use Redis & Celery if REDIS_URL set as an env variable
    from celery import Celery
    celery_app = Celery(__name__, broker=os.environ['REDIS_URL'], backend=os.environ['REDIS_URL'])
    background_callback_manager = CeleryManager(celery_app)

else:
    # Diskcache for non-production apps when developing locally
    import diskcache

app = dash.Dash(__name__,
                use_pages=True,
                # code to test without pages
                # use_pages=False,
                background_callback_manager=background_callback_manager)

app.layout = html.Div(
    page_container

    # code to test without pages
    # page.layout()
)

if __name__ == "__main__":
    app.run_server(debug=True)

```

pages/page.py
```

import dash
from dash import Input, Output, html
import time

dash.register_page(__name__, path='/page')


def layout():
    return html.Div([
        html.Div(
            [
                html.Progress(id="progress_bar", value="0"),
            ]
        ),
    ])


@dash.callback(
    background=True,
    running=[
        (
                {"visibility": "hidden"},
                {"visibility": "visible"},
        ),
        (
                {"visibility": "visible"},
                {"visibility": "hidden"},
        ),
    ],
    cancel=Input("cancel_button_id", "n_clicks"),
    prevent_initial_call=True
)
def update_progress(set_progress, n_clicks):
    total = 5
    for i in range(total + 1):
        set_progress((str(i), str(total)))
        time.sleep(1)

```

**Expected behavior**

Ability to use background callbacks AND pages

**Screenshots**

image

T4rk1n:
Fixes #2219, Fixes #2196, Fixes #2193


dash/dash.py
```diff
@@ -1208,6 +1208,8 @@ def dispatch(self):
             cb = self.callback_map[output]
             func = cb["callback"]
 
+            g.ignore_register_page = cb.get("long", False)
```
alexcjohnson, line 1211: Would it be possible to still catch `register_page` calls inside the callback functions, possibly by turning `ignore_register_page` back off again immediately before invoking the function itself, ie [here](https://github.com/plotly/dash/blob/648130685a9d4fee32e12459845997792d76b6f1/dash/long_callback/managers/celery_manager.py#L152-L157) and [here](https://github.com/plotly/dash/blob/648130685a9d4fee32e12459845997792d76b6f1/dash/long_callback/managers/diskcache_manager.py#L160-L165)?

T4rk1n: Good call :ok_hand:

dash/long_callback/managers/diskcache_manager.py
```diff
@@ -1,6 +1,9 @@
 import traceback
+from contextvars import copy_context
```
alexcjohnson, line 2: 3.7+ in stdlib?
https://pypi.org/project/contextvars/

alexcjohnson: oh it's already there, missed that.

https://github.com/plotly/dash/blob/648130685a9d4fee32e12459845997792d76b6f1/requires-install.txt#L8

alexcjohnson (approved):
💃 nice job!

PR #2237, opened 2022-09-19
Graph edit queue

alexcjohnson:
Plotly.js rendering in `dcc.Graph` can be async - more so now that we support MathJax in Dash, though there were always some async things like loading geojson. Anyway this hasn't caused any problems in core dash that I'm aware of, but in [dash-design-kit](https://plotly.com/dash/design-kit/), which wraps `dcc.Graph` and passes some information back and forth for theming purposes, `extendData` and `prependData` could cause two `render` cycles to overlap and drop the new data. This ensures the `plotly.js` method calls are sequenced properly, even if they end up delayed relative to the React renders. In my testing the changes here fix DDK (separate PR coming there) and it may fix some as-yet-undiscovered bugs in core DCC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


T4rk1n (approved):
Nice :woman_dancing:

PR #2238, opened 2022-09-19
Update LICENSE

GKI98:
Changes year of license 2021 to 2022

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2021 Plotly, Inc
+Copyright (c) 2022 Plotly, Inc
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


Kanootoko (approved):

alstn2468:
Wouldn't it be better to change all of these as well?

[@plotly/dash-component-plugins/LICENSE](https://github.com/plotly/dash/blob/e8ac94919105a91c76a966c21aca2ec7b0297e22/%40plotly/dash-component-plugins/LICENSE),
[@plotly/webpack-dash-dynamic-import/LICENSE](https://github.com/plotly/dash/blob/e8ac94919105a91c76a966c21aca2ec7b0297e22/%40plotly/webpack-dash-dynamic-import/LICENSE)

GKI98:
Got it. Updated

alstn2468 (approved):
~~LGTM~~

alstn2468 (commented):
Oh, I think there are a few more files that need to be modified.
I think it would be better to check the component folder as well.

GKI98:
Thanks, updated.

alexcjohnson:
Thanks @GKI98 @alstn2468 - we've actually discussed removing the year entirely, that's becoming more common, for example https://github.com/facebook/react/pull/13593, [Microsoft projects](https://github.com/microsoft/TypeScript/blob/main/CopyrightNotice.txt)... but since we haven't decided on that yet, thank you for updating these!

alexcjohnson (approved):

Issue #2255, opened 2022-09-30
Misleading doc string generation in typescript components


snehilvj:
Hi, I am migrating [dash-mantine-components](https://github.com/snehilvj/dash-mantine-components) to typescript. However, I have been facing some issues with the doc strings generation. The doc strings are not very accurate, in fact, for some prop types, doc string generation just fails. 

**Example 1**
Have a look at how the doc string is generated for the **_radius_** prop.

![ss1](https://aws1.discourse-cdn.com/business7/uploads/plot/original/3X/b/8/b86ac7d9e00fd202e1a4fe4c81cff7f3197b8ac8.png)

![ss2](https://aws1.discourse-cdn.com/business7/uploads/plot/original/3X/0/9/09d17793378ab5b6f8c1505497f825fc4cfecbe5.png)

The radius property will accept both strings and numbers but the doc string is misleading.

**Example 2**
For somewhat complex prop types such as _**value**_ here, the prop generation just fails with the below error.

![ss3](https://user-images.githubusercontent.com/91216500/193350954-9aeee90f-18c8-4919-a1a2-593251cf9586.png)

Error:
```
> dash_mantine_components@0.11.0a0 build:backends
> dash-generate-components ./src/ts/components dash_mantine_components -p package-info.json --r-prefix '' --jl-prefix '' --ignore \.test\.


ERROR: "__@iterator@2374" matches reserved word pattern: /^_.*$/

ERROR: "__@unscopables@2376" matches reserved word pattern: /^_.*$/
extract-meta failed

Error generating metadata in dash_mantine_components (status=1)
error: Recipe `build` failed on line 5 with exit code 1
```

Thanks
Snehil

T4rk1n:
Thanks for the report @snehilvj .

1. For values with enumeration, the implementation currently discard other types than the enumeration. Instead, when it encounter this case, it should transform the enum type into a union of the enum and the others types.
2. Adding tuple support in #2257

snehilvj:
Hi @T4rk1n Thanks for the fix on 2nd point above. Any thing going with the 1st point above? Btw loving the typescript experience so far. Migrated entire dmc to typescript in the latest 0.11.0a0 release.

tsveti22:
Hi @T4rk1n  
Currently experiencing the same issue with the dangerouslySetInnerHTML prop which includes an "__html" key. Interestingly, the component generation script runs successfully in windows, but we get this on Unix: 

```
[...]

PR #2257, opened 2022-10-03
Fix typescript tuple type.

T4rk1n:
Fix part 2 of #2255

### DIFF
--- a/dash/development/_py_components_generation.py
+++ b/dash/development/_py_components_generation.py
@@ -572,7 +572,7 @@ def map_js_to_py_types_prop_types(type_object, indent_num):
 
     def tuple_of():
         elements = [js_to_py_type(element) for element in type_object["elements"]]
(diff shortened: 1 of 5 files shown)


dash/development/_py_components_generation.py
```diff
@@ -570,6 +570,10 @@ def array_of():
             )
         return "list"
 
+    def tuple_of():
+        elements = [js_to_py_type(element) for element in type_object["elements"]]
+        return f"tuple of ({', '.join(elements)})"
```
alexcjohnson, line 575: The word `tuple` has a different meaning in Python than it has in TS, and it'll serialize as a list, so perhaps something like this?

```suggestion
        return f"list of {len(elements)} elements: [{', '.join(elements)}]"
```

alexcjohnson (approved):
💃 Excellent!

snehilvj:
💃

PR #2261, opened 2022-10-06
:sparkles: Custom filter placeholder text for data table

fraghag:
This has been asked for by multiple people in the community:
https://community.plotly.com/t/custom-text-in-datatable-filter-cell/34950/1

## Contributor Checklist

   -  [x] Add setting for custom data table column filter placeholder text
   -  [x] Update the docs

### optionals

### DIFF
--- a/components/dash-table/src/dash-table/dash/DataTable.js
+++ b/components/dash-table/src/dash-table/dash/DataTable.js
@@ -1,7 +1,7 @@
 import * as R from 'ramda';
-import React, {Component, Suspense} from 'react';
+import React, { Component, Suspense } from 'react';
 import PropTypes from 'prop-types';
-import {asyncDecorator} from '@plotly/dash-component-plugins';
+import { asyncDecorator } from '@plotly/dash-component-plugins';
 
 import LazyLoader from 'dash-table/LazyLoader';
 
@@ -313,13 +313,17 @@ export const propTypes = {
              * There are two `filter_options` props in the table.
              * This is the column-level filter_options prop and there is
              * also the table-level `filter_options` prop.
-             * These props determine whether the applicable filter relational
-             * operators will default to `sensitive` or `insensitive` comparison.
              * If the column-level `filter_options` prop is set it overrides
              * the table-level `filter_options` prop for that column.
              */
             filter_options: PropTypes.shape({
+                /**
+                 * (default: 'sensitive') Determine whether the applicable filter relational operators will default to `sensitive` or `insensitive` comparison.
+                 */
                 case: PropTypes.oneOf(['sensitive', 'insensitive']),
+                /**
+                 * (default: 'filter data...') The filter cell placeholder text.
+                 */
                 placeholder_text: PropTypes.string
             }),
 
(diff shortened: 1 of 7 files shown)


components/dash-table/tests/selenium/test_filter.py
```diff
@@ -301,3 +301,40 @@ def test_filt003_sensitivity(test, filter_options, column_filter_options):
...
+
+
+@pytest.mark.parametrize(
+    "column_placeholder_setting,table_placeholder_setting,expected_placeholder",
+    [
+        ("abc", None, "abc"),
+        (None, "def", "def"),
+        ("gah", "ijk", "gah"),
+        ("", None, ""),
+        (None, None, "filter data..."),
```
alexcjohnson, line 313: Beautiful set of test cases! One additional case I'm curious about though: what if only `filter_options.case` is set at the column level, does the table-level `filter_options.placeholder` still make it through? ie does the column-level setting merge with or replace the table-level setting? Merging would be preferred.

Vice versa too, I suppose: does table-level `case` survive if only `placeholder` is supplied by the column?

components/dash-table/src/dash-table/dash/DataTable.js
```diff
@@ -1,7 +1,7 @@
 import * as R from 'ramda';
-import React, {Component, Suspense} from 'react';
+import React, { Component, Suspense } from 'react';
 import PropTypes from 'prop-types';
-import {asyncDecorator} from '@plotly/dash-component-plugins';
+import { asyncDecorator } from '@plotly/dash-component-plugins';
```
alexcjohnson, line 4: The lint job failed, I think our settings actually prohibit these spaces. `npm run format` from `components/dash-table` should sort this out.

alexcjohnson:
Thanks @fraghag - very nicely done. Aside from my two minor comments above, just a changelog entry and I think we'll be good to go!

components/dash-table/tests/selenium/test_filter.py
fraghag: This is handled [here](https://github.com/plotly/dash/blob/2bd83a98322f95243359d846ec9e97a5b542a36e/components/dash-table/src/dash-table/dash/Sanitizer.ts#L85). Added the usecase to the tests for both placeholder_text and case.

components/dash-table/src/dash-table/dash/DataTable.js
fraghag: It was my VSCode living it's own life. Fixed it now with npm, thanks!

fraghag:
> Thanks @fraghag - very nicely done. Aside from my two minor comments above, just a changelog entry and I think we'll be good to go!

Fixed the comments now. But regarding the changelog, I don't understand where in the Changelog file I'm supposed to add this, I only see entries from already released versions. I was looking form info about this in the CONTRIBUTING file. Could you give me some pointers?

components/dash-table/tests/selenium/test_filter.py
```diff
@@ -1,6 +1,7 @@
 import dash
 
 from dash.dash_table import DataTable
+from pyparsing import col
```
alexcjohnson, line 4: Didn't end up being used? [Linter flagged](https://app.circleci.com/pipelines/github/plotly/dash/4049/workflows/ec453535-6f7e-4a75-a052-514add2cafb1/jobs/66999):

```
tests/selenium/test_filter.py:4:1: F401 'pyparsing.col' imported but unused
```

alexcjohnson:
Re: changelog - yeah we should describe that better, good point. You can add a new section at the top, right before the `[2.6.2]` heading:

```
## Unreleased

### Added

- [#2261](https://github.com/plotly/dash/pull/2261) 1-2 sentence description of the feature
```

components/dash-table/tests/selenium/test_filter.py
fraghag: Fixed!

alexcjohnson (approved):
💃 Excellent work, thanks again for the PR @fraghag!

Issue #2263, opened 2022-10-08
[BUG] Dash Pages does not support relative package imports within page modules


ned2:
Env:
```
dash                 2.6.2
dash-core-components 2.0.0
dash-html-components 2.0.0
dash-table           5.0.0
```

**Describe the bug**

I typically make my Dash apps as Python packages, using relative imports within my app. I've just discovered that Dash Pages is not happy with relative imports being used within your page modules (eg `pages/page1.py`). As far as I can tell, this is because Dash manually loads the contents of `pages` as modules using `exec_module`, but, if your app is inside a package, it does not attach the page modules to the package. So when you try to access a `utils.py` in the base of your package, eg with `from ..utils import blah` within `pages/page1.py`, you'll get:

    ImportError: attempted relative import beyond top-level package

So effectively, every module within your `pages` directory is orphaned from the package, and you won't be able to use relative imports to share utility functions across your page modules and the rest of your app.

**Minimal example**

`__init__.py` (empty)
`app.py`
```python
import dash
app = dash.Dash(__name__, use_pages=True)

app.layout = dash.html.Div(["main page", dash.page_container])

app.run_server(debug=True)
```

`pages/__init__.py` (empty)
`pages/page.py`
```python
import dash

from ..utils import blah

dash.register_page(__name__, path="/page")

layout = dash.html.Div("page")
```

`utils.py`
```python
def blah():
    pass
```

Runnning `python -m dash_pages_bug.app` from the directory above results in:

    ImportError: attempted relative import beyond top-level package

ned2:
I think I've tracked down a workaround...

The problem is that [within `_import_layouts_from_pages`](_import_layouts_from_pages), in `dash.py`, `module_name` is set to `pages.page`, which does not include a reference to the package I'm working in.  

                module_name = ".".join([pages_folder, page_filename])

                spec = importlib.util.spec_from_file_location(
                    module_name, os.path.join(root, file)
                )
                page_module = importlib.util.module_from_spec(spec)
                spec.loader.exec_module(page_module)

So we need to get the package name reflected in the module_name variable as `my_package.pages.page`. Since my standard practice is to set the `name` param of the `Dash` instance to the value of `__package__`, I can leverage that by changing the first line of the above snipped to:

     module_name = ".".join([self.config.name, pages_folder, page_filename])

[...]

PR #2271, opened 2022-10-13
 [WIP] Make Dash Pages & packages play nice & experimenting with clearing Pages registry

ned2:
_Please don't merge._ 

This is something of an experiment to try to address the problem of Dash Pages not playing nicely when used with a Dash app that's housed in a Python package, as discussed in #2263.

The solution I adopted was to prefix the modules programatically imported by Dash Pages (eg the contents of `pages/*.py`) with the `name` parameter of the `Dash` instance. All the user needs to do then is invokes their apps as `Dash(name=__package__)`. This largely worked (as far as I can tell) but broke some tests that made some assumptions about the contents of `_pages.PAGE_REGISTRY` across multiple `Dash` app instantiations.

The more experimental part of this PR is to clear the contents of the `_pages.PAGE_REGISTRY` on `__init__` of each `Dash` instance.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -172,7 +172,13 @@ no_update = _callback.NoUpdate()  # pylint: disable=protected-access
 def init_dash_globals():
     """Ensure that all Dash global state is re-initialised."""
     _pages.PAGE_REGISTRY.clear()
-    _pages.CONFIG.clear()
+
+    # this line of thinking could be extended to potentially clear the following:
+    # _pages.CONFIG
+    # _get_paths.CONFIG
+    # _callback.GLOBAL_CALLBACK_MAP
+    # _callback.GLOBAL_CALLBACK_LIST
+    # _callback.GLOBAL_INLINE_SCRIPTS
 
 
 # pylint: disable=too-many-instance-attributes
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -169,6 +169,12 @@ def _get_traceback(secret, error: Exception):
 no_update = _callback.NoUpdate()  # pylint: disable=protected-access
 
 
+def init_dash_globals():
+    """Ensure that all Dash global state is re-initialised."""
+    _pages.PAGE_REGISTRY.clear()
+    _pages.CONFIG.clear()
+
+
 # pylint: disable=too-many-instance-attributes
 # pylint: disable=too-many-arguments, too-many-locals
 class Dash:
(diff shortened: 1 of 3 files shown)


ned2:
hey @alexcjohnson, curious to get your feedback on this proposal if you have some time to review. Would be great to get Dash apps with Pages to properly support living inside Python packages. 

My solution involved clearing some `dash` module global state on app startup, which might seem like a heavy handed approach, but on reflection, I wonder if working towards doing this for _all_ global app-specific state would be a sensible move anyway (eg for callbacks). Currently, multiple Dash app instances started in the same process that uses either Dash Pages or the `dash.callback` decorator run the risk of dirty state causing conflicts, as I discovered the hard way while implementing the first part of my solution.

I had thought there was some tests that were failing, but it looks like that might be to do with some local weirdness when I try to run the full test sweet, as the tests have passed ok in CI. Though maybe there's something fishy is going on with Percy tests?

alexcjohnson:
Thanks @ned2 - I saw this PR come in (along with your extensive investigations in #2263) but haven't had a chance to look at it in detail yet. I recall when @AnnMarieW and I were discussing pages prior to its release we identified some cases where keeping this global state is important, and other cases where it's problematic. That may just have had to do with tests though, I'll need to poke around again.

Can you give me a little more context on when and why you run multiple apps in the same process, just so I can understand the use case a bit better?

> maybe there's something fishy is going on with Percy tests?

That should be cleared up next time you merge/rebase on master

ned2:
> Can you give me a little more context on when and why you run multiple apps in the same process, just so I can understand the use case a bit better?

I haven't thought of any contexts where I'd want to run multiple Dash apps in the same process, besides the motivating context of running consecutive tests. So the catalysing motivation for clearing the Page registry is to fix the clashing page keys within tests that my solution introduced, as described here: https://github.com/plotly/dash/issues/2263#issuecomment-1272540304.

I'm not totally fixed in clearing the registry on app init. I imagine there's other strategies to solve the package issue... either a different solution that doesn't introduce conflicting keys in the registry, or using the same approach, but making changes to the test harness to address the conflicts. 

I proposed clearing the registry because a) it feels like a clean solution, b) would mean Dash devs don't have to have to spend brain cycles on futzing around clearing the registry manually in tests, and b) it feels desirable to eliminate the possibility of dirty global state breaking user apps in unexpected and probably rather opaque ways.

ned2:
I've since come up with a cleaner way to address #2263 that doesn't rely on modifying any global state, so will close this and pick up in another PR.

PR #2272, opened 2022-10-14
WIP: add extra urls for deps

eff-kay:
fixes https://github.com/plotly/dekn/issues/3763

must be reviewed with https://github.com/plotly/dash-snapshots/pull/243

for the workspace containers `http://<127.0.0.1>:8050  inside the containers` == `https://><DASH_HOST_NAME>/workspace/view/workspace-name from outside the containers`

The dependencies are served at absolutes `/`. But the index layout from dash uses the `/workspace/view/workspace-name` when populating the `src` value of `<scripts>` bundle. This is fine for requests coming from outside, but for requests coming from inside, the requests generated becomes `127.0.0.1:8050/workspace/view/workspace-name/_dash-renderer`, which is not a registered url because the bundle is served at `127.0.0.1:8050/_dash-renderer`. 

So we either need to update the `<scripts>`  urls in the bundle generated , based on where the request is coming from , or we register `/workspace/view/workspace-name/_dash-renderer` as well. This PR is doing the latter.

this solves the problem.


alexcjohnson:
Not needed in the end 😌

Issue #1760, opened 2021-09-17
dcc.Slider with marks breaks with Dash 2.0


vinicvaz:
**Describe your context**
```
dash                            2.0.0     
dash-ag-grid                    1.0.0     
dash-bootstrap-components       0.13.0    
dash-cool-components            0.1.7     
dash-core-components            2.0.0     
dash-dangerously-set-inner-html 0.0.2     
dash-design-kit                 1.6.4    
dash-design-kit-docs            1.6.4     
dash-flow-example               0.0.5     
dash-html-components            2.0.0     
dash-pro-components             0.0.4  
dash-renderer                   1.9.1     
dash-table                      5.0.0     
```
    - OS: Ubuntu 20.04
    - Browser: Chrome
    - Version 92.0.4515.131

**Describe the bug**

`dcc.Slider` **marks** seems to be crashing the page rendering when using **Dash 2.0**.
When using **Dash 1.21.0** the page renders right with **marks**, but with 2.0 not.

**Expected behavior**

Page be rendered with `dcc.Slider` marks or raise some exception warning that this prop is not acceptable.

**Screenshots**
Attaching a small gif showing the behavior.

![possible_bug_dcc_slider_dash2](https://user-images.githubusercontent.com/50757601/133773279-1e7a63af-ca64-408f-be78-67230155813b.gif)

Thanks!

anders-kiaer:
We see the same change in behaviour when upgrading. In our case going from `marks` being `Dict[some_number_type, str]` to `Dict[str, str]` got it working with `dash==2.0`. 

The documentation on https://dash.plotly.com/dash-core-components/slider actually states that `marks` should be `Dict[str, str]` (even though number-type as key perhaps feels more intuitive here?). Not sure if requiring `marks` to be `Dict[str, str]` it is a recent change, or if that requirements has gotten stricter in the stack in some way 🤔🙂

AnnMarieW:
Fixed in dash V4 🎉

PR #2276, opened 2022-10-20
Add typing to component init

T4rk1n:
Add python `typing` to generated components `__init__`.

## Contributor Checklist


dash/development/_py_components_generation.py
```diff
@@ -55,7 +57,10 @@ def generate_class_string(
     _namespace = '{namespace}'
     _type = '{typename}'
     @_explicitize_args
-    def __init__(self, {default_argtext}):
+    def __init__(
+        self,
+        {default_argtext}
+    ):
```
alexcjohnson, line 63: I wonder, now that we're Py3-only, maybe we should call `black` on the generated files as part of the generation process? We already do that in the core component packages as a separate step.

tuchandra:
Hi - this is really, really cool. I would love to see this landed into Dash, and if I can help with reviewing this PR, debugging any tests, etc., please let me know. I was working on a separate effort to generate stub (.pyi) files, so I have tinkered around with this part of the codebase a little bit. I'm excited to see where this goes!

tests/unit/development/metadata_test.py
```diff
@@ -91,7 +92,31 @@ class Table(Component):
...
+    def __init__(
+        self,
+        children: typing.Union[str, int, float, Component, typing.List[typing.Union[str, int, float, Component]]] = None,
+        optionalArray: typing.List = Component.UNDEFINED,
+        optionalBool: bool = Component.UNDEFINED,
+        optionalFunc: typing.Any = Component.UNDEFINED,
+        optionalNumber: typing.Union[float, int] = Component.UNDEFINED,
```
eff-kay, line 101: @alexcjohnson @T4rk1n just wondering if people can pass in Decimals in here as well?

tests/unit/development/metadata_test.py
```diff
@@ -91,7 +92,31 @@ class Table(Component):
...
     _type = 'Table'
     @_explicitize_args
-    def __init__(self, children=None, optionalArray=Component.UNDEFINED, optionalBool=Component.UNDEFINED, optionalFunc=Component.UNDEFINED, optionalNumber=Component.UNDEFINED, optionalObject=Component.UNDEFINED, optionalString=Component.UNDEFINED, optionalSymbol=Component.UNDEFINED, optionalNode=Component.UNDEFINED, optionalElement=Component.UNDEFINED, optionalMessage=Component.UNDEFINED, optionalEnum=Component.UNDEFINED, optionalUnion=Component.UNDEFINED, optionalArrayOf=Component.UNDEFINED, optionalObjectOf=Component.UNDEFINED, optionalObjectWithExactAndNestedDescription=Component.UNDEFINED, optionalObjectWithShapeAndNestedDescription=Component.UNDEFINED, optionalAny=Component.UNDEFINED, customProp=Component.UNDEFINED, customArrayProp=Component.UNDEFINED, id=Component.UNDEFINED, **kwargs):
+    def __init__(
+        self,
+        children: typing.Union[str, int, float, Component, typing.List[typing.Union[str, int, float, Component]]] = None,
+        optionalArray: typing.List = Component.UNDEFINED,
+        optionalBool: bool = Component.UNDEFINED,
+        optionalFunc: typing.Any = Component.UNDEFINED,
```
eff-kay, line 100: there is a type called callable, https://docs.python.org/3.6/library/typing.html#typing.Callable

tests/unit/development/metadata_test.py
```diff
@@ -91,7 +92,31 @@ class Table(Component):
     _namespace = 'TableComponents'
     _type = 'Table'
     @_explicitize_args
-    def __init__(self, children=None, optionalArray=Component.UNDEFINED, optionalBool=Component.UNDEFINED, optionalFunc=Component.UNDEFINED, optionalNumber=Component.UNDEFINED, optionalObject=Component.UNDEFINED, optionalString=Component.UNDEFINED, optionalSymbol=Component.UNDEFINED, optionalNode=Component.UNDEFINED, optionalElement=Component.UNDEFINED, optionalMessage=Component.UNDEFINED, optionalEnum=Component.UNDEFINED, optionalUnion=Component.UNDEFINED, optionalArrayOf=Component.UNDEFINED, optionalObjectOf=Component.UNDEFINED, optionalObjectWithExactAndNestedDescription=Component.UNDEFINED, optionalObjectWithShapeAndNestedDescription=Component.UNDEFINED, optionalAny=Component.UNDEFINED, customProp=Component.UNDEFINED, customArrayProp=Component.UNDEFINED, id=Component.UNDEFINED, **kwargs):
+    def __init__(
+        self,
+        children: typing.Union[str, int, float, Component, typing.List[typing.Union[str, int, float, Component]]] = None,
+        optionalArray: typing.List = Component.UNDEFINED,
```
eff-kay, line 98: this should be a `List[Any]`. mypy will throw an error on this. 

also, wondering if we should use the parent type `Iterable`.

dash/development/_py_prop_typing.py
```diff
@@ -0,0 +1,77 @@
+def generate_any(_):
+    return "typing.Any"
+
+
+def generate_shape(t):
```
eff-kay, line 5: We should probably start adding annotations to these functions too.

dash/development/_py_prop_typing.py
```diff
@@ -0,0 +1,77 @@
...
+    elements = ", ".join(get_prop_typing(x.get("name"), x) for x in els)
+    return f"typing.Tuple[{elements}]"
+
+
+def generate_array_of(t):
+    typed = PROP_TYPING.get(t["value"]["name"], generate_any)(t["value"])
+    return f"typing.List[{typed}]"
+
+
+def generate_object_of(t):
```
eff-kay, line 39: Object can have any hashable type as keys, not just strs.

dash/development/_py_prop_typing.py
```diff
@@ -0,0 +1,77 @@
...
+    typed = PROP_TYPING.get(t["value"]["name"], generate_any)(t["value"])
+    return f"typing.Dict[str, {typed}]"
+
+
+def generate_type(typename):
+    def type_handler(_):
```
eff-kay, line 45: lamba function?

tests/integration/test_typing.py
```diff
@@ -0,0 +1,221 @@
...
+
+import pytest
+
+component_template = """
+from dash_generator_test_component_typescript.TypeScriptComponent import TypeScriptComponent
+
+t = TypeScriptComponent({0})
+"""
+
+
```
eff-kay, line 14: I don't think all of this should be necessary if we make pyright as part of the CI. Similar to flake8 or black.

tests/unit/development/metadata_test.py
```diff
@@ -91,7 +92,31 @@ class Table(Component):
...
+        optionalBool: bool = Component.UNDEFINED,
+        optionalFunc: typing.Any = Component.UNDEFINED,
+        optionalNumber: typing.Union[float, int] = Component.UNDEFINED,
+        optionalObject: typing.Dict = Component.UNDEFINED,
+        optionalString: str = Component.UNDEFINED,
+        optionalSymbol: typing.Any = Component.UNDEFINED,
```
eff-kay, line 104: Symbol is an identifier in Python. We should probably create a custom type for it.

tests/unit/development/metadata_test.py
```diff
@@ -91,7 +92,31 @@ class Table(Component):
...
+        optionalNode: typing.Union[str, int, float, Component, typing.List[typing.Union[str, int, float, Component]]] = Component.UNDEFINED,
+        optionalElement: Component = Component.UNDEFINED,
+        optionalMessage: typing.Any = Component.UNDEFINED,
+        optionalEnum: typing.Any = Component.UNDEFINED,
```
eff-kay, line 108: Python has Enum type.

tests/unit/development/metadata_test.py
```diff
@@ -91,7 +92,31 @@ class Table(Component):
...
+        optionalUnion: typing.Union[str, typing.Union[float, int], typing.Any] = Component.UNDEFINED,
+        optionalArrayOf: typing.List[typing.Union[float, int]] = Component.UNDEFINED,
+        optionalObjectOf: typing.Dict[str, typing.Union[float, int]] = Component.UNDEFINED,
+        optionalObjectWithExactAndNestedDescription: typing.Dict[str, typing.Union[str, typing.Union[float, int], typing.Dict[str, typing.Union[typing.List[typing.Dict], typing.Dict]]]] = Component.UNDEFINED,
+        optionalObjectWithShapeAndNestedDescription: typing.Dict[str, typing.Union[str, typing.Union[float, int], typing.Dict[str, typing.Union[typing.List[typing.Dict], typing.Dict]]]] = Component.UNDEFINED,
+        optionalAny: typing.Any = Component.UNDEFINED,
```
eff-kay, line 114: this is optional, so we should probably use `typing.Optional[typing.Any]`, the same is true of other optional params.

dash/development/_py_prop_typing.py
eff-kay, line 5: Also, not sure what this function is doing differently than `generate_object_of`, but maybe this is a nested object.

dash/development/_py_prop_typing.py
alexcjohnson: > Object can have any hashable type as keys

They're all going to turn into strings when we JSON-encode them though. Sometimes people use numbers as the keys, which has some important use cases so I guess we had better not prevent it, but it also causes problems, eg does `5.0` encode to `'5'` or `'5.0'`? https://github.com/plotly/dash/issues/1760

tests/unit/development/metadata_test.py
alexcjohnson: If [`to_json_plotly`](https://github.com/plotly/dash/blob/38cd3390ba59cc7fb2a5d1195047b704766e9079/dash/_utils.py#L18-L22) can serialize it then yes. It's pretty flexible, so probably `Decimal` is fine, but I haven't tested it.

tests/unit/development/metadata_test.py
alexcjohnson: @eff-kay note that there are types in this test that do not work in Dash, and functions are among them. Anything not JSON-serializable.

tests/unit/development/metadata_test.py
T4rk1n: `typing.List` is same as `typing.List[typing.Any]`, iterable might not json compatible, I think the List is the most safe for this typing.

dash/development/_py_prop_typing.py
T4rk1n: I try to avoid lambda's since they are harder to refactor with the different syntax, they are also suboptimal when it come to overhead.

tests/integration/test_typing.py
T4rk1n: This tests the types are right, calling on the generated component would just tells you if the component has typing errors in it.

tests/unit/development/metadata_test.py
T4rk1n: The enum type need to be generated, I was keeping that (and generating TypedDict/Model for shapes) for a later addition as there is much more work to be done on this part.

tests/unit/development/metadata_test.py
T4rk1n: An argument that has a default is automatically optional, the `typing.Optional` is more meant for when you define `TypedDict`.

dash/development/_py_prop_typing.py
T4rk1n: `ObjectOf` has a single type while shape can have more than one type for it's keys. This is also more of a placeholder, the final implementation should generate a `TypedDict` or a pydantic model.

T4rk1n:
I went with `pyright` instead of `mypy` for the static checker because it has better package/module resolution and default argument gets automatically added as an union type.

The checker and my IDE don't recognize the type of `Component` and as such put it as `Any` which breaks the checking for those types.

tests/unit/development/metadata_test.py
alexcjohnson: There are certainly a number of other iterables people frequently use though - `tuple`, `range`, `dict_keys`, `dict_values`, etc etc...  I wouldn't want to push people into casting everything to a list, that would be unnecessary overhead.

dash/development/_py_prop_typing.py
eff-kay: suboptimal?

tests/unit/development/metadata_test.py
T4rk1n: `Decimal` can serialize, I think we can put `numbers.Number` to get all the number types.

dash/development/_py_components_generation.py
```diff
@@ -1,16 +1,21 @@
...
 from dash.exceptions import NonExistentEventException
 from ._all_keywords import python_keywords
 from ._collect_nodes import collect_nodes, filter_base_nodes
+from ._py_prop_typing import get_prop_typing, enums, shapes
 from .base_component import Component
 
 
-# pylint: disable=unused-argument,too-many-locals
+# pylint: disable=unused-argument,too-many-locals,too-many-branches
 def generate_class_string(
```
eff-kay, line 19: if possible can we please change this to kw only arguments. Its difficult to see which one is which. 

```
def generate_class_string(*, typename, props, description, namespace, prop_reorder_exceptions=None, max_props=None). 
```

tests/unit/development/metadata_test.py
alexcjohnson: `numbers.Number` is too much, it includes `complex` - how about `numbers.Real`?

tests/unit/development/metadata_test.py
alexcjohnson: Still open 😎

tests/unit/development/metadata_test.py
```diff
@@ -90,8 +124,37 @@ class Table(Component):
...
+        optionalBool: bool = Component.UNDEFINED,
+        optionalFunc: typing.Any = Component.UNDEFINED,
+        optionalNumber: typing.Union[int, float, numbers.Number] = Component.UNDEFINED,
+        optionalObject: dict = Component.UNDEFINED,
+        optionalString: str = Component.UNDEFINED,
+        optionalSymbol: typing.Any = Component.UNDEFINED,
+        optionalNode: typing.Union[str, int, float, Component, typing.List[typing.Union[str, int, float, Component]]] = Component.UNDEFINED,
+        optionalElement: Component = Component.UNDEFINED,
+        optionalMessage: typing.Any = Component.UNDEFINED,
+        optionalEnum: typing.Union[str, str, OptionalEnumEnum] = Component.UNDEFINED,
```
alexcjohnson, line 146: Why does this get `str` (twice!) and not just `OptionalEnumEnum`?

tests/unit/development/metadata_test.py
```diff
@@ -90,8 +124,37 @@ class Table(Component):
...
+        children: typing.Union[str, int, float, Component, typing.List[typing.Union[str, int, float, Component]]] = None,
+        optionalArray: typing.List = Component.UNDEFINED,
+        optionalBool: bool = Component.UNDEFINED,
+        optionalFunc: typing.Any = Component.UNDEFINED,
+        optionalNumber: typing.Union[int, float, numbers.Number] = Component.UNDEFINED,
+        optionalObject: dict = Component.UNDEFINED,
+        optionalString: str = Component.UNDEFINED,
+        optionalSymbol: typing.Any = Component.UNDEFINED,
+        optionalNode: typing.Union[str, int, float, Component, typing.List[typing.Union[str, int, float, Component]]] = Component.UNDEFINED,
+        optionalElement: Component = Component.UNDEFINED,
```
alexcjohnson, line 144: Huh, never noticed this before... do we actually support `Element` as distinct from `Node`?

T4rk1n: Yes, it ends up rendering the same as node but for typing it allows only a single component to be put and not a list of string/number/components.

tests/unit/development/metadata_test.py
T4rk1n: We can add Tuple as union for untyped arrays, but `arrayOf` is really a different typing on both side.

tests/unit/development/metadata_test.py
alexcjohnson: Nice solution to deduplicate these :) But why does it need `str` at all? The `OptionalEnumEnum` by itself isn't enough to let it accept strings from the enum set?

T4rk1n: Yes, the enum only validate for the enum type and otherwise would error on the raw value.

requires-install.txt
```diff
@@ -5,3 +5,5 @@ dash_core_components==2.0.0
 dash_table==5.0.0
 importlib-metadata==4.8.3;python_version<"3.7"
 contextvars==2.4;python_version<"3.7"
+typing_extensions==4.1.1;python_version<"3.7"
+typing_extensions>=4.4.0;python_version>"3.6"
```
alexcjohnson, line 9:
```suggestion
typing_extensions>=4.4.0;python_version>="3.7"
```
Just to be more obviously symmetric with the `<"3.7"`

I wonder though, if 4.1.1 is sufficient would we be better off just collapsing these into one line? If someone on 3.7+ has a version >=4.1.1,<4.4.0 pinned for some reason is that a problem?
```
typing_extensions>=4.1.1
```

dash/development/_py_components_generation.py
```diff
@@ -181,8 +215,12 @@ def generate_class_file(
     """
     import_string = (
         "# AUTO GENERATED FILE - DO NOT EDIT\n\n"
-        + "from dash.development.base_component import "
-        + "Component, _explicitize_args\n\n\n"
+        "import typing  # noqa: F401\n"
+        "import numbers # noqa: F401\n"
+        "import enum # noqa: F401\n"
+        "from typing_extensions import TypedDict, NotRequired # noqa: F401\n"
```
alexcjohnson, line 221: Not a big deal, but in principle we can avoid `noqa: F401` if we look at `class_string` and drop any imports we don't need.

tests/unit/development/metadata_test.py
alexcjohnson: Are there any examples of this in existing components? Does it work, either as `children` or with components-as-props?

T4rk1n: Before component as props, it was kind of invalid prop as only the children would be hydrated and children would be treated as a node, component as props also doesn't differ those Element from node.

The collect_nodes make both node and element into component:
https://github.com/plotly/dash/blob/453834bbd0083d66d73752ca5d7c25b30d3583cc/dash/development/_collect_nodes.py#L1-L2

I think the only other place we have Element defined is in the typescript props:
https://github.com/plotly/dash/blob/453834bbd0083d66d73752ca5d7c25b30d3583cc/@plotly/dash-generator-test-component-typescript/src/props.ts#L31
But the prop itself was not tested as it was before component as props was supported.

tests/unit/development/metadata_test.py
T4rk1n: https://docs.python.org/3/library/numbers.html#the-numeric-tower

Number is the base class, then they extend the last one, to get the `Integral` you need all the other 3 implemented. Tested with numpy and Number is what get the most support and our `to_json` supports them, I dont think custom implementation of those works.

T4rk1n:
Discovered the shapes TypedDict are not exporting properly on the component to be used in custom types definition like callbacks inputs and returns value.

ie:
```
@callback(Input("graph", "figure"), ... )
def my_callback(figure: dcc.Graph.Figure) -> dcc.Graph.Figure:
    ...
```

For the shapes to be used in that way, the shapes must be defined on the component:

```
class C(Component):
    class Shape(TypedDict):
        ...
```

T4rk1n:
Nesting the typed dict on the class doesn't work if the `TypedDict` uses another TypedDict, need another solution

The Literal type from python 3.8, also available in `typing_extension` for python <=3.7, can be used instead of the enums to be more similar to the way it work typescript and actual value validation.

tests/unit/development/metadata_test.py
T4rk1n: It's better to go without defining type aliases that would require to be imported and would break backward compatibility for component libraries that compiled with the new typing but the user has dash locked to a previous version.

requires-install.txt
T4rk1n: 4.1.1 has fix for imports on 3.7, it's the last version that works with 3.6. I don't see anything we actually uses in the later version so we can try a single version >=4.1.1

T4rk1n:
The docstring on the components are not in the right position for IDE support, the docstring for the props should be under the `__init__` for them to appear on the tooltip when hovering/typing a component prop.

requires-install.txt
```diff
@@ -5,3 +5,5 @@ dash_core_components==2.0.0
 dash_table==5.0.0
 importlib-metadata==4.8.3;python_version<"3.7"
 contextvars==2.4;python_version<"3.7"
+typing_extensions>=4.1.1
+stringcase>=1.2.0
```
alexcjohnson, line 9: do we need `stringcase` in `requires-install`? Couldn't it go in `requires-dev`?

T4rk1n: It's needed for the old runtime component loader :disappointed:

alexcjohnson: haha ok - we should put a deprecation warning on runtime component generation, and remove it whenever we decide to do a Dash 3.0

gvwilson:
@alexcjohnson can you please review this for us? thank you - @gvwilson

dash/development/_py_prop_typing.py
```diff
@@ -0,0 +1,170 @@
...
+
+def generate_array_of(
+    type_info,
+    component_name: str,
+    prop_name: str,
+):
+    typed = get_prop_typing(
+        type_info["value"]["name"], component_name, prop_name, type_info["value"]
+    )
+    return f"typing.Union[typing.Sequence[{typed}], typing.Tuple]"
```
alexcjohnson, line 85: Why do we need the `Union` here? `Sequence` includes tuples, I believe?

dash/development/_py_prop_typing.py
```diff
@@ -0,0 +1,170 @@
...
+        "from plotly.graph_objects import Figure"
+    )
+    return "typing.Union[Figure, dict]"
+
+
+special_cases = {"dash_core_components": {"Graph": {"figure": generate_plotly_figure}}}
+
+
+PROP_TYPING = {
+    "array": generate_type("typing.Union[typing.Sequence, typing.Tuple]"),
```
alexcjohnson, line 149: Again, doesn't `Sequence` include `Tuple` already?

tests/unit/development/metadata_test.py
alexcjohnson: hmph ok, `numbers.Real` doesn't catch `Decimal` for some reason, but `numbers.Number` does. So OK, we can leave it as you have it (`int, float, numbers.Number`) - better to allow some invalid items through than to prevent some valid items.

tests/integration/test_typing.py
```diff
@@ -0,0 +1,252 @@
...
+    assert (
+        status == expected_status
+    ), f"Status: {status}\nOutput: {output}\nError: {error}"
+    for ex_out in expected_outputs:
+        assert ex_out in output, f"Invalid output:\n {output}"
+    for ex_err in expected_errors:
+        assert ex_err in error
+
+
+@pytest.mark.skipif(sys.version_info < (3, 7), reason="pyright not available on 3.6")
```
alexcjohnson, line 48:
```suggestion
```
we're 3.8+ now ☺️

alexcjohnson (approved):
Just a couple of minor comments, this looks great! 💃 

I'd still like us to put a deprecation notice on runtime-generated components, no-one should be using those anymore, right? Doesn't need to be in this PR though.

dash/development/_py_prop_typing.py
T4rk1n: Good catch, I changed from List but didn't remove the Tuple.

dash/__init__.py
```diff
@@ -53,3 +53,37 @@ def _jupyter_nbextension_paths():
             "require": "dash/main",
         }
     ]
+
+
+__all__ = [
```
gvwilson, line 58: why such a big list - was `__all__` not defined at all before?

dash/_callback.py
```diff
@@ -60,14 +61,14 @@ def is_no_update(obj):
...
-    progress_default=None,
-    running=None,
-    cancel=None,
-    manager=None,
-    cache_args_to_ignore=None,
+    background: bool = False,
+    interval: int = 1000,
+    progress: Optional[Output] = None,
+    progress_default: Any = None,
+    running: Optional[List[Tuple[Output, Any, Any]]] = None,
```
gvwilson, line 68: took me a moment to parse this :-)

dash/development/_py_components_generation.py
```diff
@@ -1,16 +1,20 @@
...
 from dash.exceptions import NonExistentEventException
 from ._all_keywords import python_keywords
 from ._collect_nodes import collect_nodes, filter_base_nodes
-from .base_component import Component
+from ._py_prop_typing import get_prop_typing, shapes, custom_imports
+from .base_component import Component, ComponentType
 
 
-# pylint: disable=unused-argument,too-many-locals
+# pylint: disable=unused-argument,too-many-locals,too-many-branches
```
gvwilson, line 17: :-)

dash/development/_py_components_generation.py
gvwilson: added issue #2965 to record this

dash/development/_py_prop_typing.py
```diff
@@ -0,0 +1,170 @@
...
+            props.append(f'        "{prop_key}": NotRequired[{typed}]')
+        else:
+            props.append(f'        "{prop_key}": {typed}')
+
+    shapes.setdefault(component_name, {})
+    shapes[component_name][name] = textwrap.indent(
+        shape_template.format(
+            name=name, values="    {\n" + ",\n".join(props) + "\n    }"
+        ),
+        "    ",
```
gvwilson, line 50: why do we need to indent an empty string at this point?

dash/development/base_component.py
```diff
@@ -406,13 +412,15 @@ def __repr__(self):
         return f"{self._type}({props_string})"
 
 
+ComponentType = typing.TypeVar("ComponentType", bound=Component)
+
+
+# This wrapper adds an argument given to generated Component.__init__
+# with the actual given parameters by the user as a list of string.
+# This is then checked in the generated init to check if required
+# props were provided.
```
gvwilson, line 421: thank you for the explanation

dash/development/component_loader.py
```diff
@@ -35,6 +36,14 @@ def load_components(metadata_path, namespace="default_namespace"):
     `type`, `valid_kwargs`, and `setup`.
     """
 
+    warnings.warn(
+        DeprecationWarning(
+            "Dynamic components loading has been deprecated and will be removed"
+            " in dash 3.0.\n"
```
gvwilson, line 42: is there an issue filed to removed dynamic component loading so we don't forget?

tests/integration/test_typing.py
```diff
@@ -0,0 +1,251 @@
...
+    assert (
+        status == expected_status
+    ), f"Status: {status}\nOutput: {output}\nError: {error}"
+    for ex_out in expected_outputs:
+        assert ex_out in output, f"Invalid output:\n {output}"
+    for ex_err in expected_errors:
+        assert ex_err in error
+
+
+@pytest.mark.parametrize(
```
gvwilson, line 48: nice

tests/unit/development/metadata_test.py
```diff
@@ -90,8 +94,66 @@ class Table(Component):
...
+            "figure": NotRequired["OptionalObjectWithShapeAndNestedDescriptionFigure"]
+        }
+    )
+
     @_explicitize_args
-    def __init__(self, children=None, optionalArray=Component.UNDEFINED, optionalBool=Component.UNDEFINED, optionalFunc=Component.UNDEFINED, optionalNumber=Component.UNDEFINED, optionalObject=Component.UNDEFINED, optionalString=Component.UNDEFINED, optionalSymbol=Component.UNDEFINED, optionalNode=Component.UNDEFINED, optionalElement=Component.UNDEFINED, optionalMessage=Component.UNDEFINED, optionalEnum=Component.UNDEFINED, optionalUnion=Component.UNDEFINED, optionalArrayOf=Component.UNDEFINED, optionalObjectOf=Component.UNDEFINED, optionalObjectWithExactAndNestedDescription=Component.UNDEFINED, optionalObjectWithShapeAndNestedDescription=Component.UNDEFINED, optionalAny=Component.UNDEFINED, customProp=Component.UNDEFINED, customArrayProp=Component.UNDEFINED, id=Component.UNDEFINED, **kwargs):
+    def __init__(
+        self,
+        children: typing.Optional[typing.Union[str, int, float, ComponentType, typing.Sequence[typing.Union[str, int, float, ComponentType]]]] = None,
```
gvwilson, line 134: whoa... :-)

gvwilson (approved):
a couple of questions for my own learning, but :+1: from me. (caution: I'm not a very informed reviewer)

dash/__init__.py
T4rk1n: With the addition of `py.typed` file, it is now expected to have a `__all__` attribute, otherwise it shows an error on every imports. The list contains everything that is "exported" here.

gvwilson: :+1: thank you

dash/_callback.py
eff-kay: I think you can just stop at a higher level `Optional[List]`, and be les concrete.

tests/unit/development/metadata_test.py
eff-kay: We can use [type-aliases](https://docs.python.org/3/library/typing.html#type-aliases) to make our lives easier here. 


```
BasicType: TypeAlias = Union[str, int, float]
ComponentOrBasic: TypeAlias = Union[BasicType, ComponentType]
ChildrenType: TypeAlias = Optional[Union[ComponentOrBasic, Sequence[ComponentOrBasic]]]



children: ChildrenType = None,
````

elomarikhalid:
# ARPA BLS Threshold Signature Network Standards

Core architecture, components, protocol, and token-economics

## Abstract

ARPA BLS Threshold Signature Network (ARPA) is a decentralized (permission-less) network consisting of multiple, dynamic groups of nodes that can securely generate BLS threshold signatures. ARPA can serve as the infrastructure of many Blockchain applications such as verifiable random beacons, secure (keyless) wallets, cross-chain bridges, decentralized custody, etc.

## 1. Introduction

### 1.1 Design Goals

| Goals | Description |
| --- | --- |
| Decentralization | Eliminate the requirement for centralized authorities or single point of failure in nodes/groups management within the network. |
| Control | Give entities, both human and non-human, the power to directly control the generation and usage of the BLS signature without the need to rely on external authorities. |
| Security | Enable sufficient security for requesting parties to depend on the generated BLS signature for their required level of assurance |
| Proof-based | Enable BLS signatures to provide cryptographic proof when used in their applications |
| Discoverability | Make it possible for entities to discover the state of nodes and the network, to learn more about or interact with the nodes and the network |
| Interoperability | Use interoperable standards so the ARPA infrastructure can make use of existing programming language, tools, and software libraries designed for interoperability |
| Portability | Be platform- and network-independent and enable entities to run a node on any operating system with internet connection |
| Simplicity | Favor a modular design where each module is simple to make the technology easier to understand, implement, and deploy |
| Extensibility | Where possible, enable extensibility provided it does not greatly hinder interoperability, portability, or simplicity. |

### 1.2 Architectural Overview

ARPA network relies on a smart-contract-capable Blockchain to manage its dynamic global states. On a high level, these dynamic global states include ___node___ information, ___group___ information, and all BLS signature tasks.

This is achieved by implementing and deploying a single "___controller___" smart contract on a particular Blockchain. And each unregistered ___node___ directly interacts with the "___controller___" to register themselves into the ARPA network and discover the information needed to communicate with other registered ___nodes___.

For higher throughput and better service availability, the ___nodes___ in the ARPA network are split into multiple ___groups___ to handle BLS signature tasks in parallel. The "___controller___" is also responsible for initiating the ___grouping___ of the ___nodes___ and storing the ___group___ information.

The ___grouping___ process is essentially a Distributed Key Generation process (DKG). A "___coordinator___" smart contract is implemented and deployed ad-hoc to coordinate a subset of ___nodes___ through the different phases of the DKG process, then submits the information of the new "___group___" to the "___controller___".

An "___adapter___" smart contract is implemented and deployed to each blockchain to provide BLS services for all smart-contract-capable Blockchains. The "___adapter___" acts as the APIs for other DApp clients to request BLS signatures.

The DApp client requests the BLS signature by calling the "___adapter___" APIs. The "___adapter___" assigns the BLS signature task to a specific ___group___. Each ___grouped node___ monitors the BLS signature task event emitted by the "___adapter___" and starts a BLS signature task if it belongs to the assigned ___group___, then submits the signature to the "___adapter___" upon completion. The "___adapter___" then returns the results to the caller DApps. A backup mechanism is also in place if the assigned ___group___ fails to fulfill the request within a reasonable time.



          +------------------------------------+                                           
          |    Blockchain (Smart Contract)     |                                           
          +------------+-------------+---------+                                           
          | Controller | Coordinator | Adapter |                                           
          +------------+-------------+---------+                                           
                                                                                           
          +------+ +------+ +------+ +------+ -+                                           
          | Node | | Node | | Node | | Node |  |                                           
          +------+ +------+ +------+ +------+  |                                           
          | Node | | Node | | Node | | Node |  +-- Group                                   
          +------+ +------+ +------+ +------+  |                                           
          | Node | | Node | | Node | | Node |  |                                           
          +------+ +------+ +------+ +------+ -+                                           

___Figure A:___ _A high-level architectural view;_
_Please note that_ ___each group___ _can have more than three_ ___nodes___.

## 2. Core

This section describes and defines the fundamental attributes of the ARPA network. It provides crucial information on what the network achieves and the general principles of how it works.

### 2.1 Network Responsibilities

- __To fulfill__ BLS threshold signature and randomness requests
- __To manage__ ___nodes___ for joining and exiting the network
- __To group__ ___nodes___ dynamically via the DKG process
- __To maintain__ the global network states
- __To sustain__ the token economics and handle rewards for the participating ___nodes___
- __To track__ the historical BLS signature tasks
- __To verify__ any BLS signature task completed by the network

### 2.2 Network Composition

The ARPA network consists of many ___groups___ of ___nodes___, and the ___groups___ are formed dynamically by the DKG process.

___Nodes___ --> ___Groups___ --> ARPA Network

The major components of the ARPA network are:

- ___Node___
- ___Controller___
- ___Coordinator___
- ___Adapter___

Note that we do not consider the ___group___ an ARPA network component because it is a transient entity consisting of many ___nodes___.

#### 2.2.1 Node Composition & Responsibilities

- Blockchain Event Listeners
  - Listen to new block events
  - Listen to recent DKG task events
  - Listen to new randomness task events
- Context Cache
  - Maintain the Blockchain information
  - Maintain the DKG private key
  - Maintain the DKG public key
  - Maintain the networking information
  - Track the unhandled events
- DKG Module
  - Run DKG process
  - Detect DKG phase
- BLS Module
  - Sign partial signature
  - Verify partial signature
  - Aggregate partial signatures
  - Verify signature
- Committer Module
  - Commit the BLS signature result

#### 2.2.2 Controller Responsibilities

- Manage network parameters & constraints
  - Node token staking amount (50,000 ARPA)
  - Node disqualification penalty amount (1,000 ARPA)
  - DKG post-process reward amount (100 ARPA)
  - Minimum signature threshold (3)
  - Number of committers (3)
  - Duration of each DKG phase (10 blocks)
  - Maximum group capacity (10 nodes)
  - Number of groups in an equilibrium stage
  - Exit cool-down period (100 blocks)
- Emit DKG task events
- Handle ___node___ registrations
- Handle ___node___ activations
- Handle ___node___ exits
- Receive DKG result commits
- Handle DKG post processes
- Handle unresponsive-group reports

#### 2.2.3 Coordinator Responsibilities

- Manage phases in each particular DKG process
- Manage DKG shares
- Manage DKG responses
- Manage DKG justifications


#### 2.2.4 Adapter Responsibilities

- Manage network parameters & constraints
  - BLS signature task reward (50 ARPA)
  - BLS signature commit reward (100 ARPA)
  - BLS signature incorrect commit penalty (1,000 ARPA)
  - BLS signature result challenge reward (300)
  - BLS signature task time window (30 blocks)
  - BLS signature failure limit (3)
- Emit BLS signature task events
- Handle reward claims
- Handle randomness requests
- Fulfill randomness requests


### 2.3 Network Stages

#### 2.3.1 Bootstrapping Stage

#### 2.3.2 Equilibrium Stage

### 2.4 Network Scaling

#### 2.4.1 Up-scaling

#### 2.4.2 Down-scaling

### 2.N Requests & Tasks

### 2.N Multi-chain Support

## N. Networking

This section describes and defines the communication protocol between the nodes in the ARPA network.

## N. Interface

This section defines the interface of interacting with the ARPA network.

## N. Application & SDKs

This section describes the possible application of the ARPA network and the example SDKs for generating randomness via a BLS signature.

## N. Testing

This section describes the methodology for testing the network and the test scenarios.

## N. Security & Attack Vectors

## N. Terminology

TODO: re-organize the words below

TODO: define each word

### threshold signature

### Boneh-Lynn-Shacham (BLS) digital signature

#### BLS signature task

### distributed key generation (DKG)

#### DKG Phase 1

#### DKG Phase 2

#### DKG Phase 3

#### DKG Phase 4

#### DKG Task

##### threshold

##### DKG task epoch

#### DKG communication private key

#### DKG communication public key

#### partial private key

#### partial public key

#### DKG group public key

#### aggregated public key

### verification

### controller (smart contract)

#### global epoch

### coordinator (smart contract)

### adapter (smart contract)

### block height

### node

#### disqualified node

#### address

#### node state

#### staking

#### node index

#### node epoch

#### committer

#### member

### group

#### group capacity

#### group state

#### group index

#### group epoch

#### grouping

#### re-balance

#### pre-grouping

#### post-grouping

#### group relay

#### group relay task

## N. Cryptographic Protocols

### Aggregate and Verifiably Encrypted Signatures from Bilinear Maps

### Short Signatures from the Weil Pairing

### Compact Multi-Signatures for Smaller Blockchains

### BLS Multi-Signatures With Public-Key Aggregation

### Secure Distributed Key Generation for Discrete-Log Based Crypto-systems

key expiration  

___Components___ are the network-level entities.
___Modules___ are the # ARPA BLS Threshold Signature Network Standards

Core architecture, components, protocol, and token-economics

## Abstract

ARPA BLS Threshold Signature Network (ARPA) is a decentralized (permission-less) network consisting of multiple, dynamic groups of nodes that can securely generate BLS threshold signatures. ARPA can serve as the infrastructure of many Blockchain applications such as verifiable random beacons, secure (keyless) wallets, cross-chain bridges, decentralized custody, etc.

## 1. Introduction

### 1.1 Design Goals

| Goals | Description |
| --- | --- |
| Decentralization | Eliminate the requirement for centralized authorities or single point of failure in nodes/groups management within the network. |
| Control | Give entities, both human and non-human, the power to directly control the generation and usage of the BLS signature without the need to rely on external authorities. |
| Security | Enable sufficient security for requesting parties to depend on the generated BLS signature for their required level of assurance |
| Proof-based | Enable BLS signatures to provide cryptographic proof when used in their applications |
| Discoverability | Make it possible for entities to discover the state of nodes and the network, to learn more about or interact with the nodes and the network |
| Interoperability | Use interoperable standards so the ARPA infrastructure can make use of existing programming language, tools, and software libraries designed for interoperability |
| Portability | Be platform- and network-independent and enable entities to run a node on any operating system with internet connection |
| Simplicity | Favor a modular design where each module is simple to make the technology easier to understand, implement, and deploy |
| Extensibility | Where possible, enable extensibility provided it does not greatly hinder interoperability, portability, or simplicity. |

### 1.2 Architectural Overview

ARPA network relies on a smart-contract-capable Blockchain to manage its dynamic global states. On a high level, these dynamic global states include ___node___ information, ___group___ information, and all BLS signature tasks.

This is achieved by implementing and deploying a single "___controller___" smart contract on a particular Blockchain. And each unregistered ___node___ directly interacts with the "___controller___" to register themselves into the ARPA network and discover the information needed to communicate with other registered ___nodes___.

For higher throughput and better service availability, the ___nodes___ in the ARPA network are split into multiple ___groups___ to handle BLS signature tasks in parallel. The "___controller___" is also responsible for initiating the ___grouping___ of the ___nodes___ and storing the ___group___ information.

The ___grouping___ process is essentially a Distributed Key Generation process (DKG). A "___coordinator___" smart contract is implemented and deployed ad-hoc to coordinate a subset of ___nodes___ through the different phases of the DKG process, then submits the information of the new "___group___" to the "___controller___".

An "___adapter___" smart contract is implemented and deployed to each blockchain to provide BLS services for all smart-contract-capable Blockchains. The "___adapter___" acts as the APIs for other DApp clients to request BLS signatures.

The DApp client requests the BLS signature by calling the "___adapter___" APIs. The "___adapter___" assigns the BLS signature task to a specific ___group___. Each ___grouped node___ monitors the BLS signature task event emitted by the "___adapter___" and starts a BLS signature task if it belongs to the assigned ___group___, then submits the signature to the "___adapter___" upon completion. The "___adapter___" then returns the results to the caller DApps. A backup mechanism is also in place if the assigned ___group___ fails to fulfill the request within a reasonable time.



          +------------------------------------+                                           
          |    Blockchain (Smart Contract)     |                                           
          +------------+-------------+---------+                                           
          | Controller | Coordinator | Adapter |                                           
          +------------+-------------+---------+                                           
                                                                                           
          +------+ +------+ +------+ +------+ -+                                           
          | Node | | Node | | Node | | Node |  |                                           
          +------+ +------+ +------+ +------+  |                                           
          | Node | | Node | | Node | | Node |  +-- Group                                   
          +------+ +------+ +------+ +------+  |                                           
          | Node | | Node | | Node | | Node |  |                                           
          +------+ +------+ +------+ +------+ -+                                           

___Figure A:___ _A high-level architectural view;_
_Please note that_ ___each group___ _can have more than three_ ___nodes___.

## 2. Core

This section describes and defines the fundamental attributes of the ARPA network. It provides crucial information on what the network achieves and the general principles of how it works.

### 2.1 Network Responsibilities

- __To fulfill__ BLS threshold signature and randomness requests
- __To manage__ ___nodes___ for joining and exiting the network
- __To group__ ___nodes___ dynamically via the DKG process
- __To maintain__ the global network states
- __To sustain__ the token economics and handle rewards for the participating ___nodes___
- __To track__ the historical BLS signature tasks
- __To verify__ any BLS signature task completed by the network

### 2.2 Network Composition

The ARPA network consists of many ___groups___ of ___nodes___, and the ___groups___ are formed dynamically by the DKG process.

___Nodes___ --> ___Groups___ --> ARPA Network

The major components of the ARPA network are:

- ___Node___
- ___Controller___
- ___Coordinator___
- ___Adapter___

Note that we do not consider the ___group___ an ARPA network component because it is a transient entity consisting of many ___nodes___.

#### 2.2.1 Node Composition & Responsibilities

- Blockchain Event Listeners
  - Listen to new block events
  - Listen to recent DKG task events
  - Listen to new randomness task events
- Context Cache
  - Maintain the Blockchain information
  - Maintain the DKG private key
  - Maintain the DKG public key
  - Maintain the networking information
  - Track the unhandled events
- DKG Module
  - Run DKG process
  - Detect DKG phase
- BLS Module
  - Sign partial signature
  - Verify partial signature
  - Aggregate partial signatures
  - Verify signature
- Committer Module
  - Commit the BLS signature result

#### 2.2.2 Controller Responsibilities

- Manage network parameters & constraints
  - Node token staking amount (50,000 ARPA)
  - Node disqualification penalty amount (1,000 ARPA)
  - DKG post-process reward amount (100 ARPA)
  - Minimum signature threshold (3)
  - Number of committers (3)
  - Duration of each DKG phase (10 blocks)
  - Maximum group capacity (10 nodes)
  - Number of groups in an equilibrium stage
  - Exit cool-down period (100 blocks)
- Emit DKG task events
- Handle ___node___ registrations
- Handle ___node___ activations
- Handle ___node___ exits
- Receive DKG result commits
- Handle DKG post processes
- Handle unresponsive-group reports

#### 2.2.3 Coordinator Responsibilities

- Manage phases in each particular DKG process
- Manage DKG shares
- Manage DKG responses
- Manage DKG justifications


#### 2.2.4 Adapter Responsibilities

- Manage network parameters & constraints
  - BLS signature task reward (50 ARPA)
  - BLS signature commit reward (100 ARPA)
  - BLS signature incorrect commit penalty (1,000 ARPA)
  - BLS signature result challenge reward (300)
  - BLS signature task time window (30 blocks)
  - BLS signature failure limit (3)
- Emit BLS signature task events
- Handle reward claims
- Handle randomness requests
- Fulfill randomness requests


### 2.3 Network Stages

#### 2.3.1 Bootstrapping Stage

#### 2.3.2 Equilibrium Stage

### 2.4 Network Scaling

#### 2.4.1 Up-scaling

#### 2.4.2 Down-scaling

### 2.N Requests & Tasks

### 2.N Multi-chain Support

## N. Networking

This section describes and defines the communication protocol between the nodes in the ARPA network.

## N. Interface

This section defines the interface of interacting with the ARPA network.

## N. Application & SDKs

This section describes the possible application of the ARPA network and the example SDKs for generating randomness via a BLS signature.

## N. Testing

This section describes the methodology for testing the network and the test scenarios.

## N. Security & Attack Vectors

## N. Terminology

TODO: re-organize the words below

TODO: define each word

### threshold signature

### Boneh-Lynn-Shacham (BLS) digital signature

#### BLS signature task

### distributed key generation (DKG)

#### DKG Phase 1

#### DKG Phase 2

#### DKG Phase 3

#### DKG Phase 4

#### DKG Task

##### threshold

##### DKG task epoch

#### DKG communication private key

#### DKG communication public key

#### partial private key

#### partial public key

#### DKG group public key

#### aggregated public key

### verification

### controller (smart contract)

#### global epoch

### coordinator (smart contract)

### adapter (smart contract)

### block height

### node

#### disqualified node

#### address

#### node state

#### staking

#### node index

#### node epoch

#### committer

#### member

### group

#### group capacity

#### group state

#### group index

#### group epoch

#### grouping

#### re-balance

#### pre-grouping

#### post-grouping

#### group relay

#### group relay task

## N. Cryptographic Protocols

### Aggregate and Verifiably Encrypted Signatures from Bilinear Maps

### Short Signatures from the Weil Pairing

### Compact Multi-Signatures for Smaller Blockchains

### BLS Multi-Signatures With Public-Key Aggregation

### Secure Distributed Key Generation for Discrete-Log Based Crypto-systems

key expiration  

___Components___ are the network-level entities.
___Modules___ are the node-level entities. entities.

Issue #2273, opened 2022-10-16
[BUG] dcc.Dropdown use outdated styles for react-virtualized-select (cursor should be pointer)


olejorgenb:
![image](https://user-images.githubusercontent.com/72201/196037000-51156c03-c2bd-4601-8796-3f56186fcc6b.png)

Cursor should be "pointer". The underlying library [has fixed this](https://github.com/bvaughn/react-virtualized-select/commit/b2c5fe394ec3145319bde37158d05b3508fbf84a) and dash use a version which includes the fix (3.1.3), but the stylesheet is ["hardcoded"](https://github.com/plotly/dash/blob/dev/components/dash-core-components/src/components/css/react-virtualized-select%403.1.0.css) to 3.1.0

```
> conda list | grep dash
dash                      2.6.1              pyhd8ed1ab_0    conda-forge
dash-bootstrap-components 1.2.1              pyhd8ed1ab_0    conda-forge
dash-core-components      2.0.0                    pypi_0    pypi
dash-daq                  0.5.0                    pypi_0    pypi
dash-html-components      2.0.0                    pypi_0    pypi
dash-table                5.0.0                    pypi_0    pypi
```

PR #2277, opened 2022-10-20
Use dropdown styles from node_modules, instead of from stored css file

nickmelnikov82:
Updated styles for dcc.Dropdown to make styles actual. Close https://github.com/plotly/dash/issues/2273 issue.

## Contributor Checklist

### optionals


components/dash-core-components/src/fragments/Dropdown.react.js
```diff
@@ -2,7 +2,7 @@ import {isNil, pluck, without, pick} from 'ramda';
 import React, {useState, useCallback, useEffect, useMemo} from 'react';
 import ReactDropdown from 'react-virtualized-select';
 import createFilterOptions from 'react-select-fast-filter-options';
-import '../components/css/react-virtualized-select@3.1.0.css';
+import '../../../dash-core-components/node_modules/react-virtualized-select/styles.css';
```
alexcjohnson, line 5:
```suggestion
import 'react-virtualized-select/styles.css';
```
does standard module resolution work with CSS files?

alexcjohnson (approved):
💃 Thanks @nickmelnikov82 - looks good! Just one small comment / question, then we need a changelog entry and this will be ready.

components/dash-core-components/src/fragments/Dropdown.react.js
T4rk1n: Yes that works :+1:

PR #2282, opened 2022-10-24
Deps update 221021

alexcjohnson:
Across-the-board depencency upgrades, including:
- plotlyjs `2.13.3` -> `2.16.1`
- browserslist (as a subdependency): by default it now declares IE11 `dead` and excludes it, meaning default output is no longer `es5`. In fact, by default it now goes all the way to `es2020`. However, because we know Dash gets used by a substantial number of people in airgapped or otherwise highly restricted environments, I opted to go for a rule based on assuming laptops have ~7 year lifespan and are at least set up with current software when they're bought. So the browserslist query I settled on is: `"last 7 years and not dead"`. That gives us `es2015` compliant code.


alexcjohnson:
Big pain solving this storybook issue https://github.com/storybookjs/storybook/issues/18276 -> I'm not sure exactly which step fixed it (reinstalling with no npm cache, using latest node/npm were the main two) but it now seems to work. Still two versions of webpack installed in the test setup, that's inherent to storybook at this point, but some important part of it flipped from 4.x to 5.x in https://github.com/plotly/dash/pull/2282/commits/ab30dbe829f885144b747f94d710abd56261ab2d and that resolved the problem.

T4rk1n (approved):
:woman_dancing:

PR #2289, opened 2022-10-26
celery: pin importlib < 5 for py3.7

alexcjohnson:
Bringing the issue noted in https://github.com/plotly/dash-snapshots/issues/248 over to dash.
We don't test dash on Py3.7 specifically, and we already pin `importlib-metadata` for Py3.6 (https://github.com/plotly/dash/pull/1872), so I don't expect this issue to manifest in our tests unfortunately...
cc @T4rk1n


T4rk1n (approved):
:woman_dancing:

Issue #2286, opened 2022-10-26
ERROR: Python 3.11.0 Installation Error


dogukankaratas:
After I updated my Python to 3.11.0, I'm no more capable of installing Dash with pip.

First I got the following error:

![image](https://user-images.githubusercontent.com/61163577/198035581-22a8e6f3-9384-46dd-984a-75906a037085.png)

Then I installed a newer version of C++ BuildTools and now I got another one;

![image](https://user-images.githubusercontent.com/61163577/198035814-9319b99e-1da0-42ac-b8dd-0c6546513855.png)

Any suggestions?

Thanks in advance.

T4rk1n:
brotli has not published wheels for python 3.11, in the meantime new wheels are published, the dependency needs to be compiled manually, for that you need to install the python-dev packages for python 3.11.

dogukankaratas:
Thanks for quick reply. I already installed vs c++ build tools but still got an error. 
![image](https://user-images.githubusercontent.com/61163577/198211765-dcc82de5-6c50-4066-9bff-c181d5b98646.png)

alexcjohnson:
Here's the root issue: https://github.com/google/brotli/issues/984
@dogukankaratas it's a bit beyond our scope to help with compiling dependent packages but maybe someone over there can help, in the event that they don't fix the issue quickly themselves by releasing Py3.11-compatible wheels.

@T4rk1n I wonder though, `compress` defaults to `False` since v2.0 (https://github.com/plotly/dash/pull/1707) - what if we removed `flask-compress` from our required dependencies and just added a note to the `compress` docs like "to use this functionality you must also install the `flask-compress` package (and a clear error if you invoke `compress` without it)? I suppose ideally we would only do that in a major release, would it be too much in a minor?

T4rk1n:
>  compress defaults to False since v2.0 

If it's false by default I think we can put it in extra require (`dash[compress]`) and message in case missing without a major version.

alexcjohnson:
Let's do that then. It'll make our installation lighter in addition to solving this problem for what might turn out to be a lot of folks if the `brotli` maintainers don't act quickly.

alexcjohnson:
#2291 is about all we can do for this from our side.

PR #2291, opened 2022-10-27
Move flask_compress dependency to extras require

T4rk1n:
Following discussion from https://github.com/plotly/dash/issues/2286#issuecomment-1293717772, this change the flask_compress dependency to an extra requires installed with `dash[compress]`. This new requirement is needed when using the `compress` argument of Dash.


alexcjohnson:
Looks great. AFAICT none of our tests load `[compress]`, which is probably a good thing, can we just add a test that triggers the new exception, mainly to verify that we are indeed running them without `flask_compress` present?

Annoying that percy won't even show comparisons when they update FF on their side. But at least this PR has nothing to do with any of them so we can blindly approve.

alexcjohnson (approved):
💃 thanks 😎

PR #2293, opened 2022-10-28
Fix Dropdown useMemo not detecting equal objects

mariusdkm:
This pull requests improves the performance of the dash core component dropdown.
The `useMemo` didn't recognize that the objects inside the `options` array of the dropdown stayed the same, so the `useMemo` got recalculated all the time.
Because of this Dropdowns with a lot of values inside (I have some with >1k) took a lot of time to render(>100ms).
Using the `useRef` the current `options`-object(and the address associated with that) are saved and compared with the new `options`-object is deep compared to that one using `isEqual` from react-fast-compare.
That way the `persistentOptions` is only changed when the options change, and that way the `useMemo` is only executed when needen.

## Contributor Checklist

### optionals


T4rk1n (approved):
:woman_dancing: Looks good and worked well for me with a dropdown of 4k uuids.

PR #2327, opened 2022-11-18
feat: with comprehension pattern

jafournier:
Just a micro-PR to use list and dict comprehension.


alexcjohnson (approved):
💃 Sure, we can do that - nice and Pythonic 🐍

PR #2332, opened 2022-11-22
Add key to wrapped children props in list.

### LINKED ISSUE #2302: [BUG] Error when passing list of components in dash component properties other than children.
```
dash                          2.7.0
dash-core-components          2.0.0
dash-html-components          2.0.0
dash-table                    5.0.0
dash-mantine-components       0.11.0a0
```

**Describe the bug**

When passing components in dash component properties other than `children`, an error is thrown if a list of components is passed. 

```python
from dash import Dash
from dash_iconify import DashIconify

app = Dash(__name__)

app.layout = dmc.Divider(label=["GitHub", DashIconify(icon="fa:github")])


if __name__ == "__main__":
    app.run_server(debug=True)
```

Error:
Screenshot 2022-11-06 at 12 10 52 AM

This behaviour is observed even if a single component is passed in the list:
```python
app.layout = dmc.Divider(label=[DashIconify(icon="fa:github")])
```

**Expected behavior**

No error should be displayed even when multiple components are passed. 

**Screenshots**

NA

T4rk1n:
- Automatically add a key to the wrapped components in a list if it doesn't have an id. Fix #2302 
- Extract memo TreeContainer component to DashWrapper for it to have a name in react dev tools.
- Refactor DashWrapper context consumer to useContext for one less component for each wrapped component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [UNRELEASED]
+
+### Fixed
+
+- [#2332](https://github.com/plotly/dash/pull/2332) Add key to wrapped children props in list.
+
 ## [2.7.0] - 2022-11-03
 
 ### Removed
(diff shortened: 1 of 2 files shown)


T4rk1n:
There is no test for the bug as generating the warning require react dev tools installed on the browser.

alexcjohnson (approved):
💃 Even updating to the `useContext` hook on the way 💪

PR #2338, opened 2022-11-29
Add playsinline attribute / prop to be used in html.Video component

bart-pywalker:
This PR adds the playsinline attribute that is used in the html Video component. Per mdn web docs, playsinline is:
"A Boolean attribute indicating that the video is to be played "inline", that is within the element's playback area. Note that the absence of this attribute does not imply that the video will always be played in fullscreen."

## Contributor Checklist

   -  [ ] test to see if playsinline applied to html video eliminates the "play" button (rotated triangle) when iPhone is on low power mode (and simply serves the first frame of the video as an image)

### optionals


alexcjohnson:
@bart-pywalker thanks for the PR! Looks to me as though it needs to be `playsInline` rather than `playsinline` though, does this work as is?

alexcjohnson:
Ah yes - if it's trying to invoke Rust that means some dependency is trying to compile, rather than installing from a wheel. You probably only need `[dev]` to get it to build and test manually, but not a big deal, we can take over the PR if you like.

Re camelCase: see https://reactjs.org/docs/dom-elements.html and specifically https://stackoverflow.com/questions/68219114/ios-video-playsinline-not-working-in-reactjs

bart-pywalker:
> Ah yes - if it's trying to invoke Rust that means some dependency is trying to compile, rather than installing from a wheel. You probably only need `[dev]` to get it to build and test manually, but not a big deal, we can take over the PR if you like.
> 
> Re camelCase: see https://reactjs.org/docs/dom-elements.html and specifically https://stackoverflow.com/questions/68219114/ios-video-playsinline-not-working-in-reactjs

Ah, I see. Sure, that is fine. I made this request before finding out that dash-player got revived and turned into an actual PIP package and am pretty happy with dash-player, but if this gets added, I could eliminate one library from the build

React, not HTML. Gotcha. yea, will need to be 'playsInline'

components/dash-html-components/scripts/data/attributes.json
```diff
@@ -526,6 +526,12 @@
             ],
             "description": "Provides a hint to the user of what can be entered in the field."
         },
+        "playsInline": {
+            "elements": [
+                "video"
+            ],
+            "description": "Sets the video to be played inline / within the element's playback area."
+        },
```
alexcjohnson, line 534: This file is actually auto-generated by `scripts/extract-attributes.js` by scraping https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes - and unfortunately `playsinline` is missing there, so when I rerun the build process your edits disappear. We could make a special case for it, but better would be to fix it upstream -> https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#see_also ("Found a problem with this page?")

bart-pywalker: Okay, i made an issue request on their github. Now it will just be a waiting game? 

Again, I am happy with dash-player, so there is no hurry on my end. Just trying to better understand the build for dash components

AnnMarieW:
Hi @bart-pywalker,

I was going through old PRs and this one works now! At some point, `playsinline` was added to https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

I'll open a new PR #3534  and reference this one (I don't think I can make changes to yours since I'm not a maintainer).  There is certainly no expectation for you to finish this up after such a long delay, but I noticed this was your first PR. :trophy:   I'll close mine if you decide you want to see yours merged.  It just needs a changelog entry and run the script to extract the attributes.  (if it doesn't do that automatically when you run the build, change to `dash/components/dash-html-components` and run `npm run extract`


Thanks for your contribution to Dash!

AnnMarieW:
closed by #3534

PR #2344, opened 2022-11-30
1519 callbacks waiting fix

eff-kay:
Fixes https://github.com/plotly/dash/issues/1519

the description of what this PR does is found in this comment

https://github.com/plotly/dash/pull/1627#issuecomment-837426077

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals

### DIFF
--- a/tests/integration/callbacks/test_multiple_callbacks.py
+++ b/tests/integration/callbacks/test_multiple_callbacks.py
@@ -6,6 +6,8 @@ import pytest
 from dash import Dash, Input, Output, State, callback_context, html, dcc, dash_table
 from dash.exceptions import PreventUpdate
 
+import dash.testing.wait as wait
+
 
 def test_cbmt001_called_multiple_times_and_out_of_order(dash_duo):
     app = Dash(__name__)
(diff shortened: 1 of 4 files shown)


eff-kay:
@alexcjohnson rebasing https://github.com/plotly/dash/pull/1627, took way longer than expected, so I simply created a new branch. This is way cleaner and easier to test. Marc-Andres comment is pretty much what I am doing to fix the issue, in that I traverse the entire tree and filter out duplicate calls.

alexcjohnson (approved):
💃 Thanks @eff-kay, looks good to me! I refactored your test a little, to sequence based on a `Lock` rather than delays and to look at number of callback invocations to ensure we aren't missing an intermediate state - but I'm pretty sure it still tests the same behavior.

eff-kay:
> 💃 Thanks @eff-kay, looks good to me! I refactored your test a little, to sequence based on a `Lock` rather than delays and to look at number of callback invocations to ensure we aren't missing an intermediate state - but I'm pretty sure it still tests the same behavior.



> 💃 Thanks @eff-kay, looks good to me! I refactored your test a little, to sequence based on a `Lock` rather than delays and to look at number of callback invocations to ensure we aren't missing an intermediate state - but I'm pretty sure it still tests the same behavior.

@alexcjohnson I verified the test, it validates the correct thing. I have one small comment on understanding how locks work, other than that feel free to merge the PR.

tests/integration/callbacks/test_multiple_callbacks.py
```diff
@@ -578,3 +580,91 @@ def callback(*args):
...
+
+    with out_lock:
+        canada_opt.click()
+
+        # all three callbacks have fired once more, but since we haven't allowed the
+        # last one to execute, the output hasn't been changed
+        wait.until(lambda: out_call_count.value == 2, timeout=3)
+        assert opts_call_count.value == 2
+        assert city_call_count.value == 2
+        assert dash_duo.find_element("#display-selected-values").text == new_york_text
```
eff-kay, line 663: how is the `lock` making sure here that the value of the text is not updated to 'Montreal is a city in Canada'. More specifically how is the lock preventing the last one to execute, as per the comment.

alexcjohnson: Only one process / thread can hold the lock at a time. We acquire the lock in the selenium process before clicking on Canada, then we wait until `out_call_count` increases, so we know we've gotten into the final callback (which means the other two have also executed, as we verify in the succeeding assertions). But because that final callback also tries to acquire the lock after incrementing its call count, it's blocked from executing. So we know that the output still says `new_york_text` when all the callbacks that _will_ be triggered have been triggered (as verified by the fact that the call counts don't increase any more below, after the output text has its final value) but the final one just hasn't been allowed to return its response to the renderer.

After that we release the lock, so the dash process is allowed to acquire it, the final callback completes, and the output updates to `canada_text` without any further callbacks triggering, so there can't have been an intermediate state.

PR #2345, opened 2022-11-30
Add license trove classifiers to component packages

johnthagen:
Add proper license [Trove classifiers](https://pypi.org/classifiers/) for component packages.

This is used by tools such as [`pip-licenses`](https://github.com/raimon49/pip-licenses) to provide a license report for users.

### DIFF
--- a/components/dash-core-components/setup.py
+++ b/components/dash-core-components/setup.py
@@ -16,4 +16,7 @@ setup(
     license=package["license"],
     description=package.get("description", package_name),
     install_requires=[],
+    classifiers=[
+        "License :: OSI Approved :: MIT License",
+    ],
 )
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Thanks @johnthagen. These `setup.py` files aren't actually used, we should just delete them. We do currently install stubs of the components as their old top-level imports at https://github.com/plotly/dash-core-components etc, just for backward compatibility with dash 1.x syntax but at this point it's probably about time to remove that.

johnthagen:
@alexcjohnson Understood. I saw these when running`pip-licenses` on an application that depended on `dash` (and brought these components in as separate PyPI packages).

Just so I understand, you're suggesting that installing these top level packages is planned to be removed (so their separate licenses will no longer matter)?

alexcjohnson:
Exactly - as soon as we're confident that all documentation, all sub-packages, and nearly all users have converted from 1.x syntax:
```
import dash_table
```
to the 2.x syntax:
```
from dash import dcc, html, dash_table
```
then we can remove the stub packages `dash_core_components==2.0.0`, `dash_html_components==2.0.0`, `dash_table==5.0.0` from the installation requirements.

johnthagen:
Sounds great. I'll go ahead and close this PR.

PR #2361, opened 2022-12-08
Dependencies upgrade

T4rk1n:


alexcjohnson:
LGTM! Can you add a changelog entry like https://github.com/plotly/dash/blob/dev/CHANGELOG.md#updated that links to each of the included plotly.js releases (2.16.2, 2.16.3, and 2.16.4) saying they "fix several bugs, particularly related to updating mapbox graphs"?

alexcjohnson (approved):
💃 Thanks!

PR #2362, opened 2022-12-09
Wrapper function to avoid polluting global namespace

frnhr:
Wrapped the clientside template JS code in an anonymous function. This prevents `clientside` and `ns` vars from polluting the global namespace.

## Contributor Checklist

   -  [ ] 🦗 
   -  [ ] 🦗🦗 

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
   - `custom_spinner`  enables using a custom component for loading messages instead of built-in spinners
   - `display`  overrides the loading status with options for "show," "hide," or "auto"
 
+## Fixed
+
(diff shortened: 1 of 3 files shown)


alexcjohnson:
Thanks @frnhr - good catch, thanks for the fix!

If you merge or rebase on the tip of `dev` the tests should pass. Then we'll just need:
- a test that these variables are indeed not in the global namespace. Maybe in [test_clientside.py](https://github.com/plotly/dash/blob/dev/tests/integration/clientside/test_clientside.py) pick any test and add a line at the end like `assert dash_duo.driver.execute_script("return !(('clientside' in window) || ('ns' in window))")`
- a changelog entry

frnhr:
PR ready for review.

Not sure what Percy the bot is on about 🤔

T4rk1n (approved):
Looks good! :woman_dancing:

PR #2363, opened 2022-12-09
Updated html attributes for ol

T4rk1n:


alexcjohnson (approved):
💃 Right, I noticed this was building differently now... thanks!

PR #2370, opened 2022-12-21
added static prop to remove event listener

AnnMarieW:
As reported on the forum,  components with event listeners can cause issues with screen readers.

https://community.plotly.com/t/how-to-remove-events-from-dash-html-components-to-avoid-misinterpretation-by-screen-readers/70477

This PR adds a `disable_n_clicks` prop to all html components, giving the ability to remove the event listener.  

By default, `n_clicks` will be enabled on any html component that has an `id`.  If `n_clicks` is not used in a callback for that component, the event listener can be removed by setting `disable_n_clicks`=True`  

For example:

- No event listener:
`html.Div()`
`html.Div(id="my-id", disable_n_clicks=True)`

- onClick event listener:
`html.Div(id="my-id")`

## Contributor Checklist

### optionals


AnnMarieW:
Needs a better name for the prop than `static` since that may imply that no callbacks can be done on the component.  Perhaps it should be `disable_n_clicks` ?

T4rk1n:
> Needs a better name for the prop than `static` since that may imply that no callbacks can be done on the component. Perhaps it should be `disable_n_clicks` ?

I like `disable_n_clicks`,  it's more explicit than `static`.

On another note, I don't think `n_clicks` changes requires a re-render, perhaps add memo with a custom handler or refactor the component back to classes and implement `shouldComponentUpdate`.

CHANGELOG.md
```diff
@@ -2,11 +2,11 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
-## Unreleased
+## [UNRELEASED]
 
-### Updated
+### Added
 
-- [#2367](https://github.com/plotly/dash/pull/2367) Updated the default `favicon.ico` to the current Plotly logo
```
T4rk1n, line 9: The changelog entry for the favicon should be there too.

components/dash-html-components/scripts/generate-components.js
```diff
@@ -263,14 +269,19 @@ const ${Component} = (props) => {
     if(props.loading_state && props.loading_state.is_loading) {
         dataAttributes['data-dash-is-loading'] = true;
     }
+    /* remove unnecessary onClick event listeners  */
+    let isStatic = typeof props.id === "undefined" ? true : false
```
T4rk1n, line 273: There is no need for the `? true : false` as it's the condition itself.
Simplified:
```javascript
const isStatic = props.disable_n_clicks || typeof props.id === "undefined";
```

AnnMarieW: oh yes, that's better.  Thanks!

components/dash-html-components/scripts/generate-components.js
```diff
@@ -263,14 +269,19 @@ const ${Component} = (props) => {
     if(props.loading_state && props.loading_state.is_loading) {
         dataAttributes['data-dash-is-loading'] = true;
     }
+    /* remove unnecessary onClick event listeners  */
+    let isStatic = props.disable_n_clicks || typeof props.id === "undefined";
+    isStatic = typeof props.disable_n_clicks === "undefined" ? isStatic : props.disable_n_clicks
```
T4rk1n, line 274: Need to remove the second line.

T4rk1n (approved):
:woman_dancing:

components/dash-html-components/scripts/generate-components.js
```diff
@@ -263,14 +269,17 @@ const ${Component} = (props) => {
     if(props.loading_state && props.loading_state.is_loading) {
         dataAttributes['data-dash-is-loading'] = true;
     }
-
+    /* remove unnecessary onClick event listeners  */
+    const isStatic = props.disable_n_clicks || typeof props.id === "undefined";
```
alexcjohnson, line 273:
```suggestion
    const isStatic = props.disable_n_clicks || !props.id;
```
NBD but we don't allow `id=""` and as a result in dash-renderer we just check if `id` is truthy for this kind of situation.

alexcjohnson (approved):
Why did the favicon change in this PR? Can you revert that?

After that this looks great! 💃

AnnMarieW:
This pull request was re-created in [2389](https://github.com/plotly/dash/pull/2389) to remove the files included in error.

Issue #2358, opened 2022-12-08
Tooltip direction auto


VelizarVESSELINOV:
**Is your feature request related to a problem? Please describe.**
With the current tooltip, direction information is difficult to define in the code during the hover data event.

**Describe the solution you'd like**
Keep the tooltip automatically in the visible region depending on the view zone.

**Describe alternatives you've considered and additional context**
```Python
direction = "bottom" if y > 1.5 else "top"
```
https://dash.plotly.com/dash-core-components/tooltip
<img width="830" alt="image" src="https://user-images.githubusercontent.com/8209940/206467245-d8fde6d0-b324-45ab-8eda-4cce96edc986.png">
As you can observe 1.5 threshold is not good in all the contexts.

Examples of similar requests:
https://community.plotly.com/t/how-to-position-custom-tooltip-within-visible-window/58819
https://stackoverflow.com/questions/72715452/plotly-dash-how-to-align-the-tooltip-to-the-top-instead-of-top-right/72717007

PR #2381, opened 2023-01-08
Autoposition for tooltip

nickmelnikov82:
Add the "auto" option for the direction props to close https://github.com/plotly/dash/issues/2358 issue.


components/dash-core-components/src/components/Tooltip.react.js
```diff
@@ -220,7 +367,13 @@ Tooltip.propTypes = {
...
      * The side of the `bbox` on which the tooltip should open.
      */
-    direction: PropTypes.oneOf(['top', 'right', 'bottom', 'left']),
+    direction: PropTypes.oneOf(['top', 'right', 'bottom', 'left', 'auto']),
+
+    /**
+     * Query selector for container in which tooltip should be visible.
+     * If not set the tooltip`s parent element will be used.
+     */
+    auto_direction_container: PropTypes.string,
```
alexcjohnson, line 376: I imagine a common ask would be just to make sure the tooltip doesn't go off the screen viewport (which typically won't change mid-hover) - is there a way we could allow that?

components/dash-core-components/src/components/Tooltip.react.js
```diff
@@ -7,20 +7,167 @@ import _JSXStyle from 'styled-jsx/style'; // eslint-disable-line no-unused-vars
...
+            },
+            {
+                direction: 'bottom',
+                value: getRealHiddenPart(
+                    tooltipRect.bottom - parentRect.bottom
+                ),
+            },
+        ];
+
+        return hiddenDirections.reduce((sum, hidden) => sum + hidden.value, 0);
```
alexcjohnson, line 52: This one doesn't use `direction`, right? So could just be an array of 4 numbers, then
```js
return sum(hiddenDirections); // import sum from 'ramda'
```

alexcjohnson: also we might want to normalize by the tooltip size in each direction, ie instead of
```js
tooltipRect.bottom - parentRect.bottom
```
use
```js
(tooltipRect.bottom - parentRect.bottom) / tooltipRect.height
```
Otherwise for example a short but wide tooltip might wind up on top or bottom even if it's mostly cut off and could be mostly visible on the left or right, where more pixels are cut off but that's a smaller fraction of its total area.

alexcjohnson:
@nickmelnikov82 really nice solution! Aside from my comments above, we should just add a test and a changelog entry.

components/dash-core-components/src/components/Tooltip.react.js
alexcjohnson: In fact I guess there are three flavors you might want: just keep it on screen, just keep it in the specified container, or keep it in the intersection of the screen and the specified container.

Not sure what API would be best for this... maybe in addition to query selectors, `auto_direction_container` could accept either `'screen'` or `'screen,<selector>'` - the latter kind of looks like a selector for both "screen" and your container.

Also maybe `'window'` is better than `'screen'`?

PR #2388, opened 2023-01-16
Fix ordering of pmc after subtree update.

### LINKED ISSUE #2368: [BUG] Pattern-Matching Callbacks `ALL` inconsistent order in dash 2.7.1
**Describe your context**

```
dash                        2.7.1
dash-core-components        2.0.0
dash-html-components        2.0.0
```

**Describe the bug**

In `Dash` 2.7.1, the order of *Pattern-Matching* `ALL` Inputs will be inconsistent with the order of the elements in the web page, here's how the same code behaves differently in versions 2.7.0 and 2.7.1:

> 2.7.1

> 2.7.0

The source code for this example is as follows:

```Python
import dash
from dash import html, dcc
from dash.dependencies import Input, Output, State, ALL

app = dash.Dash(__name__)

app.layout = html.Div(
    [
        html.Button(
            'refresh options',
            id='refresh-options'
        ),
        html.Br(),
        html.Div(
            [
                *[
                    dcc.Dropdown(
                        id={
                            'type': 'demo-options',
                            'index': i
                        },
                        placeholder=f'dropdown-{i}',
                        style={
                            'width': '200px'
                        }
                    )
                    for i in range(2)
                ],
                dcc.Dropdown(
                    id={
                        'type': 'demo-options',
                        'index': 2
                    },
                    options=[f'option2-{i}' for i in range(3)],
                    placeholder='dropdown-2',
                    style={
                        'width': '200px'
                    }
                )
            ]
        ),
        html.Br(),
        html.Pre(
            id='selected-values'
        )
    ],
    style={
        'padding': '50px'
    }
)


@app.callback(
    [
        Output(
            {
                'type': 'demo-options',
                'index': 0
            },
            'options'
        ),
        Output(
            {
                'type': 'demo-options',
                'index': 1
            },
            'options'
        )
    ],
    Input('refresh-options', 'n_clicks'),
    prevent_initial_call=True
)
def refresh_options(n_clicks):

    return [
        [f'option0-{i}' for i in range(3)],
        [f'option1-{i}' for i in range(3)]
    ]


@app.callback(
    Output('selected-values', 'children'),
    Input(
        {
            'type': 'demo-options',
            'index': ALL
        },
        'value'
    )
)
def update_selected_values(values):

    return str(values)


if __name__ == '__main__':
    app.run(debug=True)

```
@T4rk1n @alexcjohnson

T4rk1n:
Fix #2368

### DIFF
--- a/dash/dash-renderer/src/actions/paths.js
+++ b/dash/dash-renderer/src/actions/paths.js
@@ -3,9 +3,11 @@ import {
     filter,
     find,
     forEachObjIndexed,
+    insert,
     path,
     propEq,
-    props
+    props,
+    indexOf
 } from 'ramda';
 
 import {crawlLayout} from './utils';
@@ -46,7 +48,14 @@ export function computePaths(subTree, startingPath, oldPaths, events) {
                 const values = props(keys, id);
                 const keyStr = keys.join(',');
                 const paths = (objs[keyStr] = objs[keyStr] || []);
-                paths.push({values, path: concat(startingPath, itempath)});
+                const oldie = oldObjs[keyStr] || [];
+                const item = {values, path: concat(startingPath, itempath)};
+                const index = indexOf(item, oldie);
+                if (index === -1) {
+                    paths.push(item);
+                } else {
+                    objs[keyStr] = insert(index, item, paths);
+                }
             } else {
                 strs[id] = concat(startingPath, itempath);
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Unreleased
 
+## Fixed
+
+- [#2388](https://github.com/plotly/dash/pull/2388) Fix [#2368](https://github.com/plotly/dash/issues/2368) ordering or Pattern Matching ALL after update to the subtree.
+
 ### Updated
 
 - [#2367](https://github.com/plotly/dash/pull/2367) Updated the default `favicon.ico` to the current Plotly logo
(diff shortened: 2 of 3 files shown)


dash/dash-renderer/src/actions/paths.js
```diff
@@ -46,7 +48,14 @@ export function computePaths(subTree, startingPath, oldPaths, events) {
...
                 const paths = (objs[keyStr] = objs[keyStr] || []);
-                paths.push({values, path: concat(startingPath, itempath)});
+                const oldie = oldObjs[keyStr] || [];
+                const item = {values, path: concat(startingPath, itempath)};
+                const index = indexOf(item, oldie);
+                if (index === -1) {
+                    paths.push(item);
+                } else {
+                    objs[keyStr] = insert(index, item, paths);
+                }
```
alexcjohnson, line 58: I can't quite put my finger on it, but I think there's a situation here where either the index you end up inserting at isn't what it should be anymore, or the item looks new - so gets put at the end instead of in the middle - but it's just a modification of the old item.

What I think we implicitly really want is for `objs[keystr]` to wind up sorted by path (ie in increasing value of the first item in the path that differs between any two elements). And if I'm reading this correctly, the only case in which this matters is when we saved some of the old paths [(line 35 above)](https://github.com/plotly/dash/pull/2388/files#diff-c151cd2dc301d2ecf998517c7dad59aebf33b90d530b725b889a37b11ffda958R35) and added more items with the same `keyStr` here.

So perhaps we could do something like save a record of which `keyStr`'s (`oldKeys`) we pre-populated above, then when we get here if we find a `keyStr` in that set we insert the new item in the list before the first item with a later path, if any; otherwise we push the new item on the end.

T4rk1n: I am not sure there is way currently to trigger the situation you are describing, the issue this fix is that the fix for pattern matching callback with component as props reevaluate paths of the parent of the prop and thus was updating the existing pmc indexes. Otherwise currently the subtree is updated entirely from the callback and evaluated properly.

alexcjohnson: OK - I can't construct one either. Maybe partial props will expose something new... we'll see. Just keep this in mind as we go along in case any new bug reports pop up!

alexcjohnson (approved):
💃 looks good, and great test!

PR #2389, opened 2023-01-17
add disable_n_clicks prop to html components

AnnMarieW:
As reported on the forum,  components with event listeners can cause issues with screen readers.

https://community.plotly.com/t/how-to-remove-events-from-dash-html-components-to-avoid-misinterpretation-by-screen-readers/70477

This PR adds a `disable_n_clicks` prop to all html components, giving the ability to remove the event listener.  

By default, `n_clicks` will be enabled on any html component that has an `id`.  If `n_clicks` is not used in a callback for that component, the event listener can be removed by setting `disable_n_clicks`=True`  

For example:

- No event listener:
`html.Div()`
`html.Div(id="my-id", disable_n_clicks=True)`

- onClick event listener:
`html.Div(id="my-id")`

(re-created clean pull request without the botched merge commits from #2370)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Unreleased
 
+
+### Added
+
+- [#2370](https://github.com/plotly/dash/pull/2370) Added `disable_n_clicks` prop to all html components to make it possible to remove onclick event listeners
+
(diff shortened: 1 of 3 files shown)


alexcjohnson (approved):
💃 Thanks @AnnMarieW - and @T4rk1n thanks for the earlier reviews 🎉

PR #2392, opened 2023-01-23
Dash Pages bug fixes and enhancements for dash_pages param and layout module imports

### LINKED ISSUE #2263: [BUG] Dash Pages does not support relative package imports within page modules

### LINKED ISSUE #2312: [BUG] page modules not imported correctly
**Describe your context**
Dash pages for example see https://dash.plotly.com/urls

build from source using dev branch to get the latest
```
git checkout 756562bdbb5a3b7ef48197a4f9c6bfc803fb63e6
```

**Describe the bug**
In my code
```dash.page_registry[module_name]```
gives me access to many useful attributes of my page including function layout() however I can not access any custom methods or classes because the page modules are not loaded properly.

```sys.modules[module_name]```
Does not contain my module.

**Expected behavior**
One simple addition to dash.py will solve this and allow me to use page modules that have been loaded
```diff --git a/dash/dash.py b/dash/dash.py
index bb8327e6..bd43df8f 100644
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -2024,6 +2024,7 @@ class Dash:
                     _pages.PAGE_REGISTRY[module_name]["layout"] = getattr(
                         page_module, "layout"
                     )
+                    sys.modules[module_name] = page_module
 
     @staticmethod
     def _path_to_page(path_id):
```
See for example this discussion https://stackoverflow.com/questions/73060129/how-are-changes-to-sys-modules-propagated

With this fix I can then use getmembers to access my custom methods and classes
```print("functions", getmembers(sys.modules[module_name], isfunction))```

### PR DESCRIPTION
**Pull request by ned2**
# Dash Pages Improvements

## Fixes and enhancements to `dash_pages` param

* `pages_folder` param now accepts an absolute path (previously would throw an exception), bringing it in line with how `assets_folder` behaves
* `pages_folder` param now accepts a `pathlib.Path` instance (previously would throw an exception), bringing it in line with how `assets_folder` behaves
* fixed a bug where `pages_folder` being set to to custom path (ie not the default value `"path"`) did not force `use_pages` to be set to `True` (as indicated should occur in the `Dash` doc string)
* fixed a bug where `pages_folder` values containing `_`, `.`, or uppercase chars resulted in incorrectly derived default `path` values in `dash.page_registry` (eg `my_pages/page.py` would result in a path of `/my-pages/page` rather than `/page`)
* set `app.config.pages_folder` to be read-only, since this param should only be set in the `Dash.__init__()`
* replace a bare `Exception` in `dash._validate` and one in `dash._pages` with appropriate exception subclasses
* create fixture `clear_pages_state` to clear to reset all global Pages state before and after running of all tests that modify that state. this is needed to ensure that tests for Pages features are not broken by state set by other tests (and they don't break subsequent tests)

Dash Community forum thread where the question was asked if there was any reason why `pages_folder` couldn't support absolute paths: https://community.plotly.com/t/set-pages-folder-to-an-absolute-path/71347

## Fixes to imports performed by Dash Pages

The Python modules found inside the `dash_pages` directory that are programatically imported for loading layouts had two problems that this PR also fixes:
* Add imported auto Pages modules to `sys.modules`; fixes #2312 
* When running inside a package and the `name` param of the `Dash` instance is set to `__package__`, the module name of auto imported Pages modules will now be prefixed with the package name.  ie `myproj/pages/page1.py` -> `myproj.pages.page1` (previously would have been `pages.page1`). Fixes #2263

### CODE REVIEW

**Comment by alexcjohnson** [comment]
@ned2 this looks great! Really appreciate all the work you put into tests here as well.

> While working on this, I've also come up with what I think is a fairly clean solution to both https://github.com/plotly/dash/issues/2263 and https://github.com/plotly/dash/issues/2312, which doesn't involve the heavy-handed approach of clearing any global state, as I proposed in https://github.com/plotly/dash/pull/2271 (but does assume the existence of the `clear_pages_state` fixture introduced in this PR).

Meaning you'd use the `clear_pages_state` not in a test context but in a real app? That strikes me as funny, but I'm open to seeing what you have in mind, and sure, it can be part of this same PR.

**Comment by ned2** [comment]
OK @alexcjohnson, have added in the fixes for Dash Pages imports as described in the updated PR description. Ready for review!

> Meaning you'd use the `clear_pages_state` not in a test context but in a real app? That strikes me as funny, but I'm open to seeing what you have in mind, and sure, it can be part of this same PR.

Oh no, yeah that would be weird. I'm only using `clear_pages_state` within the tests, for any test using Dash Pages. The only reason I was clearing Dash internal state in that other PR was to keep the tests suite happy, where multiple invocations of `Dash` within the same process was causing issues. You can see that [in one case](https://github.com/plotly/dash/blob/dev/tests/integration/multi_page/test_pages_layout.py#L182) this was already being managed by reaching directly into the pages registry and manipulating it after a test ran. This problem only became worse with all the new tests and newly prefixed module names. 

Using the new fixture to clear state before and after the test is a little less heavy handed than actually clearing Dash state. (That said, I wonder if there is still a broader  argument to clear Dash's global state to prevent these kinds if possible gotchas)   

btw, looks like something flaky is going on with the tests. Chrome is failing to start up intermittently on some tests in this and the previous run a push triggered.

**Comment by ned2** [comment]
heya @alexcjohnson, just bumping this. would be great to get a review while it's still in my head, if you have capacity sometime soon :)

File: dash/dash.py [hunk 1]
```diff
@@ -431,7 +433,7 @@ def __init__(  # pylint: disable=too-many-statements
         _get_paths.CONFIG = self.config
         _pages.CONFIG = self.config
 
-        self.pages_folder = pages_folder
+        self.pages_folder = str(pages_folder)
         self.use_pages = True if pages_folder != "pages" else use_pages
```
alexcjohnson, line 437: Not new in this PR but since you're fixing some related edge cases... currently if you call:
```py
app = Dash(pages_folder="blah", use_pages=False)
```
the custom `pages_folder` wins and enables pages even though you explicitly said `use_pages=False`. I'm not sure why you'd do that (maybe to disable pages temporarily for debugging?), but it feels backward to me. Maybe we can switch `use_pages` to have a default of `None`, then this logic here can become something like:
```py
self.use_pages = (pages_folder != "pages") if use_pages is None else use_pages
```
Or maybe even centralize all of this logic (after changing the default `use_pages=None`) in `_configs.py`, then these lines can just draw from `config`:
```py
self.pages_folder = config.pages_folder
self.use_pages = bool(config.pages_folder)
```

tests/unit/pages/custom_pages/__init__.py
```diff
@@ -0,0 +1,2 @@
+# this directory is used for testing custom values of `pages_folder` param that
+# need to exist as actual paths.
```
alexcjohnson, line 2: You're adding a bunch of `__init__.py` files to the tests in this PR - are these now required for pages to work or did you just add them for best practice?

alexcjohnson:
@ned2 thanks for the nudge. Just two small questions, then we should be good to go! The first about explicit `use_pages=False` is nonblocking but feels like the right time to address it while we're tightening up the rest of this logic. The second I just want to make sure we aren't accidentally breaking existing apps - I don't see why we would be but worth confirming.

tests/unit/pages/custom_pages/__init__.py
ned2: In general, no, `__init__.py` are not required for Dash pages to work. However Dash Pages now detects when discovered page modules happen to be inside of a package so this it can add the package name as a prefix to the module path. See `_infer_module_name`. Note this relies on configuring Dash like so: `Dash(name=__package__)`

So in this particular case the `__init__.py` needs to be added to make the test app a package for the tests to work. The other instances I just added them in for best-practice.

dash/dash.py
ned2: ahh yep, agreed, would be good to squeeze that in. I'll get onto it!

alexcjohnson (approved):
💃 Thanks @ned2 - this will go in v2.9 🎉

dash/dash.py
ned2: Just a heads up @alexcjohnson that this suggestion didn't end up making it in, perhaps could be worth creating an issue for it?

dash/dash.py
```diff
@@ -436,8 +439,8 @@ def __init__(  # pylint: disable=too-many-statements
         _get_paths.CONFIG = self.config
         _pages.CONFIG = self.config
 
-        self.pages_folder = pages_folder
-        self.use_pages = True if pages_folder != "pages" else use_pages
+        self.pages_folder = str(pages_folder)
+        self.use_pages = (pages_folder != "pages") if use_pages is None else use_pages
```
miraculixx, line 443: this breaks backward compatibility (combined with the change in dash/_configs.py) because effectively use_pages=True behavior is inferred (by checking page_folder instead of use_pages), while the user relies on the previous use_pages=False behavior.

PR #2393, opened 2023-01-23
Fix callback crashing if getPath returns undefined

rnestler:
Fixes a crash which results in a hanging page which was caused by an `Uncaught (in promise) TypeError: paths is undefined` exception raised by `ramda.path`.

## Contributor Checklist

### optionals

### DIFF
--- a/dash/dash-renderer/src/observers/executedCallbacks.ts
+++ b/dash/dash-renderer/src/observers/executedCallbacks.ts
@@ -127,6 +127,9 @@ const observer: IStoreObserverDefinition<IStoreState> = {
                     );
 
                     const basePath = getPath(oldPaths, parsedId);
+                    if (!basePath) {
+                        return;
+                    }
                     const oldObj = path(getPath(oldPaths, parsedId), oldLayout);
 
                     const childrenProps = pathOr(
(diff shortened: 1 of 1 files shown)


T4rk1n:
This look good, not sure it's worth adding a test as the condition to make it happen is a bit convoluted involving multi-page navigation & triggering callbacks at same time.

rnestler:
> This look good, not sure it's worth adding a test as the condition to make it happen is a bit convoluted involving multi-page navigation & triggering callbacks at same time.

Yeah, I couldn't come up with a minimal example that still triggers the error. The project in which this happens uses multi-page navigation and triggers a lot of callbacks.

alexcjohnson (approved):
OK, let's move ahead. In any event short-circuiting won't be worse than throwing an error, but  unless and until we can find the real root cause this may actually be the correct way to proceed. All the code you're preventing from executing with that `return` is dedicated to finding other callbacks that need to run based on a new layout chunk, but it seems like here we're trying to replace a chunk that wasn't there to begin with (must have already been replaced by the results of a different callback), so finding other callbacks based on that chunk is meaningless.

Maybe ^^ could help us try and pin down the cause, but I'm OK with merging as is, and waiting for another bug report with more detail before we try to do better. 💃

PR #2400, opened 2023-01-27
added disable_n_clicks to page_container

AnnMarieW:
As [reported on the forum](https://community.plotly.com/t/using-dash-table-on-multipage-app-leads-to-warning/71054/10?u=annmariew), need to set `disable_n_clicks=True` in `html.Div`s included in  `dash.page_container`


T4rk1n (approved):
:woman_dancing:

PR #2414, opened 2023-02-06
Add Patch callbacks

T4rk1n:
**Add**

- `dash.Patch`: return a Patch instance to do a partial update of the target prop without transferring the previous in a `State`. This object is like a proxy of the value on the frontend, you instantiate a patch and return it in callback: `p = Patch()`. You can set the subprop with attribute access or number index for array and keys which may not be a valid Python variable name: `p.attribute = "attr"` is same as `p["attribute"] = "attr"`.   Available operations:
  - Assign: `p.assigned = "Assigned"`;
  - Append: `p.array.append()
- `allow_duplicate` argument to `Output` to allow multiple callbacks to target the same `Output`.

**Example**
```python

app = Dash(__name__)

app.layout = html.Div(
    [
        html.Div(
            [
                dcc.Input(id="set-value"),
                html.Button("Set", id="set-btn"),
            ]
        ),
        html.Div(
            [
                dcc.Input(id="append-value"),
                html.Button("Append", id="append-btn"),
            ]
        ),
        html.Div(
            [
                dcc.Input(id="prepend-value"),
                html.Button("prepend", id="prepend-btn"),
            ]
        ),
        dcc.Store(
            data={
                "value": "unset",
                "n_clicks": 0,
                "array": ["initial"],
                "delete": "Delete me",
            },
            id="store",
        ),
        html.Div(id="store-content"),
    ]
)

app.clientside_callback(
    "function(value) {return JSON.stringify(value)}",
    Output("store-content", "children"),
    Input("store", "data"),
)

@app.callback(
    Output("store", "data"),
    Input("set-btn", "n_clicks"),
    State("set-value", "value"),
    prevent_initial_call=True,
)
def on_click(_, value):
    p = Patch()
    p.value = value
    p.n_clicks += 1

    return p

@app.callback(
    Output("store", "data", allow_duplicate=True),
    Input("append-btn", "n_clicks"),
    State("append-value", "value"),
    prevent_initial_call=True,
)
def on_click(_, value):
    p = Patch()
    p.array.append(value)
    p.n_clicks += 1

    return p

@app.callback(
    Output("store", "data", allow_duplicate=True),
    Input("prepend-btn", "n_clicks"),
    State("prepend-value", "value"),
    prevent_initial_call=True,
)
def on_click(_, value):
    p = Patch()
    p.array.prepend(value)
    p.n_clicks += 1

    return p
```
patchoutput


tests/unit/test_patch.py
```diff
@@ -0,0 +1,204 @@
...
+from dash._utils import to_json
+
+
+def patch_to_dict(p):
+    return json.loads(to_json(p))
+
+
+def test_pat001_patch_assign_item():
+    p = Patch()
+    p["item"] = "item"
```
alexcjohnson, line 13: minor point... but to play devil's advocate these two `"item"` strings should be different.

tests/unit/test_patch.py
```diff
@@ -0,0 +1,204 @@
...
+    assert data["operations"][0] == {
+        "operation": "Merge",
+        "location": [],
+        "params": {"value": {"merge": "merged"}},
+    }
+
+
+def test_pat011_patch_add():
+    p = Patch()
+    p.added = p.added + 1
```
alexcjohnson, line 133: This is `__add__`, right? Do we actually want to allow this? Because it implies you could do something like `p.added = p.initial + 1` but that doesn't work, does it?

tests/unit/test_patch.py
```diff
@@ -0,0 +1,204 @@
...
+    assert data["operations"][1] == {
+        "operation": "Add",
+        "location": ["plusplus"],
+        "params": {"value": 1},
+    }
+
+
+def test_pat012_patch_sub():
+    p = Patch()
+    _ = p.sub - 1
```
alexcjohnson, line 152: That's an even weirder usage than `p.added = p.added + 1` - so the output doesn't matter at all, just the fact that you asked for `p.sub - 1` causes `prop.sub` to change? That feels unintuitive to me, so like `__add__`, I'm inclined to just drop `__sub__`.

tests/unit/test_patch.py
```diff
@@ -0,0 +1,204 @@
...
+    assert data["operations"][1] == {
+        "operation": "Sub",
+        "location": ["minusless"],
+        "params": {"value": 1},
+    }
+
+
+def test_pat013_patch_mul():
+    p = Patch()
+    _ = p.mul * 2
```
alexcjohnson, line 171: same here, and for `__truediv__`

dash/_patch.py
```diff
@@ -0,0 +1,117 @@
...
+        else:
+            # pylint: disable=consider-using-ternary
+            self._location = (parent and parent._location) or []
+        if parent is not None:
+            self._operations = parent._operations
+        else:
+            self._operations = []
+
+    def __getitem__(self, item):
+        return Patch(location=self._location + [item], parent=self)
```
alexcjohnson, line 29: `__getitem__` and `__setitem__` can receive `slice` objects - I can see that being useful, for inserting or removing elements in the middle of an array, but it generates all sorts of edge cases that might make intuitive sense but not sure we really want to support. I'm thinking of stuff like `Patch()[1::2].val += 1` which I'd interpret as "go into all the odd items and increment their `val` properties by 1"

So maybe for now the safest thing to do is just raise an error if you get here with a `slice`?

dash/_patch.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __delitem__(self, key):
+        self._operations.append(_operation("Delete", self._location + [key]))
+
+    def __add__(self, other):
+        self._operations.append(_operation("Add", self._location, value=other))
+        return _noop
+
+    def __iadd__(self, other):
+        self._operations.append(_operation("Add", self._location, value=other))
```
alexcjohnson, line 70: Does this work for extending a list, ie `Patch() += [1,2,3]`?

dash/_patch.py
```diff
@@ -0,0 +1,117 @@
...
+    def __isub__(self, other):
+        self._operations.append(_operation("Sub", self._location, value=other))
+        return _noop
+
+    def __mul__(self, other):
+        self._operations.append(_operation("Mul", self._location, value=other))
+        return _noop
+
+    def __imul__(self, other):
+        self._operations.append(_operation("Mul", self._location, value=other))
```
alexcjohnson, line 86: In principle I guess list multiplication is also a thing... I can't see that being useful though.

dash/_patch.py
alexcjohnson: Also, do `__getitem__` and `__setitem__` work with negative indices?

tests/unit/test_patch.py
```diff
@@ -0,0 +1,204 @@
...
+    data = patch_to_dict(p)
+
+    assert data["operations"][0]["location"] == ["nest_item", "nested", "deep"]
+    assert data["operations"][1]["location"] == ["nest_attr", "nested", "deep"]
+
+
+def test_pat005_patch_delete_item():
+    p = Patch()
+
+    del p["delete_me"]
```
alexcjohnson, line 64: Does it also work to remove an item from a list? ie `del p[5]` or `del p[-1]`?

dash/_patch.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __itruediv__(self, other):
+        self._operations.append(_operation("Div", self._location, value=other))
+        return _noop
+
+    def append(self, item):
+        self._operations.append(_operation("Append", self._location, value=item))
+
+    def prepend(self, item):
+        self._operations.append(_operation("Prepend", self._location, value=item))
```
alexcjohnson, line 101: `prepend` isn't actually a built-in list method, though it's useful and intuitive so I'm happy to keep it. But let's add `insert(index, item)`, which *is* a built-in method.

Other built-in list methods I don't think we need to implement, but maybe we should stub them so if someone tries to use them they at least get a reasonable error message, and if we implement them later it won't be a breaking change with existing attribute access? The ones I might put in this bucket: `clear`, `pop`, `remove`, `reverse`, `sort`.

dash/_patch.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def extend(self, item):
+        if not isinstance(item, list):
+            raise TypeError(f"{item} should be a list")
+        self._operations.append(_operation("Extend", self._location, value=item))
+
+    def merge(self, item):
```
alexcjohnson, line 108: `merge` isn't a built-in dict method, but `update` is, can we switch to `update` and mimic its behavior? What you have is the beginning of `update` but it can do more, mainly kwargs.

```
update(...) method of builtins.dict instance
    D.update([E, ]**F) -> None.  Update D from dict/iterable E and F.
    If E is present and has a .keys() method, then does:  for k in E: D[k] = E[k]
    If E is present and lacks a .keys() method, then does:  for k, v in E: D[k] = v
    In either case, this is followed by: for k in F:  D[k] = F[k]
```

Dicts have `clear` and `pop` methods that, like lists, I don't think we need to implement but could benefit from the same stub I suggested for lists. Other than that the only one I see we could potentially implement is `__ior__`, which is equivalent to `update` ie `p |= {"new": "items"}`

dash/_patch.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def append(self, item):
+        self._operations.append(_operation("Append", self._location, value=item))
+
+    def prepend(self, item):
+        self._operations.append(_operation("Prepend", self._location, value=item))
+
+    def extend(self, item):
+        if not isinstance(item, list):
+            raise TypeError(f"{item} should be a list")
```
alexcjohnson, line 105: or tuple?

tests/unit/test_patch.py
T4rk1n: Should work it's dissocPath, but let's add a test.

dash/_patch.py
T4rk1n: No this will trigger js undefined behavior, `[1,2] + [2,3] = "1,22,3"`, not sure there should really be multiple ways to do extend operations, maybe only accept numbers and strings for the numerical operations?

dash/_patch.py
T4rk1n: No negative index in js, so no.

alexcjohnson: Can we try to make that work? Feels like it would be really useful for "edit the last thing I added" type of operations.

dash/_patch.py
alexcjohnson: My feeling is we should minimize how much you need to learn - and on the assumption that you already know what Python lists and dicts can do, that means we replicate the built-in behavior of lists and dicts as much as possible (including all the redundant built-in behavior like this), and wherever we can't or don't intend to replicate the built-in behavior we should be explicit about it.

So my first preference would be to just redirect to `extend` if `other` is a list (or tuple), but if you'd rather restrict to just strings and numbers please throw an error that directs users to `extend` if they pass a sequence.

dash/_patch.py
```diff
@@ -0,0 +1,129 @@
...
+        self._operations.append(
+            _operation(
+                "Assign",
+                self._location + [key],
+                value=value,
+            )
+        )
+
+    def __delitem__(self, key):
+        self._operations.append(_operation("Delete", self._location + [key]))
```
alexcjohnson, line 70: Turns out `__delitem__` can receive a slice too, should `validate_slice(key)` here too.

```py
>>> a=[1,2,3,4,5]
>>> del a[1:3]
>>> a
[1, 4, 5]
```

dash/dash-renderer/src/actions/patch.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+function getLocationPath(location: LocationIndex[], obj: any) {
+    const current = [];
+
+    for (let i = 0; i < location.length; i++) {
+        let value = location[i];
+        if (is(Number, value) && value < 0) {
+            const previous: any = path(current, obj);
+            value = previous.length - value;
```
alexcjohnson, line 33:
```suggestion
            value = previous.length + value;
```

dash/_patch.py
T4rk1n: I think `pop` don't make sense since we can't return the value.
`clear` can be emulated with ramda empty: https://ramdajs.com/docs/#empty so let's just add it.

alexcjohnson: Oh nice, `empty` works like `clear` for both lists and dicts - let's do it! Agreed re: `pop`.

dash/_patch.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def __delitem__(self, key):
+        validate_slice(key)
+        self._operations.append(_operation("Delete", self._location + [key]))
+
+    def __iadd__(self, other):
+        if isinstance(other, (list, tuple)):
+            self.extend(other)
+        elif isinstance(other, dict):
+            self.merge(other)
```
alexcjohnson, line 77: Built-in dicts don't merge with `+`, maybe because they're not purely additive, you can overwrite keys? So we could add `__ior__` if you like but I don't think this belongs in `__iadd__`
```py
>>> a = {1:2}
>>> a += {3:4}
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +=: 'dict' and 'dict'
>>> a |= {5:6}
>>> a
{1: 2, 5: 6}
>>> a |= {5:7}
>>> a
{1: 2, 5: 7}

```

dash/_patch.py
T4rk1n: I think `sort` could have interesting use cases, not really sure what kind api we need to customize the sorting tho.
`reverse` an easy one: https://ramdajs.com/docs/#reverse so let's add it too.

dash/_patch.py
```diff
@@ -0,0 +1,140 @@
...
+    def remove(self, item):
+        """filter the item out of a list on the frontend"""
+        self._operations.append(_operation("Remove", self._location, value=item))
+
+    def update(self, E=None, **F):
+        """Merge a dict or keyword arguments with another dictionary"""
+        if E is not None:
+            value = E
+        else:
+            value = dict(F)
```
alexcjohnson, line 133: kind of odd usage, but it turns out you can use both of these together.
```py
>>> a={}
>>> a.update({1:2}, a=4,b=5)
>>> a
{1: 2, 'a': 4, 'b': 5}
```
So maybe:
```suggestion
        value = E or {}
        value.update(F)
```

dash/_patch.py
alexcjohnson: That's true, `sort` would be cool! But yeah, it's going to need a flexible API to be useful, so let's leave that out for now, we can always add it later. Then we should definitely stub it so it won't be a breaking change for anyone accessing a `sort` item inside their prop, they'll need to use `p["sort"]` from the start.

alexcjohnson (approved):
Beautiful. Can't wait to get this into peoples' hands! 💃

FatHare:
Good afternoon. 

Tested, everything works well, except for clientside_callback: there are problems when returning json - nothing happens and there are no errors.

Example of data returned by the function: [
  {"label": "one", "value": "one"},
  {"label": "two", "value": "two"}
]

image

thanks for your work, very useful functionality.

FatHare:
Good afternoon. 

Thanks for allow_duplicate, it's a very nice addition. 
Everything works well, except for **clientside_callback**. Support for clientside_callback was supposed to be? When adding allow_duplicate, an error occurs:

image

If you specify several ouputs, then even though there will be an error, the value will be updated for ouput without allow_duplicate, but not for output with allow_duplicate, example:

image

Tell me, am I doing something wrong? Thank you very much.

PR #2424, opened 2023-02-14
Terminate jobs in diskcache background callback tests teardown.

T4rk1n:
Auto terminate jobs that are still running/zombie at the end of diskcache background tests.

Build passed 10/11, there is still some flakiness with the dcc tests unrelated to background callbacks tests.


alexcjohnson (approved):
Nice solution! 💃

PR #2425, opened 2023-02-14
Move logger to dash namespace, add stream handler once.

### LINKED ISSUE #756: Dash object adds handler to logger creating extra handler
Hi, 

Just ran into this issue that stems from the two conventions of 
```
```
```
app = dash.Dash(__name__)
```
and the line 
https://github.com/plotly/dash/blob/3a1a1c6dbc6cd96040148f927af808c1a60e22f8/dash/dash.py#L217

This results in an extra handler being added to the `__name__` logger which is presumably unwanted by the user. 

I think the convention is to add at most add a `logging.NullHandler()` handler to logging library/packages.  From [Configuring Logging for a Library section](https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library) (though it's Python2 fyi), it states: 

> Note: It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

When I have a moment I would be happy to try and open a small PR  rectifying it if maintainers decide they would want to avoid adding logging handlers in Dash. 

Nice one :)

### LINKED ISSUE #1099: Allow users to override the default logger and handler [Feature Request]
I came across unexpected behavior when attaching Dash instances to a Flask application. [The logger attached to the Dash app](https://github.com/plotly/dash/blob/aac493c23366043c3e9612c3f5624ffacbb5d1e9/dash/dash.py#L249-L250) causes log messages sent to my application's logger (e.g. `current_app.logger.info('foo')`) to be output N+1 times (N=number of Dash applications attached to Flask app).

I was able to work around this by removing the handler added by Dash:

```
my_dash = dash.Dash(__name__,
    server=app,
    url_base_pathname='/my_dash/',
    assets_folder="../dashboards/foo/bar/assets"
)

if (my_dash.logger.hasHandlers()):
    my_dash.logger.handlers.clear()
```

Ideally, I'd either be able to opt out of having Dash attach the handler and/or provide a custom logger/handler via a Dash constructor kwarg. I'd be happy to take a pass at submitting a PR if this is something the maintainers would consider.

T4rk1n:
- Moved the logger namespace to `dash.dash`, as library logger it should be on that namespace instead of the user app.
- Fix multiple log handler added unconditionally to the logger, resulting in duplicate log message. 
- Add `add_log_handler=True` to Dash init, if you don't want a stream handler at all.

Fix #1099, Fix #756

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,17 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 - [#2417](https://github.com/plotly/dash/pull/2417) Add wait_timeout property to customize the behavior of the default wait timeout used for by wait_for_page, fix [#1595](https://github.com/plotly/dash/issues/1595)
 - [#2417](https://github.com/plotly/dash/pull/2417) Add the element target text for wait_for_text* error message, fix [#945](https://github.com/plotly/dash/issues/945)
+- [#2425](https://github.com/plotly/dash/pull/2425) Add `add_log_handler=True` to Dash init, if you don't want a log stream handler at all.
 
 ## Fixed
 
 - [#2417](https://github.com/plotly/dash/pull/2417) Disable the pytest plugin if `dash[testing]` not installed, fix [#946](https://github.com/plotly/dash/issues/946).
 - [#2417](https://github.com/plotly/dash/pull/2417) Do not swallow the original error to get the webdriver, easier to know what is wrong after updating the browser but the driver.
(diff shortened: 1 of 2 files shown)


alexcjohnson (approved):
Thanks, nicely done.
💃 with a changelog entry

PR #2426, opened 2023-02-15
Set default interval to 1 second for app.long_callback

### LINKED ISSUE #2418: [BUG] @app.long_callback has interval set to None
I recently upgraded our dash code (from 2.5.1 to 2.8.0). 

I noticed that our long callbacks using the original `@app.long_callback` decorator is polling our backend many times a second, instead of only once a second.

If I switch over to the new `background=True`, then the 1 second polling is restored. 

I think this is a undesired client regression.

**Describe your context**

```
dash                      2.8.0
```

**Describe the bug**

My frontend application is polling the backend excessively using `@app.long_callback` when the `interval` value is unset.

**Expected behavior**

When a long callback is created using `@app.long_callback`, the default interval of 1 second should be used.

T4rk1n:
Fix #2418

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#2417](https://github.com/plotly/dash/pull/2417) Do not swallow the original error to get the webdriver, easier to know what is wrong after updating the browser but the driver.
 - [#2425](https://github.com/plotly/dash/pull/2425) Fix multiple log handler added unconditionally to the logger, resulting in duplicate log message.
 - [#2415](https://github.com/plotly/dash/pull/2415) Fix background callbacks progress not deleted after fetch.
-- [#2426](https://github.com/plotly/dash/pull/2426) Set default interval to 1 second for app.long_callback.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,17 +15,13 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#2417](https://github.com/plotly/dash/pull/2417) Disable the pytest plugin if `dash[testing]` not installed, fix [#946](https://github.com/plotly/dash/issues/946).
 - [#2417](https://github.com/plotly/dash/pull/2417) Do not swallow the original error to get the webdriver, easier to know what is wrong after updating the browser but the driver.
 - [#2425](https://github.com/plotly/dash/pull/2425) Fix multiple log handler added unconditionally to the logger, resulting in duplicate log message.
+- [#2415](https://github.com/plotly/dash/pull/2415) Fix background callbacks progress not deleted after fetch.
+- [#2426](https://github.com/plotly/dash/pull/2426) Set default interval to 1 second for app.long_callback.
```
alexcjohnson, line 19:
```suggestion
- [#2426](https://github.com/plotly/dash/pull/2426) Set default interval to 1 second for app.long_callback, restoring the behavior it had before v2.6.0 when we introduced `backround=True` callbacks.
```

alexcjohnson (approved):
💃

PR #2429, opened 2023-02-17
Fix side effect on updating possible array children triggering callbacks that were not updated.

### LINKED ISSUE #2411: [BUG] Changing `dcc.Checklist.options` triggers callbacks with inputs `dcc.Checklist.value`
**Describe your context**

```
dash                      2.8.1
dash-bootstrap-components 1.3.1
dash-daq                  0.5.0
```
_Note that this also occurs with Dash 2.8.0, but not 2.7.0._

    - OS: Ubuntu 22.04
    - Browser: Chrome
    - Version: 109.0.5414.119

**Describe the bug**

When outputting to the `options` property of a `dcc.Checklist`, it triggers callbacks with inputs `dcc.Checklist.value`.

**Expected behavior**

Callbacks with input `dcc.Checklist.value` should only be triggered if that checklist's value is explicitly changed (either through callback or user input), _not_ if only the checklist options are changed.

**Minimal reproducible example**

I've tried to create a simple example app to show exactly what the issue is. I tried taking a screen recording though unfortunately Ubuntu is not helping out currently. 

In essence, when running with Dash <= 2.7.0, _only_ selecting/deselecting a city should increment the counter below the checklist. With Dash >= 2.8.0, however, the counter is also incremented when typing in the search bar above the checklist (not desired behaviour)

[DashBugMRE.zip](https://github.com/plotly/dash/files/10573514/DashBugMRE.zip)

T4rk1n:
Fix #2411

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Fixed
 
+- [#2429](https://github.com/plotly/dash/pull/2429) Fix side effect on updating possible array children triggering callbacks, fix [#2411](https://github.com/plotly/dash/issues/2411).
 - [#2417](https://github.com/plotly/dash/pull/2417) Disable the pytest plugin if `dash[testing]` not installed, fix [#946](https://github.com/plotly/dash/issues/946).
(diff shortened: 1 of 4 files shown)


alexcjohnson (approved):
💃 Yeah that works, nice find!

bambinobino:
Has this been included in dash 2.9.2? I am getting the same error in that version, using dbc.Checklist.

Issue #2437, opened 2023-02-28
[BUG] Path template variables not correctly parsed with dot


jowlo:
**Bug description**

Path template variables are not parsed/populated correctly when using a dot (`.`) as seperator between placeholders as in `/dot/<foo>.<bar>/`.

When using `/dot/<foo>.<bar>/` as path template, a request to `/dot/abc.def` should result in a call to `layout(foo='abc', bar='def')`.
Instead there is a call to `layout(foo='abc.de', part=None)`. _(Note the missing `e` at the end)_

Using `/dash/<foo>-<bar>/` as path template, the request url is handled as expected and variables are parsed/populated correctly.

**Expected behavior**

Variables should be parsed as expected and correctly independent of characters between seperators.

**Minimal example** 

```python
import dash
from dash import Dash, html, dcc


app = Dash(__name__, use_pages=True, pages_folder="")
dash.register_page(
    "Dot",
    layout=lambda foo=None, bar=None: html.H1([html.B(foo), html.I(bar)]),
    path_template='/dot/<foo>.<bar>'
)

dash.register_page(
    "Dash",
    layout=lambda foo=None, bar=None: html.H1([html.B(foo), html.I(bar)]),
    path_template='/dash/<foo>-<bar>'
)

app.layout = html.Div(
    [
        html.H1("App"),
        html.Div([
            html.Div(dcc.Link(f"{page['name']} - {page['path']}", href=page["path"]))
            for page in dash.page_registry.values()
        ]),
        html.Hr(),
        dash.page_container,
    ]
)

if __name__ == "__main__":
    app.run_server(debug=True)
```

Environment
```
dash                2.8.1
```

alexcjohnson:
Thanks @jowlo - nice detective work, and your proposal looks good. We would welcome a PR! Try and add a test of the specific bug you encountered with dots in the path; so far the pages tests are all integration tests (mostly in `tests/integration/multi_page`) but this could also be covered just by unit tests (would just be a new file in `tests/unit`)

PR #2440, opened 2023-03-02
Fix path_template variable parsing containing dots

jowlo:
Using dots (`.`) and other special characters as part of a `path_template` for pages is currently not supported. This is due to the fact that `path_template` is passed to Python's `regex` module without escaping first, see #2437.

The PR escapes special regex characters in the `path_template` before parsing it. It also simplifies the way the parsing is done by using `re.match`.

## Contributor Checklist

   -  [x] Fix the `path_template` parsing


jowlo:
Maybe I am missing something here, but it seems that the CI tests are quite flaky...

AnnMarieW:
I don't think it's a bug that you can't use  special characters as variable separators as described in  #2437.

The Pages path template follows the Flask convention for [path variables](https://flask.palletsprojects.com/en/2.2.x/quickstart/#variable-rules) and at this time only supports strings (any text without a slash) in each path segment.  The issue with allowing  special characters such as a dot or a dash as separators, is that those special characters cannot be used in the variable.


In your example, what would happen if `abc` was `1.23`?


> When using` /dot/<foo>.<bar>/` as path template, a request to `/dot/abc.def` should result in a call  to `layout(foo='abc', bar='def'`).


Rather than this custom parsing being handled by dash, I think it would be better to use a path template like `/dot/<foobar>/` then split foobar into two variables inside the layout function of the app.

jowlo:
> I don't think it's a bug that you can't use special characters as variable separators as described in #2437.
> 
> The Pages path template follows the Flask convention for [path variables](https://flask.palletsprojects.com/en/2.2.x/quickstart/#variable-rules) and at this time only supports strings (any text without a slash) in each path segment. The issue with allowing special characters such as a dot or a dash as separators, is that those special characters cannot be used in the variable.

But flask does support this, you can craft arbitrary URLs using flask path variables, the following splits the parameters just fine:

```python
from flask import Flask

app = Flask(__name__)

@app.route('/user/<first>.<last>')
def show_user_profile(first, last):
    # show the user profile for that user
    return f'Hello {first} {last}'

```


> In your example, what would happen if `abc` was `1.23`?
Well, you can always craft URLs that won't match to a template. Thats the same with all separators, including `/`.

> Rather than this custom parsing being handled by dash, I think it would be better to use a path template like `/dot/<foobar>/` then split foobar into two variables inside the layout function of the app.
Yes, that is possible, but leads to a lot of duplicated code if you split this everywhere. In my application I have a few flask endpoints and some dash endpoints and it just struck me as bug that they are not behaving the same.

Maybe the dash pages `path_template` parsing should leverage flask's parsing under the hood?

jowlo:
> What happens in your tests when you pass `foo.bar.test` in the url?
> It looks like Flask sticks all the extra ones in the first variable, and everything else is in reverse order based upon a split of the designated separator.


The code in the PR behaves the same as flask here. The code replaces the `<variable>` blocks with greed capture groups, so the first group will try to grab everything it can.

---

On second thought i dug a bit more into how flask handles routing, and flask uses `werkzeug`s routing underneath. Using that as described here in the [docs](https://github.com/pallets/werkzeug/blob/main/docs/routing.rst) is probably much less error prone (and faster as well).

BSd3v:
I'm wondering if switching to `werkeug` might also help with parsing variables that have been url encoded as well.

Flask can also handle mix and matched separators, heaven forbid, haha.

jowlo:
It would also make it possible to use converters such as `<pagenum:int>` further reducing repeated manual parsing at the top of each pages layout.

alexcjohnson:
I'm in favor of reusing Werkzeug behavior as much as possible. Just be aware that Flask / Werkzeug have a pretty bad record of backward compatibility in areas that we assumed were part of the public API but they choose to remove or alter in minor releases, requiring us to scramble to patch.

So if the code being reused is substantial by all means call Werkzeug, but try to find the highest-level reference possible to minimize the likelihood it moves later. If it's small, like just a regex or 2-line function, please inline it so we match their behavior without a risk of later breakage.

robertclaus:
Closing this as not planned due to inactivity.

PR #2450, opened 2023-03-10
Fix RadioItems & Checklist inline prop.

### LINKED ISSUE #2063: [BUG] `RadioItems` / `Checklist` does not display vertically as documented without stylesheet
**Describe your context**
Python 3.8.6, Windows 10

```
dash                      2.4.0
plotly                    5.8.0
```

    - OS: Windows 10
    - Browser Firefox 100.0

**Describe the bug**

In contrast to the documentation and the minimal examples provided there, the default behaviour of `RadioItems` and `Checklist` is a horizontal display of the options (https://dash.plotly.com/dash-core-components/checklist and https://dash.plotly.com/dash-core-components/radioitems describe the vertical display to be the default case). According to https://stackoverflow.com/questions/62137479/plotly-dash-dcc-radioitems-vertical-alignment a vertical behaviour is achieved using the stylesheet, yet the minimal examples given in the documentation do not incorporate this stylesheet.

Additionally, a new switch to change to horizontal behaviour has been added (`inline=True`). It would be great to either have a corresponding switch for vertical behaviour in case no stylesheet is used, or otherwise at least mention the stylesheet in the documentation.

**Expected behavior**

Display should follow minimal example code given in documentation.

T4rk1n:
Fix #2063 
Set label style `display: block` if `inline` is false in RadioItems & Checklist components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [UNRELEASED]
 
 ## Breaking
-- [#2450](https://github.com/plotly/dash/pull/2450) Set label style `display: block` if `inline` is false in RadioItems & Checklist components. To keep previous behavior, set `inline=True`.
+- [#2450](https://github.com/plotly/dash/pull/2450) Set label style `display: block` if `inline` is false in RadioItems & Checklist components. To keep previous behavior, set `inline=True`. This is already how it was described and worked in our documentation and other places with CSS stylesheets that set the default orientation of RadioItems and Checklist options to vertical (including Dash Design Kit), but for unstyled pages it is a breaking change.
 
 ## Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [UNRELEASED]
 
+## Breaking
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [UNRELEASED]
 
+## Breaking
+- [#2450](https://github.com/plotly/dash/pull/2450) Set label style `display: block` if `inline` is false in RadioItems & Checklist components. To keep previous behavior, set `inline=True`.
```
alexcjohnson, line 8:
```suggestion
- [#2450](https://github.com/plotly/dash/pull/2450) Set label style `display: block` if `inline` is false in RadioItems & Checklist components. To keep previous behavior, set `inline=True`. This is already how it was described and worked in our documentation and other places with CSS stylesheets that set the default orientation of RadioItems and Checklist options to vertical (including Dash Design Kit), but for unstyled pages it is a breaking change.
```

alexcjohnson (approved):
💃 Hopefully this change doesn't upset too many people, but it does seem like the least disruptive thing we can do at this point!

components/dash-core-components/tests/test_inline.py
alexcjohnson, line 1: Oh you know what, I don't think this test ran. I think it needs to be moved into `tests/integration/misc` to match the pattern in the CI config `circleci tests glob "tests/integration/**/test_*.py"`

PR #2471, opened 2023-03-20
Fix allow_duplicate output with clientside callback.

### LINKED ISSUE #2467: [BUG] allow_duplicate not working with clientside_callback
Thanks for allow_duplicate, it's a very nice addition. 
Everything works well, except for **clientside_callback**. Support for clientside_callback was supposed to be? When adding allow_duplicate, an error occurs:

image

If you specify several ouputs(with and without allow_duplicate), then even though there will be an error, the value will be updated for ouput without allow_duplicate, but not for output with allow_duplicate, example:

image

Tell me, am I doing something wrong? Thank you very much.

Full code sample (with one callback, but I think this is enough to show the error):

```python
import dash

from dash import Dash, html, Input, Output

app = Dash(__name__)
app.layout = html.Div(
    children=[
        html.Div(
            children=['Last pressed button: ', html.Span(id='span', children='empty')]
        ),
        html.Button(
            id='button-right',
            children='right'
        )
    ]
)

# NOT WORKING FOR "SPAN", BUT WARKING FOR BUTTON-RIGHT
dash.clientside_callback(
    """
    function(n_clicks){
        return ["right", `right ${n_clicks}`];
    }
    """,
    [
        Output('span', 'children', allow_duplicate=True),
        Output('button-right', 'children')
    ],
    Input('button-right', 'n_clicks'),
    prevent_initial_call=True
)

# WORKING EXAMPLE TO UNDERSTAND HOW IT SHOULD BE (THE ONLY DIFFERENCE IS THAT NO ALLOW_DUPLICATE)
# dash.clientside_callback(
#     """
#     function(n_clicks){
#         return ["right", `right ${n_clicks}`];
#     }
#     """,
#     [
#         Output('span', 'children'),
#         Output('button-right', 'children')
#     ],
#     Input('button-right', 'n_clicks'),
#     prevent_initial_call=True
# )

if __name__ == '__main__':
    app.run_server(debug=True, port=2414)
```

_Originally posted by @FatHare in https://github.com/plotly/dash/issues/2414#issuecomment-1473088764_

T4rk1n:
Fix #2467

### DIFF
--- a/tests/integration/test_patch.py
+++ b/tests/integration/test_patch.py
@@ -434,13 +434,13 @@ def test_pch005_clientside_duplicate(dash_duo):
     )
 
     app.clientside_callback(
-        "function onClickOne() { return 'click1';}",
+        "function() { return 'click1';}",
         Output("output", "children", allow_duplicate=True),
         Input("click1", "n_clicks"),
         prevent_initial_call=True,
     )
     app.clientside_callback(
-        "function onClickTwo(){ return 'click2';}",
+        "function() { return 'click2';}",
         Output("output", "children", allow_duplicate=True),
         Input("click2", "n_clicks"),
         prevent_initial_call=True,
--- a/tests/integration/test_patch.py
+++ b/tests/integration/test_patch.py
@@ -372,3 +372,35 @@ def test_pch003_patch_children(dash_duo):
 
     dash_duo.wait_for_text_to_equal("#new-child", "new-child")
     dash_duo.wait_for_text_to_equal("#initial", "init")
+
+
+def test_pch004_clientside_duplicate(dash_duo):
+    app = Dash(__name__)
+
+    app.layout = html.Div(
+        [
+            html.Button("Click 1", id="click1"),
+            html.Button("Click 2", id="click2"),
+            html.Div("initial", id="output"),
+        ]
+    )
+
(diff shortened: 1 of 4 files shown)


tests/integration/test_patch.py
```diff
@@ -420,3 +420,35 @@ def on_click(_):
...
+    )
+
+    app.clientside_callback(
+        "function onClickOne() { return 'click1';}",
+        Output("output", "children", allow_duplicate=True),
+        Input("click1", "n_clicks"),
+        prevent_initial_call=True,
+    )
+    app.clientside_callback(
+        "function onClickTwo(){ return 'click2';}",
```
chriddyp, line 443: Would a more comprehensive test be to have an anonymous function? To test the `uuid.uuid4.hex()` thing as above. So like `() => { return 'click2' }`?

chriddyp (approved):
💃🏼 once my comment is considered

tests/integration/test_patch.py
T4rk1n: Not really, the function name is actually discarded/unimportant when the script inline, I had fixed on the rendered first but the test still failed so I tried a few thing like adding name on the function to found out where it was getting the name and replaced with the hex.

PR #2472, opened 2023-03-21
Use object for webpack output library config.

### LINKED ISSUE #2294: replace `libraryTarget` in webpack config
Please use [output.library.type](https://webpack.js.org/configuration/output/#outputlibrarytype) instead of [output.libraryTarget](https://webpack.js.org/configuration/output/#outputlibrarytarget) as they might drop support for `output.libraryTarget` in the future.

https://github.com/plotly/dash/blob/a9eb3434023880c229882dd2aacb7d501e7eb4d2/components/dash-core-components/webpack.config.js#L53

T4rk1n:
Resolve #2294


alexcjohnson (approved):
💃 This probably exists across all our components as well, and should be updated next time we edit any of them... also the boilerplate.

PR #2473, opened 2023-03-21
Add callback id to long callback key generation.

### LINKED ISSUE #2221: [BUG] Background callbacks with different outputs not working
**Describe your context**

```
dash                      2.6.1
dash-bootstrap-components 1.2.0
```

**Describe the bug**

Background callbacks don't work if they are generated with a function/loop. For example, I've created a function `gen_callback` that creates a new callback given a css id.

```python
def gen_callback(css_id, x):
        @app.callback(
            Output(css_id, 'children'),
            Input('my-dropdown', 'value'),
            background=True,
        )
        def callback_name(value):
            print(f"Inside callback_name {self._css_id} for {org_id}")

            return int(value) + x
```

The background callback manager uses a hash function to know the key to get. This hash function just takes into account the code of the function, but not the variables used to generate that function.

https://github.com/plotly/dash/blob/c897b2b094543e930a862fe51d48abfb78057df7/dash/long_callback/managers/__init__.py#L101-L105

I think the hash should also take into account the `Output` list of the callback.

**Expected behavior**

Using a variable as the id of one of the outputs creates multiple different, valid callbacks, that should work fine with background=True.

**Possible temporary solution**

Generate a function with `exec()`, replacing parts of a template code, and decorate the compiled function.

T4rk1n:
Fix #2221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Fixed
 
-- [#2479](https://github.com/plotly/dash/pull/2479) Fix `KeyError` "Callback function not found for output [...], , perhaps you forgot to prepend the '@'?" issue when using duplicate callbacks targeting the same output. This issue would occur when the app is restarted or when running with multiple `gunicorn` workers.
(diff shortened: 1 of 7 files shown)


chriddyp (approved):
💃🏼 once a test is added

PR #2474, opened 2023-03-21
Fix some build errors not failing build.

### LINKED ISSUE #2220: [BUG] CI swallowing some errors
Specifically in lint-unit - for example [this run](https://app.circleci.com/pipelines/github/plotly/dash/3978/workflows/6436f15e-3401-4e8b-8dfc-de6d0d1c569a/jobs/65966) has two:

`dash-html-components/scripts/extract-elements.js` (during `npm run build.sequential`) fails with:

```
Error: Unexpected number of elements extracted from https://developer.mozilla.org/en-US/docs/Web/HTML/Element -
Found 123 but expected 125
Check the output and edit expectedElCount if this is intended.
```

and then `black` doesn't run:
```
> private::lint.black
> if [[ ${PYVERSION:-python39} != python36 ]]; then black dash tests --exclude metadata_test.py --check; fi

sh: 1: [[: not found
```

Why are these not resulting in the run failing, and how can we ensure all such errors DO fail the run?

T4rk1n:
Fix #2220


T4rk1n:
unrelated to the changes in this PR, es-check started failing the html build with error:
```
          ES-Check Error:
          ----
          · erroring file: dash_html_components/dash_html_components.min.js
          · error: SyntaxError: Unexpected token (1:5036)
          · see the printed err.stack below for context
          ----
```

Looking in the build at 1:5036 is `var d=n=>{`

alexcjohnson (approved):
💃 Thanks - shell scripts are always confusing, this looks like a nice way to ensure it's cross-platform and robust.

PR #2479, opened 2023-03-28
Deterministic duplicate outputs

### LINKED ISSUE #2480: [BUG] Duplicate callback outputs not working when deployed with `gunicorn`
**Describe your context**

Bug only present when app is deployed via Dash Enterprise. 

```
dash                          2.9.1
dash-ag-grid                  2.0.0a4
dash-core-components          2.0.0
dash-design-kit               1.6.8
dash-extensions               0.0.65
dash-html-components          2.0.0
dash-leaflet                  0.1.23
dash-mantine-components       0.12.0
dash-table                    5.0.0
```

**Describe the bug**

Callbacks with duplicate outputs don't work reliably when my app is deployed with `gunicorn` (ie. DE), giving errors like this: 

```
KeyError: "Callback function not found for output '..image-options.rowData...geojson.data...satellite-img.children@e635ba05b4e74958834c65e29d0a3207...classified-img.children@b954a7e5a88643d29e99412073ede694...image-options.selectedRows...delete-notify.children..', perhaps you forgot to prepend the '@'?"
```

There is definitely the appropriate decorator on my function. The callbacks behave as expected when I run the app locally, both with Python and with gunicorn. About 10% of the time the callback will run fine. 

**Expected behavior**

I'd expect this `KeyError` not to occur when the app is deployed.

T4rk1n:
Duplicate outputs we're inserted with a random uuid, when the server restarted or used with gunicorn multi workers it would have different values for callback ids resulting in KeyError for callbacks.

This PR change the behavior to be deterministic, creating md5 hash from the callback inputs to use in lieu of the uuid4.

Fix #2480

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Fixed
 
-- [#2479](https://github.com/plotly/dash/pull/2479) Fix KeyError when duplicate outputs id being different on restarts.
+- [#2479](https://github.com/plotly/dash/pull/2479) Fix `KeyError` "Callback function not found for output [...], , perhaps you forgot to prepend the '@'?" issue when using duplicate callbacks targeting the same output. This issue would occur when the app is restarted or when running with multiple `gunicorn` workers.
 
 ## [2.9.1] - 2023-03-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [UNRELEASED
+
+## Fixed
+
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [UNRELEASED
+
+## Fixed
+
+- [#2479](https://github.com/plotly/dash/pull/2479) Fix KeyError when duplicate outputs id being different on restarts.
```
chriddyp, line 9:
```suggestion
- [#2479](https://github.com/plotly/dash/pull/2479) Fix `KeyError` "Callback function not found for output [...], , perhaps you forgot to prepend the '@'?" issue when using duplicate callbacks targeting the same output. This issue would occur when the app is restarted or when running with multiple `gunicorn` workers.
```

chriddyp (approved):
💃🏼

PR #2491, opened 2023-03-31
Fix clientside inline function name.

### LINKED ISSUE #2488: [BUG] Inline clientside_callback scripts not found in Dash 2.9.2
**Environment**
```
dash                 2.9.2
```

**Describe the bug**

In Dash 2.9.2, the clientside callbacks defined inline get a random function name which seems to break. Note: the issue solves itslef when refreshing the page, however every first load triggers this issue. Tested in 2.9.1 and this does not happen, it was introduced with the random function names in 2.9.2.

Using a deterministic function name instead solves the issue:

```python
import itertools
COUNTER = itertools.count()

def register_clientside_callback(...):
    ...

    increment = next(COUNTER)
    function_name = f"inline-script-{increment:06d}"
    ...
```

**Expected behavior**

The clientside_callbacks defined inline should work from the first load.

**Screenshots**

image

T4rk1n:
Fix #2488


RenaudLN:
Nice fix, thanks :)

alexcjohnson (approved):
💃 very nice. Just needs a changelog entry 😎

T4rk1n:
The test works fine when running alone, but is unstable when running tests on the whole file.

PR #2498, opened 2023-04-05
Make `Patch` pickleable

joouha:
Hello,

This makes `Patch` objects picklable by adding `__getstate__` and `__setstate__` methods.

This allows caching of callbacks which return `Patch` objects.

Currently pickling `Patch` objects results in a `TypeError("'Patch' object is not callable")` exception.


T4rk1n (approved):

Can you add a changelog entry?

The failing tests in renderer seems unrelated

joouha:
Done!

alexcjohnson:
Oh interesting - is this for using `Patch` in a background callback? Or is there some other reason to pickle it?

joouha:
@alexcjohnson  My use case is simply so I can cache the result of a callback (as per the memoization example [in the docs](https://dash.plotly.com/performance#memoization)) - this was failing if the callback returned a `Patch`. 

But I guess this would also allow background callbacks to return `Patch`es too :-)

PR #2501, opened 2023-04-07
Update license year

T4rk1n:


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2022 Plotly, Inc
+Copyright (c) 2023 Plotly, Inc
```
alexcjohnson, line 3: I still think we should 🔪 the year entirely, but this is fine for now.

alexcjohnson (approved):
💃

PR #2508, opened 2023-04-16
Fix: validate error message

mariusdkm:
This PR fixes the error message, when the wrong number of outputs are present on a callback.

Currently
```
dash.exceptions.InvalidCallbackReturnValue: Invalid number of output values for {"index":["ALL"],"type":"output"}.children item 0.
Expected 5, got 4
output spec: [{'id': {'index': 0, 'type': 'output'}, 'property': 'children'}, {'id': {'index': 1, 'type': 'output'}, 'property': 'children'}, {'id': {'index': 2, 'type': 'output'}, 'property': 'children'}, {'id': {'index': 3, 'type': 'output'}, 'property': 'children'}]
output value: [['None'], ['None'], ['None'], ['None'], ['None']]
```

But the `Expected 5, got 4` is wrong should be the other way around (`Expected 4, got 5`)

<details><summary>Example code that throws the error</summary>
<p>

```py
import dash
from dash import dcc, html, Output, Input, ALL

app = dash.Dash(__name__)

app.layout = html.Div(
    [
        html.Div(
            id={"type": "output", "index": x},
            style={"border": "1px solid black", "width": "100px", "height": "100px"},
        ) for x in range(4)
    ] +
    [
        html.Div(dcc.Input(id="input", type="number"))
    ]
)


@app.callback(
    Output({"type": "output", "index": ALL}, "children"),
    Input("input", "value")
)
def display_output(value):
    return [[str(value)]] * 5


if __name__ == "__main__":
    app.run_server(debug=True)

```

</p>
</details> 

### optionals

### DIFF
--- a/dash/_validate.py
+++ b/dash/_validate.py
@@ -161,49 +161,49 @@ def validate_and_group_input_args(flat_args, arg_index_grouping):
     return func_args, func_kwargs
 
 
-def validate_multi_return(outputs_list, output_value, callback_id):
-    if not isinstance(output_value, (list, tuple)):
+def validate_multi_return(output_lists, output_values, callback_id):
+    if not isinstance(output_values, (list, tuple)):
         raise exceptions.InvalidCallbackReturnValue(
             dedent(
                 f"""
                 The callback {callback_id} is a multi-output.
                 Expected the output type to be a list or tuple but got:
-                {output_value!r}.
+                {output_values!r}.
                 """
             )
         )
 
-    if len(output_value) != len(outputs_list):
+    if len(output_values) != len(output_lists):
         raise exceptions.InvalidCallbackReturnValue(
             f"""
             Invalid number of output values for {callback_id}.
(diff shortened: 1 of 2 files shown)


dash/_validate.py
```diff
@@ -201,7 +201,7 @@ def validate_multi_return(outputs_list, output_value, callback_id):
                     dedent(
                         f"""
                         Invalid number of output values for {callback_id} item {i}.
-                        Expected {len(vi)}, got {len(outi)}
+                        Expected {len(outi)}, got {len(vi)}
```
T4rk1n, line 204: Thanks for catching this.

Look like these were inverted by mistake, probably because the variable name is not specific enough. Do you think you can also change the name of the `outi` and `vi` variables to `output_spec` and `output_value` like the error message ?

T4rk1n (commented):
Looks good, can you add a changelog entry? Thanks.

T4rk1n (approved):
:woman_dancing: thanks for cleaning the variables names!

PR #2513, opened 2023-04-24
Raise error when iterating over patch objects.

### LINKED ISSUE #2512: [BUG] Exception when property of patched_fig is viewed
I know that it is currently not supported to view properties of `patch_fig=Patch()` but when e.g. iterating over trace names like so:
```
 for trace in patched_fig['data']:
        print(trace['name'])
```
no exception or error message is thrown but an endless stream of 

```
...
<dash._patch.Patch object at 0x7f3b89a89b80>
<dash._patch.Patch object at 0x7f3b8305c0a0>
<dash._patch.Patch object at 0x7f3b89a89b80>
<dash._patch.Patch object at 0x7f3b8305c0a0>
<dash._patch.Patch object at 0x7f3b89a89b80>
<dash._patch.Patch object at 0x7f3b8305c0a0>
...
```
This is not exactly intended right?

I got there by trying to delete a trace of patched_fig by its name which otherwise appears not be possible (or is it?)

T4rk1n:
Fix #2512

### DIFF
--- a/dash/_patch.py
+++ b/dash/_patch.py
@@ -99,6 +99,9 @@ class Patch:
         self.update(E=other)
         return _noop
 
+    def __iter__(self):
+        raise TypeError("Iterating over patch objects is prohibited.")
+
     def append(self, item):
         """Add the item to the end of a list"""
         self._operations.append(_operation("Append", self._location, value=item))
--- a/dash/_patch.py
+++ b/dash/_patch.py
@@ -102,6 +102,9 @@ class Patch:
     def __iter__(self):
         raise TypeError("Patch objects are write-only, you cannot iterate them.")
 
+    def __repr__(self):
+        return f"<write-only dash.Patch object at {self._location}>"
+
     def append(self, item):
         """Add the item to the end of a list"""
         self._operations.append(_operation("Append", self._location, value=item))
--- a/dash/_patch.py
+++ b/dash/_patch.py
@@ -100,7 +100,7 @@ class Patch:
         return _noop
 
     def __iter__(self):
-        raise TypeError("Iterating over patch objects is prohibited.")
+        raise TypeError("Patch objects are write-only, you cannot iterate them.")
 
     def append(self, item):
         """Add the item to the end of a list"""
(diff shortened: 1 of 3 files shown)


dash/_patch.py
```diff
@@ -99,6 +99,9 @@ def __ior__(self, other):
         self.update(E=other)
         return _noop
 
+    def __iter__(self):
+        raise TypeError("Iterating over patch objects is prohibited.")
```
alexcjohnson, line 103:
```suggestion
        raise TypeError("Patch objects are write-only, you cannot iterate them.")
```
Just to make clear _why_ it's prohibited.

alexcjohnson:
@T4rk1n what's the issue with raising on `__repr__`? Not seeing a problem in a simple test but I haven't run the full test suite on it. Anyway raising on `__iter__` is nice, we should probably also do `__contains__` (currently `5 in Patch().x` is an infinite loop) and maybe `__bool__` (a patch is always truthy), `__eq__`, `__ne__` (it will not equal the value it has in the actual prop) - ... but those are all very specific usage cases, it would be nice to do something better for folks just trying to print a piece of the prop.

Actually perhaps better than raising on `__repr__` would be just a more descriptive string, something like:
```
<write-only dash.Patch object at ['layout', 'title', 'font', 'color']>
```

Another related thing I'm noticing, if someone tries to use part of a prop to set a different part, like this:
```py
patched_figure = Patch()
# seems reasonable, make the figure square?
patched_figure["layout"]["height"] = patched_figure["layout"]["width"]
return patched_figure
```
it leads to infinite recursion inside `clean_to_json_compatible` in the `plotly.py` JSON encoder - any way we can break that?

T4rk1n:
> it leads to infinite recursion inside clean_to_json_compatible in the plotly.py JSON encoder - any way we can break that?

Think that will require more thought, maybe check the value in setattr if it's another patch and prevent that?

T4rk1n:
> and maybe __bool__

Equality comparators may be used with a filtering system so let's leave them out.

alexcjohnson (approved):
💃 Probably more we could do here, but you're right we should be cautious in case we add other functionality later.

PR #2516, opened 2023-04-28
Lock flask version

T4rk1n:
Recent versions of flask/werkzeug have introduced breaking bugs in dash. This PR lock the versions of flask/werkzeug installed in order to prevent further breakage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#2508](https://github.com/plotly/dash/pull/2508) Fix error message, when callback output has different length than spec
 - [#2207](https://github.com/plotly/dash/pull/2207) Fix object of components support.
 
+## Changed
+
+- [#2516](https://github.com/plotly/dash/pull/2516) Lock Flask/Werkzeug version to 2.2.2 for Python>=3.7 and 2.0.3 for Python < 3.7.
+
 ## [2.9.3] - 2023-04-13
 
 ## Fixed
--- a/requires-install.txt
+++ b/requires-install.txt
@@ -1,4 +1,7 @@
-Flask>=1.0.4
+Flask==2.2.2;python_version>="3.7"
+Werkzeug==2.2.2;python_version>="3.7"
+Flask==2.0.3;python_version<"3.7"
+Werkzeug==2.0.3;python_version<"3.7"
 plotly>=5.0.0
 dash_html_components==2.0.0
 dash_core_components==2.0.0


alexcjohnson:
Forum post discussing the motivations for this PR: https://community.plotly.com/t/dash-is-planning-to-lock-flask-werkzeug-versions/74829

moorepants:
I saw the recent forum post and just wanted to express that I don't think this is a good idea. The general reason is explained in this blog post: https://iscinumpy.dev/post/bound-version-constraints/. Dash is used as a library when we build applications and we need to be able to mix versions of dependencies as flexibly as possible. Making a decision like this may seem like a good idea from the perspective of a dash developer, but for users it will cause headaches. I think a better approach is to use the same `>=` pin you used before but bump the version of flask up and make dash robust to a set of flask versions. I maintain the dash conda package and this will also make it virtually impossible to install dash alongside other packages that depend on flask.

alexcjohnson:
Thanks @moorepants - I pointed to exactly that post last year in https://github.com/plotly/jupyter-dash/issues/87#issuecomment-1087871645 when this same issue came up with Flask/Werkzeug 2.1. We've really tried hard to stick to that stance, but it's now been three minor versions in a row that caused different breaking changes for Dash. We need to acknowledge that the great majority of Dash users don't care what version of Flask they're using - many don't even know they're using Flask. So this is causing lots of users' code to break unnecessarily, causing us ongoing headaches trying to help them and scramble to update our code, and only helping the small fraction of Dash users who use Dash within another Flask app.

We're investigating what we can do to make life easier for our majority use case while still allowing Flask to float for the Dash-in-a-Flask-app use case (with the caveat that you're on your own if you end up with a version different from the officially supported one). It might mean a separate package (`dash-core`?) that doesn't pin Flask, while the primary `dash` package does. We're open to any ideas you have, but one way or another the status quo is not sustainable.

moorepants:
The idea I have is to follow the same method as most all other Python libraries. Pin with `>=` and if a new release of a dependency breaks Dash, then update Dash to deal with the change. If a Dash user experiences a break when they update dependencies in their environment, then they can make an `==` or `<=` pin there until Dash is patched. If Dash users don't want their app to break, they shouldn't update dependencies.

alexcjohnson:
That’s exactly what we have been doing until now. But with regards to Flask specifically that philosophy is not working.

T4rk1n:
Discovered the shapes TypedDict are not exporting properly on the component to be used in custom types definition like callbacks inputs and returns value.

ie:
```
@callback(Input("graph", "figure"), ... )
def my_callback(figure: dcc.Graph.Figure) -> dcc.Graph.Figure:
    ...
```

For the shapes to be used in that way, the shapes must be defined on the component:

```
class C(Component):
    class Shape(TypedDict):
        ...
```

T4rk1n:
Closing in favor of #2538

PR #2530, opened 2023-05-15
Merge JupyterDash with Dash

T4rk1n:
Merge [JupyterDash](https://github.com/plotly/jupyter-dash) functionality in dash to render dash apps in jupyter notebooks.

Simply run app inside a jupyter cell to render automatically on the cell output:
jupdash

## Changed from jupyter dash

- Default display mode changed to `inline`.
- Add mode `tab`, automatically open the app in a new browser tab.
- Set the default mode and inline exception options on the jupyter_dash instance:
```python
from dash import jupyter_dash

jupyter_dash.default_mode = "external"
jupyter_dash.inline_exceptions = False
```


janosh:
Exciting work! Thanks @T4rk1n, looking forward to using this.

@plotly/dash-jupyterlab/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@plotly/dash-jupyterlab",
+  "version": "0.4.3",
+  "description": "A JupyterLab extensions for rendering Plotly Dash apps",
```
alexcjohnson, line 4:
```suggestion
  "description": "A JupyterLab extension for rendering Plotly Dash apps",
```

@plotly/dash-jupyterlab/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@plotly/dash-jupyterlab",
```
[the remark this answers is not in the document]
T4rk1n: It uses the tar bundle included in the Python package, that needs to be built the first time you open jupyterlab. Not sure if the npm package is used for anything, the jupyterlab docs says it can be directly to jupyterlab with npm but I am not sure the user really want to do that. https://jupyterlab.readthedocs.io/en/latest/extension/extension_tutorial.html#publishing-your-extension

I changed the name to avoid conflict if jupyter-dash is installed.

dash/_jupyter.py
```diff
@@ -0,0 +1,496 @@
...
+    import nest_asyncio
+
+    import requests
+
+    _dash_comm = Comm(target_name="dash")
+    _dep_installed = True
+except ImportError:
+    _dep_installed = False
+    _dash_comm = None
+    get_ipython = None
```
alexcjohnson, line 37:
```suggestion
    get_ipython = lambda: None
```
Would simplify [usage](https://github.com/plotly/dash/pull/2530/files#diff-bf535dc399d2e2b82f75acf026efb0491402489814f432b433c3698bc61c2539R206) especially if we drop the bare [`import IPython`](https://github.com/plotly/dash/pull/2530/files#diff-bf535dc399d2e2b82f75acf026efb0491402489814f432b433c3698bc61c2539R26) (which I'm surprised the linter didn't complain about?) and reuse it in the [other](https://github.com/plotly/dash/pull/2530/files#diff-bf535dc399d2e2b82f75acf026efb0491402489814f432b433c3698bc61c2539R96) places we [call it](https://github.com/plotly/dash/pull/2530/files#diff-bf535dc399d2e2b82f75acf026efb0491402489814f432b433c3698bc61c2539R115)

dash/__init__.py
```diff
@@ -39,5 +39,17 @@
     page_container,
 )
 from ._patch import Patch  # noqa: F401,E402
+from ._jupyter import jupyter_dash  # noqa: F401,E402
```
alexcjohnson, line 42: Do we need this in the main namespace? Will users ever need to use it directly?

dash/__init__.py
```diff
@@ -39,5 +39,17 @@
...
 
 ctx = callback_context
+
+
+def _jupyter_nbextension_paths():
```
alexcjohnson, line 47: What is this function for?

T4rk1n: It's for loading the basic extension that gets the proxy settings from comm. Not really sure if it's useful, I didn't cases in testing where it was actually needed, maybe some older hub version.

dash/__init__.py
alexcjohnson: cc @LiamConnors to make sure these options get documented

dash/__init__.py
alexcjohnson: OK, we can leave it there for the time being... and if we don't find a purpose, it starts with `_` so we can remove it later without calling it breaking.

alexcjohnson (approved):
💃 just a style point https://github.com/plotly/dash/pull/2530#discussion_r1228611136, this looks great!

MP-MaximilianLattka:
Is an additional extension required to be able to open dash as a jupyterlab-tab? 

On a clean setup, I'm trying `app.run(jupyter_mode="jupyterlab")`  without success. To clarify, I'm not running jupyter-lab v3, but already v4.

https://dash.plotly.com/dash-in-jupyter#display-modes

I'm uncertain whether to open an issue for my concern.

Coding-with-Adam:
hi @MP-MaximilianLattka 
We highlight recommend using Dash 2.11 or higher with JupyterLab 3. We have encountered a few issues running Dash with JupyterLab version 4.

bjlittle:
Hi @Coding-with-Adam,

As @MP-MaximilianLattka mentions above, I'm also having similar issues with `dash` and the `jupyter_mode="jupyterlab"` mode, which executes but doesn't open a new tab.

I've simply been attempting to get the [minimal dash app](https://dash.plotly.com/minimal-app) to work in this mode, but sadly to no avail.

I've pinned back `jupyterlab<4` but this makes no difference.

Can you recommend a combination other than `conda create -n dash -c conda-forge dash pandas "jupyterlab<4"` that will work for the `minimal dash app` and for the `jupyter_mode=jupyterlab` mode?

Many thanks

P.S., Are there other similar live issues that you know of that I can track?

Coding-with-Adam:
This might be a bug. 
@T4rk1n can you please take a quick look? I installed Jupyterlab 3 and ran the app code below. All jupyter modes work except for the `jupyter_mode="jupyterlab"` . Even after I restart the kernel, it doesn't do anything. It remains stuck as if the code cannot be executed.

```
from dash import Dash, html, dcc, callback, Output, Input
import plotly.express as px
import pandas as pd

df = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/gapminder_unfiltered.csv')

app = Dash(__name__)

app.layout = html.Div([
    dcc.Dropdown(df.country.unique(), 'Canada', id='dropdown-selection'),
    dcc.Graph(id='graph-content')
])

@callback(
    Output('graph-content', 'figure'),
    Input('dropdown-selection', 'value')
)
def update_graph(value):
    dff = df[df.country==value]
    return px.line(dff, x='year', y='pop')

if __name__ == '__main__':
    app.run(jupyter_mode="jupyterlab")
```

bjlittle:
Also, you'll notice that once you start the app, no jupyter tab is opened, but you can hit the URL of the app in another browser window and it connects and renders.

PR #2555, opened 2023-06-01
Fix browser back button with multiple dcc.Location components in layout

### LINKED ISSUE #1312: [BUG] No callback to dcc.Location when two or more dcc.Location are placed in layout

jowlo:
Multiple dcc.Location were replacing the others event handlers for 'popstate'. If any dcc.Location was unmounted, there was no event handler left

fix #1312

## Contributor Checklist

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [2.10.3] - 2032-06-XX
+
+## Fixed
+
+- [#2555](https://github.com/plotly/dash/pull/2555) Fix browser back button when removing one of multiple location components from layout, fix [#1312](https://github.com/plotly/dash/issues/1312)
+
+
 ## [2.10.2] - 2023-05-31
 
 ## Changed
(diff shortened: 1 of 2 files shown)


T4rk1n (approved):
Looks good :woman_dancing:

PR #2556, opened 2023-06-01
Allow to load pages from package

rnestler:
This allows to load the pages from an imported python package which is recursively scanned. This is useful if one compiles an application using Cython where the modules are then not .py files anymore, but compiled .so files.

In our application we compile everything with Cython in the end, to get some performance gains. Since the current implementation of the pages feature only searches for `.py` files in the `pages` folder I worked around it in our application using this approach.

Would this be a feature of general interest? IMO it is nice, since the implementation is way simpler and it uses built in tools to load the modules.

## Contributor Checklist

   -  [ ] Agree that this is a useful feature
   -  [ ] Agree on the naming of the new parameter
   -  [ ] Add documentation for the new parameter
   -  [ ] Add more tests

### optionals


gvwilson:
Related: #2968 2968

rnestler:
So any feedback on this?

T4rk1n:
@AnnMarieW  Can you take a look at this?

AnnMarieW:
Would the solution in this PR work?  https://github.com/plotly/dash/pull/2392

It's been available for a while but didn't get documented.

> When running inside a package and the name param of the Dash instance is set to __package__, the module name of auto imported Pages modules will now be prefixed with the package name. ie myproj/pages/page1.py -> myproj.pages.page1 (previously would have been pages.page1). Fixes https://github.com/plotly/dash/issues/2263

rnestler:
> Would the solution in this PR work? https://github.com/plotly/dash/pull/2392

I don't think so. It still seems to search for `.py` files, just loading them differntly. In my case there *aren't* any `.py` files available, since we compile it with Cython.

> I think we could use the walk for the regular pages and not add another pages_package in the dash init if that works for both regular files.

I think this should work as long as there is an `__init__.py` in the folder where the pages are.

But I'm not currently actively working on the project were we needed this, so I don't have a lot of capacity to work on this.

Issue #2564, opened 2023-06-14
[BUG] wrong sorting of registered pages by ordering


gothicVI:
Thank you so much for helping improve the quality of Dash!

We do our best to catch bugs during the release process, but we rely on your help to find the ones that slip through.

**Describe your context**
Please provide us your environment, so we can easily reproduce the issue.
```
dash                              2.10.2
dash-ag-grid                      2.1.0
dash-bootstrap-components         1.4.1
dash-core-components              2.0.0
dash-html-components              2.0.0
dash-table                        5.0.0
```

**Describe the bug**
Hi,

I have an app where I add pages via:
```
register_page(
    __name__,
    path='/xxx',
    order=n,
    name="yyy",
    title="zzz"
)
```
where everything works fine as long as I have only 10 pages, i.e., n = 0,...,9.

As soon as I add an eleventh page and n turns into a two digit number, the new pages is entered as the third one, i.e., the sorting becomes: 0, 1, 10, 2, ..., 9.
This can also be seen by printing the page_registry in the global layout where I generate the navigation bar.

The error is due to the string cast here: https://github.com/plotly/dash/blob/dev/dash/_pages.py#L361

A workaround is to replace the above line with
```
key=lambda i: (str(i.get("order", i["module"])).zfill(9), i["module"]),
```
which prepends the order integer return with zeros to ensure correct sorting.
It is a dirty workaround as it would brake again when we have more then 100000000 pages but who would do that...

PR #2565, opened 2023-06-14
Update _pages.py - fix sorting for > 10 pages

gothicVI:
Fix https://github.com/plotly/dash/issues/2564

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals


dash/_pages.py
```diff
@@ -358,7 +358,7 @@ def register_page(
     # Sort by order and module, then by module
     for page in sorted(
         PAGE_REGISTRY.values(),
-        key=lambda i: (str(i.get("order", i["module"])), i["module"]),
+        key=lambda i: (str(i.get("order", i["module"])).zfill(9), i["module"]),
```
alexcjohnson, line 361: Thanks @gothicVI! This seems like it would fix a lot of common cases, feels a little arbitrary though, and IDK if anyone does this but it would also prevent using fractional orders, in case you want to squeeze a new page between two existing pages.

Also, I just had a chat with @AnnMarieW about this, if you provide a string for `order` it will interleave with `module` names - is that what we want? Or do we want everything that has a defined `order` to come before anything that doesn't? @AnnMarieW points out that if we change that it could be considered a breaking change, but I think we could argue this is also a bug and we should fix it.

My proposal then would be we order as:
- all defined numeric orders
- then all defined string orders
- with module as a secondary sort only when orders are equal.

which we could write as perhaps:
```suggestion
        key=lambda i: (
            "order" not in i,  # False (order given) sorts before True
            i["order"] if isinstance(i.get("order"), (int, float)) else float("inf"),
            str(i.get("order")),
            i["module"],
        ),
```
Whatever we end up with, let's extend [test_pages_order](https://github.com/plotly/dash/blob/dev/tests/integration/multi_page/test_pages_order.py) to lock in the fixed behavior.

gothicVI: Hi @alexcjohnson, hi @AnnMarieW,

sure that would work as well.
You should be able to add it - if not I'll commit the suggestion.

How to go about the change in https://github.com/plotly/dash/blob/dev/tests/integration/multi_page/test_pages_order.py?
Do we just add more pages? - I'll suggest something :smile:

gothicVI:
@alexcjohnson @AnnMarieW
I've included the proposed change and added some test pages for

* \>10 pages
* fractional ordering
* no order given

I hope these cover the changes well.

tests/integration/multi_page/test_pages_order.py
```diff
@@ -24,6 +23,35 @@ def test_paor001_order(dash_duo, clear_pages_state):
...
+    dash.register_page(
+        "multi_layout14",
+        layout=html.Div("text for multi_layout14", id="text_multi_layout14"),
+        id="multi_layout14",
+    )
+    dash.register_page(
+        "multi_layout15",
+        layout=html.Div("text for multi_layout15", id="text_multi_layout15"),
+        id="multi_layout15",
+    )
```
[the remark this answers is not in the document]
gothicVI: I've just pushed a test for the first point you raised.
I'm not sure how to go about the second one. Isn't that sort of already done by `multi_layout14` and `multi_layout15`?
These don't have an `order` defined but I'm not sure I understand what you'd like to see with the three modules you mentioned... Maybe you could quickly add them yourself?

tests/integration/multi_page/test_pages_order.py
```diff
@@ -24,6 +23,42 @@ def test_paor001_order(dash_duo, clear_pages_state):
...
+    # Test string for order
+    dash.register_page(
+        "multi_layout16",
+        layout=html.Div("text for multi_layout16", id="text_multi_layout16"),
+        order="2a",
+        id="multi_layout16",
+    )
```
alexcjohnson, line 61:
```suggestion
        "first_string_order",
        layout=html.Div("text for string_order", id="text_string_order"),
        order="2a",
        id="first_string_order",
    )
    dash.register_page(
        "aaa_first_with_no_order",
        layout=html.Div("text for aaa", id="text_aaa"),
        id="aaa_first_with_no_order",
    )
    dash.register_page(
        "yyy_last_string_order",
        layout=html.Div("text for yyy", id="text_yyy"),
        order="zzz",
        id="yyy_last_string_order",
    )
```

tests/integration/multi_page/test_pages_order.py
```diff
@@ -48,6 +83,20 @@ def test_paor001_order(dash_duo, clear_pages_state):
...
+        "multi_layout5",
+        "multi_layout6",
+        "multi_layout7",
+        "multi_layout8",
+        "multi_layout9",
+        "multi_layout10",
+        "multi_layout11",
+        "multi_layout12",
+        "multi_layout13",
+        "multi_layout16",
```
alexcjohnson, line 97:
```suggestion
        "first_string_order",
        "yyy_last_string_order",
        "aaa_first_with_no_order",
```

tests/integration/multi_page/test_pages_order.py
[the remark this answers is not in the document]
alexcjohnson: Added a couple more, hopefully more or less self-documenting.

@AnnMarieW all of these `register_page` calls include an `id` that matches `module` but I can't tell what purpose this serves?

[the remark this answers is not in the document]
alexcjohnson: interesting, so it's actually [not behaving as I thought it would](https://app.circleci.com/pipelines/github/plotly/dash/4654/workflows/e86d52a8-d1b7-44d9-9a78-ea71969f373e/jobs/78065/tests):
```
  -  'yyy_last_string_order',
     'aaa_first_with_no_order',
     'multi_layout14',
     'multi_layout15',
     'pages.defaults',
     'pages.metas',
     'pages.not_found_404',
     'pages.page1',
     'pages.page2',
     'pages.path_variables',
     'pages.query_string',
     'pages.redirect',
  +  'yyy_last_string_order',
```
ie pages with strings for `order` appear to be still interleaving with the module names. Maybe we're OK with that and can adjust the test to match, or maybe we need to fix the behavior - either way we should make this a conscious decision.

dash/_pages.py
```diff
@@ -358,7 +358,12 @@ def register_page(
     # Sort by order and module, then by module
     for page in sorted(
         PAGE_REGISTRY.values(),
-        key=lambda i: (str(i.get("order", i["module"])), i["module"]),
+        key=lambda i: (
+            "order" not in i,  # False (order given) sorts before True
+            i["order"] if isinstance(i.get("order"), (int, float)) else float("inf"),
+            str(i.get("order")),
```
alexcjohnson, line 364: OK I see what was happening, the entry in `page_registry` always has `order`, it's just `None` if it wasn't provided - so `"order" not in i` was always `False`.
```suggestion
            i["order"] is None,  # False (order given) sorts before True
            i["order"] if isinstance(i["order"], (int, float)) else float("inf"),
            str(i["order"]),
```

dash/_pages.py
```diff
@@ -358,7 +358,12 @@ def register_page(
     # Sort by order and module, then by module
```
alexcjohnson, line 358:
```suggestion
    # Sort numeric orders first, then string orders, then no order,
    # finally by module name for matching orders
```

tests/integration/multi_page/test_pages_order.py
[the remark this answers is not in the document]
alexcjohnson: Fixed in https://github.com/plotly/dash/pull/2565/commits/ae8ab8d16d782002496dd9014af9270beb83842f - as described in my now-hidden comment:
> OK I see what was happening, the entry in `page_registry` always has `order`, it's just `None` if it wasn't provided - so `"order" not in i` was always `False`.

@AnnMarieW to my mind this is better behavior, and can be considered 100% bugfix, would you agree?

@gothicVI the last thing we'll need in this PR is a changelog entry.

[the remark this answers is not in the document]
gothicVI: @alexcjohnson done - also squashed all of the commits to allow for a proper merge

AnnMarieW:
@gothicVI Thanks for posting about this issue on the forum and creating this pull request.  It looks great!

@alexcjohnson  regarding the question about [id in the page_registry:](https://github.com/plotly/dash/pull/2565#discussion_r1231750424)

The `id` is not used in this test, but it is used in others in this folder.

Typically when assigning an `id` to components like buttons and links by looping through the `dash.page_registry` I like to use an `id` prop rather than using `module`.  I think it makes the code more readable, and also if you change a file name in the pages folder it won’t break callbacks.  

However, for tests, I think it would have been simpler - and better to just use the `module`.  I could make a separate PR to refactor this if you like.

alexcjohnson:
Thanks for the explanation of `id` @AnnMarieW - not a big deal, you're welcome to change it but it's not hurting anything.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [2.10.3] - 2023-xx-yy
```
alexcjohnson, line 5:
```suggestion
## [UNRELEASED]
```
Will probably be 2.11.0 given the other things about to merge, but best to leave with no number until it's actually in a release regardless.

alexcjohnson (approved):
💃 Thanks very much @gothicVI!

PR #2573, opened 2023-06-21
Use `julia --project` in `JuliaRunner`

etpinard:
Currently the `JuliaRunner` spawns a Dash.jl server using the `julia` executable with no option. This means the `JuliaRunner` expects the Julia test dependencies to be installed in the default environment, located e.g. in `~/.julia/environments/v1.9/`.

While installing the Julia test dependencies in the default environment might be ok in a test container, it is not the [recommended way to develop Julia packages](https://pkgdocs.julialang.org/v1/environments/#Creating-your-own-environments). 

Using the `julia --project` in the server-spawning command, makes the `JuliaRunner` use the test dependencies from the [Dash.jl package environment](https://github.com/plotly/Dash.jl/blob/dev/Project.toml), the same test dependencies we used in the Dash.jl unit tests.

## Contributor Checklist

### optionals


alexcjohnson (approved):
💃 Please add a changelog entry and we'll merge :)

PR #2589, opened 2023-07-06
Bugfix: scope CSS of input elements

KoolADE85:
This PR adds a class name to `input` elements so that the CSS for those inputs can be contained within the dash app and not affect non-Dash inputs.

## Contributor Checklist

### optionals


components/dash-core-components/src/components/Input.react.js
```diff
@@ -66,13 +66,15 @@ export default class Input extends PureComponent {
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
+                className={`dash-input ${this.props.className || ''}`}
```
T4rk1n, line 69: This leave an extra space if there is no className.

components/dash-core-components/tests/integration/input/test_input_basics.py
```diff
@@ -72,3 +73,34 @@ def test_inbs002_user_class(dash_dcc):
...
+    """
+
+    app.layout = html.Div(
+        className="test-input-css",
+        children=[dcc.Input(id="DashInput", required=True, className="unittest")],
+    )
+
+    dash_dcc.start_server(app)
+
+    external_input = dash_dcc.driver.find_element(By.ID, "ExternalInput")
```
T4rk1n, line 100: You can use `dash_dcc.find_element("#ExternalInput")` instead of the driver and By object.

components/dash-core-components/src/components/css/input.css
```diff
@@ -1,7 +1,7 @@
-input:invalid {
+input.dash-input:invalid {
     outline: solid red;
 }
 
-input:valid {
+input.dash-input:valid {
     outline: none black;
 }
```
T4rk1n, line 7: Was already there, but we could add a newline.

components/dash-core-components/src/components/Input.react.js
KoolADE85: Addressed in 745f0f3

components/dash-core-components/tests/integration/input/test_input_basics.py
KoolADE85: Addressed in 745f0f3

components/dash-core-components/src/components/Input.react.js
```diff
@@ -66,13 +66,15 @@ export default class Input extends PureComponent {
                 data-dash-is-loading={
                     (loading_state && loading_state.is_loading) || undefined
                 }
+                className={['dash-input', this.props.className].join(' ')}
```
T4rk1n, line 69: This still puts an empty space at the end if undefined. The most proper way I can see to do it, would be to instantiate an array with `['dash-input']` and add `className` if not undefined then join the array or filter is probably enough for such a case with only 2 elements.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## UNRELEASED
+
+## Fixed
+
+- [dash-embedded-component #83](https://github.com/plotly/dash-embedded-component/issues/83) CSS for input elements not scoped to Dash application
```
T4rk1n, line 9: We usually also add the PR link to the changelog

KoolADE85: Done! 👍

T4rk1n (approved):
:woman_dancing: Looks good.

components/dash-html-components/scripts/data/attributes.json
```diff
@@ -54,15 +54,6 @@
             ],
             "description": "Indicates whether controls in this form can by default have their values automatically completed by the browser."
         },
-        "autoFocus": {
```
alexcjohnson, line 57: weird, `autoFocus` disappeared from the reference docs? Yes: https://github.com/mdn/content/pull/27734 - the PR mentions this page: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes which (a) we don't inspect at all during the build process, (b) implies `autoFocus` applies to a lot more than just these four element types.

This seems like it would be a breaking change - I don't actually know if this attribute works in a Dash app, we may be adding our elements too late in the page lifecycle for this to be honored, but even if it doesn't work we shouldn't break an app that already includes it.

More broadly, I don't like the fact that in every PR our tests are at the mercy of MDN's changes, we should really make "update the HTML spec from MDN" an optional thing that we do occasionally, maybe when we're updating dependencies or maybe even less frequently. @T4rk1n thoughts?

components/dash-core-components/src/components/Input.react.js
```diff
@@ -61,18 +61,22 @@ export default class Input extends PureComponent {
...
                 }
+                className={className}
                 ref={this.input}
                 onBlur={this.onBlur}
                 onChange={this.onChange}
                 onKeyPress={this.onKeyPress}
                 {...valprops}
                 {...omit(
                     [
+                        'className',
```
alexcjohnson, line 79: FYI we use this a lot but I'd love to get rid of this `...omit` pattern pretty much wherever we use it. It causes various problems, like I bet we're currently improperly forwarding `persistence`/`persistence_type`/`persisted_props` here, and some other places we still forward `loading_state` so the DOM ends up with `loading_state="[object Object]". Instead we should figure out which props we DO want and [pick](https://ramdajs.com/docs/#pick) them, or just individually forward them. I thought I had made an issue for this but not seeing it now 🤔

components/dash-html-components/scripts/data/attributes.json
T4rk1n: I think we should change the generation to be more stable. The test allow us to see there is change, but maybe it should be a CI job that run once a month or so and create a PR if there is change.

An alternative could be to change the scraper and instead use the React types that are contained in `JSX.IntrinsicElements`. That would change only when the react version is changed and should more stable.

KoolADE85: It looks like the changes to `attributes.json` don't actually need to be committed now. I reverted them, and the build/tests still function without them.
However, the change to `extract-elements.js` does need to be here in order to complete a build.
So I'd propose decoupling `attributes.json` from this PR, and we can think about this as a separate fix.

alexcjohnson: I think we'll need to address this prior to a release regardless, as that process will call the full rebuild. But yeah for the sake of this PR we can ignore it.

alexcjohnson (approved):
💃 This looks great, please either before or after merging make new issues for both removing `omit` for prop forwarding and for decoupling our everyday build process from MDN

KoolADE85:
> woman_dancing This looks great, please either before or after merging make new issues for both removing `omit` for prop forwarding and for decoupling our everyday build process from MDN

Done! #2600 #2601

PR #2593, opened 2023-07-10
Feature: dcc.Input accepts a number for its debounce argument

KoolADE85:
This PR changes dcc.Input so that it accepts a "number of seconds" for the debounce argument.
When passing a number, the callbacks to the Dash server are debounced accordingly.
This does not change the behaviour of a boolean debounce argument.

With this PR in place, we will be able to address [this issue](https://github.com/plotly/dashboard-engine/issues/1051) in the dbe project.

## Contributor Checklist

### optionals

### DIFF
--- a/components/dash-core-components/src/components/Input.react.js
+++ b/components/dash-core-components/src/components/Input.react.js
@@ -26,6 +26,7 @@ export default class Input extends PureComponent {
         this.onChange = this.onChange.bind(this);
         this.onEvent = this.onEvent.bind(this);
         this.onKeyPress = this.onKeyPress.bind(this);
+        this.debounceEvent = this.debounceEvent.bind(this);
         this.setInputValue = this.setInputValue.bind(this);
         this.setPropValue = this.setPropValue.bind(this);
     }
@@ -123,13 +124,30 @@ export default class Input extends PureComponent {
         }
     }
 
+    debounceEvent() {
+        const {value} = this.input.current;
+        let {debounce} = this.props;
+        debounce = Number.isFinite(debounce) ? debounce * 1000 : 500;
+
+        window.clearTimeout(this.state?.pendingEvent);
+        const pendingEvent = window.setTimeout(() => {
+            this.onEvent();
+        }, debounce);
+
+        this.setState({
+            ...this.state,
+            value,
+            pendingEvent,
+        });
+    }
+
     onBlur() {
         this.props.setProps({
             n_blur: this.props.n_blur + 1,
             n_blur_timestamp: Date.now(),
         });
         this.input.current.checkValidity();
-        return this.props.debounce && this.onEvent();
+        return this.props.debounce === true && this.onEvent();
     }
 
     onKeyPress(e) {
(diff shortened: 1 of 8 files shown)


components/dash-core-components/src/components/Input.react.js
```diff
@@ -26,12 +26,17 @@ export default class Input extends PureComponent {
...
         this.onEvent = this.onEvent.bind(this);
         this.onKeyPress = this.onKeyPress.bind(this);
+        this.debounceEvent = this.debounceEvent.bind(this);
         this.setInputValue = this.setInputValue.bind(this);
         this.setPropValue = this.setPropValue.bind(this);
     }
 
     UNSAFE_componentWillReceiveProps(nextProps) {
         const {value} = this.input.current;
+        if (this.state?.pendingEvent) {
```
T4rk1n, line 36: State should be defined in constructor.

components/dash-core-components/src/components/Input.react.js
```diff
@@ -121,6 +126,23 @@ export default class Input extends PureComponent {
         } else {
             this.props.setProps({value});
         }
+        this.setState({pendingEvent: undefined});
+    }
+
+    debounceEvent(time = 0.5) {
+        const {value} = this.input.current;
+        const MILLISECONDS = 1000;
```
T4rk1n, line 134: Move to global or use 1000 directly.

components/dash-core-components/src/components/Input.react.js
```diff
@@ -140,14 +162,22 @@ export default class Input extends PureComponent {
...
+            if (Number.isFinite(debounce)) {
+                this.debounceEvent(debounce);
+            }
+            if (this.props.type !== 'number') {
+                this.setState({value: this.input.current.value});
+            }
+        } else {
             this.onEvent();
-        } else if (this.props.type !== 'number') {
-            this.setState({value: this.input.current.value});
```
T4rk1n, line 150: Was this moved somewhere else?

components/dash-core-components/src/components/Input.react.js
KoolADE85: Good call. Changed in 0198e20

components/dash-core-components/src/components/Input.react.js
KoolADE85: Nope, this is still here. Since I added another condition to the "if/else", I just inverted the logic to keep it readable.

components/dash-core-components/src/components/Input.react.js
KoolADE85: I would prefer to use 1000 directly, but there is a linter rule preventing that :( 
Since this number is only relevant to this function, I'm not sure it makes sense to move it higher.

components/dash-core-components/src/components/Input.react.js
```diff
@@ -121,6 +131,23 @@ export default class Input extends PureComponent {
         } else {
             this.props.setProps({value});
         }
+        this.setState({pendingEvent: undefined});
+    }
+
+    debounceEvent(time = 0.5) {
+        const {value} = this.input.current;
+        const MILLISECONDS = 1000;
+        time = time * MILLISECONDS;
```
alexcjohnson, line 140: Better not to modify a var in place if we don't need to, just multiply it in the `setTimeout` call. And to your other question, you're welcome to add `1000` to the [already long list of exclusions in the linter](https://github.com/plotly/dash/blob/a7a12d180e16eac0a84b88e2b8dc8f7c7601cbaf/components/dash-core-components/.eslintrc#L120), I agree that it's a little silly to create a const just to satisfy the linter. But this would be a little more self-documenting if the var were called `seconds` or some such instead of `time`.

components/dash-core-components/tests/integration/input/test_debounce.py
```diff
@@ -0,0 +1,56 @@
...
+    # expect that a long debounce does not call back in a short amount of time
+    elem = dash_dcc.find_element("#input-slow")
+    elem.send_keys("unit test slow")
+    with pytest.raises(TimeoutException):
+        WebDriverWait(dash_dcc.driver, timeout=1).until(
+            lambda d: d.find_element(By.XPATH, "//*[text()='unit test slow']")
+        )
+
+    # but do expect that it is eventually called
+    assert dash_dcc.wait_for_text_to_equal(
```
alexcjohnson, line 19: I see you're following a pattern we have in a few other places... but these `dash_duo.wait_for_*` methods (which `dash_dcc` inherits) don't need `assert`, they either succeed or raise a timeout error.

alexcjohnson:
LGTM, just a couple of minor nonblocking comments, but we do still need a changelog entry :)

alexcjohnson (approved):
💃

PR #2610, opened 2023-08-02
Load plotly.js from plotly.py

T4rk1n:
Load plotly.js bundle from installed plotly.py.

This has the following effects:
- Version of plotly.js from dcc.Graph is no longer locked to the plotly.js version that dash was build on.
- To get the new version of plotly.js, you will now have to specify the version if you already installed plotly.py in your environment.


components/dash-core-components/dash_core_components_base/__init__.py
```diff
@@ -121,15 +122,6 @@
...
         },
-        {
-            "relative_package_path": "dcc/plotly.min.js",
-            "external_url": (
-                "https://unpkg.com/dash-core-components@{}"
-                "/dash_core_components/plotly.min.js"
-            ).format(__version__),
-            "namespace": "dash",
-            "async": "eager",
-        },
```
alexcjohnson, line 132: Can we also delete the following two entries, `async-plotlyjs.js` and its `.map`? And I guess we can remove the `prebuild:js` command in dcc's `package.json` that creates the `plotly.min.js` file referenced here, right?

components/dash-core-components/package.json
```diff
@@ -49,7 +48,6 @@
     "mathjax": "^3.2.2",
     "moment": "^2.29.4",
     "node-polyfill-webpack-plugin": "^2.0.1",
-    "plotly.js-dist-min": "2.25.2",
```
alexcjohnson, line 52: Good call!

alexcjohnson (approved):
💃 Love it!

PR #2614, opened 2023-08-06
Bump Flask & Werkzeug in requires-install.txt

### LINKED ISSUE #2592: Be compatible with Flask 2.3
dash dependency of end of support **flask** branch
```Flask>=1.0.4,<2.3.0```
since https://github.com/plotly/dash/commit/7bd5b7ebec72ffbfca85a57d0d4c19b595371a5a

The 2.3.x branch is now the supported fix branch, the 2.2.x branch will become a tag marking the end of support for that branch.
https://github.com/pallets/flask/releases

gothicVI:
Fixes #2592

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## Changed
 
 - [#2593](https://github.com/plotly/dash/pull/2593) dcc.Input accepts a number for its debounce argument
+- [#2614](https://github.com/plotly/dash/pull/2614) Bump Flask & Werkzeug in requires-install.txt, closes [#2592](https://github.com/plotly/dash/issues/2592)
 
 ## Updated
 
--- a/requires-install.txt
+++ b/requires-install.txt
@@ -1,5 +1,5 @@
-Flask>=1.0.4,<2.3.0
-Werkzeug<2.3.0
+Flask>=1.0.4,<2.4.0
+Werkzeug<2.4.0
 plotly>=5.0.0
 dash_html_components==2.0.0
 dash_core_components==2.0.0


alexcjohnson:
Thanks @gothicVI - Flask 2.3 is ok with most of the dash ecosystem but there are some packages (including some on our enterprise side that unfortunately are not visible open-source) that we still need to update before we'll be ready to allow Flask 2.3. Thank you for your patience!

Moisan:
@alexcjohnson could you point out which open source packages that are not currently compatible? Maybe we could open a PR for them. 

For the non-open source ones, wouldn't it be possible to limit the flask version on your side? The current requirements are quite open (`Flask>=1.0.4,<2.3.0`), I suspect forcing `Flask<2.3.0` on a project depending on Dash, even after this PR is merged, would work.

gothicVI:
I'd opt for removing the upper version limit entirely:
https://iscinumpy.dev/post/bound-version-constraints/

alexcjohnson:
> I'd opt for removing the upper version limit entirely

That’s what we used to do, see https://github.com/plotly/dash/pull/2538 for context on why we added an upper bound

> could you point out which open source packages are not currently compatible?

I’m not sure. The open-source ones maintained by Plotly are good AFAIK, but I would look at 3rd-party packages that alter the back end. 

> For the non-open source ones, wouldn't it be possible to limit the flask version on your side?

Plausible, assuming we can identify ahead of time where the potential issues are.

gothicVI:
> That’s what we used to do, see #2538 for context on why we added an upper bound

Which from my point of view is the wrong approach. That should be dealt with by the user using `venv`s or `requirement.txt`s but artificially breaking install dependencies for other packages on a package level is leading us into hell at some point (see the linked article).
If every package comes with lower and upper bounds we'll soo be unable to install anything because it'll break some dependencies...
Just my two cents though ;)

alexcjohnson:
I linked to exactly that article in #2538 and explained why we feel our decision is consistent with it. It’s not every package. It’s just Flask, which has consistently broken something in Dash with every minor update. Thus an upper bound here is just a statement that we expect this to be true in the future as well.  No other package has such a history, hence we don’t specify any other upper bounds.

I’m sorry that it is taking us a while to find and fix the other Flask 2.3 issues but we feel strongly that this is better than uncontrolled breakage.

alexcjohnson:
superseded by #2674

PR #2617, opened 2023-08-10
add setuptools as a dependency

aleivag:
this might sound silly, but setuptools is actually [required to run this library](https://github.com/plotly/dash/blob/dev/dash/dash.py#L22C1-L22C58), so explicitly declaring it in req files. we had issues about using this library where the solution has been installing setuptools (we try to only include declared dependencies)

### DIFF
--- a/requires-all.txt
+++ b/requires-all.txt
@@ -47,3 +47,4 @@ requests[security]>=2.21.0
 selenium>=3.141.0
 waitress>=1.4.4
 webdriver-manager>=3.5.1
+setuptools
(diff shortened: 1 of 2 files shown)


alexcjohnson (approved):
Thanks @aleivag - I wonder how this never caused issues before - seems like it's always there, to the extent that I assumed it was part of the std lib. But you're right, we should declare it. 💃

aleivag:
i guess you need setuptools to install application so it never came up... but at meta we use buck to build packages (bypassing setup.py) so unless is specified it will fail to work... i guess it cant hurt to have it there as a req, and for most folks will be a noop..

Thanks for merging this so fast!

PR #2625, opened 2023-08-16
Fix background callbacks missing cancel error.

### LINKED ISSUE #2624: [BUG] Server fails to setup when using background callbacks without cancel argument
**Describe your context**

```
dash                                     2.12.0
dash-bootstrap-components                1.4.2
dash-bootstrap-templates                 1.0.8
dash-core-components                     2.0.0
dash-html-components                     2.0.0
dash-table                               5.0.0
```

**Describe the bug**

Server fails to setup when using background callbacks without cancel argument.
[This condition](https://github.com/plotly/dash/blob/909e39cad3041ded960432d29fa493e34cd90dab/dash/_callback.py#L169) is not triggered, causing a KeyError [here](https://github.com/plotly/dash/blob/909e39cad3041ded960432d29fa493e34cd90dab/dash/dash.py#L1324).

**Expected behavior**

Server setups properly. Adding None as default to the offending pop seemed to fix the issue in my quick test.

T4rk1n:
Fix #2624


alexcjohnson (approved):
💃 🚀

PR #2630, opened 2023-08-22
Feature: Add hooks for layout "pre" and "post"

KoolADE85:
This PR adds new renderer hooks for users to write code that reads & updates the page layout.
In the same style as `request_pre` and `request_post`, there is now a `layout_pre` and a `layout_post`.

## Contributor Checklist

### optionals


tests/integration/renderer/test_request_hooks.py
```diff
@@ -295,3 +295,46 @@ def wrap(*args, **kwargs):
...
+            {%favicon%}
+            {%css%}
+        </head>
+        <body>
+            {%app_entry%}
+            <footer>
+                {%config%}
+                {%scripts%}
+                <script id="_dash-renderer" type"application/json">
+                    const renderer = new DashRenderer({
```
T4rk1n, line 317: This api usage is not something I think we should expand upon without improvements, it's really hard to use and has boilerplate not required for the feature.

I suggest we add parameter(s) that get them in the main index without the need to override the index string.

KoolADE85: oh sure, I was just following the same style as the other tests in this file.
I agree it's odd to overwrite the entire index_string. I can also do this instead:

```
app.renderer = """
    new DashRenderer({
        ... 
    });
"""
```

which would resemble what I'm actually doing in EnterpriseDash

T4rk1n: No I meant we should add a parameter to init like:
```python
app = Dash(hooks={"pre_layout": "() => null"})
```
This way you don't need to override the index values and the boilerplate is minimal.

KoolADE85: As discussed, I've added the ability to pass these hooks as an argument to the constructor.

dash/dash.py
```diff
@@ -373,6 +375,7 @@ def __init__(  # pylint: disable=too-many-statements
         long_callback_manager=None,
         background_callback_manager=None,
         add_log_handler=True,
+        hooks: Union[RendererHooks, None] = None,
```
alexcjohnson, line 378: I think w should still mention this in the docstring, even though it's primarily meant for internal use. Something like "a dict of javascript functions for modifying behavior of the Dash renderer"

dash/_utils.py
```diff
@@ -267,3 +269,15 @@ def coerce_to_list(obj):
...
+
+
+def hooks_to_js_object(hooks: Union[RendererHooks, None]) -> str:
+    if hooks is None:
+        return ""
+    hook_str = reduce(
+        lambda reduced, hook: f"{reduced}{hook[0]}: {hook[1]},",
+        hooks.items(),
+        "",
+    )
```
alexcjohnson, line 281: heh writing Python like it's JS 😎 
```suggestion
    hook_str = ",".join(f"{key}: {val}" for key, val in hooks.items())
```

alexcjohnson (approved):
💃 Looks good, just had a couple of minor comments!

dash/dash.py
KoolADE85: Good call, done!

dash/_utils.py
KoolADE85: Spot my native language :) 
Addressed in 5e3688b

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 - [#2610](https://github.com/plotly/dash/pull/2610) Load plotly.js bundle/version from plotly.py
 
+## UNRELEASED
+
```
alexcjohnson, line 11:
```suggestion
```
(cruft from updating the branch via GUI)

PR #2635, opened 2023-08-30
Refactors

T4rk1n:
- Pick props instead of keys for `dcc.Input`, `dcc.Textarea`, `dcc.Slider`, `dcc.RangeSlider`
- Batch redux dispatches in APIController.


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 orbs:
   win: circleci/windows@2.4.0
   percy: percy/agent@0.1.3
-  browser-tools: circleci/browser-tools@1.4.3
+  browser-tools: circleci/browser-tools@1.4.5
```
alexcjohnson, line 6: There's a 1.4.6 now, but we can wait on that if this one works.

alexcjohnson: oh maybe that would let us remove the chrome 116 pin?

tests/integration/devtools/test_props_check.py
```diff
@@ -216,3 +218,5 @@ def display_content(pathname):
             assert dash_duo.get_logs(), tc
         else:
             assert dash_duo.get_logs() == [], tc
+
+        time.sleep(0.1)
```
alexcjohnson, line 222: what effect does this sleep have, coming after all the assertions?

alexcjohnson (approved):
💃 LGTM. If you don't want to investigate `browser-tools@1.4.6` now, please make an issue to come back and remove the chrome pin later.

tests/integration/devtools/test_props_check.py
T4rk1n: This test was failing but I am not sure that it needs this.

.circleci/config.yml
T4rk1n: Yes I think that was fixed and there is also now a proper chromedriver version I think that was the cause for dvpc001 failures.

T4rk1n: I unpinned the chrome version and the dvpc was failing again.

alexcjohnson: OK - please make an issue to come back to this

PR #2647, opened 2023-09-22
Routing callback inputs

RenaudLN:
Allows to add new Inputs/States to the pages routing callback, to pass additional data to the layout functions.

This allows things like (non-exhaustive list):
* A language dropdown that will be passed to every layout function, for internationalisation
* Serialising the state in URL hashes without reloading the page on every input update, and using the hash on first load / refresh
* Passing a global app data store on page load

The syntax for this parameter is a dict of State objects: `routing_callback_inputs={"language": Input("language", "value")}`

## Example 1: Internationalisation

_app.py_
 ```python
from dash import Dash, Input, html, page_container

app = Dash(
    __name__,
    use_pages=True,
    routing_callback_inputs={
        # The language will be passed as a `layout` keyword argument to page layout functions
        "language": Input("language", "value"),
    },
)
app.layout = html.Div(
    [
        html.Div(
            [
                "My app",
                html.Div(
                    [
                        dcc.Link("Home", href="/"),
                        dcc.Link("Page 1", href="/page-1"),
                        dcc.Dropdown(
                            id="language",
                            options=[
                                {"label": "English", "value": "en"},
                                {"label": "Français", "value": "fr"},
                            ],
                        ),
                    ],
                    style={"marginLeft": "auto", "display": "flex", "gap": 10}
                )
            ],
            style={"background": "#CCC", "padding": 10, "marginBottom": 20, "display": "flex"},
        ),
        page_container,
    ]
)
```

_pages/home.py_
```python
from dash import register_page, html

register_page(__name__, "/")

translations = {
    "en": {
        "title": "Hello world",
        "subtitle": "This is the home page.",
    },
    "fr": {
        "title": "Bonjour le monde",
        "subtitle": "Ceci est la page d'accueil.",
    }
}

def layout(language: str = "en"):
    return [
        html.H1(translations[language]["title"]),
        html.H2(translations[language]["subtitle"]),
    ]
```

## Example 2: State serialisation in URL

_app.py_
```python
from dash import Dash, State, html, page_container
from dash.dash import _ID_LOCATION

app = Dash(
    __name__,
    use_pages=True,
    routing_callback_inputs={
        # The app state is serialised in the URL hash without refreshing the page
        # This URL can be copied and then parsed on first load
        "state": Input(_ID_LOCATION, "hash"),
    },
)
app.layout = html.Div(
    [
        html.Div(
            [
                "My app",
                html.Div(
                    [
                        dcc.Link("Home", href="/"),
                        dcc.Link("Page 1", href="/page-1"),
                    ],
                    style={"marginLeft": "auto", "display": "flex", "gap": 10}
                )
            ],
            style={"background": "#CCC", "padding": 10, "marginBottom": 20, "display": "flex"},
        ),
        page_container,
    ]
)
```

_pages/home.py_
```python
import base64
import json
from dash import ALL, Input, Output, callback, html, dcc, register_page, ctx
from dash.dash import _ID_LOCATION

register_page(__name__, "/")

def layout(state: str):
    # Define default values
    defaults = {"country": "France", "year": 2020}
    # Decode the state from the hash
    state = defaults | (json.loads(base64.b64decode(state)) if state else {})

    return [
        html.Div(
            [
                dcc.Dropdown(
                    id={"type": "control", "id": "country"},
                    value=state.get("country"),
                    options=["France", "USA", "Canada"],
                ),
                dcc.Dropdown(
                    id={"type": "control", "id": "year"},
                    value=state.get("year"),
                    options=[{"label": str(y), "value": y} for y in range(1980, 2021)],
                ),
            ],
        ),
        html.Div(contents(**state), id="contents")
    ]

def contents(country: str, year: int):
    return f"Country: {country}, Year: {year}"


# Update the hash in the URL Location component
@callback(
    Output(_ID_LOCATION, "hash", allow_duplicate=True),
    Input({"type": "control", "id": ALL}, "value"),
    prevent_initial_call=True,
)
def update_hash(_values):
    return "#" + base64.b64encode(
        json.dumps(
            {inp["id"]["id"]: inp["value"] for inp in ctx.inputs_list[0]}
        )
        .encode("utf-8")
    ).decode("utf-8")


# Update the contents when the dropdowns are updated
@callback(
    Output("contents", "children"),
    Input({"type": "control", "id": "country"}, "value"),
    Input({"type": "control", "id": "year"}, "value"),
    prevent_initial_call=True,
)
def update_contents(country, year):
    return contents(country, year)
```

## Contributor Checklist

### optionals


dash/dash.py
```diff
@@ -346,6 +347,17 @@ class Dash:
...
     javascript functions. To hook into the layout, use dict keys "layout_pre" and
     "layout_post". To hook into the callbacks, use keys "request_pre" and "request_post"
+
+    :param routing_callback_inputs: When using Dash pages (use_pages=True), allows to
+    add new States to the routing callback, to pass additional data to the layout
+    functions. The syntax for this parameter is a dict of State objects:
+    `routing_callback_inputs={"language": Input("language", "value")}`
+    This allows things like (non-exhaustive list):
+    * A language dropdown that will be passed to every layout function,
+      for internationalisation
```
T4rk1n, line 357: While localization might be a use case, this allows for more so we can delete this from the docstring.

T4rk1n (commented):
Looks good, just need to add a test.

RenaudLN:
Test added 👍

requires-all.txt
```diff
@@ -9,6 +9,7 @@ flake8==3.9.2
 flaky==3.7.0
 flask-talisman==0.8.1
 isort==4.3.21;python_version<"3.7"
+importlib-metadata
```
T4rk1n, line 12: This file is not in the actual requirements and may be outdated, it is generated by CI to create a hash for requirements. If this is a new requirements it should go in `requires-install.txt` or the other `requires-*` for extras.

tests/integration/multi_page/test_pages_layout.py
```diff
@@ -179,3 +180,67 @@ def test_pala004_no_layout_exception(clear_pages_state):
...
+            dcc.Dropdown(id="language", options=["en", "fr"], value="en"),
+            dash.page_container,
+        ]
+    )
+    return app
+
+
+def test_pala005_routing_inputs(dash_duo, clear_pages_state):
+    dash_duo.start_server(get_routing_inputs_app())
+    dash_duo.wait_for_page(url=f"http://localhost:{dash_duo.server.port}#123")
```
T4rk1n, line 227: You can get the url with `dash_duo.server_url`

T4rk1n (approved):
:woman_dancing: Looks good

requires-all.txt
alexcjohnson: What @T4rk1n is saying is this whole file is useless, [CI creates it](https://github.com/plotly/dash/blob/d9b8d17e3fae5c7f9f342e480a85e782f6289fe8/.circleci/config.yml#L45) but it shouldn't really be in the repo. Shall we just delete it and `.gitignore` it?

That said there's a weird situation with this package: we pin it for Py3.6:

https://github.com/plotly/dash/blob/d9b8d17e3fae5c7f9f342e480a85e782f6289fe8/requires-install.txt#L7

but we don't require it directly for later Python versions, perhaps we should add the line:
```
importlib-metadata;python_version>="3.7"
```
to `requires_install.txt`?

RenaudLN: Updated 👍

RenaudLN:
Docs update PR link: https://github.com/plotly/ddk-dash-docs/pull/2055

requires-all.txt
Akay7: > perhaps we should add the line:
> ```
> importlib-metadata;python_version>="3.7"
> ```

But it's backport for older versions of python. I believe with the latest version of python there's no requirement of `importlib-metadata` at all:
 
* Maybe we should remove that line?
* Or maybe we should remove `importlib-metadata` at all because seems it useful only for deprecated versions of python.

PR #2649, opened 2023-09-29
dynamic callback creation

T4rk1n:
Add a new `app.callback` argument: `_allow_dynamic_callbacks` to add new callbacks defined inside a callback.

- Only `app.callback` for the same will be registered.
- Need to return a value from the callback, `PreventUpdate` not supported in the callback creator.
- Only the user that created the callback will receive the updated callbacks, other users would need to refresh.

**_:warning: Intended for internal usage, use only with a local server for 1 user per app._**

### DIFF
--- a/tests/integration/callbacks/test_dynamic_callback.py
+++ b/tests/integration/callbacks/test_dynamic_callback.py
@@ -1,0 +1,39 @@
+from dash import html, Dash, Input, Output
+
+
+def test_dync001_dynamic_callback(dash_duo):
+    app = Dash()
+
+    app.layout = html.Div(
+        [
+            html.Div(id="output"),
+            html.Button("create", id="create"),
+            html.Div("initial", id="output-2"),
+            html.Button("dynamic", id="dynamic"),
+        ]
+    )
+
+    @app.callback(
+        Output("output", "children"),
+        Input("create", "n_clicks"),
+        _allow_dynamic_callbacks=True,
+        prevent_initial_call=True,
+    )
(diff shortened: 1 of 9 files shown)


tests/integration/callbacks/test_dynamic_callback.py
```diff
@@ -0,0 +1,39 @@
...
+
+        return f"creator {n_clicks}"
+
+    dash_duo.start_server(app)
+
+    dash_duo.wait_for_element("#create").click()
+    dash_duo.wait_for_text_to_equal("#output", "creator 1")
+
+    dash_duo.wait_for_element("#dynamic").click()
+    dash_duo.wait_for_text_to_equal("#output-2", "Dynamic clicks 1")
```
alexcjohnson, line 39:
```suggestion
    dash_duo.wait_for_element("#dynamic").click()
    dash_duo.wait_for_element("#create").click()
    dash_duo.wait_for_text_to_equal("#output", "creator 1")
    dash_duo.wait_for_text_to_equal("#output-2", "initial")

    dash_duo.wait_for_element("#dynamic").click()
    dash_duo.wait_for_text_to_equal("#output-2", "Dynamic clicks 2")
```
Just to make it crystal clear that the dynamic callback was not attached at first.

alexcjohnson (approved):
💃 Looks great, feel free to merge after considering my comment on the test, and (as discussed) adding a note in the changelog and in the test case about how this feature is dangerous, has numerous ways it can break, and is not for production, multi-user, or multi-process use.

PR #2672, opened 2023-10-25
Fix get_caller_name

T4rk1n:
Fix `get_caller_name` by always taking the third stack frame from the function call. (0 is `get_caller_name`, `__init__` is 1 then next one is the `app = Dash()` call)

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -396,7 +396,7 @@ class Dash:
     ):
         _validate.check_obsolete(obsolete)
 
-        caller_name = get_caller_name(self.__class__.__name__)
+        caller_name = get_caller_name()
 
         # We have 3 cases: server is either True (we create the server), False
         # (defer server creation) or a Flask app instance (we use their server)
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -396,7 +396,7 @@ class Dash:
     ):
         _validate.check_obsolete(obsolete)
 
-        caller_name = get_caller_name()
+        caller_name = None if name else get_caller_name()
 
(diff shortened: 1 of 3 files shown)


T4rk1n:
Taking by index doesn't work in the case of a custom dash class.

Think we can do with a check on the function name, it is `get_caller` then a chain of `__init__` and we are looking for `<module>` which has a name.

dash/dash.py
```diff
@@ -396,7 +396,7 @@ def __init__(  # pylint: disable=too-many-statements
     ):
         _validate.check_obsolete(obsolete)
 
-        caller_name = get_caller_name(self.__class__.__name__)
+        caller_name = get_caller_name()
```
alexcjohnson, line 399: To limit the impact on existing apps, can we skip this if there's already a `name` (in which case we aren't going to use it anyway)?
```suggestion
        caller_name = None if name else get_caller_name()
```

alexcjohnson (approved):
💃 This looks good to me!

Note that there's at least one case where this can give a different result from `app = Dash(__name__)`: if you create `app` in a function that you call from a different module. The inferred root will be the last time you were at the module scope before calling a function. You could construct some very strange situations here - a function imports a module, and that module calls a function that creates the dash app... but it feels to me as though in all those situations you should really be passing in `__name__` from the user code anyway, so what you have is the right solution.

T4rk1n:
> but it feels to me as though in all those situations you should really be passing in __name__

Yes I think in those cases you would want to give what is appropriate by either passing the `__name__` from the module or in a function argument.

Issue #1549, opened 2021-02-08
Allow for "null" / no output callbacks


tr8dr:
Dash callbacks currently require at least one output.   There are scenarios where one would like to act on an event from the UI, producing no output as a side-effect.   Would like to be able to do:

```
@app.callback(None, [Input("A","value"), Input("B", "value")]
def handler(a, b):
    <some application logic>
    return
```

Right now one has to create a dummy container (such as a div) in the DOM and use it as a "fake" output sink:

```
    ...
    html.Div(id='dummy1'),
    ...

@app.callback(Output("dummy1", "children"), [Input("A","value"), Input("B", "value")]
def handler(a, b):
    <some application logic>
    return None
```

jeslinmx:
This is all the more relevant with clientside callbacks; an example is using them as some sort of post-update hook to create clientside-only behaviour:

```
app.clientside_callback(
    """
    function(fig) {
        document.querySelector("#graph .js-plotly-plot").on("plotly_hover", event => {
            ...
        })
        return null // we can't even use dash.no_update in here
    }
    """,
    Output('dummy', 'children'), # ideally we should not need this
    Input('graph', 'figure')
)
```

kenshima4:
Hi guys, I'm new here. I'd like to tackle this if possible. I've had that need myself before.

alexcjohnson:
@kenshima4 you're more than welcome to give it a shot. It would be a great feature but it's not one Plotly staff currently have in our roadmap. Just be aware that it may have a number of complications - the ones that occur to me offhand are validation, determining when to trigger, `MATCH` pattern-matching inputs, and inclusion in the devtools callback graph - and will require quite a few tests as a result to give us confidence in the implementation. One piece of good news is that since we added `allow_duplicate=True` we have the machinery to support multiple no-output callbacks, as long as we implicitly set this flag in this case.

kenshima4:
Hi @alexcjohnson I just wanted to say that I can still try and solve it but maybe at a later stage (if somebody else wants to give it a shot that's ok). My apologies but I have a few other projects that just came my way. Don't have too much time now.

yashjha123:
It's finally possible! You can now create callbacks without providing any outputs!! Just clone [this fork](https://github.com/yashjha123/dash/tree/support_for_no_output_callbacks)] and run your code without any issues. 
[...]

PR #2682, opened 2023-10-31
Support for callbacks with no outputs

yashjha123:
This PR allows callbacks to have no return values. This is the most requested feature [on this repo.](https://github.com/plotly/dash/issues/1549). Personally, I can see multiple examples where this feature can be used. It can be used to, let's say, make a post request to a website we don't need the response for, for eg. tracking clicks. Other examples, for eg, opening a new website (using clientside callbacks) and many applications.

Here's a basic example of how this might work.
```python
from dash import Dash, html, callback, Output, Input

app = Dash(__name__)

app.layout = html.Div([
    html.Div(id="result"),
    html.Button(id='output-button',children="This button gives an output"),
    html.Button(id='no-output-button',children="no output button"),

])

@callback(
    Output("result","children"),
    Input("output-button","n_clicks")
)
def make_result(n_clicks):
    return f"Button with output looks like this {n_clicks}"

@callback(
    Input("no-output-button","n_clicks"),
    force_no_output=True
)
def make_result(n_clicks):
    print("Ignored",n_clicks)
    return None


if __name__ == '__main__':
    app.run(debug=True)
```
This new feature is built on the existing logic used for duplicate outputs. I've added conditions to the error statements that overwrite the behaviors specifically for no-output callbacks.
I imagine the final version will have warnings on the terminal and more rigid statements. However, this implementation is convincing to me to be in the production env. What do you folks think? What other changes should we make?
## Contributor Checklist

   -  [X] Basic MVP Working
   -  [ ] A generalize function wrappers(?) => 0 outputs automatically forces the attribute.
   - [ ] More Tests needs (also rewrite old tests)


alexcjohnson:
Thanks @yashjha123, looks like a very promising start! Three things before we dig too deeply into this:
- Would you mind rebasing or recreating this PR off the `dev` branch instead of `master`, and ensuring none of the build artifacts (`dash/dash-renderer/build/*`) are committed?
- As mentioned in https://github.com/plotly/dash/issues/1549#issuecomment-1668401201 from @T4rk1n we should be able to do this without users needing to set a flag in the callback definition, just by looking at how many outputs the callback defines. If you want to generate that flag behind the scenes and pass it to the renderer that's fine, but we shouldn't make users do that.
- Tests will be very important here, and we'll need them to cover quite a lot of situations: direct initial call, indirect initial call (ie the input is the output of another initial callback), direct user input, indirect user input, pattern matching with `MATCH` input, pattern matching with `ALL` input... we can start reviewing prior to tests but we'll need them before we can merge. Because these callbacks are only expected to have server-side effects, the best pattern to use is probably `multiprocessing.Value` as for example in [cbmt001](https://github.com/plotly/dash/blob/dev/tests/integration/callbacks/test_multiple_callbacks.py#L22) - but this feature deserves a new test file in the same folder.

dash/dash-renderer/webpack.base.config.js
```diff
@@ -31,7 +31,7 @@ const defaults = {
             {
                 test: /\.ts(x?)$/,
                 exclude: /node_modules/,
-                use: ['babel-loader', 'ts-loader'],
```
alexcjohnson, line 34: Interesting, I wasn't aware this could be omitted. Sounds to me as though it's safer to keep it though, to keep type checking. Per https://webpack.js.org/guides/typescript/#loader:

> Note that if you're already using [babel-loader](https://github.com/babel/babel-loader) to transpile your code, you can use [@babel/preset-typescript](https://babeljs.io/docs/en/babel-preset-typescript) and let Babel handle both your JavaScript and TypeScript files instead of using an additional loader. Keep in mind that, contrary to ts-loader, the underlying [@babel/plugin-transform-typescript](https://babeljs.io/docs/en/babel-plugin-transform-typescript) plugin does not perform any type checking.

T4rk1n:
Thank you for opening this PR, I adapted part of the code found here to not require setting a flag and allow for multiples no output callbacks.

Closing as implemented in #2822

PR #2693, opened 2023-11-13
Adding support for bringing your own driver to the Browser

BSd3v:
This is a proof-of-concept of being able to bring your own driver to the Browser object of dash.testing and be able to utilize predefined dash_duo tests.

Looking for feedback and usability.

### DIFF
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -41,6 +41,7 @@ class Browser(DashPageMixin):
     def __init__(
         self,
         browser,
+        driver=None,
         remote=False,
         remote_url=None,
         headless=False,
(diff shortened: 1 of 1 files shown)


dash/testing/browser.py
```diff
@@ -65,8 +66,11 @@ def __init__(
         self._percy_run = percy_run
         self._pause = pause
 
-        self._driver = until(self.get_webdriver, timeout=1)
-        self._driver.implicitly_wait(2)
+        if driver:
+            self._driver = driver
+        else:
+            self._driver = until(self.get_webdriver, timeout=1)
+            self._driver.implicitly_wait(2)
```
alexcjohnson, line 73: Do we want the `implicitly_wait` inside here or would it be better to always call that even if you provide your own `driver`?

BSd3v: Lets set the `implicitly_wait(2)` for all, since there is no easy way to confirm if the driver already has one set.

PR #2721, opened 2024-01-08
Remove ansi2html.

### LINKED ISSUE #2713: [Feature Request] Remove dependency on LGPL licensed `ansi2html`

Dash currently depends on [`ansi2html`](https://pypi.org/project/ansi2html/)

https://github.com/plotly/dash/blob/7527383050b6fe59a11833bbeac6f1a741436084/requires-install.txt#L13

`ansi2html` is licensed under LGPL:

- https://github.com/pycontribs/ansi2html/blob/main/LICENSE

This is not a truly permissive license, such as MIT, BSD, or Apache, which can cause headaches when distributing some private code that depends on Dash (and transitively depends on `ansi2html`).

LGPL makes it challenging for including Dash inside of a binary due to the restriction that users need to be able to have the ability to swap out the LGPL library for another of their choosing, something infeasible/very tricky for a binary distribution.

For example, this affects users of [Pyinstaller](https://www.pyinstaller.org/), [Nuitka](https://nuitka.net/), [PyOxdizer](https://pyoxidizer.readthedocs.io/en/stable/), etc. that want to package their Python code and dependencies into a stand-alone binary.

Dash only depends on fully permissive licensed packages (such as MIT, BSD, Apache).

**Related Historical Work**

For a long time `chardet` was an LGPL package depended on by popular packages, such as `requests`:

- https://github.com/chardet/chardet/issues/36

The community has migrated to a permissively licensed package [`charset_normalizer`](https://github.com/Ousret/charset_normalizer).

One of the motivations was to ensure that the full dependency tree was permissively licensed.

T4rk1n:
Resolve #2713 by removing the ansi2html package, upon investigation the stacktrace weren't colored anyway.

### DIFF
--- a/dash/_jupyter.py
+++ b/dash/_jupyter.py
@@ -5,7 +5,6 @@ import logging
 import os
 import queue
 import uuid
-import re
 import sys
 import threading
 import time
@@ -22,7 +21,6 @@ try:
     from IPython.core.display import HTML
     from IPython.core.ultratb import FormattedTB
     from retrying import retry
-    from ansi2html import Ansi2HTMLConverter
     from ipykernel.comm import Comm
     import nest_asyncio
 
@@ -471,21 +469,7 @@ class JupyterDash:
             if self.inline_exceptions:
                 print(ansi_stacktrace)
 
-            # Use ansi2html to convert the colored ANSI string to HTML
-            conv = Ansi2HTMLConverter(scheme="ansi2html", dark_bg=False)
-            html_str = conv.convert(ansi_stacktrace)
-
-            # Set width to fit 75-character wide stack trace and font to a size the
-            # won't require a horizontal scroll bar
-            html_str = html_str.replace(
-                "<html>", '<html style="width: 75ch; font-size: 0.86em">'
-            )
-
-            # Remove explicit background color so Dash dev-tools can set background
-            # color
-            html_str = re.sub("background-color:[^;]+;", "", html_str)
-
-            return html_str, 500
+            return ansi_stacktrace, 500
 
     @property
     def active(self):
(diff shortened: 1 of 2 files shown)


dash/_jupyter.py
```diff
@@ -471,21 +469,7 @@ def _wrap_errors(error):
...
-            html_str = html_str.replace(
-                "<html>", '<html style="width: 75ch; font-size: 0.86em">'
-            )
-
-            # Remove explicit background color so Dash dev-tools can set background
-            # color
-            html_str = re.sub("background-color:[^;]+;", "", html_str)
-
-            return html_str, 500
+            return ansi_stacktrace, 500
```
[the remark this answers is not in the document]
T4rk1n: My bad, I didn't see the color, without `dev_tools_prune_errors=False` I only had one line and the color was dark red, in the devtools error had the escaped ansi symbols.

There is a NoColor mode that should do.

requires-ci.txt
```diff
@@ -7,7 +7,7 @@ flake8==3.9.2
 flaky==3.7.0
 flask-talisman==1.0.0
 isort==4.3.21;python_version<"3.7"
-mimesis
+mimesis<=11.1.0
```
alexcjohnson, line 10: NBD since this is only in requires-ci, but why is this lock needed?

alexcjohnson (approved):
💃 Thanks!

requires-ci.txt
T4rk1n: The latest version introduced a union type like `str | int` and that is only supported in recent python versions.

alexcjohnson: OK yeah weird:
> To work with Mimesis on Python versions 3.8 and 3.9, the final compatible version is Mimesis 11.1.0. Install this specific version to ensure compatibility.

but even v12.1.0 says
> Requires: Python >=3.8, <4.0

Wonder why they didn't just bake that first statement into the `python_requires` 🤔

PR #2723, opened 2024-01-10
Add Slider/RangeSlider tooltip.format and tooltip.style.

### LINKED ISSUE #1846: [Feature Request] `dcc.RangeSlider` tooltip - allow displaying something else than `value`
# Problem
Currently, the tooltip of `dcc.RangeSlider` can be used to display the `value` property of the slider. It would be nice if it could display an arbitrary string. E.g. using a dictionary, where keys are the `value`s of the slider and values are the strings to be displayed in the tooltip.

# Solution(?)
Apparently, such functionality is present in the react slider.

[http://react-component.github.io/slider/?path=/story/rc-slider–handle](http://react-component.github.io/slider/?path=/story/rc-slider–handle)

```html
<div style={wrapperStyle}>
      <p>Range with custom tooltip</p>
      <Range min={0} max={20} defaultValue={[3, 10]} tipFormatter={value => `${value}%`} />
</div>
```

See the `tipFormatter` property. Could this property be added to `dcc.RangeSlider`?

# Example use-case 
Imagine that your slider `value`s are actually indexes of a list of the _real_ values. You want to present the _real_ values to the user instead of the indexes. You cannot use the `marks` property of the slider because there are too many values and it would look messy. Thus, you want to use the tooltip to display **something else than the `value` property of the slider**.

T4rk1n:
Resolve #1846 

Add two new props to dcc Slider and RangeSlider:

- `tooltip.format` a string for the format template, `{value}` will be formatted with the actual value.
- `tooltip.style` a style object to give to the div of the tooltip.


components/dash-core-components/src/fragments/Slider.react.js
```diff
@@ -72,17 +73,29 @@ export default class Slider extends Component {
...
              */
-            tipProps = assoc('visible', tooltip.always_visible, tooltip);
-            delete tipProps.always_visible;
-        } else {
-            tipProps = tooltip;
+            tipProps = pipe(
+                assoc('visible', tooltip.always_visible),
+                dissoc('always_visible'),
+                dissoc('format'),
+                dissoc('style')
```
alexcjohnson, line 87:
```suggestion
                omit(['always_visible', 'format', 'style'])
```

components/dash-core-components/src/components/RangeSlider.react.js
```diff
@@ -125,6 +125,29 @@ RangeSlider.propTypes = {
...
+         * This can be added in a scrip in the asset folder.
+         *
+         * For example, in `assets/tooltip.js`:
+         * ```
+         * window.dccFunctions = window.dccFunctions || {};
+         * window.dccFunctions.multByTen = function(value) {
+         *     return value * 10;
+         * }
+         * ```
+         * Then in the component `tooltip={'transform': 'multByTen'
```
AnnMarieW, line 148:
```suggestion
         * Then in the component `tooltip={'transform': 'multByTen'}
```

Is there any way to call this "function" instead of "transform" so it's consistent with dash-ag-grid?

components/dash-core-components/src/components/RangeSlider.react.js
```diff
@@ -125,6 +125,29 @@ RangeSlider.propTypes = {
             'bottomLeft',
             'bottomRight',
         ]),
+        /**
+         * Format to apply to the tooltip
+         * The string must contain `{value}`
+         */
+        format: PropTypes.string,
```
alexcjohnson, line 132: I don't know about `format` - based on that I'd expect you to be able to choose the precision, whether or not to use exponential notation, etc, but it's really only about prefixing and suffixing the default string representation (or the transformed representation) of the number. So I might change it to `template` instead, and say something like "Template string to display the tooltip in. Must contain `{value}`, which will be replaced with either the default string representation of the value or the result of the `transform` function if there is one"

components/dash-core-components/src/components/RangeSlider.react.js
```diff
@@ -125,6 +125,29 @@ RangeSlider.propTypes = {
...
+        /**
+         * Custom style for the tooltip.
+         */
+        style: PropTypes.object,
+        /**
+         * Reference to a function in the `window.dccFunctions` namespace.
+         * This can be added in a scrip in the asset folder.
```
alexcjohnson, line 139:
```suggestion
         * This can be added in a script in the asset folder.
```

components/dash-core-components/src/components/RangeSlider.react.js
alexcjohnson:
```suggestion
         * Then in the component `tooltip={'transform': 'multByTen'}`
```
Also wants a trailing backtick ;)

I'd be ok with calling this `function`, though for dash-ag-grid we need a generic word since it gets used in a lot of different contexts, whereas here it's specifically transforming the data value into the display value.

Like in `format`/`template` above, we should say something about how these two interact, perhaps "If both `template` and `function` are used, the result of `function` is inserted into the `template`. Alternatively you can just use `function` to construct the final string to display".

AnnMarieW: It would be consistent with this pattern in AG Grid - providing a function that returns a formatted string.
```
 "valueFormatter": {"function": "d3.format('($,.2f')(params.value)"},

```
In fact, if you were already using AG Grid in the app, could these d3 functions be available to the tooltip as well?  That would be pretty cool

alexcjohnson: We could if we brought dash-ag-grid into the core, otherwise we wouldn’t want to couple these two packages together like that

alexcjohnson: Actually, the implementation here is quite different from that in `dash-ag-grid`, in that all we accept here is the name of the function to look for in `window.dccFunctions`, not a whole expression to be parsed. I think this makes sense, because these are much simpler functions than we need to support in `dash-ag-grid` where there's this huge parameter object you may need to access, and it's better to keep dcc simpler and lighter. So it might even be good to not try for naming consistency as that could confuse people into thinking they have feature parity.

components/dash-core-components/src/components/Slider.react.js
```diff
@@ -105,6 +107,31 @@ Slider.propTypes = {
...
+         * This can be added in a scrip in the asset folder.
+         *
+         * For example, in `assets/tooltip.js`:
+         * ```
+         * window.dccFunctions = window.dccFunctions || {};
+         * window.dccFunctions.multByTen = function(value) {
+         *     return value * 10;
+         * }
+         * ```
+         * Then in the component `tooltip={'transform': 'multByTen'
```
alexcjohnson, line 132:
```suggestion
         * Then in the component `tooltip={'transform': 'multByTen'}`
```

components/dash-core-components/src/components/Slider.react.js
```diff
@@ -105,6 +107,31 @@ Slider.propTypes = {
...
+         * transform function if there is one.
+         */
+        template: PropTypes.string,
+        /**
+         * Custom style for the tooltip.
+         */
+        style: PropTypes.object,
+        /**
+         * Reference to a function in the `window.dccFunctions` namespace.
+         * This can be added in a scrip in the asset folder.
```
alexcjohnson, line 123:
```suggestion
         * This can be added in a script in the asset folder.
```

alexcjohnson (approved):
💃

Issue #2670, opened 2023-10-23
Poor docstring formatting in vscode


chriddyp:
Here's what it looks like for getting the docstring within `@callback`. Lots of white space and odd newline breaks. Could be nicer!

<img width="730" alt="image" src="https://github.com/plotly/dash/assets/1280389/288da817-8db5-40a2-9be3-36b00c867e32">

woefulpines:
Hello, I'm a little new to open source development. I wanted to try this issue to learn a thing or two. Whats the process to be assigned it?

alexcjohnson:
Welcome @woefulpines! We don't normally bother with assignments in the Plotly open-source repos, you can just get started. Feel free to open a PR as soon as you have some progress to commit and mention this issue, so we know you're working on it and nobody duplicates your efforts.

woefulpines:
I was looking through the repository. I saw that some of the python generation had proper doc string formatting, I was a little confused as the doc string in the image provided doesn't match the template in the component generation in dash/development/ python component code gen file. I'll continue looking but any other perspective would be appreciated.

T4rk1n:
@woefulpines I think the components docstring are ok, it's in the callback docstring they are indented too much:
https://github.com/plotly/dash/blob/508cce6422c952c4ec13e037aacfa2973b1f3540/dash/_callback.py#L84-L134

woefulpines:
I'm sorry for the slow speed on this, I think I have a solution, but I was confused when it comes to opening a pull request. Would I need to apply for certain permissions to create a branch?

sofiemor:
Hello! Is this available to work on?

LiamConnors:
This looks like it is now resolved in VS Code

<img width="734" alt="image" src="https://github.com/user-attachments/assets/7d5d7da2-222c-4511-b664-0ca0de0247e9">

There were a number of bug fixes here that dealt with docstring formatting
https://github.com/microsoft/pylance-release/blob/main/CHANGELOG.md#20246100-13-june-2024-prerelease

PR #2728, opened 2024-01-21
Fixing poor docstring formatting for @callback

woefulpines:
It seems like some reStructuredText elements don't render well in vscodes docstring preview window #2670. I replaced feature lists with elements that were supported according to this example in [pandas contributing guidelines](https://pandas.pydata.org/docs/development/contributing_docstring.html). In terms of content, nothing is added or changed, just formatted it differently. 

   -  [X] Set up local repo according to contributing guidelines. 
   -  [X] Figure out why the doc string was not formatting properly.
   -  [X] Replace problematic reStructure elements with more agreeable elements.
   -  [X] Conform additions to code style, eliminating linter errors.

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -81,12 +81,15 @@ def callback(
     not to fire when its outputs are first added to the page. Defaults to
     `False` and unlike `app.callback` is not configurable at the app level.
 
-    :Keyword Arguments:
-        :param background:
+    Keyword Arguments
+    -----------------
+    
+    param background:
             Mark the callback as a long callback to execute in a manager for
             callbacks that take a long time without locking up the Dash app
             or timing out.
-        :param manager:
+            
+    param manager:
             A long callback manager instance. Currently, an instance of one of
             `DiskcacheManager` or `CeleryManager`.
             Defaults to the `background_callback_manager` instance provided to the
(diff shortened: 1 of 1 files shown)


Coding-with-Adam:
Thank you for this PR, @woefulpines . We appreciate the time you put into this.

alexcjohnson:
Thanks @woefulpines! The main thing I'm concerned about here is how it works with the dash docs reference pages: https://dash.plotly.com/reference#dash.callback is auto-generated from this docstring. Frankly I'd prefer to move away from reStructuredText entirely, in favor of something more human-centered, perhaps looking more like component docstrings (which are themselves autogenerated). FYI @LiamConnors

woefulpines:
Out of curiosity, how do you generate your documentation? Is there some resource that talks about it?

woefulpines:
Also I realized now that dash/dash.py has an example of feature lists working in the vscode doc popup window so I can try that for the time being and have some changes that may be more inline with your standards.

alexcjohnson:
Unfortunately our docs repo is private, just because we use our enterprise packages in it. But we’ve been discussing some ways to potentially open it up, which would make this kind of thing a lot easier!

woefulpines:
I added feature lists back in. I wasn't able to use blank lines to format the docstring, I hope newlines wont give the document generator any trouble. Looking at dash/dash.py. The comment for the dash class is allowed to use blank lines and it also has type annotations in the docstring for each parameter. So if those were things you would like in the callback docstring, I can add them in. Thank for your patience with this. I really appreciate it.

LiamConnors:
Yeah the formatting here does break it in the Dash docs
image
Checking to see what part might be breaking it

gvwilson:
@LiamConnors can you please either merge or close? thank you - @gvwilson

LiamConnors:
This seems to be resolved in VS Code. 
See https://github.com/plotly/dash/issues/2670#issuecomment-2389510596

PR #2730, opened 2024-01-23
Load asset files with ending ".mjs" as js modules

### LINKED ISSUE #1904: [Feature Request] Import javascript files as modules
I am attempting to write some larger clientside callbacks that share logic between them. To keep the code clean, I am separating callbacks into separate .js files. For the callbacks, there is some common logic that I would like to keep in a third file e.g. common.js, that I would import as `import * as common from './common.js';`. 

Unfortunately, this causes an "Uncaught SyntaxError: Cannot use import statement outside a module" when the scripts are loaded in the default way i.e. by storing them in the assets folder.

The issue would be resolved if the scripts were loaded as modules, i.e. 
`<script type="module" src="/assets/foo.js></script>`
rather than
`<script src="/assets/foo.js></script>`.

This is currently possible to do for external scripts, since those allow you to provide a dict containing tag attributes. Would it be possible to enable something similar for scripts hosted in assets?

jowlo:
When a script file in the `assets` folder ends in `.mjs`, the `type="module"` attribute is added to the `<script>` tag.

Fixes #1904

## Contributor Checklist

### optionals

- [x] If this PR needs a follow-up in **dash docs**, **community thread**, I have mentioned the relevant URLS as follows


alexcjohnson:
@jowlo this looks great, thanks! Can you add a test of this in [test_clientside.py](https://github.com/plotly/dash/blob/dev/tests/integration/clientside/test_clientside.py), with a `.mjs` file in the corresponding [assets folder](https://github.com/plotly/dash/tree/dev/tests/integration/clientside/assets)?

insilications:
Hope this gets upstreamed!

alexcjohnson:
@insilications we’re just blocked on a test, maybe you’d like to write one so we can merge this?

jowlo:
Sorry for the delay here, this issue slowly settled down to the bottom of my todo list...

I added a simple tests that just checks whether any script tag with `module=true` was loaded. It registers a simple display callback just like the most basic test for non-module clientside callbacks.

I moved the display function to its own module to make sure it loads other modules that are imported, which would not be possible if `callbacks.mjs` was not loaded as module itself.

tests/integration/clientside/assets/clientside.mjs
```diff
@@ -0,0 +1,5 @@
+import {display} from "./clientsideModule.mjs";
+
+window.dash_clientside.clientside_module = Object.assign({}, window.dash_clientside, {
+	display
+});
```
alexcjohnson, line 5: Won't this put whatever else is already in `dash_clientside` into `dash_clientside.clientside_module`? ie:
```js
display = 5;
dash_clientside={something: {a: 1}};
window.dash_clientside.clientside_module = Object.assign({}, window.dash_clientside, {display});
console.log(JSON.stringify(dash_clientside,null,2));

{
  "something": {
    "a": 1
  },
  "clientside_module": {
    "something": {
      "a": 1
    },
    "display": 5
  }
}
```

Also just to ensure we're testing with the same patterns we recommend to users, the order of asset loading isn't guaranteed so each file that updates `dash_clientside` should be prepared to create it, something like:
```js
const dc = window.dash_clientside = (window.dash_clientside || {});
dc.clientside_module = {display};
// or if you imagine other modules contributing to clientside_module as well,
// use the same pattern again:
const cm = dc.clientside_module = (dc.clientside_module || {});
Object.assign(cm, {display});
```

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## Changed
 - [#2652](https://github.com/plotly/dash/pull/2652) dcc.Clipboard supports htm_content and triggers a copy to clipboard when n_clicks are changed
 - [#2721](https://github.com/plotly/dash/pull/2721) Remove ansi2html, fixes [#2613](https://github.com/plotly/dash/issues/2713)
+- [#2730](https://github.com/plotly/dash/pull/2721) Load script files with `.mjs` ending as js modules
```
alexcjohnson, line 33: Ah, now this needs to move up since we made a release after you opened this PR. Also I think I'd call it either `Added` or `Fixed`, could be either depending on your perspective!

alexcjohnson (approved):
💃 @jowlo this looks great, beautiful test! Just made two minor comments, re: the test module and the changelog but then this is ready to go 🎉

tests/integration/clientside/assets/clientside.mjs
jowlo: Good catch, yes! This was a copy-paste error and should be the same object that is read and assigned to, so:
```javascript
window.dash_clientside.clientside_module = Object.assign({}, window.dash_clientside.dash_clientside_module, {display});
```

The intention was to merge the (potentially) existing properties of `window.dash_clientside.dash_clientside_module` and  `{display}` to a new object (that is then assigned to `window.dash_clientside.clientside_module`).

I fixed it in the test, i found this to be the most concise way of writing it - but my javascript is a bit rusty.

CHANGELOG.md
jowlo: Sure, moved it up!

tests/integration/clientside/assets/clientside.mjs
```diff
@@ -0,0 +1,5 @@
+import {display} from "./clientsideModule.mjs";
+
+window.dash_clientside.clientside_module = Object.assign({}, window.dash_clientside.clientside_module, {
```
alexcjohnson, line 3:
```suggestion
window.dash_clientside = window.dash_clientside || {};
window.dash_clientside.clientside_module = Object.assign({}, window.dash_clientside.clientside_module, {
```
Perhaps my previous comment got lost or wasn't clear - I just don't want to depend on the order in which these scripts execute, to assume that `window.dash_clientside` already exists.

jowlo: Oh, yeah, sorry. Thanks for finding this again!
I was only thinking of not overwriting `window.dash_clientside.clientside_module`, not `window.dash_clientside` itself.

PR #2732, opened 2024-01-24
Sanitize html props with xss vulnerability.

### LINKED ISSUE #2729: Validate url to prevent XSS attacks
I've identified a Cross-Site Scripting (XSS) vulnerability in 'dash-core-components'

**Vulnerability Details:**
- **Severity**: High/Critical
- **Description**: There's a risk of malicious script execution when the href of the a tag is controlled by an adversary.

**Steps to Reproduce:**
In a React.js project:
```
import { Link } from 'dash-core-components'

<Link href={`javascript:alert(1)`} />
```
Then the malicious code alert(1) will be executed. Any React.js application using this package may be vulnerable to XSS.

**Suggested Fix or Mitigation:**
https://github.com/plotly/dash/blob/000ec180be049d511df04be63fa4be39c2d5a808/components/dash-core-components/src/components/Link.react.js#L64-L94
It is best practice for a React.js components package to sanitize the href attribute before passing it to an <a> tag. React.js and many popular libraries such as react-router-dom and Next.js also ensure the safety of href attributes. For instance, React.js issues warnings about URLs starting with javascript: and is planning to block these in future versions, as indicated in [this pull request](https://github.com/facebook/react/pull/15047).

 Please consider validating the href to resolve this vulnerability, thanks!

T4rk1n:
- Sanitize html props that are vulnerable to xss vulnerability if user data is inserted. Fix #2729 
  - `dcc.Link.href`
  - `html.A.href`
  - `html.Iframe.src`
  - `html.ObjectEl.data`
  - `html.Embed.src`
  - `html.Button.formAction`
  - `html.Form.action`
- Add special key `_dash_error` to setProps, allowing component developers to send error without throwing in render. Usage `props.setProps({_dash_error: new Error("custom error")})`

### DIFF
--- a/components/dash-core-components/package.json
+++ b/components/dash-core-components/package.json
@@ -35,6 +35,7 @@
   "maintainer": "Alex Johnson <alex@plotly.com>",
   "license": "MIT",
   "dependencies": {
+    "@braintree/sanitize-url": "^7.0.0",
     "@fortawesome/fontawesome-svg-core": "1.2.36",
     "@fortawesome/free-regular-svg-icons": "^5.15.4",
     "@fortawesome/free-solid-svg-icons": "^5.15.4",
--- a/components/dash-core-components/src/components/Link.react.js
+++ b/components/dash-core-components/src/components/Link.react.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 
-import React, {Component} from 'react';
-
+import React, {useEffect, useMemo} from 'react';
+import {sanitizeUrl} from '@braintree/sanitize-url';
 import {isNil} from 'ramda';
 
 /*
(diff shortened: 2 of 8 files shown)


components/dash-html-components/scripts/generate-components.js
```diff
@@ -280,8 +319,8 @@ const ${Component} = (props) => {
                 n_clicks_timestamp: Date.now()
             })
             })}
-            {...omit(['n_clicks', 'n_clicks_timestamp', 'loading_state', 'setProps', 'disable_n_clicks'], props)}
-            {...dataAttributes}
+            {...omit(['n_clicks', 'n_clicks_timestamp', 'loading_state', 'setProps', 'disable_n_clicks', 'href'], props)}
```
alexcjohnson, line 322: not always `href` though, right?

T4rk1n: Actually not necessary to omit the prop since it is redefined in extraProps it will take precedence.

components/dash-core-components/package.json
```diff
@@ -92,6 +92,7 @@
     "webpack": "^5.90.0",
     "webpack-cli": "^5.1.4"
   },
+  "optionalDependencies": { "fsevents": "*" },
```
alexcjohnson, line 95: I see the commit title "chokidar fix" but don't know what that means or what this does, can you explain?

components/dash-core-components/src/components/Link.react.js
```diff
@@ -52,49 +60,45 @@ export default class Link extends Component {
...
+                _dash_error: new Error(`Dangerous link detected:: ${href}`),
+            });
+        }
+    }, [href, sanitizedUrl]);
+
+    /*
+     * ideally, we would use cloneElement however
+     * that doesn't work with dash's recursive
+     * renderTree implementation for some reason
+     */
```
alexcjohnson, line 84: Does this comment still have any relevance? It's there from the very first commit creating `dcc.Link` but this pattern looks entirely orthodox to me so IDK what it was talking about.

alexcjohnson (approved):
💃 Very nice. Love the test!

components/dash-core-components/package.json
T4rk1n: The error message mention that weird name:

> npm ERR! fsevents not accessible from chokidar

components/dash-core-components/src/components/Link.react.js
T4rk1n: I don't think it's relevant, probably some remnant of the first builds.

PR #2735, opened 2024-01-31
Drop 3.6 and 3.7 and support 3.11 and 3.12

graingert-coef:
Fixes https://github.com/plotly/dash/issues/2736

## Contributor Checklist

   -  [x] update ci to support 3.12
   -  [x] update ci to support 3.8
   -  [x] update package metdata to support 3.8-3.12

### optionals


graingert:
@alexcjohnson when running the integration tests locally I get: `ModuleNotFoundError: No module named 'dash_generator_test_component_nested'` do you know how I can install those?

.circleci/config.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           name: 🏁 Build Component Packages & Update Dependencies/Artifacts
           command: |
             python -m venv venv && . venv/bin/activate
-            pip install --upgrade pip wheel
+            pip install --upgrade pip wheel setuptools
```
graingert, line 52: ok the reason this works took me a few days to workout, but basically celery prepends the current working directory to sys.path if it's not already present. This means the local source tree takes priority over installed modules so "import tests" works but "import dash" fails with a missing attribute on dash.dcc

When setuptools and wheel is installed pip will use "setup.py develop" which adds the current sys.path to site-packages/easy-install.pth which *appends* the current directory so the local source tree is available but installed modules take priority, this means "import tests" works and so does "import dash"

graingert-coef:
@alexcjohnson finally all the tests are passing! What do you think?

graingert-coef:
@alexcjohnson should I rebase this off of dev?

alexcjohnson:
> @alexcjohnson should I rebase this off of dev?

Yep! Then I'll give it a final review 🎉

graingert-coef:
@alexcjohnson ok I've rebased

.pylintrc312
```diff
@@ -3,7 +3,7 @@
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
-extension-pkg-whitelist=
+extension-pkg-allow-list=dash.html,dash.dash_table
```
graingert-coef, line 6:
```suggestion
extension-pkg-allow-list=
```

.circleci/config.yml
graingert-coef: adding setuptools here is just a workaround that causes pip to fallback to legacy editable project mode, I have a fix for modern editables here https://github.com/plotly/dash/pull/2749

components/dash-core-components/package.json
```diff
@@ -15,7 +15,7 @@
     "private::format.black": "black dash_core_components_base/ tests/ setup.py",
     "private::format.eslint": "eslint src --fix",
     "private::format.prettier": "prettier --config .prettierrc --write src/**/*.js",
-    "private::lint.black": "node -e \"if ((process.env.PYVERSION || 'python310') !== 'python36'){process.exit(1)} \" || black --check dash_core_components_base/ tests/ setup.py",
+    "private::lint.black": "node -e \"if ((process.env.PYVERSION || 'python312') !== 'python38'){process.exit(1)} \" || black --check dash_core_components_base/ tests/ setup.py",
```
alexcjohnson, line 18: I suppose we can leave this as you have it for now, but way better would be to pick a version of `black` that supports all the same Python versions we do and always run it. The only reason we didn't do that from the start is that when we first introduced `black` we still supported Py2.7 and `black` was always Py3-only.

graingert: I'll do this in a follow up PR

requires-ci.txt
```diff
@@ -3,24 +3,24 @@ black==22.3.0;python_version>="3.7"
...
 flaky==3.7.0
 flask-talisman==1.0.0
 isort==4.3.21;python_version<"3.7"
 mimesis<=11.1.0
 mock==4.0.3
-numpy<=1.25.2
+numpy<=1.26.3
 orjson==3.5.4;python_version<"3.7"
-orjson==3.6.7;python_version>="3.7"
+orjson==3.9.12;python_version>="3.7"
```
alexcjohnson, line 14: We can get rid of all the `python_version>="3.7"` and `python_version>="3.8"` specifiers, and the entire lines for `<"3.7"` and `<"3.8"`

graingert-coef: @alexcjohnson ok done!

alexcjohnson (approved):
💃 Beautiful!

PR #2734, opened 2024-01-30
configure ci for python 3.10

graingert-coef:
Fixes #1863

switch 3.9 CI to 3.10 CI

## Contributor Checklist

- [ ] I have broken down my PR scope into the following TODO tasks
   -  [ ] task 1
   -  [ ] task 2
- [ ] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))
- [ ] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR

### optionals

- [ ] I have added entry in the `CHANGELOG.md`
- [ ] If this PR needs a follow-up in **dash docs**, **community thread**, I have mentioned the relevant URLS as follows
    -  [ ] this GitHub [#PR number]() updates the dash docs
    -  [ ] here is the show and tell thread in Plotly Dash community

PR #2746, opened 2024-02-06
remove python version number from Percy snapshot names

alexcjohnson:
Following up on https://github.com/plotly/dash/pull/2734#discussion_r1480085144 - #2734 hard-coded the python version number so we could clearly see that the Percy snapshots didn't change, now removing that number and we'll blindly approve the "new" snapshots


alexcjohnson:
(FYI @graingert-coef)

alexcjohnson:
haha ok I'll let you have it in #2747 😉

PR #2747, opened 2024-02-06
remove misleading py3.9 from snapshot names

graingert-coef:
## Contributor Checklist

   -  [x] remove misleading py3.9 from snapshot names

### optionals

### DIFF
--- a/dash/testing/browser.py
+++ b/dash/testing/browser.py
@@ -160,10 +160,7 @@ class Browser(DashPageMixin):
         if widths is None:
             widths = [1280]
 
-        # py3.9 hardcoded here to keep snapshot names the same accorss
-        # future python upgrades
-        snapshot_name = f"{name} - py3.9"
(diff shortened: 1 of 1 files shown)


alexcjohnson:
@T4rk1n FYI I guess this is going to change the behavior for any other components that use `percy_snapshot` - next time they upgrade Dash (or next time they test, if they build Dash from the dev branch) their snapshots won't have the Python version in the name. I think this is the right way to go, we long ago stopped taking snapshots in all the Python versions we test, and anyone that still does want this version number in their snapshots can add it themselves pretty easily.

alexcjohnson (approved):
💃

alexcjohnson:
Ha all the tests passed first try except one of the Py3.6 jobs that's now on its third try. Since we're about to remove 3.6 entirely, if it doesn't pass this time I'll just merge anyway

PR #2749, opened 2024-02-07
Support modern editable installs

graingert-coef:
pip is switching to pyproject based editable installs which breaks the celery tests see https://github.com/plotly/dash/pull/2735#discussion_r1480010572

## Contributor Checklist

   -  [x] move dash source code to src/dash
   -  [x] configure the project with new src layout

### optionals


gvwilson:
@graingert-coef thank you again for this PR - we're rethinking our build environments right now, so we will revisit it in the next cycle.

marthacryan:
Thank you for the PR! Could you undo the change that moves files into the `src` directory?

robertclaus:
Closing due to age, inactivity, and the fact that this should be resolved in https://github.com/plotly/dash/pull/3336

PR #2753, opened 2024-02-07
Edit two error messages to suggest that users check for trailing commas

red-patience:
Based off my own experience getting tripped up by the layout error. Thanks @AnnMarieW (can't seem to find you in the list of reviewers for some reason).

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals


dash/dash-renderer/src/utils/TreeContainer.ts
```diff
@@ -61,8 +61,8 @@ export function validateComponent(componentDefinition: any) {
     if (type(componentDefinition) === 'Array') {
         throw new Error(
             'The children property of a component is a list of lists, instead ' +
-                'of just a list. ' +
-                'Check the component that has the following contents, ' +
+                'of just a list. Check for trailing commas, or ' +
```
AnnMarieW, line 64: This isn't really an "or" condition.  If a tuple is supplied as children, the content  will be shown in the error message details.

How about something like:

> The children property of a component is a list of lists, instead of just a list. Review the component with the following content and remove one of the levels of nesting.  Note that tuples are converted to a list of lists.  Check for trailing commas.

AnnMarieW: Thanks for doing this PR @red-patience.  I know I found these error messages very confusing when I first started learning Dash. This is going to be so helpful!

alexcjohnson: > Note that tuples are converted to a list of lists.

That's not quite right - a tuple is converted to a list, it's just that implicit tuple creation due to a trailing comma can give you an extra layer of nesting you weren't expecting

red-patience: Would this be better?

`The children property of a component is a list of lists, instead of just a list. This can sometimes be due to a trailing comma. Check the component that has the following contents... `

red-patience: @alexcjohnson I saw that layout list support may be coming down the pipeline. Is this still valuable, especially if we'd need to remember to revert it?

alexcjohnson: Still valuable and will not need reverting! Layout list will only apply to the top level and this situation can arise anywhere if it happens in a callback return value for example.

red-patience: Updated!

alexcjohnson (approved):
💃 Great!

Weird test failure, we'll see if it happens again after merging dev, if not I'll sort it out.

Issue #951, opened 2019-10-02
Ability to target individual properties in the loading states


chriddyp:
e.g. for a `dcc.Graph` - ability to only display the loading state / loading component if `figure` is updating, not if `hoverData` is updating. there might be some more specificity with respect to a component being an "input" component vs an "output" component as well (e.g. `hoverData` will always be an "input" property not an output property, so we might consider omitting the loading state for it.)

More details here https://community.plot.ly/t/disable-dcc-loading-animation-on-chart-with-hoverdata-callback/29466

ethanopp:
@chriddyp thanks!

rredburn:
I think I'm having the same issue with a selectData callback.  It appears this prevents shift-click from working correctly when clickmode='event+select' as selections don't accumulate.  If I remove the graph from dcc.Loading everything works as expected.

luggie:
I'm having the same issue with intervals that in my case should not trigger loading

ashraf97it:
I'm have the same issue my graph still reload on every update

MarcoNastasi:
Ran into this issue with a graph with hover data. I saw that the related issue https://github.com/plotly/dash-core-components/issues/709 was meant to be included with a prior release, but seems that got dropped?

Issue #2696, opened 2023-11-20
[Feature Request] Make `dcc.Loading` toggle-able attribute for enabled/disabled


matt-sd-watson:
I have created an application where various `dcc.Store` elements are used to hold user-imported hash tables for images of varying sizes. These images can then be blended together and viewed in a `dcc.Graph`.

Based on the size and number of images, I would like to be able to wrap these store elements in `dcc.Loading` components whose visibility/functionality is toggle-able i.e. if the user wishes, he/she can either enable or disable the loading visibility for the stores using a Dash toggle switch. Right now, I don't see an option for this because there is no `visibility` or `enabled` attribute for the Loading component. The result is that the visibility for the stores must be set when the app is executed but cannot be dynamically changed within the session. It would be very beneficial to have an attribute for enabled or visible, similar to how some other dash components have, in order to dynamically toggle the visibility of loadings within the app on the fly.

Coding-with-Adam:
great idea, @matt-sd-watson, thank you. 
Would you be interested in creating a pull request for this feature?

JamesKunstle:
This would be a hugely beneficial feature for our project. In React projects this is super easy, especially with TanQuery 'isLoading' props being available.

Coding-with-Adam:
hi @JamesKunstle .
It doesn't look like @matt-sd-watson was able to make the PR. Is this something you would like to take a jab at?

JamesKunstle:
Yeah @Coding-with-Adam I can give it an effort. Would also love some help triaging #2725 . It's more complicated than this one.

AnnMarieW:
@JamesKunstle I'm about to do a PR to add some features to `dcc.Loading`  Would you like to collaborate on this project?

JamesKunstle:
@AnnMarieW Yeah I would love that! I started reading the dash-renderer code and the dcc components code to prepare to make an intelligent PR. @Coding-with-Adam is sharing some resources with me as well, but I would love to work with you!

AnnMarieW:
Great!  I saw you post on the forum recently - I'll send you a PM there

AnnMarieW:
Closed by #2760

Issue #2690, opened 2023-11-13
Loading Indicator during plotly rendering


Yokogawa-Akshay:
I have created a simple dash app, which renders a scatter plot using plotly express, I have used the dcc.Loading to wrap the component which displays the plot upon a simple button click.

But during the initial rendering the indicator stops displaying once the component is ready with its data, but the plot is not yet rendered and displayed to the user.

Is there any way to show a loading spinner or any loading indicator which can be displayed until the plot is rendered?

I have searched around the plotly documentation and seem to not find anything suitable.

The code below

`import dash
from dash import dcc, html
from dash.dependencies import Input, Output
import plotly.express as px
import pandas as pd
import numpy as np

# Read files
df = pd.read_csv(
    'https://raw.githubusercontent.com/kuanrongchan/vaccine-studies/main/Ad5_seroneg.csv',
    index_col=0)

# Calculate log2FC and -logP values
df['log2FC_7d'] = np.log2(df['ratio_7d'])
df['negative_log_pval_7d'] = np.log10(df['qval_7d']) * (-1)

# Dash App
app = dash.Dash(__name__)

app.layout = html.Div([
    html.Button('Update Plot', id='update-button', n_clicks=0),
    html.Div([
        dcc.Loading(
            id="loading",
            type="default",
            children=[
                html.Div(id='graph-container')
            ]
        )
    ])
])

@app.callback(
    Output('graph-container', 'children'),
    [Input('update-button', 'n_clicks')]
)
def update_plot(n_clicks):
    if n_clicks > 0:
        # Generate the updated plot using plotly.express
        fig = px.scatter(df, x='log2FC_7d', y='negative_log_pval_7d', text=df.index,
                         labels={'log2FC_7d': 'Log2FC', 'negative_log_pval_7d': '-LogP'},
                         title='Volcano plot for seronegatives (7Days)')
        fig.update_traces(textposition='top center')

        return dcc.Graph(id='volcano-plot', figure=fig)

    # Return an empty div if the button has not been clicked
    return html.Div()

if __name__ == '__main__':
    app.run_server(debug=True)
`
The CSV file being consumed is around 2.5mb and the delay after the loading indicator stops and the plot to render is around a second or 2, but this might increase once the files being consumed are also increasing in size.

Issue #2422, opened 2023-02-13
[BUG] Dash DCC.Dropdown using Dynamic Options, nested within dcc.Loading returns empty list


urifeigin:
```
dash                      2.7.1
dash-ag-grid              0.0.1
dash-bootstrap-components 1.3.0
dash-core-components      2.0.0
dash-html-components      2.0.0
dash-table                5.0.0
```
    - OS: Windows 11
    - Browser: Edge
    - Version 110.0.1587.41 

**Describe the bug**

Using Dash DCC.Dropdown using Dynamic Options, nested within dcc.Loading returns empty list.

Upon inspection the callback fires twice, once retuning the options list for the dropdown as expected and once returns and empty list even though prevent update is used. This happens on any single key pressed in the search box of the dropdown. removing the DCC.Dropdown from being nested in dcc.loading resolves the issue.

The code example used is exactly as in the [dash documentation](https://dash.plotly.com/dash-core-components/dropdown#dynamic-options), except for the dcc.loading wrapper. 

                                                            dcc.Loading(
                                                                [
                                                                    dcc.Dropdown(
                                                                        id="add_new_wave_modal_entity_sel_id",
                                                                    ),
                                                                ]
                                                            ),

@callback(
    Output("add_new_wave_modal_entity_sel_id", "options"),
    Input("add_new_wave_modal_entity_sel_id", "search_value"),
)
def update_options(search_value):
    if not search_value:
        raise PreventUpdate
    return [o for o in glb.wave_entity_signals_dd_options if search_value in o["label"]]

**Expected behavior**

The dcc.dropdown list will populate according to the typed filter on the fly. the spinner of the dcc.loading will appear if the list takes time to appear.

**Screenshots**
![Screenshot 2023-02-13 020120](https://user-images.githubusercontent.com/7108242/218348954-51f12682-7e92-480d-a0cc-4411d1f4779c.jpg)

Coding-with-Adam:
closed by #2760

PR #2760, opened 2024-02-15
New features for dcc.Loading

AnnMarieW:
This PR adds features to `dcc.Loading`
- adds `delay_show` and `delay_hide` props to prevent flickering when the loading spinner shows only for a short time. (same as in the Dash Bootstrap Components dbc.Spinner)
- adds `overlay_style` prop so you can do things like make children visible during loading and add opacity.
- adds `target_components` prop to specify which component/props can trigger the loading spinner
- adds `custom_spinner` prop so you can provide your own spinner rather than using the built-in options
- adds `display` prop to override the loading status.  Can set to "show", "hide", or "auto"
- refactored component to functional instead of a class

Closes:
https://github.com/plotly/dash/issues/951
#2147 will be fixed by using `delay_show`
#1922 will be fixed by using `target_components`
#1802 will be fixed by using `delay_show`
#2422 will be fixed by using `delay_show`
#879 will be possible by using `custom_spinner` and `overlay_style`
#736.  can't replicate the issue

Closes from the old dcc repo
https://github.com/plotly/dash-core-components/issues/873  - possible with `custom_spinner` and `delay_show`
https://github.com/plotly/dash/issues/1541 - can't replicate the issue
https://github.com/plotly/dash-core-components/issues/709 - possible with `target_components`

## Contributor Checklist

- [ ] To Do / Questions
   -  [x] Controlling the visibility of the component being loaded with the opacity and background color of the spinner would be  a breaking change.  Probably need to find a better way to make it possible to add opacity to the component .  **Update:** added `overlay_style` prop.
   -  [x] Add ability to manually trigger loading as requested in #2696  .  It's currently called "mode" with options "on", "off", "auto".  
      
   - [ ] Would setting a `delay_hide` time solve the issue where there is a lag between the callback finishing and a figure rendering with large data sets?  Or might that time be too variable?  If this is a solution, I'd need to update the `delay_hide` timer because it currently sets a minimum time for the timer to display rather than extending the display time.    See  https://github.com/plotly/dash/issues/2690  
 

--------------------------------

<br />

## Example 1 `delay_hide` and `delay_show` prop

The `delay_hide` and `delay_show` props can be used to prevent flickering when the loading spinner shows only for a short time.

This callback runs for 500ms.  This example shows how to:
- prevent the spinner from showing by setting   `delay_show=700`.  
-  make the spinner show for a minimum amount of time by setting `delay_hide=2000`

A good use-case for `delay_show` is when `hoverData` is used in a callback. A good demo is this [cross filter example in the docs](https://dash.plotly.com/interactive-graphing#update-graphs-on-hover)  Try wrapping the app layout with  `dcc.Loading([<example app layout> ], delay_show=500)`

dcc_loading1

```python

import time

import dash
from dash import Dash, Input, Output, State, html, dcc

app=Dash()
app.layout = html.Div(
    [
        html.Button("Load", id="loading-button", n_clicks=0),
        html.Div("delay_show (ms)"),
        dcc.Input(type="number", value=0, id="delay-show", debounce=True),
        html.Div("delay_hide (ms)"),
        dcc.Input(type="number", value=0, id="delay-hide", debounce=True),
        html.Hr(),
        dcc.Loading(html.Div(id="loading-output"), id="loading"),
    ]
)


@app.callback(
    Output("loading-output", "children"),
    Input("loading-button", "n_clicks"),
)
def load_output(n):
    if n:
        time.sleep(.5)
        return f"Output loaded {n} times"
    return "Output not reloaded yet"

@app.callback(
    Output("loading", "delay_show"),
    Output("loading", "delay_hide"),
    Input("delay-show", "value"),
    Input("delay-hide", "value")
)
def update_delay_show_hide(show, hide):
    if show is None or hide is None:
        return dash.no_update
    return int(show), int(hide)

app.run(debug=True)

```

--------------------------------

<br />

## Example 2 `target_components` prop
Use the `target_components` prop to specify which component/props can trigger the loading spinner.
By default, Loading fires when any child element enters loading state. This makes loading opt-in: Loading animation only enabled when one of target components/props enters loading state.

`target_components` is a dict where the key is the component id, and the value is the prop name | list of prop names | "*"

examples - the following will trigger the loading spinner:  

Any prop in the "grid" component
```
target_components ={"grid": "*"}
```
Either the grid's `rowData` or `columnDefs`
```
target_components ={"grid": ["rowData", "columnDefs"]}
```
Only the grid's `rowData`

```
target_components ={"grid": "rowData"}
```

dcc_loading2

```python
import time

import dash
from dash import Dash, Input, Output, State, html, dcc

app=Dash()
app.layout = html.Div(
    [
        html.Button("Load div 1", id="loading-button1", n_clicks=0),
        html.Button("Load div 2", id="loading-button2", n_clicks=0),
        html.Hr(),
        dcc.Loading([
            html.Div(id="loading-output1"),
            html.Div(id="loading-output2"),
        ], target_components={"loading-output1": "children"}),
    ]
)


@app.callback(
    Output("loading-output1", "children"),
    Input("loading-button1", "n_clicks"),
)
def load_output(n):
    if n:
        time.sleep(2)
        return f"Output loaded {n} times.  This callback triggers the loading spinner"
    return "Callback 1 output not reloaded yet"



@app.callback(
    Output("loading-output2", "children"),
    Input("loading-button2", "n_clicks"),
)
def load_output(n):
    if n:
        time.sleep(.5)
        return f"Output loaded {n} times.  No  loading spinner"
    return "Callback 2 output not reloaded yet"

app.run(debug=True)

```

--------------------------------

<br />

### Default:  content is hidden while loading
dcc_loading3b

### Styled with `overlay_style={"visibility":"visible", "opacity": .5, "backgroundColor": "white"}`  

This keeps the content visible while loading and adds opacity  

dcc_loading3a

```python
import time

import dash
from dash import Dash, Input, Output, State, html, dcc
import plotly.express as px
data_canada = px.data.gapminder().query("country == 'Canada'")
app=Dash()
app.layout = html.Div(
    [
        html.Button("Start", id="loading-button", n_clicks=0),
        html.Hr(),
        dcc.Loading(
            [dcc.Graph(id="loading-output", figure=px.line(data_canada, x="year", y="pop"))],
            overlay_style={"visibility":"visible", "opacity": .5, "backgroundColor": "white"},
            color="red"
        ),
    ]
)


@app.callback(
    Output("loading-output", "figure"),
    Input("loading-button", "n_clicks"),
)
def load_output(n):
    if n:
        time.sleep(1)
        return px.bar(data_canada, x="year", y="pop")
    return dash.no_update


app.run(debug=True)
```

--------------------------------

<br />

## Example 4 `custom_spinner` prop
Instead of using one of the built-in spinner component, you can provide your own.

A `custom_spinner` can be used to remove/hide the spinner.  This is a convenient way to remove the spinner when using nested `dcc.Loading` components.
It's possible to create a component  with custom loading messages using any Dash components.

This example uses:
```python
custom_spinner=html.H2(["My Custom Spinner", dbc.Spinner(color="danger")])
```
dcc_loading4

```python
import time

import dash
from dash import Dash, Input, Output, State, html, dcc
import plotly.express as px

data_canada = px.data.gapminder().query("country == 'Canada'")

app=Dash(external_stylesheets=[dbc.themes.BOOTSTRAP])
app.layout = dbc.Container(
    [
        dbc.Button("Start", id="loading-button", n_clicks=0),
        html.Hr(),
        dcc.Loading(
            [dcc.Graph(id="loading-output", figure=px.line(data_canada, x="year", y="pop"))],
            overlay_style={"visibility":"visible", "opacity": .5, "backgroundColor": "white"},
            custom_spinner=html.H2(["My Custom Spinner", dbc.Spinner(color="danger")])
        ),
    ]
)


@app.callback(
    Output("loading-output", "figure"),
    Input("loading-button", "n_clicks"),
)
def load_output(n):
    if n:
        time.sleep(1)
        return px.bar(data_canada, x="year", y="pop")
    return dash.no_update


app.run(debug=True)
```

--------------------------------

<br />

## Example 5 `display` prop

Use the `display` prop to manually display or hide the loading spinner.  Set to "show", "hide" or "auto"(the default)

dcc_loading5

```python
import time

import dash
from dash import Dash, Input, Output, State, html, dcc


app=Dash()
app.layout = html.Div(
    [
        html.Div("Select mode"),
        html.Button("Start", id="loading-button", n_clicks=0),
        dcc.Dropdown(["auto", "show", "hide"], value="auto", id="display", style={"marginBottom": 100}),
        html.Hr(),
        dcc.Loading(
            html.Div("Demo of manually controlling  loading status", id="loading-output"),
            id="loading"


        ),
    ]
)

@app.callback(
    Output("loading", "display"),
    Input("display", "value")
)
def update_display(display):
    return display


@app.callback(
    Output("loading-output", "children"),
    Input("loading-button", "n_clicks"),
)
def load_output(n):
    if n:
        time.sleep(5)
        return f"Updated content {n}"
    return dash.no_update


app.run(debug=True)

```


components/dash-core-components/src/components/Loading.react.js
```diff
@@ -164,4 +234,36 @@ Loading.propTypes = {
...
+
+    /**
+     * Whether the Spinner should show on app start-up before the loading state
+     * has been determined. Default True.
+     */
+    show_initially: PropTypes.bool,
+
+    /**
+     * Specify component and prop to trigger showing the loading spinner
+     * example: `[{"output-container": "children"}]`
```
JamesKunstle, line 257: Does this mean 'while pending'? e.g. while server-side callback is executing?

components/dash-core-components/src/components/Loading.react.js
```diff
@@ -7,66 +7,121 @@ import CircleSpinner from '../fragments/Loading/spinners/CircleSpinner.jsx';
...
+            const [component_name, prop_name] = Object.entries(component)[0];
+            return (
+                loading_state.component_name === component_name &&
+                loading_state.prop_name === prop_name
+            );
+        });
+        return isMatchingComponent;
+    };
+
+    const [showSpinner, setShowSpinner] = useState(show_initially);
```
JamesKunstle, line 71: good use of distinction between useRef and useState

JamesKunstle:
@AnnMarieW Could you give an example of when target_component is used?

delay_show and delay_hide seem reasonable and the implementation makes sense to me.

components/dash-core-components/src/components/Loading.react.js
AnnMarieW: That bit of code came directly from the dash-bootstrap-component library :slightly_smiling_face:

components/dash-core-components/src/components/Loading.react.js
AnnMarieW: Yes - the `dcc.Loading` component can wrap the whole layout - but maybe you only want the loading spinner to show when only certain components are being updates.  I'll add some examples here shortly

components/dash-core-components/src/components/Loading.react.js
```diff
@@ -7,66 +7,121 @@ import CircleSpinner from '../fragments/Loading/spinners/CircleSpinner.jsx';
...
-const coveringSpinner = {
-    visibility: 'visible',
-    position: 'absolute',
-    top: '0',
-    height: '100%',
-    width: '100%',
-    display: 'flex',
-    justifyContent: 'center',
-    alignItems: 'center',
+const spinnerComponents = {
```
JamesKunstle, line 10: maybe `spinnerComponentOptions`

AnnMarieW: or `spinnerType` like it was originally

components/dash-core-components/src/components/Loading.react.js
JamesKunstle: Oh I see. I'm going to suggest a comment then above.

components/dash-core-components/src/components/Loading.react.js
```diff
@@ -7,66 +7,121 @@ import CircleSpinner from '../fragments/Loading/spinners/CircleSpinner.jsx';
...
+        height: '100%',
+        width: '100%',
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+        backgroundColor: backgroundColor,
+        opacity: opacity,
+    };
+
+    const isTarget = () => {
```
JamesKunstle, line 57: I'd recommend a comment along these lines to clarify:

Overrides default `Loading` behavior if `target_components` is set. By default, Loading fires when any recursive child enters loading state. This makes loading opt-in: Loading animation only enabled when one of `target` components enters loading state.

components/dash-core-components/src/components/Loading.react.js
JamesKunstle: I'm 100% splitting hairs here: each spinner is a React component and not a type definition, so on cursory read-through I feel like calling them components like you initially suggested is right, but adding 'Options' is mildly more ergonomic from a code-to-english perspective.

components/dash-core-components/src/components/Loading.react.js
JamesKunstle: ++

JamesKunstle:
@AnnMarieW This is really exciting, the loading opacity update is really clever, and the demonstrations are really helpful to understand the scope of these updates.

JamesKunstle:
@AnnMarieW re: 'mode' prop- 

What you implemented matches the Loading behavior that I'd like to have.

I'd consider changing the name, however. 'mode' seems like a execution context rather than 'on or off.' I'd perhaps use 'mode' to be 'auto' or 'manual' and 'manual' would disable targeted loading state and consideration of the loading state of children, and instead would rely on another prop like 'setLoading' or something.

This would clarify that the Loading component is either driven by state that can be inferred from the renderer (those components whose state are waiting for backend information) or directly specified by the user (indirectly related components, like a search option that should be put into a loading state while something else is executing).

AnnMarieW:
@JamesKunstle I'm not sure what you mean re the mode prop.  Can you clarify or supply some code?

JamesKunstle:
Yes certainly:

So `mode` prop could just be `manual-mode` and be a boolean, True or False.

When not in `manual-mode`, the Loading component behaves exactly as it does now, changing as an effect of other components waiting on a response from the backend.

In `manual` mode, the Loading component would honor the additional prop `set-loading` (naming not final), a boolean that would be under full manual control from the user.

```python
        if (manual_mode) {
            setShowSpinner(set_loading);
            return;
        }
```

Here's a concrete example:

Imagine that I have a dcc.Interval component that fires every 5 seconds, making a request to the backend to check whether an asynchronous resource is available. Another component, like a Dropdown, indirectly relies on the availability of that resource, so it should be in a forced Loading state that stays enabled even when the dcc.Interval is waiting to fire again.

If I could use manual mode, a final successful return from the callback that handles the Interval logic could disable loading for the Dropdown.

This way, Loading becomes a developer-controllable state that is managed by application logic rather than by the dash renderer.

components/dash-core-components/src/components/Loading.react.js
```diff
@@ -164,4 +236,41 @@ Loading.propTypes = {
...
+     * has been determined. Default True.  Use when also setting `delay_show`.
+     */
+    show_initially: PropTypes.bool,
+
+    /**
+     * Specify component and prop to trigger showing the loading spinner
+     * example: `[{"output-container": "children"}]`
+     *
+     */
+    target_components: PropTypes.arrayOf(PropTypes.object),
```
alexcjohnson, line 267: I wonder if this would be better as
```js
objectOf(arrayOf(string))
```
or perhaps
```js
objectOf(oneOfType([string, arrayOf(string)]))
```
so you could do something like:
```py
target_components={"my_table": ["data", "columns"], "my_dropdown": "options"}
```

components/dash-core-components/src/components/Loading.react.js
```diff
@@ -7,66 +7,130 @@ import CircleSpinner from '../fragments/Loading/spinners/CircleSpinner.jsx';
...
+        height: '100%',
+        width: '100%',
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+    };
+    const hiddenContainer = mergeRight(
+        {visibility: 'hidden', position: 'relative'},
+        overlay_style
+    );
```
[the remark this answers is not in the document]
AnnMarieW: I thought about that, but I expect most people would tweak the opacity and background color  depending on the theme (light/dark).   Or do know some option that would work in most cases?

alexcjohnson:
re: `mode` - agreed that the name is a little ambiguous, but I do like having a single prop with three values, rather than a second prop that only applies to one of the values of the other.

What about borrowing from CSS: `display: "show" | "hide" | "auto"`?

emilykl:
Re: api for enabling/disabling, what feels most intuitive to me is `disabled` for consistency with `dcc.Interval` and the input components, and also it's a more transparent name -- doesn't lend itself super easily to 3 values though.

components/dash-core-components/src/components/Loading.react.js
AnnMarieW: I added this, plus it's possible to use a wildcard.  This will trigger loading for any prop in the `"my_table"` component
```
target_components={"my_table": "*", "my_dropdown": "options"}
```

JamesKunstle:
> Re: api for enabling/disabling, what feels most intuitive to me is `disabled` for consistency with `dcc.Interval` and the input components, and also it's a more transparent name -- doesn't lend itself super easily to 3 values though.

That, for me, feels ergonomically less nice for the same reason that it does for dcc.Interval - negating 'disabled' in my head to set the correct boolean value takes a few more cycles than I want, and I normally create global variables to solve it, like "INTERVAL_DISABLED" = True, "INTERVAL_ENABLED" = False, etc.

> re: `mode` - agreed that the name is a little ambiguous, but I do like having a single prop with three values, rather than a second prop that only applies to one of the values of the other.
> 
> What about borrowing from CSS: `display: "show" | "hide" | "auto"`?

Passing a string parameter for 'display' seems reasonable

alexcjohnson:
Ok great, let’s go with `display: "show" | "hide" | "auto"`

AnnMarieW:
> Ok great, let’s go with `display: "show" | "hide" | "auto"`

Done

components/dash-core-components/src/components/Loading.react.js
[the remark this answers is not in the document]
ndrezn: Could probably pick a neutral grey? I've found `#7F8487` is a good middle ground grey.

ndrezn (approved):
💃 Super excited about these changes. I think it'd be nice to have a dedicated `opacity` prop but I understand concerns about color... so I don't feel super strongly about this, as long as it's well-documented.

On that note... is there a corresponding docs PR?

components/dash-core-components/src/components/Loading.react.js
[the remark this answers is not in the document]
AnnMarieW: I can try that - but still want to give the ability for people to use any CSS in `overlay_style`.  Are you suggesting another prop, perhaps  that sets `{"visibility":"visible", "opacity": .5, "backgroundColor": "#7F8487"}`?
Then what if people set both props?

If there are a couple examples in the docs, it might work OK to just leave it as-is?

AnnMarieW:
I can work on the docs PR next.
Note that I can't merge PRs here, so can you do that for me?

components/dash-core-components/src/components/Loading.react.js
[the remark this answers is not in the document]
ndrezn: Yeah, maybe let's leave as-is for now with good examples in the docs, and see if there is community feedback from there?

emilykl:
Super excited for these!

T4rk1n:
> I can work on the docs PR next.
Note that I can't merge PRs here, so can you do that for me?

Can we hold on merging while I work on a fix for #2775

T4rk1n (approved):
:woman_dancing:

AnnMarieW:
Here's the PR for the docs https://github.com/plotly/ddk-dash-docs/pull/2570

schultz77:
Hello AnnMarieW,

thank you very much for implementing the features delay_show and delay_hide to dcc.Loading.
Do I have to update dash to have access to it?

Thank You very much in Advance

> TypeError: The `dcc.Loading` component (version 2.14.1) received an unexpected keyword argument: `delay_show`
> Allowed arguments: children, className, color, debug, fullscreen, id, loading_state, parent_className, parent_style, style, type
> 

dash==2.14.1
dash-bootstrap-components==1.5.0
dash-core-components==2.0.0

schultz77:
Ok! Thank You for your quick reply. In the meantime I'll keep using dbc.Spinner...

PR #2762, opened 2024-02-16
[WIP] Dynamic loading of component libraries.

T4rk1n:
Change the loading of component libraries to be based on the layout and loaded asynchronously instead of a script inserted on page index. 

Adds:
- `dash.Dash(dynamic_loading=True)` Enable/Disable dynamic loading.
- `dash.Dash(preloaded_libraries=['dash_design_kit'])` Preload this list of dash component libraries on the index.

Todos:
- [ ] Add test.
- [ ] Add changelog entry.


dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -508,8 +512,37 @@ function handleServerside(
...
 
                     if (!long || data.response !== undefined) {
-                        completeJob();
-                        finishLine(data);
+                        const newLibs: string[] = [];
+                        Object.values(data.response as any).forEach(
+                            (newData: any) => {
+                                Object.values(newData).forEach(newProp => {
+                                    crawlLayout(newProp, (c: any) => {
+                                        if (!libraries.includes(c.namespace)) {
```
alexcjohnson, line 520: minor perf point - wouldn't it be better to leave `libraries` as a map?
```js
const {libraries} = getState();
...
if (!(c.namespace in libraries))
```

dash/dash-renderer/src/APIController.react.js
```diff
@@ -46,6 +53,10 @@ const UnconnectedContainer = props => {
     if (!events.current) {
         events.current = new EventEmitter();
     }
+
+    const [libraryReady, setLibraryReady] = useState(false);
+    const onLibraryReady = useCallback(() => setLibraryReady(true), []);
```
alexcjohnson, line 58: curious, what does `useCallback` do for us here? Prevents unnecessary rerender of `LibraryManager`?

@plotly/dash-test-components/src/components/StyledComponent.js
```diff
@@ -15,7 +15,7 @@ StyledComponent.propTypes = {
     /**
      * The style
      */
-    style: PropTypes.shape,
+    style: PropTypes.any,
```
alexcjohnson, line 18: `PropTypes.object` perhaps?

dash/dash-renderer/src/APIController.react.js
T4rk1n: Yes, it doesn't need to be redefined every render, just the first.

dash/dash-renderer/src/actions/callbacks.ts
T4rk1n: I think the perf gain would be minimal here since the number of libraries should never be high, I added another condition that was missing and I think it's more readable if both conditions are the same.

alexcjohnson (approved):
💃 Amazing - really excited about this one!

PR #2768, opened 2024-02-22
Ignore component files based on their filename or parent directory 

HammadTheOne:
This PR updates the ignore pattern check we run as we're gathering component files during `dash-generate-components`. 

As an example, if the filepath is in the following common form, `src/lib/components/_MyComponent.react.js`, the RegEx doesn't find anything matching the default `^_` pattern, or any other pattern that might match the beginning of a file or directory unless it's at the top-level - the directory passed in via `components-source` argument. 

With this change, we would ignore the component file if the filename or any other directories along the path contain the ignore pattern, which I believe fits more in line with the intended function.

### optionals

### DIFF
--- a/dash/extract-meta.js
+++ b/dash/extract-meta.js
@@ -166,7 +166,11 @@ function gatherComponents(sources, components = {}) {
     const filepaths = [];
 
     const gather = filepath => {
-        if (ignorePattern && ignorePattern.test(filepath)) {
+
+        // Ignore file names that match the ignore pattern.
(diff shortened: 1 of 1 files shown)


alexcjohnson:
On the other hand this way you wouldn't be able to use a pattern involving the whole path, only the segments. So maybe it would be better to update the default pattern to achieve the same goal, something like `r"(^|\\|/)_"` (or whatever is needed to get the proper escaping through to `extract_meta.js`)?

PR #2770, opened 2024-02-27
Add running to regular callbacks.

### LINKED ISSUE #2559: [Feature Request] Enable running functionality in non-background callbacks
The running keyword in callbacks currently only works when `background=True` is set for callbacks. However, sometimes you have a normal callback, that takes a while to execute, that you don't want to turn in a background callback, but which you do want to have running functionality for. 

Currently, I solve this by manually creating a callback loop, i.e. by having an entry-exit callback that disables/enables the button which then triggers a second callback that does the actual calculation. But that is a bit cumbersome to set up every time, since you have to be careful to make sure that the loop properly closes. For example, if you run the same calculation twice consecutively, the exit callback might not be triggered the second time because the calculation return data didn't change compared to the first run and therefore Dash won't update the component that triggers the loop exit condition. (The solution is to add extra data that will always update, e.g. a calculation counter)

Having this functionality enabled via the running keyword that already works for background callbacks would be amazing.

T4rk1n:
Running now works with regular callbacks and not just background callbacks.

Resolve #2559


tests/integration/callbacks/test_basic_callback.py
```diff
@@ -791,3 +791,31 @@ def on_click(_):
...
+    )
+
+    @app.callback(
+        Output("output", "children"),
+        Input("start", "n_clicks"),
+        running=[[Output("running", "children"), "on", "off"]],
+        prevent_initial_call=True,
+    )
+    def on_click(_):
+        time.sleep(1.0)
```
alexcjohnson, line 814: Rather than a `sleep` can we do this test with a `Lock`, like in [`cbsc013`](https://github.com/plotly/dash/blob/4822b7b61915ce21bf8c0367492e70489fd4d02b/tests/integration/callbacks/test_basic_callback.py#L593-L612) for example? It's both faster and more robust that way.

alexcjohnson (approved):
💃 Fantastic feature @T4rk1n! Just one small comment on the test, then it's ready to go!

alexcjohnson:
Oh and a changelog entry 🎉

tests/integration/callbacks/test_basic_callback.py
T4rk1n: I couldn't make it work with the lock, it updates too fast for the driver.

aGitForEveryone:
Amazing that you picked this up, such awesomeness!

PR #2771, opened 2024-03-01
Dynamic loading: map namespace to package name in case they differ.

T4rk1n:
Fix cases where the namespace of a component package differ from it's import module.


alexcjohnson (approved):
💃 Looks great, can you just update one of the test component packages so it makes use of this?

alexcjohnson (approved):

PR #2780, opened 2024-03-04
Fixes for dynamic loading.

### LINKED ISSUE #2775: [BUG] Weird component redraw issue in dash 2.16
Consider the following demo component:

```js
import { useEffect } from 'react';

const DemoComponent = (props) => {

    useEffect(() => {
        console.log('========================')
        console.log('DemoComponent is redrawning');
    })

    return <div >DemoComponent</div>;
}

DemoComponent.propTypes = {
    /**
     * Component id.
     */
    id: PropTypes.string,

    loading_state: PropTypes.shape({
        /**
         * Determines if the component is loading or not
         */
        is_loading: PropTypes.bool,
        /**
         * Holds which property is loading
         */
        prop_name: PropTypes.string,
        /**
         * Holds the name of the component that is loading
         */
        component_name: PropTypes.string
    }),

    /**
     * Dash-assigned callback that should be called to report property changes
     * to Dash, to make them available for callbacks.
     */
    setProps: PropTypes.func
};

DemoComponent.defaultProps = {
}

export default DemoComponent;
```

In the following case, useEffect is executed a lot more times in `2.16 ` than `2.15`:

```python
app.layout = html.Div(
    [
        'Dash: '+dash.__version__,
        html.Br(),
        html.Button('test', id='input'),
        html.Div(
            DemoComponent(),
            id='output'
        )
    ],
    style={
        'padding': 50
    }
)


@app.callback(
    Output('output', 'children'),
    Input('input', 'n_clicks'),
    prevent_initial_call=True
)
def demo(n_clicks):

    return DemoComponent()


if __name__ == '__main__':
    app.run(debug=True)
```

in `2.15`:

demo215

in `2.16`:

demo216

@T4rk1n @alexcjohnson

### LINKED ISSUE #2778: [BUG] dash_mantine_components can't be found in dash 2.16.0
**Describe your context**
When running a dash app with the latest dash version, we're currently getting these console errors (see screenshot) and the app doesn't seem to render anymore. 

```
dash                           2.16.0
dash_ag_grid                   31.0.1
dash-bootstrap-components      1.5.0
dash-core-components           2.0.0
dash-daq                       0.5.0
dash-html-components           2.0.0
dash-mantine-components        0.12.1
dash-table                     5.0.0
dash-testing-stub              0.0.2
```

Not sure if it's a bug on dash side or on the dash_mantine_library, so let me know if I should post the issue on the dash_mantine_library instead!

Thank you! Any help appreciated 🙏 

**Describe the bug**

With the latest dash version our app seems to break. Seems like the dependencies can't be loaded in (see screenshot of the output)

**Expected behavior**

App running as normally (works with older dash versions)

**Screenshots**
Output when running the app in debug mode now:

Screenshot 2024-03-04 at 12 12 15

T4rk1n:
- Fix #2775, additional redraws.
- Fix #2778, dynamic loading and pages.

### DIFF
--- a/dash/dash-renderer/src/libraries/LibraryComponent.tsx
+++ b/dash/dash-renderer/src/libraries/LibraryComponent.tsx
@@ -1,5 +1,4 @@
-import React, {useContext, useEffect} from 'react';
-import {LibrariesContext} from './librariesContext';
+import React from 'react';
 import Registry from '../registry';
 import {CheckedComponent} from '../CheckedComponent.react';
 import {createLibraryElement} from './createLibraryElement';
@@ -15,16 +14,6 @@ type LibraryComponentProps = {
 
 const LibraryComponent = (props: LibraryComponentProps) => {
     const {props_check, namespace, type, ...rest} = props;
-
-    const context = useContext(LibrariesContext);
-
-    useEffect(() => {
-        context.addToLoad(namespace);
-    }, []);
-
-    if (!context.isLoaded(namespace)) {
-        return <></>;
-    }
     const element = Registry.resolve({namespace, type});
     if (props_check) {
         return (
--- a/dash/dash-renderer/src/libraries/LibraryManager.tsx
+++ b/dash/dash-renderer/src/libraries/LibraryManager.tsx
@@ -7,6 +7,7 @@ import {isEmpty} from 'ramda';
 type LibrariesManagerProps = {
     children: JSX.Element;
     requests_pathname_prefix: string;
+    validation_layout: any;
     onReady: () => void;
     ready: boolean;
     layout?: any;
@@ -35,7 +36,7 @@ const LibraryProvider = (props: LibrariesManagerProps) => {
 };
 
 const LibraryManager = (props: LibrariesManagerProps) => {
-    const {children, ready, layout} = props;
+    const {children, ready, layout, validation_layout} = props;
 
     const [initialLibraries, setInitialLibraries] = useState<string[] | null>(
         null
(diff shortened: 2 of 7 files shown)


AnnMarieW:
FYI - I tried this fix with with the example in https://github.com/plotly/dash/issues/2781, but it didn't works, so it's probably a different issue.

dash/dash-renderer/src/libraries/LibraryComponent.tsx
```diff
@@ -15,16 +14,6 @@ type LibraryComponentProps = {
 
 const LibraryComponent = (props: LibraryComponentProps) => {
     const {props_check, namespace, type, ...rest} = props;
-
-    const context = useContext(LibrariesContext);
```
alexcjohnson, line 19: So I guess this was left over from a previous iteration but not needed the way you finally implemented the feature?

dash/dash-renderer/src/libraries/LibraryManager.tsx
```diff
@@ -44,14 +45,18 @@ const LibraryManager = (props: LibrariesManagerProps) => {
...
-            crawlLayout(layout, (child: any) => {
+            const addLib = (child: any) => {
                 if (child.namespace && !libraries.includes(child.namespace)) {
                     libraries.push(child.namespace);
                 }
-            });
+            };
+            crawlLayout(layout, addLib);
+            if (validation_layout) {
+                crawlLayout(validation_layout, addLib);
```
alexcjohnson, line 55: Doesn't this defeat the purpose though? The difference between `validation_layout` and `layout` is just items that we're not going to show on the initial view.

ndrezn:
@T4rk1n do you still want to review this or should this PR be closed after #2783 is merged?

T4rk1n:
Can be closed but I'll add the redraw test later.

PR #2784, opened 2024-03-06
Version 2.16.1

T4rk1n:


dash/_dash_renderer.py
```diff
@@ -1,6 +1,6 @@
 import os
 
-__version__ = "1.19.0"
+__version__ = "1.19.1"
```
ndrezn, line 3: This is interesting, I didn't realize we independently bump `dash_renderer`.

ndrezn (approved):
💃

dash/_dash_renderer.py
T4rk1n: It's automatically updated by the build from the package.json

PR #2789, opened 2024-03-11
Add library loading capacity to `_allow_dynamic_callbacks`.

T4rk1n:
When `_allow_dynamic_callbacks=True`, callback can now import a component library and automatically loaded on the callback returns. Same restrictions as adding callbacks during callbacks, only guaranteed to work for a single user, meant to be used for development purposes.

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -360,6 +360,7 @@ def register_callback(
             callback_ctx = kwargs.pop(
                 "callback_context", AttributeDict({"updated_props": {}})
             )
+            app = kwargs.pop("app", None)
             callback_manager = long and long.get("manager", app_callback_manager)
             error_handler = on_error or kwargs.pop("app_on_error", None)
             original_packages = set(ComponentRegistry.registry)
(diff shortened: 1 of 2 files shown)


dash/dash-renderer/src/utils/libraries.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (resource.type === '_js_dist') {
+        const element = document.createElement('script');
+        element.src = resource.url;
+        element.async = true;
+        prom = new Promise<void>((resolve, reject) => {
+            element.onload = () => {
+                resolve();
+            };
+            element.onerror = error => reject(error);
+        });
```
syl-oh, line 18: since the way you are creating the promise is the same for both, you could extract that into a function to avoid duplication. Like with a `makePromise(element)` which would do that.

ndrezn:
@T4rk1n: From @robertclaus, this change was tested in App Studio and works as expected 🥳 . See: https://github.com/plotly/notebook-to-app/issues/401#issuecomment-2269443734

gvwilson (approved):

PR #2795, opened 2024-03-13
Allow layout to be a list of components.

T4rk1n:
Add the ability to set the layout as a list of components.

Example:

```
from dash import html, dcc, Input, Output, Dash

app = Dash()

app.layout = [
    html.Div("one", id="one"),
    html.Div("two", id="two"),
]
```

### DIFF
--- a/dash/_validate.py
+++ b/dash/_validate.py
@@ -403,7 +403,7 @@ def validate_layout_type(value):
     ):
         raise exceptions.NoLayoutException(
             """
-            Layout must be a single dash component, a list of dash components
+            Layout must be a single dash component, a list of dash components,
             or a function that returns a dash component.
             """
         )
--- a/dash/_validate.py
+++ b/dash/_validate.py
@@ -398,12 +398,13 @@ def validate_index(name, checks, index):
 
 
 def validate_layout_type(value):
-    if not isinstance(value, (Component, patch_collections_abc("Callable"))):
+    if not isinstance(
+        value, (Component, patch_collections_abc("Callable"), list, tuple)
+    ):
         raise exceptions.NoLayoutException(
             """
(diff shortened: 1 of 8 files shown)


T4rk1n:
@AnnMarieW 
> So all this does is to make it unnecessary to use html.Div( ) as the container of app.layout?

Yes that is all it does, reduce the number of characters and remove a mostly useless div in many cases. :slightly_smiling_face:

alexcjohnson:
You can also think of it as changing from one existing rule to another - now layout looks like a `children` prop

dash/_validate.py
```diff
@@ -398,12 +398,13 @@ def validate_index(name, checks, index):
...
 
 def validate_layout_type(value):
-    if not isinstance(value, (Component, patch_collections_abc("Callable"))):
+    if not isinstance(
+        value, (Component, patch_collections_abc("Callable"), list, tuple)
+    ):
         raise exceptions.NoLayoutException(
             """
-            Layout must be a single dash component
+            Layout must be a single dash component, a list of dash components
```
ndrezn, line 406:
```suggestion
            Layout must be a single dash component, a list of dash components,
```
missing comma

ndrezn (approved):
Although I see @AnnMarieW 's point as a "new rule" I think generally this introduces better parity with `children`, and to me more importantly means that the "Hello World!" Dash examples _do not_ require `html.Div` but require instead the far more understandable Python list -- ofc you can always do it the traditional way but it'll be a nice simplication of our early docs that I expect won't take much away in terms of pedagogy (hopefully soon we can come to a place where Dash devs don't need to learn HTML concepts to build cool projects!).

Nice work Phillipe and nice test as well! 💃

Issue #2799, opened 2024-03-16
[Feature Request] Expose `dash.NoUpdate` for Type Hinting


2Ryan09:
Enable more comprehensive type hinting of dash callback methods by exposing `dash._callback.NoUpdate`.

Given `callback()` and `clientside_callback()` are already imported up, is there any reason not to also expose `NoUpdate`?

[A Minimal Dash App](https://dash.plotly.com/minimal-app) with comprehensive python typing:
```python
# import NoUpdate alongside any other top-level import
from dash import Dash, NoUpdate, html, dcc, callback, Output, Input
import plotly.express as px
import pandas as pd

# TYPE_CHECKING block
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from plotly.graph_objects import Figure

df = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/gapminder_unfiltered.csv')

app = Dash(__name__)

app.layout = html.Div([
    html.H1(children='Title of Dash App', style={'textAlign':'center'}),
    dcc.Dropdown(df.country.unique(), 'Canada', id='dropdown-selection'),
    dcc.Graph(id='graph-content')
])

@callback(
    Output('graph-content', 'figure'),
    Input('dropdown-selection', 'value')
)
def update_graph(value: str | None) -> Figure | NoUpdate:  # Complete Input and Output typing, no room for error
    dff = df[df.country==value]
    return px.line(dff, x='year', y='pop')

if __name__ == '__main__':
    app.run(debug=True)
```

alexcjohnson:
Thanks @2Ryan09 - makes sense, but I’m worried about confusion between the `no_update` value and the `NoUpdate` class (of which normally `no_update` is the lone instance). Is there a relevant convention we can use to make this distinction clear? Rename the class `NoUpdateType` or something, ala `None` / `NoneType`?

2Ryan09:
Unfortunately I don't believe there is any accept conventions for this other than the [variable names themselves be styled as CapWords](https://peps.python.org/pep-0008/#type-variable-names). 

`NoUpdateType` seems perfectly reasonable solution to me. Will update the PR.

2Ryan09:
On a higher note, should `NoUpdate` be a [`TypedDict`](https://typing.readthedocs.io/en/latest/spec/typeddict.html#typeddict)

alexcjohnson:
[...]

PR #2800, opened 2024-03-16
expose NoUpdate

2Ryan09:
Expose `NoUpdate` at top-level dash.

The primary application is for type hinting, allowing for comprehensive type checking of dash callbacks.

See: #2799

## Contributor Checklist

- [X] Expose `NoUpdate` from `dash._callback`

### DIFF
--- a/dash/__init__.py
+++ b/dash/__init__.py
@@ -19,7 +19,8 @@ from . import html  # noqa: F401,E402
 from . import dash_table  # noqa: F401,E402
 from .version import __version__  # noqa: F401,E402
 from ._callback_context import callback_context  # noqa: F401,E402
-from ._callback import callback, clientside_callback, NoUpdate  # noqa: F401,E402
+from ._callback import callback, clientside_callback  # noqa: F401,E402
+from ._callback import NoUpdate as NoUpdateType  # noqa: F401,E402
 from ._get_app import get_app  # noqa: F401,E402
 from ._get_paths import (  # noqa: F401,E402
     get_asset_url,
--- a/dash/__init__.py
+++ b/dash/__init__.py
@@ -19,8 +19,7 @@ from . import html  # noqa: F401,E402
 from . import dash_table  # noqa: F401,E402
 from .version import __version__  # noqa: F401,E402
 from ._callback_context import callback_context  # noqa: F401,E402
-from ._callback import callback, clientside_callback  # noqa: F401,E402
-from ._callback import NoUpdate as NoUpdateType  # noqa: F401,E402
+from ._callback import callback, clientside_callback, NoUpdateType  # noqa: F401,E402
 from ._get_app import get_app  # noqa: F401,E402
 from ._get_paths import (  # noqa: F401,E402
     get_asset_url,
--- a/dash/__init__.py
+++ b/dash/__init__.py
@@ -19,7 +19,7 @@ from . import html  # noqa: F401,E402
 from . import dash_table  # noqa: F401,E402
 from .version import __version__  # noqa: F401,E402
 from ._callback_context import callback_context  # noqa: F401,E402
-from ._callback import callback, clientside_callback  # noqa: F401,E402
+from ._callback import callback, clientside_callback, NoUpdate  # noqa: F401,E402
 from ._get_app import get_app  # noqa: F401,E402
 from ._get_paths import (  # noqa: F401,E402
     get_asset_url,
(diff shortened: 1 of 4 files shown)


dash/__init__.py
```diff
@@ -20,6 +20,7 @@
 from .version import __version__  # noqa: F401,E402
 from ._callback_context import callback_context  # noqa: F401,E402
 from ._callback import callback, clientside_callback  # noqa: F401,E402
+from ._callback import NoUpdate as NoUpdateType  # noqa: F401,E402
```
alexcjohnson, line 23: Might be better do rename it inside `_callback.py`, otherwise there might be situations where the same thing shows up as either name. I wouldn't consider that a breaking change, this hasn't ever been part of the exposed API.

ndrezn: @KoolADE85 brought up this point as well.

Compare:
```python
type(no_update) # <class 'dash._callback.NoUpdate'>
```
whereas with `None`, it's
```python
type(None)      # <class 'NoneType'>
```

We'll want this `type` to be discoverable & renaming rather than aliasing enables that.

T4rk1n: I'd rather it not be postfixed.

ndrezn: @T4rk1n how come?

2Ryan09:
I just pushed the minimal possible change in my mind: renaming `NoUpdate` to `NoUpdateType` and exposing it. 

This does not implement the behavior @ndrezn mentions as it will still be of type `dash._callback.NoUpdateType`, but is there value in it being `dash.NoUpdate`? It's the only class as the other two imported into `dash` are functions, so maybe `NoUpdate` is just declared in the wrong module?

Given `NoUpdate` can be entirely described by `{"_dash_no_update": "_dash_no_update"}` (see context in #2800), can it not somehow inherit from `Dict` or `TypedDict`? This could make its behavior much more predictable and pythonic.

dash/__init__.py
T4rk1n: What we have here is not a type but a class. Since everything is a class in Python it feel meaningless to have a postfix of `Type` which it is not.

T4rk1n:
> Given `NoUpdate` can be entirely described by `{"_dash_no_update": "_dash_no_update"}` (see context in #2800), can it not somehow inherit from `Dict` or `TypedDict`? This could make its behavior much more predictable and pythonic.

It does not equate to a typed dict, it's a singleton object, the dict representation is for internal use.

2Ryan09:
In an attempt to keep the implementation of this PR at a minimum and, if I'm interpreting the conversation correctly up until now, is there any objection to the current state of this PR?

ndrezn (approved):
💃 for me -- thanks for making those changes. @T4rk1n , any lingering comments from you?

T4rk1n:
> any lingering comments from you?

I'd rather we fix the cases where using `NoUpdate` would be creating an error instead of renaming the variable in hope it won't be used.
I think the only case is actually just returning the class like this:

```
from dash import NoUpdate

def cb(...):
    return NoUpdate
```
intead of `return NoUpdate()`

Think we can add the check like this:

```
    @staticmethod
    def is_no_update(obj):
        return obj is NoUpdate or isinstance(obj, NoUpdate) or (
            isinstance(obj, dict) and obj == {"_dash_no_update": "_dash_no_update"}
        )
```

Then change all our checks to `NoUpdate.is_no_update(obj)`:

- https://github.com/plotly/dash/blob/15e97b85d24461dfce907d038b7788db75d1f005/dash/_callback.py#L474
- https://github.com/plotly/dash/blob/15e97b85d24461dfce907d038b7788db75d1f005/dash/_callback.py#L479

2Ryan09:
> > any lingering comments from you?
> 
> I'd rather we fix the cases where using `NoUpdate` would be creating an error instead of renaming the variable in hope it won't be used. I think the only case is actually just returning the class like this:
> 
> ```
> from dash import NoUpdate
> 
> def cb(...):
>     return NoUpdate
> ```
> 
> intead of `return NoUpdate()`
> 
> Think we can add the check like this:
> 
> ```
>     @staticmethod
>     def is_no_update(obj):
>         return obj is NoUpdate or isinstance(obj, NoUpdate) or (
>             isinstance(obj, dict) and obj == {"_dash_no_update": "_dash_no_update"}
>         )
> ```
> 
> Then change all our checks to `NoUpdate.is_no_update(obj)`:
> 
> * https://github.com/plotly/dash/blob/15e97b85d24461dfce907d038b7788db75d1f005/dash/_callback.py#L474
> * https://github.com/plotly/dash/blob/15e97b85d24461dfce907d038b7788db75d1f005/dash/_callback.py#L479

I agree with this. Specifically, `return NoUpdate` seems like a more desirable behavior than `return no_update`.

However, such a change seems adjacent to the topic of this PR which is to simply expose the value. I am more than happy to open another PR suggesting this change.

Issue #2697, opened 2023-11-20
[BUG] implementation of hashlib in _utils.py fails in FIPS environment 


caplinje-NOAA:
**Describe your context**
CentOS FIPS environment
Python 3.11.5

```
dash
dash-bootstrap-components
dash-leaflet
numpy
pandas
plotly
pyproj
scipy
xarray
tables
```

**Describe the bug**
When deployed using rstudio-connect, the app fails to initialize with the following error:
`...dash/_utils.py", line 144, in _concat
hashed_inputs = hashlib.md5(
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS`

The error can be replicated simply with python3 in a similar environment, as was done in this redhat bug report:
https://bugzilla.redhat.com/show_bug.cgi?id=1744670

There is a fair amount of discussion about this issue in other projects (where reporters on average do a better job of describing the issue):

https://github.com/dask/dask/issues/8608
https://github.com/Linuxfabrik/lib/issues/30
https://github.com/PrefectHQ/prefect/issues/7615

Where the suggested fix is to use a different hasher (e.g. blake3) or using the hashilb.md5 optional flag used_for_security=False (python>=3.9) if you're not hashing sensitive data. I for some reason did not see this error while in python 3.8.  I believe this is a necessary fix to make dash compatible with FIPS environments.

alexcjohnson:
Thanks @caplinje-NOAA - FIPS is new to me, but I don't see any reason we can't switch to `sha256` or some such. That's probably easier than trying to use `used_for_security=False` since we still support back to Py3.6, though I don't believe there are any security implications of our use of hashes here, we just need repeatability and uniqueness.

Feel like making a PR? For completeness here are all the places I see us using `hashlib`. The `dash.py` usage is fine, and if we're OK with `sha1` so is the `long_callbacks/managers/__init__.py` usage but I see no harm converting that to `sha256` as well.

```py
> ag --py hashlib
development/_jl_components_generation.py
9:import hashlib
419:        hex=u.hex[:-12] + hashlib.md5(package_name.encode("utf-8")).hexdigest()[-12:]

_callback.py
2:import hashlib
530:        function_name = hashlib.md5(clientside_function.encode("utf-8")).hexdigest()

long_callback/managers/__init__.py
3:import hashlib
76:        return hashlib.sha1(str(hash_dict).encode("utf-8")).hexdigest()
105:        return hashlib.sha1(

dash.py
15:import hashlib
[...]

PR #2817, opened 2024-03-26
Fix to hashing compatability issues [2697] with some environments (FIPS)

caplinje-NOAA:
This PR resolves [2697](https://github.com/plotly/dash/issues/2697), and is mostly a find and replace of hashlib.md5 and hashlib.sha1 with hashlib.sha256.  I did not replace one instance of sha1 in testing.  sha256 is slightly slower than md5, but testing did not show a measurable change in performance for creating callback ids.  Further, though they are generally extremely unlikely, sha256 has a reduced risk of collisions. 

## Contributor Checklist

   -  [x] swapped hashing algorithms md5 and sha1 to sha256
- [x] I have run the tests locally and one test did indeed fail (test_inin027_multi_page_without_pages_folder). Apparently this is not an issue as the tests ran fine for others locally and in the CI. 
   -  [x] tested performance of create_callback_id for md5, sha256, blake2b, and sha512.  There was no measurable impact to performance, even when creating ids for thousands of components, despite md5 being known to be somewhat faster. This means hashing ids is a negligible portion of the total runtime associated with creating callbacks. 
   -  [x] tested that switching to sha256 solves the compatibility issue in my deployment environment, by simply overloading the method in the current dash release.  
   -  Both additional tests exist here: [dash_md5_patch_testing](https://github.com/caplinje-NOAA/dash_md5_patch_testing)
   

### optionals

### DIFF
--- a/dash/development/build_process.py
+++ b/dash/development/build_process.py
@@ -97,8 +97,7 @@ class BuildProcess:
             logger.info("bundles in %s %s", folder, copies)
 
             for copy in copies:
-                # note md5 has been replaced with sha256, leaving string 'MD5 (hash)' as is because impacts are unclear
-                payload[f"MD5 ({copy})"] = compute_hash(self._concat(folder, copy)) 
+                payload[f"SHA256 ({copy})"] = compute_hash(self._concat(folder, copy)) 
 
         with open(self._concat(self.main, "digest.json"), "w", encoding="utf-8") as fp:
             json.dump(payload, fp, sort_keys=True, indent=4, separators=(",", ":"))
--- a/dash/development/build_process.py
+++ b/dash/development/build_process.py
@@ -8,7 +8,7 @@ import coloredlogs
 import fire
 import requests
 
-from .._utils import run_command_with_process, compute_md5, job
+from .._utils import run_command_with_process, compute_hash, job
 
 logger = logging.getLogger(__name__)
 coloredlogs.install(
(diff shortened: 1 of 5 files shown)


emilykl:
All looks good to me. Thanks for your work on this, @caplinje-NOAA .

The failing test you mentioned is working in the CI, and also on my local machine, so I don't see an issue there.

dash/development/build_process.py
```diff
@@ -97,7 +97,8 @@ def digest(self):
             logger.info("bundles in %s %s", folder, copies)
 
             for copy in copies:
-                payload[f"MD5 ({copy})"] = compute_md5(self._concat(folder, copy))
+                # note md5 has been replaced with sha256, leaving string 'MD5 (hash)' as is because impacts are unclear
+                payload[f"MD5 ({copy})"] = compute_hash(self._concat(folder, copy))
```
emilykl, line 101: @T4rk1n @alexcjohnson Can this string be safely changed? By my understanding it shouldn't cause any issues, but looking for confirmation.

T4rk1n: It's only used for manual comparison:
> `renderer digest` computes the content hash of each asset in `dash_renderer` folder, prints out the result in logs, and dumps into a JSON file `digest.json`. Use this when you have a doubt about the current assets in `dash_renderer`, and compare it with previous result in one shot by this command.

It should contains the new algorithm.

emilykl: Thanks @T4rk1n .

@caplinje-NOAA , can you change this string to `SHA256`?

caplinje-NOAA: @emilykl I corrected the payload string relative to this, let me know if any additional edits/corrections are required. Thanks.

dash/development/build_process.py
```diff
@@ -97,7 +97,7 @@ def digest(self):
             logger.info("bundles in %s %s", folder, copies)
 
             for copy in copies:
-                payload[f"MD5 ({copy})"] = compute_md5(self._concat(folder, copy))
+                payload[f"SHA256 ({copy})"] = compute_hash(self._concat(folder, copy))
```
T4rk1n, line 100:
```suggestion
                payload[f"SHA256 ({copy})"] = compute_hash(self._concat(folder, copy))
```

T4rk1n (approved):
:woman_dancing:

emilykl (approved):
LGTM

Issue #2820, opened 2024-03-29
Allow dcc.Dropdown with multi=True to stay open when item is selected


WWakker:
When you have a dropdown with `multi=True`, a user might want to select multiple items at the same time. However, `dcc.Dropdown` closes when you select one item, and to select another item you have to open it again. 

It would be much more user-friendly if `dcc.Dropdown` can stay open when an item is selected, or at least have it as an option. With `rshiny` this is already possible, see the [multi-select example here](https://shiny.posit.co/r/gallery/widgets/selectize-examples/).

There seems to be a way to [hack the javascript code to get it working](https://community.plotly.com/t/dash-dropdown-select-multi-keep-options-open-after-selection/34576/6), but a proper solution would be to have a boolean option in dcc.Dropdown, e.g. `close_on_select`, or `multi_close_on_select`.

TillerBurr:
Duplicate of #2669.
This is something that I was working on when I have time.

WWakker:
I forgot I already had opened that issue!

@TillerBurr Do you see any difficulties in the implementation? I don't know how far you got already but I could also try to have a look.

TillerBurr:
@WWakker No worries. It actually wasn't that bad implementing. The only thing that was slightly challenging were the integration tests, mainly making sure the menu stayed open. 

I had forgotten that I had started working on this a couple weeks ago and this issue reminded me that I should figure out those tests.

celia-lm:
Related! https://github.com/plotly/dash/issues/1934

antons-soundskrit:
The dropdown should revert to the previous scroll state when reopened, so that you don't have to scroll all the way back down to the position you were. This is especially useful when you have hundreds of dropdown options. The dropdown with `multi=False` does this, but with `multi=True`, the scroll always reverts to the top. This is even more convenient than leaving the dropdown open after a selection and should be standard behavior.

A workaround to revert back to the previous scroll position for a dropdown with `multi=True` is to create a `assets/scroll.js` file, then put the  ids of your dash dropdowns in `document.querySelector("#id1,#i2,...")` below.

```
var scrollPos = 0;

document.addEventListener('DOMContentLoaded', () => {
    
    function setup_click_listener() {
        var dropdownContainer = document.querySelector("#dropdown, #dropdown-directionality-measurements"); // put dash dropdown ids here
        if (dropdownContainer == null) { // Try again until the dropdown exists
            setTimeout(setup_click_listener, 200);
            return;
        }
[...]

PR #2821, opened 2024-03-29
[Feature] Allow Dropdown to Remain Open for `multi==True`

### LINKED ISSUE #2669: [Feature Request] Add option not to close dcc.Dropdown when multi=True
Currently a dropdown where you can select multiple values ( `dcc.Dropdown(['a', 'b', 'c'], multi=True)`) closes when you select a value. This can be inconvenient if a user want to select multiple values straight away.

It would be great if an option could be added that allows for keeping the dropdown open when you click on a value.

TillerBurr:
Closes #2669 (and #2820, a duplicate). This PR adds the ability to set the prop `close_on_select`. If `multi` is `True` and this prop is `False`, the menu selection remains open. If this new prop is `True`, it is the same behavior. The `close_on_select` prop has no effect when `multi` is `False`, it well essentially ignore the value. I did it that way, as I don't see the value changing after the component is created, similar to `multi`.

This PR will probably have a need to update the dash docs for the [dropdown](https://dash.plotly.com/dash-core-components/dropdown).

## Contributor Checklist

   -  [x] Add the prop `close_on_select` to the dropdown component
   -  [x] Add tests

### optionals

### DIFF
--- a/components/dash-core-components/src/fragments/Dropdown.react.js
+++ b/components/dash-core-components/src/fragments/Dropdown.react.js
@@ -24,6 +24,7 @@ const TOKENIZER = {
 
 const RDProps = [
     'multi',
+    'close_on_select',
     'clearable',
     'searchable',
     'search_value',
@@ -41,6 +42,7 @@ const Dropdown = props => {
         clearable,
         searchable,
         multi,
+        close_on_select,
         options,
         setProps,
         style,
@@ -53,7 +55,9 @@ const Dropdown = props => {
     if (!persistentOptions || !isEqual(options, persistentOptions.current)) {
         persistentOptions.current = options;
     }
-
+    if (!multi) {
+        setProps({close_on_select: true});
+    }
     const [sanitizedOptions, filterOptions] = useMemo(() => {
         let sanitized = sanitizeOptions(options);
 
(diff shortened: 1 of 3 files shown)


TillerBurr:
On second thought, it would make more sense to not write over the `close_on_select` prop. If someone were to change multi, the prop would be lost and it would always close.

components/dash-core-components/src/fragments/Dropdown.react.js
```diff
@@ -155,6 +157,7 @@ const Dropdown = props => {
                 filterOptions={filterOptions}
                 options={sanitizedOptions}
                 value={value}
+                closeOnSelect={multi ? close_on_select : true}
```
T4rk1n, line 160: > On second thought, it would make more sense to not write over the `close_on_select` prop. If someone were to change multi, the prop would be lost and it would always close.

:+1:

WWakker:
@T4rk1n I’m willing to help out for this request. Is there something that still needs to be done? Probably the docs need to be updated for the new parameter but this doesn’t seem to be open source

T4rk1n:
@WWakker Just need to change this: `closeOnSelect={multi ? close_on_select : true}` for something that is more user friendly. Either we can set the `close_on_select` default prop to true or invert the prop condition name to something like `stay_open` and give `closeOnSelect(!stay_open}`. I think the first option is closer to the wrapped dropdown api so might be better.

WWakker:
@T4rk1n I think this is already addressed by https://github.com/plotly/dash/pull/2821/commits/bc1d0d29bd8edefc57a629b4750c31da3bb1763f. The default `close_on_select` prop is true, and it can only by false if multi is true which I believe is the desired behavior?

robertclaus:
Closing as resolved in https://github.com/plotly/dash/pull/3202

PR #2822, opened 2024-03-29
Support Arbitrary callbacks

### LINKED ISSUE #1549: Allow for "null" / no output callbacks

### LINKED ISSUE #2111: [Long Callback] Support pattern matching for long callbacks side updates (progress/cancel/running)
Long callbacks `progress`, `running` and `cancel` dependencies should support pattern matching ids.

### LINKED ISSUE #2803: [Feature Request] Global set_props in backend callbacks.
Add a global `dash.set_props` to be used in callbacks to set arbitrary props not defined in the callbacks outputs, similar to the clientside `dash_clientside.set_props`.

Example:

```
app.layout =  html.Div([
    html.Div(id="output"),
    html.Div(id="secondary-output"),
    html.Button("click", id="clicker"),
])

@app.callback(
    Output("output", "children"),
    Input("clicker", "n_clicks"),
    prevent_initial_call=True,
)
def on_click(n_clicks):
    set_props("secondary-output", {"children": "secondary"})
```

T4rk1n:
- Allow no output callback, resolve #1549 
- Add global set_props, resolve #2803
- Fix side update pattern ids, fix #2111

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -426,7 +426,7 @@ max-public-methods=40
 max-returns=6
 
 # Maximum number of statements in function / method body
-max-statements=50
+max-statements=75
 
 # Minimum number of public methods for a class (see R0903).
 min-public-methods=2
--- a/.pylintrc
+++ b/.pylintrc
@@ -426,7 +426,7 @@ max-public-methods=40
 max-returns=6
 
 # Maximum number of statements in function / method body
-max-statements=75
+max-statements=50
 
 # Minimum number of public methods for a class (see R0903).
 min-public-methods=2
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -500,7 +500,7 @@ def register_callback(  # pylint: disable=R0914
             else:
                 if output_value is not None:
                     raise InvalidCallbackReturnValue(
-                        f"No output callback received return value: {output_value}"
+                        f"No-output callback received return value: {output_value}"
                     )
                 output_value = []
                 flat_output_values = []
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -226,6 +226,7 @@ def insert_callback(
     manager=None,
     running=None,
     dynamic_creator=False,
+    no_output=False,
 ):
     if prevent_initial_call is None:
         prevent_initial_call = config_prevent_initial_callbacks
@@ -234,7 +235,7 @@ def insert_callback(
         output, prevent_initial_call, config_prevent_initial_callbacks
     )
 
-    callback_id = create_callback_id(output, inputs)
+    callback_id = create_callback_id(output, inputs, no_output)
     callback_spec = {
         "output": callback_id,
         "inputs": [c.to_dict() for c in inputs],
@@ -248,6 +249,7 @@ def insert_callback(
             "interval": long["interval"],
         },
         "dynamic_creator": dynamic_creator,
+        "no_output": no_output,
     }
     if running:
         callback_spec["running"] = running
@@ -262,6 +264,7 @@ def insert_callback(
         "raw_inputs": inputs,
         "manager": manager,
         "allow_dynamic_callbacks": dynamic_creator,
+        "no_output": no_output,
     }
     callback_list.append(callback_spec)
 
@@ -283,10 +286,12 @@ def register_callback(  # pylint: disable=R0914
         # Insert callback with scalar (non-multi) Output
         insert_output = output
         multi = False
+        no_output = False
     else:
         # Insert callback as multi Output
         insert_output = flatten_grouping(output)
         multi = True
+        no_output = len(output) == 0
 
     long = _kwargs.get("long")
     manager = _kwargs.get("manager")
@@ -315,6 +320,7 @@ def register_callback(  # pylint: disable=R0914
         manager=manager,
         dynamic_creator=allow_dynamic_callbacks,
         running=running,
+        no_output=no_output,
     )
 
     # pylint: disable=too-many-locals
@@ -333,7 +339,8 @@ def register_callback(  # pylint: disable=R0914
             app_callback_manager = kwargs.pop("long_callback_manager", None)
             callback_ctx = kwargs.pop("callback_context", {})
             callback_manager = long and long.get("manager", app_callback_manager)
-            _validate.validate_output_spec(insert_output, output_spec, Output)
+            if not no_output:
+                _validate.validate_output_spec(insert_output, output_spec, Output)
 
             context_value.set(callback_ctx)
 
@@ -443,6 +450,9 @@ def register_callback(  # pylint: disable=R0914
                         NoUpdate() if NoUpdate.is_no_update(r) else r
                         for r in output_value
                     ]
+                updated_props = callback_manager.get_updated_props(cache_key)
+                if len(updated_props) > 0:
+                    response["sideUpdate"] = updated_props
 
                 if output_value is callback_manager.UNDEFINED:
                     return to_json(response)
@@ -452,7 +462,10 @@ def register_callback(  # pylint: disable=R0914
             if NoUpdate.is_no_update(output_value):
                 raise PreventUpdate
 
-            if not multi:
+            if no_output:
+                output_value = []
+                flat_output_values = []
+            elif not multi:
                 output_value, output_spec = [output_value], [output_spec]
                 flat_output_values = output_value
             else:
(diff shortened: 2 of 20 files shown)


ndrezn:
Going to add a few more tests, then we should be almost there!

dash/_utils.py
```diff
@@ -149,6 +149,12 @@ def _concat(x):
             _id += f"@{hashed_inputs}"
         return _id
 
+    if no_output:
+        # No output will hash the inputs.
+        return hashlib.sha256(
+            ".".join(str(x) for x in inputs).encode("utf-8")
+        ).hexdigest()
```
emilykl, line 156: Any way to avoid duplicating this code from the above? Could `_concat` instead be modified to handle the no-output case?

dash/_callback.py
```diff
@@ -464,23 +481,29 @@ def add_context(*args, **kwargs):
...
-                    continue
-                for vali, speci in (
-                    zip(val, spec) if isinstance(spec, list) else [[val, spec]]
-                ):
-                    if not isinstance(vali, NoUpdate):
-                        has_update = True
-                        id_str = stringify_id(speci["id"])
-                        prop = clean_property_name(speci["property"])
-                        component_ids[id_str][prop] = vali
+            if not no_output:
```
syl-oh, line 485: `if not no_output:` that double negation 😅 is a bit hard on me.
Since it's a boolean maybe we could have something like `is....` or `should...`, but I am not sure what the output means in this context.

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -324,19 +324,37 @@ async function handleClientside(
...
+            }
+            dispatch(updateComponent(componentId, {[propName]: value}));
+        } else {
+            if (id.startsWith('{')) {
+                componentId = JSON.parse(id);
+            } else {
+                componentId = id;
+            }
+            dispatch(updateComponent(componentId, value));
+        }
```
syl-oh, line 357: I think we could remove some of the checks and method calls that happens twice if we do something like:

```js
  toPairs(outputs).forEach(([id, value]) => {
    let componentId, propName = value;
    if (id.includes('.')) {
      [componentId, propName] = id.split('.');
    } else {
      componentId = id;
    }

    if (componentId.startsWith('{')) {
      componentId = JSON.parse(componentId);
    }

    const props = propName === value ? value : {[propName]: value};
    dispatch(updateComponent(componentId, props));
  });
```

We always parse if it starts by `{` and if we split then we know the propName is not equal to `value` from the `let`.

dash/_callback.py
```diff
@@ -331,9 +337,12 @@ def wrap_func(func):
         def add_context(*args, **kwargs):
             output_spec = kwargs.pop("outputs_list")
             app_callback_manager = kwargs.pop("long_callback_manager", None)
-            callback_ctx = kwargs.pop("callback_context", {})
+            callback_ctx = kwargs.pop(
+                "callback_context", AttributeDict({"updated_props": {}})
```
emilykl, line 341: Why does `callback_ctx` always need to contain an `updated_props` key?

dash/_callback_context.py
```diff
@@ -247,5 +248,20 @@ def using_outputs_grouping(self):
...
+        if isinstance(component_id, dict):
+            ctx_value.updated_props[json.dumps(component_id)] = props
+        else:
+            ctx_value.updated_props[component_id] = props
+
 
 callback_context = CallbackContext()
+
+
+def set_props(component_id: typing.Union[str, dict], props: dict):
```
emilykl, line 263: Is there a reason why the function signature for `set_props` isn't structured as 3 inputs: `component_id`, `prop_name`, `value`?

That to me seems more consistent with the rest of the Dash API.

dash/dash-renderer/src/actions/dependencies.js
```diff
@@ -238,8 +234,22 @@ function validateDependencies(parsedDependencies, dispatchError) {
             });
         });
 
-        findDuplicateOutputs(outputs, head, dispatchError, outStrs, outObjs);
-        findMismatchedWildcards(outputs, inputs, state, head, dispatchError);
+        if (hasOutputs) {
```
emilykl, line 237: re: @syl-oh 's comments about the boolean named `no_update` in the Python code, maybe `has_outputs` would be a good name for that boolean.

dash/long_callback/_proxy_set_props.py
```diff
@@ -0,0 +1,7 @@
+class ProxySetProps(dict):
```
emilykl, line 1: What's the purpose of `ProxySetProps`? Maybe add a comment at the top of the file.

dash/dash-renderer/src/actions/dependencies.js
syl-oh: indeed :)

emilykl:
Aside from my comments, LGTM. ✨ 

In general there are a number of places where you've added separate code paths for `no_output=True` vs. `no_output=False` and I wonder whether some of them can be condensed down to a single code path... for example if we need to run some processing on the list of outputs, plausibly the same code should work on an empty list of outputs without the need for an if/else (or could be made to work with some small tweaks).

dash/_callback.py
T4rk1n: I remember there is an error with some mocked tests.

dash/_callback_context.py
T4rk1n: It is the same signature as the frontend `set_props`, we considered `set_props("id", prop=value)` but decided to stay with same API for frontend and backend.

dash/_utils.py
T4rk1n: Refactored in a function the hashing to be reused. :ok_hand:

.pylintrc
```diff
@@ -426,7 +426,7 @@ max-public-methods=40
 max-returns=6
 
 # Maximum number of statements in function / method body
-max-statements=50
+max-statements=75
```
alexcjohnson, line 429: Unless we really agree that the rules in this file should change, I'd rather we first try to adjust the code to conform (in this case, break up the offending function into smaller functions), then if that's not possible override the rule just for that function. That way we're being intentional about exceptions while still encouraging small modular code everywhere else.

dash/_callback_context.py
```diff
@@ -247,5 +248,20 @@ def using_outputs_grouping(self):
     def timing_information(self):
         return getattr(flask.g, "timing_information", {})
 
+    @has_context
+    def set_props(self, component_id: typing.Union[str, dict], props: dict):
+        ctx_value = _get_context_value()
+        if isinstance(component_id, dict):
+            ctx_value.updated_props[json.dumps(component_id)] = props
```
alexcjohnson, line 255: Does plain `json.dumps` work here? It's not quite the same as we have for stringifying component IDs in the callback definition:

https://github.com/plotly/dash/blob/9a4a47933e541ad0a61fd377314b576f93a2ea37/dash/dependencies.py#L46-L59

For `set_props` we don't need to handle wildcards, but I'd have thought we still want to sort keys and compress spaces.
(though looks like `_dump` there is unused and should be removed)

dash/_callback_context.py
alexcjohnson: It's true, this pattern is a little more javascripty, but consistency is good and it could be convenient to be able to set multiple props of the same component in one call.

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -324,19 +324,35 @@ async function handleClientside(
...
-            notifyObservers({id: componentId, props: {[propName]: value}})
-        );
+        dispatch(notifyObservers({id: component_id, props}));
+    };
+}
+
+function sideUpdate(outputs: any, dispatch: any) {
+    toPairs(outputs).forEach(([id, value]) => {
+        let componentId, propName;
+        if (id.includes('.')) {
```
alexcjohnson, line 344: Can't a dict ID contain `.`?

dash/dash.py
```diff
@@ -1312,6 +1312,12 @@ def dispatch(self):
             inputs_state = inputs + state
             inputs_state = convert_to_AttributeDict(inputs_state)
 
+            if cb.get("no_output"):
+                outputs_list = []
+            elif not outputs_list:
+                # FIXME Old renderer support?
```
alexcjohnson, line 1318: Why do we want to support old renderer? Embedded? We should rebuild & test embedded with this update anyway.

dash/_callback_context.py
T4rk1n: There is already a `stringify_id` here: https://github.com/plotly/dash/blob/9a4a47933e541ad0a61fd377314b576f93a2ea37/dash/_utils.py#L169-L172

I think the output is the same and is used in the base component, we should use the same everywhere.

dash/long_callback/_proxy_set_props.py
T4rk1n: Yes, the app is updated async like with the set_progress, it should be possible to update the same prop as the output.

dash/_callback.py
```diff
@@ -467,27 +467,28 @@ def add_context(*args, **kwargs):
             if NoUpdate.is_no_update(output_value):
                 raise PreventUpdate
 
-            if no_output:
+            component_ids = collections.defaultdict(dict)
+
+            if not has_output:
```
syl-oh, line 472: nit: you could reverse the `if` to make it `if has_output:` since there's an else after.

dash/_callback.py
```diff
@@ -452,35 +467,49 @@ def add_context(*args, **kwargs):
...
+                        zip(val, spec) if isinstance(spec, list) else [[val, spec]]
+                    ):
+                        if not isinstance(vali, NoUpdate):
+                            has_update = True
+                            id_str = stringify_id(speci["id"])
+                            prop = clean_property_name(speci["property"])
+                            component_ids[id_str][prop] = vali
+            else:
+                if output_value is not None:
+                    raise InvalidCallbackReturnValue(
```
emilykl, line 502: This gives a weird error message: 
Screen Shot 2024-04-30 at 5 47 28 PM

Not crucial to fix as part of this PR but would be nice if it's an easy fix.

dash/_callback.py
```diff
@@ -452,35 +467,49 @@ def add_context(*args, **kwargs):
...
+                        f"No output callback received return value: {output_value}"
```
emilykl, line 503: nit: should be "No-output" rather than "No output"

dash/_utils.py
```diff
@@ -131,24 +131,31 @@ def first(self, *names):
...
 
 
-def create_callback_id(output, inputs):
+def create_callback_id(output, inputs, no_output=False):
     # A single dot within a dict id key or value is OK
     # but in case of multiple dots together escape each dot
     # with `\` so we don't mistake it for multi-outputs
     hashed_inputs = None
 
+    def _hash_inputs():
```
emilykl, line 140: 👍

emilykl (approved):
LGTM @T4rk1n , no more comments from me. 🎉 

Could you double-check with @alexcjohnson whether his questions have been resolved before merging?

dash/dash.py
T4rk1n: I'll look into it when I'll update embedded next.

PR #2823, opened 2024-03-30
Fix: `None` in "wait" Methods Causing Incorrectly Passing Tests

### LINKED ISSUE #2818: [BUG] Dash Testing: `wait_for_text_to_equal` may incorrectly succeed when used with text `"None"`
**Describe your context**

```
dash                            2.16.1         
dash-core-components            2.0.0
dash-dangerously-set-inner-html 0.0.2
dash-flow-example               0.0.5
dash-html-components            2.0.0
dash-table                      5.0.0
dash-testing-stub               0.0.2
```

**Describe the bug**

When `wait_for_text_to_equal` is used to wait for the text `"None"`, the function will often succeed even when you would reasonably expect it to fail. 

I think this is part of the reason why the regression in #2733 wasn't caught by the tests.

This behavior is demonstrated by the following test case:

```python
import dash
from dash import html


def test_wait_for_text_to_equal_none(dash_duo):
    app = dash.Dash(__name__)
    app.layout = html.Div(id="my-div", children="Hello world")
    dash_duo.start_server(app)
    dash_duo.wait_for_text_to_equal("#my-div", "None", timeout=4)
```

**Expected behavior**

The test should fail because the contents of the `#my-div` div are never equal to `None` or `"None"`.

**Actual behavior**

The test passes.

**Explanation**

This happens because `wait_for_text_to_equal` checks not only the text content of the element, but also the value of the `value` attribute. ([see here](https://github.com/plotly/dash/blob/f7f8fb4c5893506e35cdeaec141310a95fe1486a/dash/testing/wait.py#L110C13-L113C14)).

If `value` is not defined we get a value of `None`, which is then converted to a string and therefore matches the string `"None"`.

So `dash_duo.wait_for_text_to_equal("#my-div", "None")` _always_ succeeds unless the target element has a defined `value`.

**Proposed solutions**

IMO the cleanest solution would be to modify `wait_for_text_to_equal` to check _only_ the element's text, and add a new function `wait_for_value_to_equal` which checks the value (or a generalized `wait_for_attr_to_equal` function). This would break backwards compatibility.

Alternatively we could have `wait_for_text_to_equal` ignore `value` if value is not defined, or issue a warning when used with the text `"None"`.

TillerBurr:
Closes #2818. This is my take on solving the issue. This method ignores the value prop if it is `None`. I think there is merit to adding a `wait_for_value_to_equal`, but this solves the immediate problem, while causing as little friction as possible.

As mentioned in #2818, there are also potential issues in the other methods in the `dash.testing.wait` namespace. Not sure if those cases need to be handled here.

## Contributor Checklist

   -  [x] Check if the value attribute is None and don't compare if it is.
   -  [x] Add tests to make sure the `None` isn't cast to a string.

### optionals

### DIFF
--- a/dash/testing/wait.py
+++ b/dash/testing/wait.py
@@ -64,8 +64,8 @@ class contains_text:
             elem = driver.find_element(By.CSS_SELECTOR, self.selector)
             logger.debug("contains text {%s} => expected %s", elem.text, self.text)
             value = elem.get_attribute("value")
-            return str(elem.text) in self.text or (
-                value is not None and str(value) in self.text
+            return self.text in str(elem.text) or (
+                value is not None and self.text in str(value)
             )
         except WebDriverException:
             return False
--- a/dash/testing/wait.py
+++ b/dash/testing/wait.py
@@ -1,5 +1,6 @@
 # pylint: disable=too-few-public-methods
 """Utils methods for pytest-dash such wait_for wrappers."""
+
 import time
 import logging
 from selenium.common.exceptions import WebDriverException
@@ -108,7 +109,9 @@ class text_to_equal:
             elem = self._get_element(driver)
             logger.debug("text to equal {%s} => expected %s", elem.text, self.text)
             value = elem.get_attribute("value")
-            return str(elem.text) == self.text or (value is not None and str(value) == self.text)
+            return str(elem.text) == self.text or (
+                value is not None and str(value) == self.text
+            )
         except WebDriverException:
             return False
 
--- a/dash/testing/wait.py
+++ b/dash/testing/wait.py
@@ -62,10 +62,9 @@ class contains_text:
         try:
             elem = driver.find_element(By.CSS_SELECTOR, self.selector)
             logger.debug("contains text {%s} => expected %s", elem.text, self.text)
-            if value := elem.get_attribute("value") is not None:
+            if (value := elem.get_attribute("value")) is not None:
                 return self.text in str(elem.text) or self.text in str(value)
-            else:
-                return self.text in str(elem.text)
+            return self.text in str(elem.text)
         except WebDriverException:
             return False
 
@@ -108,10 +107,9 @@ class text_to_equal:
         try:
             elem = self._get_element(driver)
             logger.debug("text to equal {%s} => expected %s", elem.text, self.text)
-            if value := elem.get_attribute("value") is not None:
+            if (value := elem.get_attribute("value")) is not None:
                 return str(elem.text) == self.text or str(value) == self.text
-            else:
-                return str(elem.text) == self.text
+            return str(elem.text) == self.text
         except WebDriverException:
             return False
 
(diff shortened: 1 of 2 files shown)


TillerBurr:
Tests are failing due to bad test that is fixed in https://github.com/plotly/dash/pull/2816. Once that PR is merged and this merged into this PR, the tests will all pass.

tests/integration/test_duo.py
```diff
@@ -27,10 +32,33 @@ def test_duo001_wait_for_text_error(dash_duo):
         == "text -> invalid not found inside element within 1.0s, found: Content"
     )
 
+    with pytest.raises(TimeoutException) as err:
+        dash_duo.wait_for_contains_text("#content", "None", timeout=1.0)
+
+    assert (
```
emilykl, line 38: 👍

dash/testing/wait.py
```diff
@@ -107,10 +107,9 @@ def __call__(self, driver):
         try:
             elem = self._get_element(driver)
             logger.debug("text to equal {%s} => expected %s", elem.text, self.text)
-            return (
-                str(elem.text) == self.text
-                or str(elem.get_attribute("value")) == self.text
-            )
+            if (value := elem.get_attribute("value")) is not None:
+                return str(elem.text) == self.text or str(value) == self.text
+            return str(elem.text) == self.text
```
emilykl, line 112:
```suggestion
            value = elem.get_attribute("value")
            return str(elem.text) == self.text or (value is not None and str(value) == self.text)
```

IMO better to avoid the walrus operator here as it's clearer without. Same goes for the other function.

emilykl:
I like this solution. Need to think through possible consequences though as this is technically a breaking change in Dash Testing.

dash/testing/wait.py
```diff
@@ -62,9 +63,9 @@ def __call__(self, driver):
         try:
             elem = driver.find_element(By.CSS_SELECTOR, self.selector)
             logger.debug("contains text {%s} => expected %s", elem.text, self.text)
-            return self.text in str(elem.text) or self.text in str(
-                elem.get_attribute("value")
-            )
+            if (value := elem.get_attribute("value")) is not None:
```
emilykl, line 66: @TillerBurr Can you change the logic here as well to remove the walrus operator? Once that's done, this PR looks good to me.

TillerBurr: Oops. That one is updated now.

T4rk1n (approved):
:woman_dancing:

dash/testing/wait.py
```diff
@@ -62,8 +63,9 @@ def __call__(self, driver):
         try:
             elem = driver.find_element(By.CSS_SELECTOR, self.selector)
             logger.debug("contains text {%s} => expected %s", elem.text, self.text)
-            return self.text in str(elem.text) or self.text in str(
-                elem.get_attribute("value")
+            value = elem.get_attribute("value")
+            return str(elem.text) in self.text or (
```
emilykl, line 67: This is swapped -- should be `self.text in str(elem.text)` and `self.text in str(value)`

TillerBurr: Thanks. Got confused on the order. Fixed.

emilykl (approved):
Looks good to me @TillerBurr -- thanks for the contribution! ✨ 

There's some failing tests I want to investigate but I'm pretty sure they're not related to your changes.

PR #2829, opened 2024-04-05
implementing `callback_fallback` 

BSd3v:
implementing `callback_fallback` on `Dash` and `callback` to allow for universal error handling of callbacks.

This manipulates a new variable in `_callback` `GLOBAL_CALLBACK_FALLBACK` in order to pass the configuration to all other callbacks from the `Dash()` registration

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -53,6 +53,7 @@ class NoUpdate:
 GLOBAL_CALLBACK_LIST = []
 GLOBAL_CALLBACK_MAP = {}
 GLOBAL_INLINE_SCRIPTS = []
+GLOBAL_CALLBACK_FALLBACK = None
 
 
 # pylint: disable=too-many-locals
@@ -145,6 +146,7 @@ def callback(
     )
     callback_map = _kwargs.pop("callback_map", GLOBAL_CALLBACK_MAP)
     callback_list = _kwargs.pop("callback_list", GLOBAL_CALLBACK_LIST)
+    callback_fallback = _kwargs.pop("callback_fallback", GLOBAL_CALLBACK_FALLBACK)
 
     if background:
         long_spec = {
@@ -185,6 +187,7 @@ def callback(
         long=long_spec,
         manager=manager,
         running=running,
+        callback_fallback=callback_fallback
     )
 
 
@@ -291,6 +294,7 @@ def register_callback(  # pylint: disable=R0914
     long = _kwargs.get("long")
     manager = _kwargs.get("manager")
     running = _kwargs.get("running")
+    callback_fallback = _kwargs.get('callback_fallback')
     if running is not None:
         if not isinstance(running[0], (list, tuple)):
             running = [running]
@@ -316,6 +320,7 @@ def register_callback(  # pylint: disable=R0914
         dynamic_creator=allow_dynamic_callbacks,
         running=running,
     )
+
 
     # pylint: disable=too-many-locals
     def wrap_func(func):
@@ -495,7 +500,9 @@ def register_callback(  # pylint: disable=R0914
             return jsonResponse
 
         callback_map[callback_id]["callback"] = add_context
-
+        if callback_fallback:
+            callback_map[callback_id]['callback'] = callback_fallback(insert_output)(
+                callback_map[callback_id]['callback'])
         return func
 
     return wrap_func
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -16,7 +16,7 @@ import hashlib
 import base64
 import traceback
 from urllib.parse import urlparse
-from typing import Dict, Optional, Union
+from typing import Dict, Optional, Union, Callable
 
 import flask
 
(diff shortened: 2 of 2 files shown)


ndrezn:
Related: https://community.plotly.com/t/error-handling-for-callbacks-and-layouts/83586/2

ndrezn:
@BSd3v could you add an example of using this new `callback_fallback` functionality as implemented here to get a sense of use case/how to document?

BSd3v:
Here is an example app:

```
from dash import *
from dash._utils import to_json
import traceback
from dash import ctx
import json


def callback_fallback(output=None):
    error_message = "error in callback"

    def decorator(func):
        def wrapper(*args, **kwargs):
            try:
                return func(*args, **kwargs)
            except Exception as e:
                if str(e) == "":
                    raise PreventUpdate
            alertError(f"{error_message}", f"{output}\n {traceback.format_exc()}")
            resp = {
                "app_notification": {
                    "children": json.loads(
                        to_json(
                            notification(
                                "error", "there was an issue, IT has been notified", ctx
                            )
                        )
                    )
                }
            }
            return json.dumps({"multi": True, "response": resp})

        return wrapper

    return decorator

def callback_fallback2(output=None):
    error_message = "error in callback"

    def decorator(func):
        def wrapper(*args, **kwargs):
            try:
                return func(*args, **kwargs)
            except Exception as e:
                if str(e) == "":
                    raise PreventUpdate
            alertError(f"{error_message}", f"{output}\n {traceback.format_exc()}")
            resp = {
                "app_notification": {
                    "children": json.loads(
                        to_json(
                            notification(
                                "error", "I'm sorry Dave, I'm afraid I can't do that", ctx
                            )
                        )
                    )
                }
            }
            return json.dumps({"multi": True, "response": resp})

        return wrapper

    return decorator


app = Dash(
    __name__,
    callback_fallback=callback_fallback,
)

def notification(type, msg, ctx=None):
    if type == 'error':
        return html.Div(
            children=msg + (f' - {ctx.triggered[0]["value"]}' if ctx else ''),
            style={'color': 'red'}
        )
    return ''


def alertError(subject, message):
    print(subject)
    print(message)
    pass

@callback(
    Output("children", "children"),
    Input("click", "n_clicks"),
    State("testChecked", "value"),
    prevent_initial_call=True,
)
def partialFailingCall(n, c):
    if c:
        return rawr
    return f"I ran properly - {n}"

@callback(
    Output("children2", "children"),
    Input("click2", "n_clicks"),
    State("testChecked2", "value"),
    callback_fallback=callback_fallback2,
    prevent_initial_call=True,
)
def partialFailingCall(n, c):
    if not c:
        return rawr
    return f"I ran properly - {n}"


app.layout = html.Div(
                children=[html.Div(id="app_notification"),
                          html.Div([
                            html.Div("When checked, the callback will fail"),
                                  html.Button("test callback", id="click"),
                                  dcc.Checklist([True], id="testChecked"),
                                  html.Div(id="children")
                              ]
                            ),
                          html.Div([
                            html.Div("When not checked, the callback will fail, callback handler is different"),
                                  html.Button("test callback", id="click2"),
                                  dcc.Checklist([True], id="testChecked2"),
                                  html.Div(id="children2")
                              ]
                            )
                          ]
            )

app.run(debug=True)
```

Notice that in the second callback I am overriding the default callback_fallback in the callback definition.

ndrezn (commented):
I think this pattern is a bit confusing. We're passing a decorator as an argument: With this API I would expect to pass a function as an argument. I guess in general I feel like there's still some API design to be done here.

What is the ideal way to pass in an error fallback for a Dash callback? It's not immediately obvious to me, but I think we can definitely simplify the pattern from having to pass a decorator. @T4rk1n might have some other comments... but we might want to spend some time in an issue ticket discussing this feature more broadly before starting implementation?

BSd3v:
> ```python
> alertError(f"{error_message}", f"{output}\n {traceback.format_exc()}")
>             resp = {
>                 "app_notification": {
>                     "children": json.loads(
>                         to_json(
>                             notification(
>                                 "error", "I'm sorry Dave, I'm afraid I can't do that", ctx
>                             )
>                         )
>                     )
>                 }
>             }
>             return json.dumps({"multi": True, "response": resp})
> ```

So just pass something like this instead?

```
alertError(f"{error_message}", f"{output}\n {traceback.format_exc()}")
            resp = {
                "app_notification": {
                    "children": json.loads(
                        to_json(
                            notification(
                                "error", "I'm sorry Dave, I'm afraid I can't do that", ctx
                            )
                        )
                    )
                }
            }
            return json.dumps({"multi": True, "response": resp})
```

AnnMarieW:
Here's another use-case for this feature from the forum:
 https://community.plotly.com/t/errors-showing-up-when-run-locally-but-not-when-deployed-to-gcp/85311/2

BSd3v:
Closed with this PR: [#2903](https://github.com/plotly/dash/pull/2903)

PR #2831, opened 2024-04-10
[Discussional PR]Update the merge strategy for accepting additional parameter features

insistence:
This is a discussional PR. First of all, I apologize for not discovering the issue in #2819 in a timely manner. Compared to #2819, this PR only changes the strategy of merging additional parameters. Therefore, the test case of #2819 is also applicable to this PR. I initiated this PR to discuss which merger strategy is better. I adopted a more conservative merger strategy in #2819. The merging strategy in #2819 is that when additional parameters and subcomponent parameters are the same, the priority of subcomponent parameters is higher. One issue with this strategy is that it is not possible to control the parameters of the subcomponent by setting the parameters of the parent component when the same parameters occur and the subcomponent itself has values. This behavior looks a bit strange. Therefore, I adjusted the priority of subcomponent parameters in this pr, and set the priority of additional parameters higher after excluding the important parameter id. However, I am not sure if this will lead to some unknown issues. After all, the additional parameters are unknown. Which merger strategy does the official think is better?

### DIFF
--- a/dash/dash-renderer/src/TreeContainer.js
+++ b/dash/dash-renderer/src/TreeContainer.js
@@ -270,8 +270,8 @@ class BaseTreeContainer extends Component {
             _dashprivate_config
         );
         let props = mergeRight(
-            _extraProps,
-            dissoc('children', _dashprivate_layout.props)
+            dissoc('children', _dashprivate_layout.props),
+            _extraProps
         );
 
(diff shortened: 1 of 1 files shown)


insistence:
@T4rk1n Excuse me，what do you think?

T4rk1n (approved):
I think this is better too. Thank you.

:woman_dancing:

PR #2832, opened 2024-04-16
Add startup route setup

danton267:
Add the ability for packages to add routes that will be set up at the end of Dash init.

example:

```python
my-package.py

from dash import Dash

def my_route_f():
   return {} # serializable object

Dash.add_startup_route("my-package-route", my_route_f, ["POST"])
```

`_dash_startup_route/my-package-route` will be created at the `app=Dash()`, and can be then used by the package features


dash/dash.py
```diff
@@ -1626,6 +1628,38 @@ def display_content(path):
...
+            raise ValueError(f"Route name '{name}' is already in use.")
+
+        Dash.STARTUP_ROUTES.append((name, view_func, methods))
+
+    def setup_startup_routes(self):
+        """
+        Initialize the startup routes stored in STARTUP_ROUTES.
+        """
+        for _name, _view_func, _methods in self.STARTUP_ROUTES:
+            self._add_url(f"_dash_startup_route/{_name}", _view_func, _methods)
```
T4rk1n, line 1661: I think there should be a clear/reset of the STARTUP_ROUTES after this.

PR #2833, opened 2024-04-16
Allow data urls

### LINKED ISSUE #2764: Dangerous link detected error after upgrading to Dash 2.15.0
```
dash                      2.15.0
dash-bootstrap-components 1.5.0
dash-extensions           1.0.12
dash-iconify              0.1.2
dash-mantine-components   0.12.1
```

After upgrading to Dash 2.15.0, I have apps that are now breaking with `Dangerous link detected` errors, which are emitted when I use `Iframe` components to display embedded data, e.g. PDF files.  Here is a small example,

```
import base64
import requests
from dash import Dash, html

# Get a sample PDF file.
r = requests.get("https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf")
bts = r.content
# Encode PDF file as base64 string.
encoded_string = base64.b64encode(bts).decode("ascii")
src = f"data:application/pdf;base64,{encoded_string}"
# Make a small example app.
app = Dash()
app.layout = html.Iframe(id="embedded-pdf", src=src, width="100%", height="100%")

if __name__ == '__main__':
    app.run_server()
```

I would expect the app would continue to work, displaying the PDF, like it did in previous versions. 

I guess the issue is related to the fixing of XSS vulnerabilities as mentioned in #2743 . However, I am not sure why it should be considered a vulnerability to display an embedded PDF file.

T4rk1n:
- Allows data urls protocol in links.
- Add `window.dash_clientside.clean_url` function to sanitize url, used between html and dcc.
   - Disable the sanitation by overriding clean_url in an asset file: `window.dash_clientside.clean_url = (url) => url;`

Fixes #2764

### DIFF
--- a/dash/dash-renderer/src/utils/clientsideFunctions.ts
+++ b/dash/dash-renderer/src/utils/clientsideFunctions.ts
@@ -34,7 +34,8 @@ const clean_url = (url: string, fallback = 'about:blank') => {
     const cleaned = url
         .replace(newLines, '')
         .replace(ctrlChars, '')
-        .replace(htmlEntities, '');
+        .replace(htmlEntities, (_, dec) => String.fromCharCode(dec))
+        .trim();
     if (invalidProtocols.test(cleaned)) {
         return fallback;
     }
--- a/dash/dash-renderer/src/utils/clientsideFunctions.ts
+++ b/dash/dash-renderer/src/utils/clientsideFunctions.ts
@@ -18,6 +18,30 @@ const set_props = (id: string | object, props: {[k: string]: any}) => {
     }
 };
 
+// Clean url code adapted from https://github.com/braintree/sanitize-url/blob/main/src/constants.ts
+// to allow for data protocol.
+const invalidProtocols = /^([^\w]*)(javascript|vbscript)/im;
+const newLines = /&(tab|newline);/gi;
+
+// eslint-disable-next-line no-control-regex
+const ctrlChars = /[\u0000-\u001F\u007F-\u009F\u2000-\u200D\uFEFF]/gim;
+const htmlEntities = /&#(\w+)(^\w|;)?/g;
+
+const clean_url = (url: string, fallback = 'about:blank') => {
+    if (url === '') {
+        return url;
+    }
+    const cleaned = url
+        .replace(newLines, '')
+        .replace(ctrlChars, '')
+        .replace(htmlEntities, '');
+    if (invalidProtocols.test(cleaned)) {
+        return fallback;
+    }
+    return url;
+};
+
 const dc = ((window as any).dash_clientside =
     (window as any).dash_clientside || {});
 dc['set_props'] = set_props;
+dc['clean_url'] = dc['clean_url'] === undefined ? clean_url : dc['clean_url'];
--- a/tests/integration/security/test_xss.py
+++ b/tests/integration/security/test_xss.py
@@ -9,7 +9,19 @@ def test_xss001_banned_protocols(dash_duo):
             dcc.Link("dcc-link", href="javascript:alert(1)", id="dcc-link"),
             html.Br(),
             html.A(
-                "html.A", href='javascr\nipt:alert(1);console.log("xss");', id="html-A"
+                "html.A",
+                href='javascr\n\nipt:alert(1);console.log("xss");',
+                id="html-A",
+            ),
+            html.A(
+                "html.A.escape",
+                id="html-A-escape",
+                href="""javascript\x09\x0a:alert(1)""",
+            ),
+            html.A(
+                "html.A.encoded",
+                id="html-A-encoded",
+                href="&#0000106&#0000097&#0000118&#0000097&#0000115&#0000099&#0000114&#0000105&#0000112&#0000116&#0000058&#0000097&#0000108&#0000101&#0000114&#0000116&#0000040&#0000039&#0000088&#0000083&#0000083&#0000039&#0000041",
             ),
             html.Br(),
             html.Form(
@@ -35,6 +47,8 @@ def test_xss001_banned_protocols(dash_duo):
     for element_id, prop in (
         ("#dcc-link", "href"),
         ("#html-A", "href"),
+        ("#html-A-escape", "href"),
+        ("#html-A-encoded", "href"),
         ("#iframe-src", "src"),
         ("#object-data", "data"),
         ("#embed-src", "src"),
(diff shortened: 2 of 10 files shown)


dash/dash-renderer/src/utils/clientsideFunctions.ts
```diff
@@ -18,6 +18,30 @@ const set_props = (id: string | object, props: {[k: string]: any}) => {
...
+
+// eslint-disable-next-line no-control-regex
+const ctrlChars = /[\u0000-\u001F\u007F-\u009F\u2000-\u200D\uFEFF]/gim;
+const htmlEntities = /&#(\w+)(^\w|;)?/g;
+
+const clean_url = (url: string, fallback = 'about:blank') => {
+    if (url === '') {
+        return url;
+    }
+    const cleaned = url
```
emilykl, line 34: Do we need to include the `decodeHtmlCharacters` step as in `braintree/sanitize-url`?

tests/integration/security/test_xss.py
```diff
@@ -58,3 +59,21 @@ def test_xss002_blank_href(dash_duo):
...
...kC7eZdzZNYIRlPGoMWUOCO/ao99M81CxKTO2jMG6eAwixIbOZgoWppCiY/EUfMec+knC+QlZQPqoZkGcjz4UkT1Ph0AYUQ0zbXxU3t0vfkDzr0FOjJmM93uu84a8s7dhxI9lPO2kieinXMyLtjMkvlP94h8SADyydmRdnC9IYsnFQDscHJJ0vpKbCEE/58sDwFIxhmJnVpplHeIxUKJOxjWH16gt+3lJ5PY6NGIPn+RRR6LIegHm5q7jzz51laClxV7mW/g4JhWsP4JB46FKvNTuPVHAwyGAwnmHY7ne6UJlg6/O/YRqxv96O4k3s9hZnzJGdBdoGrEusVwtZIS2GTVb0thWRSwxj77EM+w6kUcvYOwhjl+Uj87BOyZzvJk4PTgnykvD757rpEUyuiwMvhll5Nbp5n624KRE+t0+YvIurDFlHwJKegalVPwQluyFOdaTAoabFdISzR1OZs5yLR7UCZqSpS5RR8h59aXEel0Bi25ZsCdT6ftLmMfH3Odl6X5L/UxE2tggLirJUMhA0x7pAtxraYxh7na8i8pUiX5b7ZZ7sAo00eSpynpvgdPVKxLm5qBRSszjjws3scP+jL9oxTu/lyZko4a9LkYgs1hwU5pR7C2ZPz+WqEDWX62qT9ANXDTW1UQAElDxqOTWxjczzUhA8mvm1PVTOH3Q0H2zZ2NroLbNeZ9dMnc1afiR+TLx/RFaTaNG1Bq6cwN2I4GAyKsqQHRe5jcs/NnmUoPTjkfr/Z2Rgjm76a06/NUKiKQiXymha3JtUljpfqHlEIhZ/HBfWZj+4VhRPhcKcmsYtMq6NAf0u53PaXEEfBr/gXnbJWRyE7KvArDBrQH8X57dOaxI+oVkehMTkyfdyv8PdjJ23km8i2Z1X/wFzRCik0BtGw2t1q6A2jSV1iR7xWSQFUJKBQBwGFOggo1EFAoQ4CCnXQEincmNIKKpI+vWXEKCBtVc7tP2VV7dW1eu3EzhLb3I5yQg8Pff4vOjF7BzdWBZXThBytm0KAhyQpCGjY34syBXvTBGj4M8q3UdRMs4XxsRBxHljp4azfFiMOw3Za1W4JVXu+Oa9aryjl/51ONuzmsfzjLjfqN/2Rkn27YbeVqmHUP/JnMBJAIBAIBAKBQCAQCAQCgUAgEAgEAoFAIBAIBAKBQLfV/wD2/u2SfXLkQQAAAABJRU5ErkJggg==",
```
emilykl, line 71: IMO it would be more readable to commit the image file and then read and base64 encode it inside the test

tests/integration/security/test_xss.py
```diff
@@ -42,9 +42,10 @@ def test_xss001_banned_protocols(dash_duo):
     ):
 
         element = dash_duo.find_element(element_id)
+        prop_value = element.get_attribute(prop)
         assert (
-            element.get_attribute(prop) == "about:blank"
-        ), f"Failed prop: {element_id}.{prop}"
+            prop_value == "about:blank"
```
emilykl, line 47: If we are inlining the sanitization logic to our codebase, probably makes sense to add a few more tests along the lines of [these](https://github.com/braintree/sanitize-url/blob/0d5140d7f7489f46472a545635be99f8282f83e4/src/__tests__/index.test.ts#L102)

T4rk1n: I added more tests cases, some of the sanitization in that was done in the lib were not valid as dash prop.

emilykl:
LGTM at a high level. The workflow of overriding `window.dash_clientside.clean_url` to disable the sanitization feels a little weird... couldn't we just add another prop to the affected components to disable sanitization? How many places would we want to allow users to disable?

T4rk1n:
> LGTM at a high level. The workflow of overriding `window.dash_clientside.clean_url` to disable the sanitization feels a little weird... couldn't we just add another prop to the affected components to disable sanitization? How many places would we want to allow users to disable?

I just thought it would be convenient to have it available for multiple components library (dcc & html in this case), if a community library want to use the same clean url pattern for it's url it's now available. Overriding the function to disable it is an afterthought pattern that emerged from that, but I think it does well as a global disable feature that would be something only a user that really want to do that.

ndrezn:
@emilykl re: prop on component level, see: https://github.com/plotly/dash/issues/2764#issuecomment-2023154681. We don't want to allow users to disable sanitization at the component level.

emilykl:
Understood re: why we don't want to disable sanitization at the component level. 👍

emilykl (approved):
🚀

ndrezn:
One thing I would like to validate before merging. We know this pattern is considered sufficient by Mozilla, but is there any documentation for other browsers (Safari, Chrome) that is in agreement? I started looking but did not find anything. If it exists I imagine it's buried in dev docs somewhere.

T4rk1n:
> One thing I would like to validate before merging. We know this pattern is considered sufficient by Mozilla, but is there any documentation for other browsers (Safari, Chrome) that is in agreement? I started looking but did not find anything. If it exists I imagine it's buried in dev docs somewhere.

I found this for Safari: https://developer.apple.com/forums/thread/650034?answerId=624675022#624675022
Seems like it was the last one to allow it and no longer does as from 4 years ago.

ndrezn (approved):
💃

PR #2841, opened 2024-04-27
fix some typing issues

DanielIskandar:

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals


dash/dash.py
```diff
@@ -375,40 +378,43 @@ class Dash:
...
+        extra_hot_reload_paths: Optional[List[str]] = None,
+        plugins: Optional[List[Dict[str, Any]]] = None,
+        title: str = "Dash",
+        update_title: str = "Updating...",
+        long_callback_manager: Optional[
+            Any
+        ] = None,  # Type should be specified if possible
+        background_callback_manager: Optional[
+            Any
+        ] = None,  # Type should be specified if possible
```
T4rk1n, line 413: Both of those should be of type `dash.long_callback.managers.BaseLongCallbackManager`

dash/dash.py
```diff
@@ -375,40 +378,43 @@ class Dash:
...
+        include_pages_meta: bool = True,
+        url_base_pathname: Optional[str] = None,
+        requests_pathname_prefix: Optional[str] = None,
+        routes_pathname_prefix: Optional[str] = None,
+        serve_locally: bool = True,
+        compress: Optional[bool] = None,
+        meta_tags: Optional[List[Dict[str, Any]]] = None,
+        index_string: str = _default_index,
+        external_scripts: Optional[List[Union[str, Dict[str, Any]]]] = None,
+        external_stylesheets: Optional[List[Union[str, Dict[str, Any]]]] = None,
```
T4rk1n, line 400: Those `Any` in the `Dict` would be `str`

dash/dash.py
```diff
@@ -375,40 +378,43 @@ class Dash:
     STARTUP_ROUTES: list = []
```
T4rk1n, line 379: we can add `server: flask.Flask` here.

T4rk1n (commented):
This look good overall and a good addition, just need to address my previous comments and add a changelog entry.

gvwilson:
@DanielIskandar are you interested in updating this one so that we can merge it? thanks - @gvwilson

T4rk1n (approved):

PR #2854, opened 2024-05-09
Fix dropdown

### LINKED ISSUE #2850: Dropdown changes in Dash 2.17 causing page loading issues in Dash docs
Our docs tests test different paths:

https://github.com/plotly/ddk-dash-docs/blob/main/tests/integration/test_bad_paths.py

With Dash 2.17, some of these tests fail because the page never loads

The page ends up stuck in a state like this

image

Seems to be happening on page where the dropdown is updated based on the URL and vice versa. Seems to work without this change: https://github.com/plotly/dash/pull/2816

To recreate:

Run the /ddk-dash-docs tests using Dash 2.17

T4rk1n:
Fix dropdown with a value of `""` resetting to None. Fix #2850

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## UNRELEASED
+
+## Fixed
+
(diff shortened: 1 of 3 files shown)


emilykl:
Wow, great detective work @T4rk1n .

Could you explain the issue and the fix in a bit more detail?

T4rk1n:
We discovered the value was getting reset to null and triggering callbacks that ends up breaking the docs in some odd testing case. The empty string should have been considered the same as null and not trigger the reset.

emilykl (approved):
Looks good!

PR #2856, opened 2024-05-10
Fix/multi set props

### LINKED ISSUE #2852: [BUG] set_props called multiple times only keep the last props.
For regular callbacks, when multiple call of `set_props` to the same component id, only the last call is saved.

Example:
```
from dash import Dash, Input, html, set_props

app = Dash()

app.layout = [
    html.Button("start", id="start"),
    html.Div("initial", id="output"),
]

@app.callback(
    Input("start", "n_clicks"),
)
def on_click(_):
    set_props("output", {"children": "changed"})
    set_props("output", {"style": {"background": "red"}})

if __name__ == "__main__":
    app.run(debug=True)

```

Clicking on the start button only set the background red, the text stays at "initial". The props should be merged and both updated.

T4rk1n:
Fix #2852

### DIFF
--- a/dash/_callback_context.py
+++ b/dash/_callback_context.py
@@ -252,7 +252,11 @@ class CallbackContext:
     def set_props(self, component_id: typing.Union[str, dict], props: dict):
         ctx_value = _get_context_value()
         _id = stringify_id(component_id)
-        ctx_value.updated_props[_id] = props
+        updates = ctx_value.updated_props.get(_id)
+        if updates is not None:
+            ctx_value.updated_props[_id] = {**updates, **props}
+        else:
+            ctx_value.updated_props[_id] = props
 
 
 callback_context = CallbackContext()
--- a/dash/_callback_context.py
+++ b/dash/_callback_context.py
@@ -252,9 +252,9 @@ class CallbackContext:
     def set_props(self, component_id: typing.Union[str, dict], props: dict):
         ctx_value = _get_context_value()
         _id = stringify_id(component_id)
-        updates = ctx_value.updated_props.get(_id)
-        if updates is not None:
-            ctx_value.updated_props[_id] = {**updates, **props}
+        existing = ctx_value.updated_props.get(_id)
+        if existing is not None:
+            ctx_value.updated_props[_id] = {**existing, **props}
         else:
             ctx_value.updated_props[_id] = props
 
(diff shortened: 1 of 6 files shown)


dash/_callback_context.py
```diff
@@ -252,7 +252,11 @@ def timing_information(self):
     def set_props(self, component_id: typing.Union[str, dict], props: dict):
         ctx_value = _get_context_value()
         _id = stringify_id(component_id)
-        ctx_value.updated_props[_id] = props
+        updates = ctx_value.updated_props.get(_id)
+        if updates is not None:
+            ctx_value.updated_props[_id] = {**updates, **props}
```
gvwilson, line 257: why `{**updates, **props}` rather than `{**props, **updates}`?
```
props = {"left": 1, "right": 2}
updates = {"left": 3}

# updates then props doesn't update
as_written = {**updates, **props}
assert as_written == {"left": 1, "right": 2}

# props then updates does the update
inverting_order = {**props, **updates}
assert inverting_order == {"left": 3, "right": 2}
```

dash/long_callback/_proxy_set_props.py
```diff
@@ -7,6 +7,12 @@ class ProxySetProps(dict):
     def __init__(self, on_change):
         super().__init__()
         self.on_change = on_change
+        self._data = {}
 
     def __setitem__(self, key, value):
         self.on_change(key, value)
```
gvwilson, line 13: why is the `on_change` call no longer needed?

tests/integration/callbacks/test_arbitrary_callbacks.py
```diff
@@ -165,3 +165,26 @@ def on_click(clicked):
...
+        html.Button("start", id="start"),
+        html.Div("initial", id="output"),
+    ]
+
+    @app.callback(
+        Input("start", "n_clicks"),
+    )
+    def on_click(_):
+        set_props("output", {"children": "changed"})
+        set_props("output", {"style": {"background": "rgb(255,0,0)"}})
```
gvwilson, line 183: this example doesn't overwrite an existing property, so if `{**updates, **props}` is actually a bug, this test won't trigger it.

gvwilson (changes requested):
I have a question about `{**updates, **props}` - if I'm misunderstanding what it's doing, please let me know.

dash/_callback_context.py
T4rk1n: I changed the variable `updates` to `existing` for less confusion, we want to merge the latest call onto the existing props.

gvwilson: thank you - much appreciated.

gvwilson (approved):

PR #2867, opened 2024-05-23
Fix clientside no outputs

### LINKED ISSUE #2866: [BUG] No output callbacks with clientside_callback
If a clientside callback has no output, it generate `ID not found in layout` error:
```
clientside_callback(
    """
    function(_) {
        window.location.reload();
    }
    """,
    Input("reset-button", "n_clicks"),
    prevent_initial_call=True,
)
```

T4rk1n:
Fix #2866

### DIFF
--- a/tests/integration/callbacks/test_arbitrary_callbacks.py
+++ b/tests/integration/callbacks/test_arbitrary_callbacks.py
@@ -1,7 +1,15 @@
 import time
 from multiprocessing import Value
 
-from dash import Dash, Input, Output, html, set_props, register_page
+from dash import (
+    Dash,
+    Input,
+    Output,
+    html,
+    set_props,
+    register_page,
+    clientside_callback,
+)
 
 
 def test_arb001_global_set_props(dash_duo):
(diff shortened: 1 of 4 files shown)


tests/integration/callbacks/test_arbitrary_callbacks.py
```diff
@@ -188,3 +196,39 @@ def on_click(_):
     dash_duo.wait_for_style_to_equal(
         "#output", "background-color", "rgba(255, 0, 0, 1)"
     )
+
+
+def test_arb007_clientside_no_output(dash_duo):
```
emilykl, line 201: @T4rk1n just a sanity check here since I'm not deeply familiar with the Dash tests -- this test fails without your fix, right?

emilykl (approved):
LGTM!

tests/integration/callbacks/test_arbitrary_callbacks.py
T4rk1n: Yes it generates an error and the clientside function is not called.

PR #2876, opened 2024-06-05
Fix pattern matching in side updates (running/progress/set_props)

### LINKED ISSUE #2863: [BUG] running does not support wildcards in ids
**Describe your context**

```
dash                                     2.17.0
dash-core-components                     2.0.0
dash-html-components                     2.0.0
dash-table                               5.0.0
dash-bootstrap-components                1.6.0

```

**Describe the bug**

Using wildcards in ids for Outputs in the running kwarg does leads to `itempath is undefined` error.

**Expected behavior**

Wildcards are properly resolved and matching components are updated while the callback is running.

**MVE**

```
from time import sleep

from dash import MATCH, Input, Output, callback

layout = [
    dbc.Button(
        "Test1",
        id={"component": "button", "index": "1"},
    ),
    dbc.Button(
        "Test2",
        id={"component": "button", "index": "2"},
    ),
]


@callback(
    Output({"component": "button", "index": MATCH}, "color"),
    Input({"component": "button", "index": MATCH}, "n_clicks"),
    running=[
        (Output({"component": "button", "index": MATCH}, "children"), "running", "finished"),
    ],
    prevent_initial_call=True,
)
def test(_) -> str:
    sleep(3)
    return "warning"
```

T4rk1n:
Fix #2863

Fixes for cases where the given id has a `MATCH`, `ALL`, `ALLSMALLER` wildcard in callback `running` argument, progress for background callbacks and `set_props`.

### DIFF
--- a/dash/dash-renderer/src/actions/callbacks.ts
+++ b/dash/dash-renderer/src/actions/callbacks.ts
@@ -1,7 +1,5 @@
 import {
     concat,
-    dissoc,
-    equals,
     flatten,
     intersection,
     keys,
@@ -33,7 +31,8 @@ import {
     IPrioritizedCallback,
     LongCallbackInfo,
     CallbackResponse,
-    CallbackResponseData
+    CallbackResponseData,
+    SideUpdateOutput
 } from '../types/callbacks';
 import {isMultiValued, stringifyId, isMultiOutputProp} from './dependencies';
 import {urlBase} from './utils';
@@ -46,6 +45,8 @@ import {handlePatch, isPatch} from './patch';
 import {getPath} from './paths';
 
 import {requestDependencies} from './requestDependencies';
+import {parsePMCId} from './patternMatching';
+import {replacePMC} from './patternMatching';
 
 export const addBlockedCallbacks = createAction<IBlockedCallback[]>(
     CallbackActionType.AddBlocked
@@ -342,19 +343,14 @@ function updateComponent(component_id: any, props: any) {
     };
 }
 
-function parsePMCId(id: string) {
-    let componentId, propName;
-    const index = id.lastIndexOf('}');
-    if (index + 2 < id.length) {
-        propName = id.substring(index + 2);
-        componentId = JSON.parse(id.substring(0, index + 1));
-    } else {
-        componentId = JSON.parse(id);
-    }
-    return [componentId, propName];
-}
-
-function sideUpdate(outputs: any, cb: ICallbackPayload) {
+/**
+ * Update a component props with `running`/`progress`/`set_props` calls.
+ *
+ * @param outputs Props to update.
+ * @param cb The originating callback info.
+ * @returns
+ */
+function sideUpdate(outputs: SideUpdateOutput, cb: ICallbackPayload) {
     return function (dispatch: any, getState: any) {
         toPairs(outputs)
             .reduce((acc, [id, value], i) => {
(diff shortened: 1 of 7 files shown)


gvwilson (approved):
thank you

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -340,33 +342,108 @@ function updateComponent(component_id: any, props: any) {
     };
 }
 
-function sideUpdate(outputs: any, dispatch: any) {
-    toPairs(outputs).forEach(([id, value]) => {
-        let componentId = id,
-            propName;
+function parsePMCId(id: string) {
```
emilykl, line 345: There's a lot of magic numbers in this function (`+2`, `+1`) -- could you add a comment explaining some of the string manipulations here?

T4rk1n: It's just the indexes of the strings, I think it's pretty readable if you know the format:
`{"key":"value"}.propName` or just the dictionary, so index refers to position of `}`, if the `index + 2  < id.length` means there is more data after the index and `.`, the propname starts at index + 2 and componentId is 0 to index + 1.

emilykl: got it, that makes sense, thanks! Could you add that format as a comment in the code?

alexcjohnson: This looks duplicative of:

https://github.com/plotly/dash/blob/09252f8d2f690480cc468b2e015f9e2417dc90ad/dash/dash-renderer/src/actions/dependencies.js#L93

tests/integration/callbacks/test_wildcards.py
```diff
@@ -552,3 +553,65 @@ def update_selected_values(values):
...
+                Output({"component": "button", "index": MATCH}, "children"),
+                "running",
+                "finished",
+            ),
+            (Output({"component": "button", "index": ALL}, "disabled"), True, False),
+        ],
+        prevent_initial_call=True,
+    )
+    def on_click(_) -> str:
+        time.sleep(1)
```
alexcjohnson, line 593: Can we do this with a lock instead of a delay?

dash/dash-renderer/src/actions/callbacks.ts
T4rk1n: This handle case where the prop name is not attached to the key, which comes with `set_props` so it searches for `}` instead of a dot.

T4rk1n:
I notice the running is not available/working with clientside callbacks, it could be working with a bit of refactor and I think could be useful when the clientside callback returns a promise.

dash/dash-renderer/src/actions/callbacks.ts
alexcjohnson: OK - can we at least put these in the same location - maybe move them both to a new file so it's easier to find them and we avoid making any more of them?

My main concern with doing this in slightly different ways is accidentally introducing additional constraints on what constitutes a valid ID, and/or causing an error with certain IDs. I think your logic here is OK, as you only get to searching for `}` after you already know the ID starts with `{` which is an [invalid character](https://github.com/plotly/dash/blob/b1c6efb08d5f80e873d95f489e0c8d65335aed84/dash/_validate.py#L107) for string IDs - but that's something we need to check carefully every time we modify the ID handling code.

ndrezn (approved):
💃 🚀

gvwilson:
💯

dash/dash-renderer/src/actions/callbacks.ts
alexcjohnson: Thanks for putting the new function in a new file. The existing function is still in its old location though, so at some point please let's bring that one into the new file too.

Issue #2878, opened 2024-06-07
[BUG] `id` passed through `dcc.Loading` not visible in DOM


huong-li-nguyen:
**Describe your context**
Hello guys 👋 

I am currently trying to pass an `id` to the dcc.Loading component or its parent container and I would like the `id` to be visible in the DOM such that I can target the CSS of the components inside the `dcc.Loading` via ID.

Please provide us your environment, so we can easily reproduce the issue.

-  replace the result of `pip list | grep dash` below
```
dash                      2.17.0
dash-bootstrap-components 1.5.0
dash-core-components      2.0.0
dash-html-components      2.0.0
```
-  if frontend related, tell us your Browser, Version and OS

    - Browser: Chrome

**Describe the bug**
Let's take the example app below - what I would have expected is that there would be an html div visible with a className="loading" and an id="loading-id". However, if I provide the `className="loading"` I see a div but it does not have the className="loading" in the DOM nor does it have the id="loading-id" in the DOM. 

When I switch this to `parent_className="loading"`, now I see a div with the className="loading", but I cannot attach an id to this parent container.

I am not a react expert, but from the source I can see that the `id` doesn't seem to be passed on in the return of the react component and is therefore not visible in the DOM? Is there any reason for that?

https://github.com/plotly/dash/blob/09252f8d2f690480cc468b2e015f9e2417dc90ad/components/dash-core-components/src/components/Loading.react.js#L128-L133

```
from dash import Dash, html, dcc, callback, Output, Input
import plotly.express as px
import pandas as pd

df = pd.read_csv('https://raw.githubusercontent.com/plotly/datasets/master/gapminder_unfiltered.csv')

app = Dash()

app.layout = [
    html.H1(children='Title of Dash App', style={'textAlign':'center'}),
    dcc.Dropdown(df.country.unique(), 'Canada', id='dropdown-selection'),
    dcc.Loading(dcc.Graph(id='graph-content'), color='grey', id="loading-id", parent_className="loading")
]

@callback(
    Output('graph-content', 'figure'),
    Input('dropdown-selection', 'value')
)
def update_graph(value):
    dff = df[df.country==value]
    return px.line(dff, x='year', y='pop')

if __name__ == "__main__":
    app.run(debug=True)
```

**Expected behavior**

I would expect the `id` being passed on to the react component and visible in the DOM, so having a <div class="loading" id="loading-id" </div> visible in the DOM.

**Screenshots**
[...]

PR #2888, opened 2024-06-15
Adds id to dcc.Loading DOM

leeagustin:
Hi!👋 This is my first time contributing to an open-source project. I hope it's okay if I give the issue (#2878) a try.

I didn't add a test as I was unsure if it was needed for this (small) change; however, I'd be more than happy to add one if needed!

## Contributor Checklist

   -  [x] Adds id to dcc.Loading DOM

### optionals

### DIFF
--- a/components/dash-core-components/src/components/Loading.react.js
+++ b/components/dash-core-components/src/components/Loading.react.js
@@ -145,11 +145,10 @@ const Loading = ({
             >
                 {children}
             </div>
-            <div style={showSpinner ? coveringSpinner : {}}>
+            <div id={id} style={showSpinner ? coveringSpinner : {}}>
                 {showSpinner &&
                     (custom_spinner || (
                         <Spinner
-                            id={id}
                             className={className}
                             style={style}
                             status={loading_state}
--- a/components/dash-core-components/src/components/Loading.react.js
+++ b/components/dash-core-components/src/components/Loading.react.js
@@ -24,6 +24,7 @@ const Loading = ({
     loading_state,
     display,
     color,
+    id,
     className,
     style,
     parent_className,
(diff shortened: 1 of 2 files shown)


huong-li-nguyen:
Congrats on your first PR in an open-source project! 🎊 

I forgot about raising the PR 😬 but happy that it lets you create your first PR! 👏

components/dash-core-components/src/components/Loading.react.js
```diff
@@ -148,6 +149,7 @@ const Loading = ({
                 {showSpinner &&
                     (custom_spinner || (
                         <Spinner
+                            id={id}
```
T4rk1n, line 152: The id should be on the outer div,  the spinners https://github.com/plotly/dash/blob/4692acf61f3e282950a8d1014b152176addb2aec/components/dash-core-components/src/fragments/Loading/spinners/ don't have the id prop so it wouldn't render on the DOM.

T4rk1n (commented):
I don't think we need to add new tests, but the existing ones could be modified to include an id here: https://github.com/plotly/dash/blob/ba27733e6ecedbaf600e6d91051234ecb00e1234/components/dash-core-components/tests/integration/loading/test_loading_component.py#L23

components/dash-core-components/src/components/Loading.react.js
leeagustin: Ah I see, good catch 😄 

Changed in 9dcf243

leeagustin:
> I don't think we need to add new tests, but the existing ones could be modified to include an id here:
> 
> https://github.com/plotly/dash/blob/ba27733e6ecedbaf600e6d91051234ecb00e1234/components/dash-core-components/tests/integration/loading/test_loading_component.py#L23

Sounds good!

Changed in ca5d66a

T4rk1n (approved):
Looks good, thank you :woman_dancing:

PR #2898, opened 2024-06-21
Allowing defining potentially non existent components in the running keyword

### LINKED ISSUE #2897: [BUG] Callback running keyword doesn't work when component is not in layout
When using the running keyword in a regular callback, the renderer throws an error if the component defined in running does not exist. In my case that happened, because I am targeting a button that is not always part of the layout. The callback that is triggered by clicking the button can also be triggered by a silent background process when the button is not in the layout. 

The error that is thrown is: 
```
TypeError: Cannot read properties of undefined (reading 'concat')

    at getInputHistoryState (http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:9446:78)

    at http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:9475:26

    at http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:9519:12

    at dispatch (http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:44147:22)

    at http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:43826:16

    at dispatch (http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:44550:28)

    at http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:689:5

    at http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:43822:18

    at dispatch (http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:44550:28)

    at http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_17_1m1718952684.dev.js:744:7
```

Here is a small example to reproduce the bug. Click the button for the error to appear. Note the component id used in running, it refers to a non existing component:
```python
from dash import Dash, html, Input, Output, callback, set_props
import time

app = Dash(__name__)


app.layout = html.Div(
    [
        html.H1("Updating Component While Callback is Running"),
        html.Div(
            "Data saved and downloaded successfully!",
            id="alert",
            style={"display": "none"},
        ),
        html.Button("Save as CSV", id="save-as-csv"),
    ]
)


@callback(
    Input("save-as-csv", "n_clicks"),
    running=[(Output("non_existing_component", "disabled"), True, False)],
)
def update_output_div(_):
    print("Saving data...")
    time.sleep(3)
    set_props("alert", {"style": {"display": "block"}})


if __name__ == "__main__":
    app.run(debug=True)

```

A solution for the bug is to add a check to see if the `componentPath` that was fetched is `"undefined"` in the function `updateComponent`.

```javascript
function updateComponent(component_id, props) {
  return function (dispatch, getState) {
    var paths = getState().paths;
    var componentPath = (0,_paths__WEBPACK_IMPORTED_MODULE_7__.getPath)(paths, component_id);
    if (typeof component_id === 'undefined') {
      // Can't find the component that was defined in the running keyword,
      // Let's skip the component to prevent the dashboard from crashing.
      return;
    }
    ... rest of code ...
```

The entry point for this function is `handleServerside (with if (running) == True -> sideUpdate -> updateComponent`
As far as I understood the code, `updateComponent` is the first point in this flow where it is actually checked whether the component exists.

In the proposed solution the component update is simply skipped if the path is undefined.

I will create a PR to propose this solution. One question I have here is whether the updateComponent function is the proper place to add the check.

aGitForEveryone:
closes #2897

When using the running keyword in (regular) callbacks, the app crashes if a component is used that does not exist in the layout. Added a check to only continue with component update it the component was found in the dashboard. Full explanation in #2897. 

## Contributor Checklist

   -  [x] Added check to only allow component updates in running if component is defined
   -  [x] Added test to verify correct behavior
- [ ] I have run the tests locally and they passed. I can't run the full test suite locally. My PC can't handle it. 

### optionals

### DIFF
--- a/dash/dash-renderer/src/actions/callbacks.ts
+++ b/dash/dash-renderer/src/actions/callbacks.ts
@@ -334,6 +334,11 @@ function updateComponent(component_id: any, props: any) {
     return function (dispatch: any, getState: any) {
         const paths = getState().paths;
         const componentPath = getPath(paths, component_id);
+        if (typeof componentPath === 'undefined') {
+            // Can't find the component that was defined in the running keyword,
+            // Let's skip the component to prevent the dashboard from crashing.
+            return;
+        }
         dispatch(
             updateProps({
                 props,
--- a/tests/integration/callbacks/test_basic_callback.py
+++ b/tests/integration/callbacks/test_basic_callback.py
@@ -823,3 +823,38 @@ def test_cbsc019_callback_running(dash_duo):
 
     dash_duo.wait_for_text_to_equal("#output", "done")
     dash_duo.wait_for_text_to_equal("#running", "off")
+
+
+def test_cbsc020_callback_running_non_existing_component(dash_duo):
+    lock = Lock()
+    app = Dash(__name__)
+
+    app.layout = html.Div(
+        [
+            html.Button("start", id="start"),
+            html.Div(id="output"),
+        ]
+    )
+
+    @app.callback(
+        Output("output", "children"),
+        Input("start", "n_clicks"),
+        running=[
+            [
+                Output("non_existent_component", "children"),
+                html.B("on", id="content"),
+                "off",
+            ]
+        ],
+        prevent_initial_call=True,
+    )
+    def on_click(_):
+        with lock:
+            pass
+        return "done"
+
+    dash_duo.start_server(app)
+    with lock:
+        dash_duo.find_element("#start").click()
+
+    dash_duo.wait_for_text_to_equal("#output", "done")


dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -334,6 +334,11 @@ function updateComponent(component_id: any, props: any) {
     return function (dispatch: any, getState: any) {
         const paths = getState().paths;
         const componentPath = getPath(paths, component_id);
+        if (typeof componentPath === 'undefined') {
+            // Can't find the component that was defined in the running keyword,
+            // Let's skip the component to prevent the dashboard from crashing.
```
T4rk1n, line 339: I think it would be better to throw an error unless `config.suppress_callback_exceptions=True`. It is in state.config, could refactor line 335 to `const {paths,config} = getState();`

aGitForEveryone: I added a check for that and threw an error when suppress_callback_exceptions=False

T4rk1n (approved):
:woman_dancing: Looks good

Issue #2894, opened 2024-06-18
Remove LogoutButton


AnnMarieW:
Since dcc.LogoutButton is no longer recommended, it should be removed.

The docs page will be redirected to html.Button.
https://github.com/plotly/ddk-dash-docs/issues/2603

This is a great "first issue" for anyone looking to contribute, and I'm happy to help you get started! If no one takes it on within about a week, I'll handle it as part of the docs PR I'm working on,

shawon-majid:
Hi, I'm new open source, sorry if this is silly question, But the given link is not working.

AnnMarieW:
Hi @shawon-majid and welcome!
That link is to the dash docs library, which is private.  The relevant content  is only about how the LogoutButton should be removed, which lead to opening this issue.

Is this something you are interested in working on?

AnnMarieW:
Note - This might be considered a breaking change - may need to deprecate this component, then remove it in dash 3.0

T4rk1n:
> Note - This might be considered a breaking change - may need to deprecate this component, then remove it in dash 3.0

Not sure we have an easy way of deprecating components, maybe we can check the class attribute (`_namespace`, `_type`) is not in a list and emit a deprecation warning in here: https://github.com/plotly/dash/blob/685dcbcb8ddb19727df87e42e655c8e33f84648f/dash/_validate.py#L412

shreya1728:
Hi, I've worked on a dash app in my internship so I'm quite familiar with it. I'm new to open-source and would love to contribute to this issue. Could you please assign it to me?

T4rk1n:
Hi @shreya1728 there is already a PR for this issue, #2899

Gannu456:
@AnnMarieW I would love to contribute to other issues....could you please give me a chance?

andre996:
The logout button has already been removed in dash-3.0 branch.  https://github.com/plotly/dash/pull/3137#issuecomment-2631546974

PR #2899, opened 2024-06-22
Deprecate LogoutButton

leeagustin:
PR will output a `DeprecationWarning` when the `LogoutButton` is used (#2894).

I tried making the changes from https://github.com/plotly/dash/issues/2894#issuecomment-2176661569, but it seems that it only validates the initial page layout and not the components used in callbacks.

As seen in the modified test in this PR, the test doesn't output a `DeprecationWarning` and fails when the LogoutButton is in the callback but not in the initial page layout. On the other hand, it passes when `LogoutButton` is in the initial page layout.

I'm assuming we would also want to show a `DeprecationWarning` if `LogoutButton` is used in a callback. What would be a good way to approach this? Or would it be simpler to remove the `LogoutButton` component entirely?

## Contributor Checklist

   -  [x] Outputs DeprecationWarning if LogoutButton is used in layout
   -  [ ] Outputs DeprecationWarning if LogoutButton is used in callback

### optionals

### DIFF
--- a/components/dash-core-components/tests/integration/location/test_location_logout.py
+++ b/components/dash-core-components/tests/integration/location/test_location_logout.py
@@ -5,7 +5,8 @@ import pytest
 import time
 
 
-def test_llgo001_location_logout(dash_dcc):
+@pytest.mark.parametrize("add_initial_logout_button", [False, True])
+def test_llgo001_location_logout(dash_dcc, add_initial_logout_button):
     app = Dash(__name__)
 
     @app.server.route("/_logout", methods=["POST"])
@@ -14,9 +15,14 @@ def test_llgo001_location_logout(dash_dcc):
         rep.set_cookie("logout-cookie", "", 0)
         return rep
 
-    app.layout = html.Div(
-        [html.H2("Logout test"), dcc.Location(id="location"), html.Div(id="content")]
-    )
+    layout_children = [
+        html.H2("Logout test"),
+        dcc.Location(id="location"),
+        html.Div(id="content"),
+    ]
+    if add_initial_logout_button:
+        layout_children.append(dcc.LogoutButton())
+    app.layout = html.Div(layout_children)
 
     @app.callback(Output("content", "children"), [Input("location", "pathname")])
     def on_location(location_path):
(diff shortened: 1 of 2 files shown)


dash/_validate.py
```diff
@@ -422,6 +423,15 @@ def validate_layout(layout, layout_value):
     component_ids = set()
 
     def _validate(value):
+        def _validate_type(comp):
+            deprecated_types = ["LogoutButton"]
```
T4rk1n, line 427: Just to make sure there is no collision, we should also check the `_namespace` attribute. Also could add a custom deprecation message to the component.

I would refactor the `deprecated_types` to a `collections.defaultdict(lambda: collections.defaultdict(dict))` then set:
```
deprecated_components["dash_core_components"]["LogoutButton"] = "''
The Logout Button is no longer used with Dash Enterprise and can be replaced with a html.Button or html.A.
eg: html.A(href=os.getenv('DASH_LOGOUT_URL'))
"""
```
The for the check you could have:

```
_type = getattr(comp, "_type",  "")
_ns = getattr(comp, "_namespace", "")
deprecation_message = deprecated_components[_ns][_type]
if deprecation_message:
    ...
```

leeagustin: Changed in 87385a9

components/dash-core-components/tests/integration/location/test_location_logout.py
```diff
@@ -33,15 +40,19 @@ def _insert_cookie(rep):
 
             return dcc.LogoutButton(id="logout-btn", logout_url="/_logout")
 
-    dash_dcc.start_server(app)
-    time.sleep(1)
-    dash_dcc.percy_snapshot("Core Logout button")
+    with pytest.warns(
```
T4rk1n, line 43: Need to add a condition `if add_initial_logout_button` for testing the deprecation notice otherwise it fails. Maybe better to just test the deprecation notice in a new test.

leeagustin: Right. My worry is that `DeprecationWarning` is not being output when the `LogoutButton` is being used inside a callback—only when the `LogoutButton` is defined in `app.layout`.

This is the reason I was checking for the warning in the two separate scenarios. However, if we don't need to output a `DeprecationWarning` when the `LogoutButton` is used inside a callback, then I can just create a new test.

gvwilson: @leeagustin can you please let us know if you added or are interested in adding the new test? thanks - @gvwilson

PR #2900, opened 2024-06-22
Allow strings in layout list

### LINKED ISSUE #2890: [BUG] No Layout Exception when using string elements when the layout is a list
In all versions of Dash this works:

```python 
app.layout = html.Div(["Select City", dcc.Dropdown()])
```

In dash >= 2.17, this will throw an error:

```python 
app.layout = ["Select City", dcc.Dropdown()]
```

Here is a full example.  When I run the app, I see the following error.  After refreshing the screen it works fine.

```python
from dash import Dash, dcc
app = Dash(__name__)

app.layout = ["Select City", dcc.Dropdown()]

if __name__ == '__main__':
    app.run(debug=True)


```

image

leeagustin:
Fixes #2890.

The problem was that strings were not allowed when the page layout was defined as a list. The issue only happens the first time—and not the second one after a refresh—because the layout validation using `validate_layout()` is only done the first time the page is accessed, resulting in the error `dash.exceptions.NoLayoutException: List of components as layout must be a list of components only.`

The solution is simply to allow strings to be included when the page layout is a list.

## Contributor Checklist

   -  [x] Allows strings if layout is a list

### optionals

### DIFF
--- a/dash/_validate.py
+++ b/dash/_validate.py
@@ -445,7 +445,7 @@ def validate_layout(layout, layout_value):
                 _validate(component)
             else:
                 raise exceptions.NoLayoutException(
-                    "List of components as layout must be a list of strings and components only."
+                    "Only strings and components are allowed in a list layout."
                 )
     else:
         _validate(layout_value)
--- a/dash/_validate.py
+++ b/dash/_validate.py
@@ -439,11 +439,13 @@ def validate_layout(layout, layout_value):
 
     if isinstance(layout_value, (list, tuple)):
         for component in layout_value:
+            if isinstance(component, (str,)):
+                continue
             if isinstance(component, (Component,)):
                 _validate(component)
             else:
                 raise exceptions.NoLayoutException(
-                    "List of components as layout must be a list of components only."
+                    "List of components as layout must be a list of strings and components only."
                 )
     else:
         _validate(layout_value)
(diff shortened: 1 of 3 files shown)


dash/_validate.py
```diff
@@ -439,11 +439,13 @@ def _validate_id(comp):
...
     if isinstance(layout_value, (list, tuple)):
         for component in layout_value:
+            if isinstance(component, (str,)):
+                continue
             if isinstance(component, (Component,)):
                 _validate(component)
             else:
                 raise exceptions.NoLayoutException(
-                    "List of components as layout must be a list of components only."
+                    "List of components as layout must be a list of strings and components only."
```
T4rk1n, line 448:
```suggestion
                    "Only strings and components are allowed in a list layout."
```

T4rk1n (commented):
Looks good just missing a changelog entry.

dash/_validate.py
leeagustin: Modified in d9d963f

leeagustin:
> Looks good just missing a changelog entry.

Added in ba30a4d

T4rk1n (approved):
:woman_dancing: Thank you

leeagustin:
It seems that one of the tests failed (`tests/integration/renderer/test_persistence.py:209 test_rdps005_persisted_props`), although these changes shouldn't have influenced it. The test is also passing on my machine. Is this what's blocking the merge?

leeagustin:
> 💃 Thank you

No problem! 💯 

It seems that one of the tests failed (tests/integration/renderer/test_persistence.py:209 test_rdps005_persisted_props), although these changes shouldn't have influenced it. The test is also passing on my machine. Is this what's blocking the merge?

leeagustin:
No problem @T4rk1n! Thank you for reviewing my changes! 🙏 

It seems that one of the tests failed (tests/integration/renderer/test_persistence.py:209 test_rdps005_persisted_props), although these changes shouldn't have influenced it. The test is also passing on my machine. Is this what's blocking the merge?

leeagustin:
Ah okay, sounds good. Thanks for the clarification!

ndrezn:
Nice fix!

leeagustin:
Thank you @ndrezn! 🙏

T4rk1n:
> test_rdps005_persisted_props

Think there is something wrong with the CI image/container because that test fails consistently on Python 3.8 image. I will merge the PR in another branch then in dev just to be sure.

PR #2901, opened 2024-06-24
Set progress no update

### LINKED ISSUE #2846: [BUG] no_update is not allowed as an output to set_progress for background progress
**Describe your context**
An easy way to replicate te code is using the provided example from [Dsah Plotly Website](https://dash.plotly.com/background-callbacks#example-4:-progress-bar:~:text=debug%3DTrue)-,Example%204%3A%20Progress%20Bar,-This%20example%20uses). I slightly modified to code by adding a no_update as a return value

```
dash                 0.42.0
dash-core-components 0.47.0
dash-html-components 0.16.0
dash-renderer        0.23.0
dash-table           3.6.0
```

**Describe the bug**

Returning `dash.no_update` as a value in set_progress is not allowed.

**Expected behavior**

`dash.no_update` should function as normally intended in callback functions

**Screenshots**

image

```python
import time
import os

from dash import Dash, DiskcacheManager, CeleryManager, Input, Output, html, callback, no_update

if 'REDIS_URL' in os.environ:
    # Use Redis & Celery if REDIS_URL set as an env variable
    from celery import Celery
    celery_app = Celery(__name__, broker=os.environ['REDIS_URL'], backend=os.environ['REDIS_URL'])
    background_callback_manager = CeleryManager(celery_app)

else:
    # Diskcache for non-production apps when developing locally
    import diskcache

app = Dash(__name__, background_callback_manager=background_callback_manager)

app.layout = html.Div(
    [
        html.Div(
            [
                html.Progress(id="progress_bar", value="0"),
            ]
        ),
    ]
)

@callback(
    background=True,
    running=[
        (
            {"visibility": "hidden"},
            {"visibility": "visible"},
        ),
        (
            {"visibility": "visible"},
            {"visibility": "hidden"},
        ),
    ],
    cancel=Input("cancel_button_id", "n_clicks"),
    prevent_initial_call=True
)
def update_progress(set_progress, n_clicks):
    total = 5
    for i in range(total + 1):
        set_progress((no_update, str(total)))
        time.sleep(1)

if __name__ == "__main__":
    app.run(debug=True)
```

leeagustin:
Fixes #2846.

While I wasn't able to replicate the error from the issue, I did find that `set_progress((no_update, str(total)))` sets the progress to 0:
image

This is because the progress bar's value is being set to `"[object Object]"`, which is probably being interpreted as 0. The progress bar's HTML being: `<progress id="progress_bar" value="[object Object]" max="5" style="visibility: visible;"></progress>`.

I modified it so that if `set_progress` has `no_update` for a component's output, that component will not be updated, but the other components—assuming they are not `no_update`—will be updated.

## Contributor Checklist

   -  [x] Allows for no_update in background callback set_progress

### optionals


tests/integration/long_callback/app_progress_delete.py
```diff
@@ -36,7 +36,9 @@
 )
 def on_bg_progress(set_progress, _):
     set_progress("start")
-    time.sleep(2)
+    time.sleep(1)
```
T4rk1n, line 39: Might be better to change this to lock to make sure it's not reset.

leeagustin: I'm assuming you meant to write _look_ instead of _lock_. Would 7437e2b work?

The final results for this test (_tests/integration/long_callback/test_basic_long_callback014.py_) would be:
- Using `dev`
image
- Using this PR
image

In `dev`, the `#progress-counter` is _3_ because the value gets incremented every time `set_progress` is called, but it's correctly being left as _2_ with this PR's changes.

As for why `#progress-output` is set back to _start_ in `dev`, I am not sure 🤔. However, it's correctly being left as _stop_ in this PR.

alexcjohnson: I think @T4rk1n was referring to `multiprocessing.Lock`, which you can find in some of the other Dash tests. This can remove the need for any explicit `sleep` statements, which not only makes the tests faster but also more robust on CI where you never know what else is running on the same hardware and can lead to unexpected delays that will intermittently break tests based on `sleep`.

T4rk1n: Yes, I was talking about a `multiprocessing.Lock`.

Basically you would do in the callback

```
with lock:
    set_progress(no_update)
```

Then in the test you can also request the lock and do the assertions, so you can be sure the call to set_progress is done before the assertions.

leeagustin: Ohh I see, that makes sense. I added the lock in 21bc4b2 🔒

tests/integration/long_callback/app_progress_delete.py
```diff
@@ -35,9 +36,12 @@
     prevent_initial_call=True,
 )
 def on_bg_progress(set_progress, _):
-    set_progress("start")
-    time.sleep(2)
-    set_progress("stop")
+    with lock:
+        set_progress("start")
+        time.sleep(0.5)
```
gvwilson, line 41: should we pull `0.5` out into a setting somewhere, or is it unlikely to ever change?

gvwilson (approved):
one question about style but otherwise good to go

robertclaus:
Closing since the PR hasn't had activity in a long time. Please re-open and resolve conflicts if still relevant!

PR #2903, opened 2024-06-26
Add callback on_error handler

### LINKED ISSUE #2348: [Feature Request] Callback Errors on production environments with Debug=False
When errors occur during callbacks on a production environment, with debug=False. This breaks the production and more than likely restarts the worker.

When starting up the server, allow for a mailbox item like smtp to be passed in order to send error alerts with traceback messages to the identified email address. ie `app.run(mail=mail, error_to=emailaddress)` Then, when an error occurs, raise PreventUpdate or some other standard error message, along the lines of "IT has been notified of the error that just occurred."

Wrapping all callbacks with the try except clause where an alert is sent to my email.

T4rk1n:
Add two error handler to call when a callback raise an exception, the return value of the handler is instead used as output. The callback context can be used to differentiate the erroring callbacks and access the original inputs and states.

A global error handler can be added with a function reference to the Dash constructor:  eg:  `app = Dash(on_error=handler)`.
And a local handler for particular callback can be added to individual callback: eg: `@callback(..., on_error=handler)`.

The signature of the handler function is: `def func(error: Exception) -> None`. `set_props` can be used to set extra props or the original output.

Example:
```python
def global_callback_error_handler(err):
    set_props("output-global": {"children": "error: {err}")

app = Dash(on_error=global_callback_error_handler)

app.layout = [
    html.Button("start", id="start-local"),
    html.Button("start-global", id="start-global"),
    html.Div(id="output"),
    html.Div(id="output-global"),
]

def on_callback_error(err):
    set_props("output": {"children": "callback error: {err}")

@app.callback(
    Output("output", "children"),
    Input("start-local", "n_clicks"),
    on_error=on_callback_error,
    prevent_initial_call=True,
)
def on_start(_):
    raise Exception("local error")

@app.callback(
    Output("output-global", "children"),
    Input("start-global", "n_clicks"),
    prevent_initial_call=True,
)
def on_start_global(_):
    raise Exception("global error")
```

Partially based of #2829 to catch exception and call a function instead of a decorator pattern and also works with background callbacks.

Closes #2348

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -414,7 +414,7 @@ max-bool-expr=5
 max-branches=15
 
 # Maximum number of locals for function / method body
-max-locals=20
+max-locals=25
 
 # Maximum number of parents for a class (see R0901).
 max-parents=7
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -475,6 +475,8 @@ def register_callback(
             else:
                 try:
                     output_value = _invoke_callback(func, *func_args, **func_kwargs)
+                except PreventUpdate as err:
+                    raise err
                 except Exception as err:  # pylint: disable=broad-exception-caught
                     if on_error:
                         output_value = on_error(err)
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -233,7 +233,7 @@ def insert_callback(
     long=None,
     manager=None,
     running=None,
-    dynamic_creator=False,
+    dynamic_creator: Optional[bool] = False,
     no_output=False,
 ):
     if prevent_initial_call is None:
@@ -277,6 +277,14 @@ def insert_callback(
     callback_list.append(callback_spec)
 
     return callback_id
+
+
+def _set_side_update(ctx, response) -> bool:
+    side_update = dict(ctx.updated_props)
+    if len(side_update) > 0:
+        response["sideUpdate"] = side_update
+        return True
+    return False
 
 
 # pylint: disable=too-many-branches,too-many-statements
@@ -350,7 +358,7 @@ def register_callback(
                 "callback_context", AttributeDict({"updated_props": {}})
             )
             callback_manager = long and long.get("manager", app_callback_manager)
-            app_on_error = kwargs.pop("app_on_error", None)
+            error_handler = on_error or kwargs.pop("app_on_error", None)
 
             if has_output:
                 _validate.validate_output_spec(insert_output, output_spec, Output)
@@ -361,7 +369,7 @@ def register_callback(
                 args, inputs_state_indices
             )
 
-            response = {"multi": True}
+            response: dict = {"multi": True}
             has_update = False
 
             if long is not None:
@@ -413,7 +421,6 @@ def register_callback(
                             triggered_inputs=callback_ctx.triggered_inputs,
                             ignore_register_page=True,
                         ),
-                        on_error=on_error or app_on_error,
                     )
 
                     data = {
@@ -451,10 +458,19 @@ def register_callback(
                     isinstance(output_value, dict)
                     and "long_callback_error" in output_value
                 ):
-                    error = output_value.get("long_callback_error")
-                    raise LongCallbackError(
+                    error = output_value.get("long_callback_error", {})
+                    exc = LongCallbackError(
                         f"An error occurred inside a long callback: {error['msg']}\n{error['tb']}"
                     )
+                    if error_handler:
+                        error_handler(exc)
+                        output_value = NoUpdate()
+                        # set_props from the error handler uses the original ctx
+                        # instead of manager.get_updated_props since it runs in the
+                        # request process.
+                        has_update = _set_side_update(callback_ctx, response)
+                    else:
+                        raise exc
 
                 if job_running and output_value is not callback_manager.UNDEFINED:
                     # cached results.
@@ -478,15 +494,14 @@ def register_callback(
                 except PreventUpdate as err:
                     raise err
                 except Exception as err:  # pylint: disable=broad-exception-caught
-                    if on_error:
-                        output_value = on_error(err)
-                    elif app_on_error:
-                        output_value = app_on_error(err)
+                    if error_handler:
+                        error_handler(err)
+                        if not multi:
+                            output_value = NoUpdate()
+                        else:
+                            output_value = [NoUpdate for _ in output_spec]
                     else:
                         raise err
-
-            if NoUpdate.is_no_update(output_value):
-                raise PreventUpdate
 
             component_ids = collections.defaultdict(dict)
 
@@ -508,12 +523,12 @@ def register_callback(
                 )
 
                 for val, spec in zip(flat_output_values, output_spec):
-                    if isinstance(val, NoUpdate):
+                    if NoUpdate.is_no_update(val):
                         continue
                     for vali, speci in (
                         zip(val, spec) if isinstance(spec, list) else [[val, spec]]
                     ):
-                        if not isinstance(vali, NoUpdate):
+                        if not NoUpdate.is_no_update(vali):
                             has_update = True
                             id_str = stringify_id(speci["id"])
                             prop = clean_property_name(speci["property"])
@@ -527,10 +542,7 @@ def register_callback(
                 flat_output_values = []
 
             if not long:
-                side_update = dict(callback_ctx.updated_props)
-                if len(side_update) > 0:
-                    has_update = True
-                    response["sideUpdate"] = side_update
+                has_update = _set_side_update(callback_ctx, response) or has_update
 
             if not has_update:
                 raise PreventUpdate
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -463,12 +463,17 @@ def register_callback(
                         f"An error occurred inside a long callback: {error['msg']}\n{error['tb']}"
                     )
                     if error_handler:
-                        error_handler(exc)
-                        output_value = NoUpdate()
+                        output_value = error_handler(exc)
+
+                        if output_value is None:
+                            output_value = NoUpdate()
                         # set_props from the error handler uses the original ctx
                         # instead of manager.get_updated_props since it runs in the
                         # request process.
-                        has_update = _set_side_update(callback_ctx, response)
+                        has_update = (
+                            _set_side_update(callback_ctx, response)
+                            or output_value is not None
+                        )
                     else:
                         raise exc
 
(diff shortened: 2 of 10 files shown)


dash/_callback.py
```diff
@@ -67,6 +68,7 @@ def callback(
     cancel=None,
     manager=None,
     cache_args_to_ignore=None,
+    on_error: Optional[Callable[[Exception], Any]] = None,
```
emilykl, line 71: Type hints <3 <3

.pylintrc
```diff
@@ -414,7 +414,7 @@ max-bool-expr=5
 max-branches=15
 
 # Maximum number of locals for function / method body
-max-locals=20
+max-locals=25
```
emilykl, line 417: Is it possible to move some of the logic to a sub-function rather than increasing this limit?

dash/_callback.py
```diff
@@ -272,8 +279,8 @@ def insert_callback(
     return callback_id
 
 
-# pylint: disable=R0912, R0915
-def register_callback(  # pylint: disable=R0914
+# pylint: disable=too-many-branches,too-many-statements
```
emilykl, line 282: Same question as above -- can this function be lightened a bit by moving some logic to sub-functions?

dash/_callback.py
```diff
@@ -462,7 +473,15 @@ def add_context(*args, **kwargs):
                 if output_value is callback_manager.UNDEFINED:
                     return to_json(response)
             else:
-                output_value = _invoke_callback(func, *func_args, **func_kwargs)
+                try:
```
emilykl, line 476: 💯

dash/_callback.py
```diff
@@ -462,7 +473,15 @@ def add_context(*args, **kwargs):
...
+                    output_value = _invoke_callback(func, *func_args, **func_kwargs)
+                except Exception as err:  # pylint: disable=broad-exception-caught
+                    if on_error:
```
emilykl, line 479: Why not do something like `on_error_func = on_error or app_on_error` further up in the code, to avoid this extra `elif`

emilykl:
@T4rk1n This rocks!!!

I'm going to play around with this feature tomorrow and try to break it. 

Couple of questions in the meantime:

- Does the global error handler also need to have the same number of return values as callback outputs? How does that work with apps with callbacks with different numbers of outputs?

- Regarding the API: Did you consider adding something like an `error_output` argument, which lets you specify an output that gets used only in case of error? Would that be possible?

.pylintrc
T4rk1n: Oh I forgot to mention in the PR description, this rule is broken cannot be concatenated with other rule disabling so i had to change this instead.

T4rk1n:
> @T4rk1n This rocks!!!
> 
> I'm going to play around with this feature tomorrow and try to break it.
> 
> Couple of questions in the meantime:
> 
>     * Does the global error handler also need to have the same number of return values as callback outputs? How does that work with apps with callbacks with different numbers of outputs?
> 
>     * Regarding the API: Did you consider adding something like an `error_output` argument, which lets you specify an output that gets used only in case of error? Would that be possible?

The global error handler needs the same outputs, to know the outputs the callback_context can be used. I just noticed that `ctx.output_list` has deprecation notice but I don't think it's justified and we can remove the notice.

BSd3v:
@T4rk1n,

If the `on_error` needs the same outputs as the target callback, wouldnt you encounter an error if these were misconfigured and fall silently when in production?

If this is also the case, then this is not really a global error handler as it can fail unexpectedly and is no different than placing error handling inside of the callback in the first place?

---

I think the `on_error` needs to return the same outputs as the renderer is expecting in order to allow for the callback to finish the `updates` on a component and close the `running`, and is not necessarily a restriction of the `on_error`. Wouldnt it make sense to do something like a prevent update for the original callback props?

T4rk1n:
> If the on_error needs the same outputs as the target callback, wouldnt you encounter an error if these were misconfigured and fall silently when in production?

The handler return values goes thru the same validation as the original callback function, the exception is raised.

> Wouldnt it make sense to do something like a prevent update for the original callback props?

Can still do `raise PreventUpdate` or return `[no_update for _ in ctx.outputs_list]` in the handler to prevent the outputs from updating if not required. I think leaving this the same return value for the error handler allow for more possibilities than if we automatically prevent update. cc @ndrezn

emilykl:
> Can still do raise PreventUpdate or return [no_update for _ in ctx.outputs_list] in the handler to prevent the outputs from updating if not required. I think leaving this the same return value for the error handler allow for more possibilities than if we automatically prevent update. cc @ndrezn

I like the idea of adding a `PreventUpdate` or `[no_update for _ in ctx.outputs_list]` _if_ the error handler doesn't have a return value -- would that be feasible @T4rk1n ?

BSd3v:
Another thing, is if someone does `raise PreventUpdate`, this flags as an error and would trigger the `on_error`, is this correct?

emilykl:
> Another thing, is if someone does `raise PreventUpdate` in a legit callback return, this flags as an error and would trigger the `on_error`, is this correct?

Just tried this out and it looks like currently it does (cc @T4rk1n). I suppose the fix would be to modify the try/except so that it catches every type of error _except_ `PreventUpdate`.

T4rk1n:
>  would that be feasible @T4rk1n ?

Yes, I think that would be less confusing, the outputs would still be able to be changed via set_props.

gvwilson:
@emilykl can you please give this one final look and approve the merge?

dash/_callback.py
```diff
@@ -462,10 +489,19 @@ def add_context(*args, **kwargs):
...
+                    output_value = _invoke_callback(func, *func_args, **func_kwargs)
+                except PreventUpdate as err:
+                    raise err
+                except Exception as err:  # pylint: disable=broad-exception-caught
+                    if error_handler:
+                        error_handler(err)
+                        if not multi:
+                            output_value = NoUpdate()
+                        else:
+                            output_value = [NoUpdate for _ in output_spec]
```
emilykl, line 502:
```suggestion
                    if error_handler:
                        output_value = error_handler(err)
                        if not multi:
                            output_value = output_value or NoUpdate()
                        else:
                            output_value = output_value or [NoUpdate() for _ in output_spec]
```

@T4rk1n I feel that we should still allow the dev to return outputs from the error handler -- we should supply the `NoUpdate()` outputs _only_ if the error handler doesn't return anything. (I realize the dev could still update the outputs via `set_props()`, but I don't see any downside to also allowing them to output the "normal" way.) Happy to discuss further if you disagree.

dash/_callback.py
```diff
@@ -462,10 +489,19 @@ def add_context(*args, **kwargs):
...
                     return to_json(response)
             else:
-                output_value = _invoke_callback(func, *func_args, **func_kwargs)
-
-            if NoUpdate.is_no_update(output_value):
-                raise PreventUpdate
+                try:
+                    output_value = _invoke_callback(func, *func_args, **func_kwargs)
+                except PreventUpdate as err:
+                    raise err
```
emilykl, line 495: 👍

dash/_callback.py
```diff
@@ -440,10 +458,19 @@ def add_context(*args, **kwargs):
...
                 ):
-                    error = output_value.get("long_callback_error")
-                    raise LongCallbackError(
+                    error = output_value.get("long_callback_error", {})
+                    exc = LongCallbackError(
                         f"An error occurred inside a long callback: {error['msg']}\n{error['tb']}"
                     )
+                    if error_handler:
+                        error_handler(exc)
+                        output_value = NoUpdate()
```
emilykl, line 467: see https://github.com/plotly/dash/pull/2903/files#r1669336763

```suggestion
                        output_value = error_handler(exc)
                        output_value = output_value or NoUpdate()
```

dash/_callback.py
emilykl: @T4rk1n any thoughts on ^ ? Not to be a stickler and it's not a blocker for merging this PR but would love to see this function simplified if possible rather than disabling the linter

emilykl:
@T4rk1n See my comments regarding outputs from the error handler. Otherwise LGTM.

dash/_callback.py
BSd3v: For this wouldn't the dev just handle the error inside of the callback itself, as no two outputs are going to be the same?

Also, I thought that this would return a regular output? The no_update seems to be only supplied if there was no returned value?

BSd3v:
Just going to throw this here:

This PR, once merged should close this PR: https://github.com/plotly/dash/pull/2829

And also close this issue: https://github.com/plotly/dash/issues/2348

dash/_callback.py
T4rk1n: Doing it like this could lead to situation where you wanted to output something that evaluate to false but it would get a `no_update` instead. see https://github.com/plotly/dash/issues/2906#issuecomment-2209086283, a components which has no children props will always be false.

Best we could do is `if output_value is not None`, but that brings back the age old discussion of undefined vs none, which python lacks an undefined type and None could be a valid prop.

BSd3v: What about this:
```
                    if error_handler:
                        output_value = error_handler(err)
                        if output_value is not None or output_value is False:
                            # Code to execute if a value is returned
                            '''output no change'''
                        else:
                            # Code to execute if no value is returned
                            output_value = [NoUpdate() for _ in output_spec] if multi else NoUpdate()
                            ```

Then have it documented that this will pass `NoUpdate` in error if you are really trying to pass a `None` as a single output. If this is the case, then the dev should instead place: `[None]`

emilykl: > For this wouldn't the dev just handle the error inside of the callback itself, as no two outputs are going to be the same?

For the global error handler, yes, this is true -- I can't imagine a situation where you'd want to return specific outputs from the global error handler. But for a callback-specific error handler the dev might want to return specific outputs.

> python lacks an undefined type and `None` could be a valid prop.

Ugh, this is a good point. Thanks Python. I think I would still be OK with the solution of checking `if output_value is not None`, and then documenting that in the specific edge case where you do actually want to return `None`, you have to use `set_props()`.

BSd3v: > callback-specific error handler

unless they are reusing, they could still handle inside of the callback, but our syntax is still easier I guess, haha. Rather than them having to wrap the whole function in a try except clause.

dash/_callback.py
T4rk1n: I didn't disable anything in this PR, just changed to the rulename instead of number. The settings for these rules is kinda low imo,
- too-many-branches=15 is kinda high, but this function is wrapped multiple times in decorators which is why it's triggered here.
- too-many-statements=50 is kinda low for a project this size and keep getting triggered everywhere, we also have functions within functions which actually separate the code inside but the linter isn't too smart here and the top function accumulates all the statements.

dash/_callback.py
T4rk1n: Changed to NoOutput only if `output_value` is None, the output could be set to `[None]` if the dev want to set the output prop to null. cc @LiamConnors

dash/_callback.py
emilykl: Oh I see now, thanks for the explanation. 👍

dash/_callback.py
emilykl: > the output could be set to [None] if the dev want to set the output prop to null.

FWIW, does this actually work? Typically I have gotten errors when wrapping a single output in a list.

emilykl (approved):
@T4rk1n Could you maybe update the test cases so there's one test case with an error handler function which returns callback outputs?

In any case, looks good! 🚀

dash/_callback.py
BSd3v: Yes, it works, but only would work if the output definition in the callback is `[Output(id, prop)]`

PR #2908, opened 2024-06-28
Evaluate HOST, PORT, and PROXY env vars when Dash.run() is invoked

### LINKED ISSUE #2902: [BUG] Environment variables from .env file ignored in Dash.run
**Describe your context**

```
dash                      2.17.1
dash-bootstrap-components 1.6.0
```

**Describe the bug**

Environment variables `HOST`, `PORT` and `DASH_PROXY` are ignored if defined in a .env file and loaded using `python-dotenv`.

In `Dash.run`, environment variables used as default argument values are evaluated at the time the method is defined, not when it is called. This can lead to discrepancies if the environment variables change between definition and execution.

Specifically, this means that environment variables specified in a .env file and loaded using the python-dotenv package will be ignored if they are loaded after the method definition. Since load_dotenv() is typically called after importing Dash, this issue affects many standard usage patterns.

The [docs](https://dash.plotly.com/reference#app.run) for `app.run` say "If a parameter can be set by an environment variable, that is listed too. Values provided here take precedence over environment variables." It doesn't mention that an .env file cannot be used.

**Expected behavior**

Arguments which can be set using environment variables should reflect variables from the .env file (loaded using `python-dotenv`).

**Proposed solution**

In `Dash.app` (`dash.py`), set `host`, `port` and `proxy` to `None` by default.

```python
   def run(
        self,
        host=None,
        port=None,
        proxy=None,
        ...
```

Then use

```python
        host = host or os.getenv("HOST", "127.0.0.1")
        port = port or os.getenv("PORT", "8050")
        proxy = proxy or os.getenv("DASH_PROXY", None)
```

at the start of the function itself to read the env vars when the method is called.

martian0x80:
Fixes #2902 

Made the changes as proposed in #2902. Nothing major, just re-evaluates environment variables when `Dash.run()` is invoked.

## Contributor Checklist

   -  [x] Evaluate the environment variable after Dash.run() is called

### optionals

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -2109,6 +2109,12 @@ class Dash:
             dev_tools_prune_errors,
         )
 
+        # Evaluate the env variables at runtime
+
+        host = os.getenv("HOST", host)
+        port = os.getenv("PORT", port)
+        proxy = os.getenv("DASH_PROXY", proxy)
+
         # Verify port value
         try:
             port = int(port)
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1981,8 +1981,8 @@ class Dash:
 
     def run(
         self,
-        host=None,
-        port=None,
+        host="127.0.0.1",
+        port="8050",
         proxy=None,
         debug=None,
         jupyter_mode: JupyterDisplayMode = None,
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1981,9 +1981,9 @@ class Dash:
 
     def run(
         self,
-        host=os.getenv("HOST", "127.0.0.1"),
-        port=os.getenv("PORT", "8050"),
-        proxy=os.getenv("DASH_PROXY", None),
+        host=None,
+        port=None,
+        proxy=None,
         debug=None,
         jupyter_mode: JupyterDisplayMode = None,
         jupyter_width="100%",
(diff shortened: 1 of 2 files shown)


T4rk1n (changes requested):
Looks good but also need to set the parameters to `None` on line 1984-1986

dash/dash.py
```diff
@@ -1981,9 +1981,9 @@ def delete_resource(resources):
 
     def run(
         self,
-        host=os.getenv("HOST", "127.0.0.1"),
-        port=os.getenv("PORT", "8050"),
-        proxy=os.getenv("DASH_PROXY", None),
+        host=None,
+        port=None,
```
T4rk1n, line 1985:
```suggestion
        host="127.0.0.1",
        port="8050",
```

T4rk1n (commented):
Sorry we lost the default arguments by setting the parameters to None, they need the previous default instead.

Also need to add a changelog entry then it's good to merge.

dash/dash.py
martian0x80: I was planning to ask the same, because they were going to be overwritten anyway, and defaults were already set. Will make the changes.

T4rk1n (approved):
Thank you :woman_dancing:

PR #2909, opened 2024-06-28
Rehighlights dcc.Markdown when it is updated

### LINKED ISSUE #2895: Markdown does not update code component properly after callback update.
I have a dcc.Markdown that I want to update based on user input in a Textarea. Everything works correctly besides the code component, which when edited stops showing the correct colors (and upon further inspection does not assign the correct classes in the html code).

Expected:
The markdown should update the code component properly.

Environment:
```
dash==2.17.1
dash-core-components==2.0.0
dash-html-components==2.0.0
dash-table==5.0.0
```
    - OS: Windows, Linux
    - Browser: Chrome, Firefox 
    - Chrome Version 126.0.6478.62, Firefox Version 127.0 (Windows)

Minimal reproducible example:
```python

app = Dash(__name__)

app.layout = html.Div([
    dcc.Textarea(value='''def sum(a, b):
    return a+b

sum(2,2)''', id='codeta'),
    dcc.Markdown(id='markdown'),
])

@app.callback(
    Output('markdown', 'children'),
    Input('codeta', 'value')
)
def update_markdown(value) -> str:
    return f'```python\n\n{value}\n\n```'

if __name__ == '__main__':
    app.run(debug=True)
```

[markdown_bug](https://github.com/plotly/dash/assets/34633356/cd7be0e3-9191-4acc-aaaa-538835f4fa10)

leeagustin:
Fixes #2895.

The problem is that highlight.js was changed in https://github.com/highlightjs/highlight.js/commit/3987abe437dbf962d64a51da6282d9c9bc20fc13, which prevents re-highlighting an element.

The attribute `data-highlighted="yes"` flags an element as highlighted. Hence, the solution is to remove this attribute before the code is re-highlighted.

## Contributor Checklist

   -  [x] Re-highlights `dcc.Markdown` after it is updated

### optionals


KMORGAN008 (approved):

gvwilson:
@T4rk1n can you please have a look and let us know if we can merge?

T4rk1n (approved):

PR #2915, opened 2024-07-06
Fixes 'AttributeError' when layout is a function that returns a list of components in dash layout setter

### LINKED ISSUE #2905: Layout as list of components does not work when layout is a function
**Describe your context**

```
dash                       2.17.1
dash_ag_grid               31.2.0
dash-bootstrap-components  1.6.0
dash-core-components       2.0.0
dash-html-components       2.0.0
dash-mantine-components    0.12.1
dash-table                 5.0.0
dash-testing-stub          0.0.2
```

**Describe the bug**

https://github.com/plotly/dash/pull/2795 enabled you to pass a list of components to `app.layout`. However, this does not work when `app.layout` is set to a function:

```
import datetime

from dash import Dash, html


def layout():
    return [html.H1(datetime.datetime.now())]


app = Dash(__name__)
app.layout = layout

app.run()
```

Gives:
```
  File "/Users/antony_milne/Library/Application Support/JetBrains/PyCharm2024.1/scratches/scratch_51.py", line 11, in <module>
    app.layout = layout
    ^^^^^^^^^^
  File "/Users/antony_milne/Library/Application Support/hatch/env/virtual/vizro/fm3ubPZu/vizro/lib/python3.11/site-packages/dash/dash.py", line 734, in layout
    [simple_clone(c) for c in layout_value._traverse_ids()],
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'list' object has no attribute '_traverse_ids'
```

I see that the main motivation for #2795 was that it simplified examples for beginners, so this is not such an important case because using a function for `layout` is already more advanced. But still I think this should probably work.

The root of the problem is that the code in `Dash.layout` that handles the case that layout is a function still expect it to return a single Dash component. One possible fix would be to alter this line:
https://github.com/plotly/dash/blob/bbd013cd9d97d8c41700b240dd95c32e6b875998/dash/dash.py#L729
This expects a single Dash component to be returned rather than a list so it works if you just wrap it in an `html.Div`.

martian0x80:
Fixes #2905
Changes as the proposed solution suggested, layout setter throws an `AttributeError` when layout is a function that returns a list of components. The solution checks if the `layout_value` is an instance of `list` or `tuple`, if it is, it wraps it in a `html.Div` component, otherwise proceeds as usual.

## Contributor Checklist

   -  [x] Handle the case when layout is a function that returns a list of components

### optionals


T4rk1n:
@alexcjohnson I see this [comment](https://github.com/plotly/dash/blob/bbd013cd9d97d8c41700b240dd95c32e6b875998/dash/dash.py#L713-L715) from a couple years ago mention this fix is mostly to empty out children for Python 2. I don't see any other uses, think we can remove the simple_clone function ?

dash/dash.py
```diff
@@ -727,6 +727,10 @@ def simple_clone(c, children=None):
                 return cls(**props)
 
             layout_value = self._layout_value()
+
+            if isinstance(layout_value, (list, tuple)):
+                layout_value = html.Div(children=layout_value)
```
T4rk1n, line 732: We can  delete the function [`simple_clone`](https://github.com/plotly/dash/blob/bbd013cd9d97d8c41700b240dd95c32e6b875998/dash/dash.py#L711-L727) and assign the `self.validation_layout = layout_value`  instead.

T4rk1n (changes requested):

dash/dash.py
martian0x80: thanks, will make requested changes

martian0x80: done

T4rk1n (commented):
Looks good, seems like there is some formatting error to fix (`npm run format`) then we can merge :tada:

martian0x80:
@T4rk1n 

Should I commit the other files that were formatted as well?
I avoided the commit hooks that invoke the black formatter with `--no-verify` because it wouldn't let me commit since other files that I did not change had formatting problems. So, should I commit all the format changes or just the `dash.py`?

These are the changes. 
image

T4rk1n:
> Should I commit the other files that were formatted as well?

No, I don't think it should have formatted all those files. npm probably took the wrong version of black, make sure the venv is activated before running `npm run format` as it may takes the `black` from a global env.

T4rk1n (approved):
:woman_dancing: Look good, thank you.

martian0x80:
> > Should I commit the other files that were formatted as well?
> 
> No, I don't think it should have formatted all those files. npm probably took the wrong version of black, make sure the venv is activated before running `npm run format` as it may takes the `black` from a global env.

Thanks, that was indeed the problem. Sorry, for being clueless.

PR #2922, opened 2024-07-15
Fixes hash_function and error handler missing call for multi

### LINKED ISSUE #1885: [BUG] long_callback fails when bundling python code
**Describe your context**

```
```
**Describe the bug**

**Expected behavior**

T4rk1n:
- Fix #1885, fallback on function name if the source is not available.
- Fix missing `NoUpdate()` for multi output callbacks error handler.

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -507,7 +507,7 @@ def register_callback(
                             if not multi:
                                 output_value = NoUpdate()
                             else:
-                                output_value = [NoUpdate for _ in output_spec]
+                                output_value = [NoUpdate() for _ in output_spec]
                     else:
                         raise err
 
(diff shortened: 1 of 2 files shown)


gvwilson:
@marthacryan can you please review?

marthacryan:
When I try to follow the link for #1885, it gives me a "Page not found" error. Is that because of permissions? Was it deleted?

gvwilson:
that is strange - @T4rk1n any ideas?

marthacryan (approved):
LGTM!

corebit-nl:
@T4rk1n , could you apply the try-except logic from https://github.com/plotly/dash/blob/dev/dash/long_callback/managers/__init__.py#L110 to https://github.com/plotly/dash/blob/dev/dash/long_callback/managers/__init__.py#L58 as well? In the current situation, long callbacks still fail for packed applications (eg. Pyinstaller). Fixing this would close https://github.com/Nuitka/Nuitka/issues/2988 as well.

PR #2926, opened 2024-07-22
Update default props

### LINKED ISSUE #2919: React 18 warning defaultProps will be removed from function components
Running an app with React 18 emit a warning that defaultProps for functions components are deprecated.

Warning message:
`Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.`

We use the defaultProps for our own components generator in either the generated react-docgen metadata or the typescript transpiler. We do not support newer version of react-docgen (5.4.3 is our last supported version) that may have a fix for this in their code. The typescript transpiler needs to be adapted.

Affected core components:
- All html components.
- dcc
  - Link
  - Loading
  - Location
  - Tab
  - Tooltip

AnnMarieW:
closes #2919 

I updated all the components listed in #2919 except for `dcc.Location` because it's still a class component.

- [ ] validate if it works with the typescript components

### optionals


AnnMarieW:
Another dependency to update - our version of eslint does not seem to support the spread operator.  https://github.com/eslint/eslint/issues/10307

```
> eslint src scripts

/home/circleci/dash/components/dash-html-components/src/components/A.react.js
  12:52  error  Parsing error: Unexpected token ..
```
For example, this error is from defining the defaults like this in `A.react.js`
```
const A = ({n_clicks = 0, n_clicks_timestamp = -1, ...props}) => {

```

gvwilson:
@AnnMarieW can you please let us know if you're still working on this one or is it ready for review? thanks - @gvwilson

AnnMarieW:
Hi @gvwilson 

I'm afraid, I've hit a roadblock on this one.  Here's the current status:

-  The components listed in issue 2919 are updated so that the defaults are defined in the function signature. I verified that the default props are set correctly in the component and the docstrings are generated correctly to show the defaults.
-  The [sample typescript component](@plotly/dash-generator-test-component-typescript/src/components/TypeScriptComponent.tsx) is updated.  I verified that the defaults are set correctly in the component, however the docstrings do not show the default values.

I found this project that [parses the default props](https://github.com/mrginglymus/react-docgen-typescript/blob/7136ce3f5c7acf39ac260cdcd0c2c66f3fc4fcb1/src/parser.ts#L962) from the function signature in typescript components, however, I'm stuck trying to integrate that logic into [extract-meta.js](https://github.com/plotly/dash/blob/e37ce5ea9aba9ef77b81968659b93f9362b747cf/dash/extract-meta.js#L508)  

If someone could help with that piece, I could probably finish  the rest (ie tests)  However, I'm traveling for a few weeks, and would be happy to turn this project over to someone if you'd like to get it wrapped up sooner.

gvwilson:
@T4rk1n please have a look at this one and help @AnnMarieW move it forward when you can. thanks - @gvwilson

@plotly/dash-generator-test-component-typescript/src/props.ts
```diff
@@ -32,7 +32,7 @@ export type TypescriptComponentProps = {
     array_elements?: JSX.Element[];
 
     string_default?: string;
-    number_default?: string;
+    number_default?: number;
```
gvwilson, line 35: this seems kind of important :-)

components/dash-core-components/.eslintrc
```diff
@@ -43,7 +44,7 @@
     "import/named": ["off"],
     "import/namespace": ["off"],
     "import/no-duplicates": ["error"],
-    "import/no-named-as-default": ["error"],
+    "import/no-named-as-default": ["off"],
```
gvwilson, line 47: apologies, but can you please explain what the impact of this one will be?

components/dash-core-components/src/components/Loading.react.js
```diff
@@ -22,8 +22,8 @@ const getSpinner = spinnerType =>
 const Loading = ({
     children,
     loading_state,
-    display,
-    color,
+    display = 'auto',
+    color = '#119DFF',
```
gvwilson, line 26: do we have colors and other constants scattered elsewhere through the code as well or is there a way to consolidate them all in one place? (very low priority, not worth holding up this PR for)

components/dash-core-components/src/components/Tooltip.react.js
```diff
@@ -6,25 +6,35 @@ import _JSXStyle from 'styled-jsx/style'; // eslint-disable-line no-unused-vars
 /**
  * A tooltip with an absolute position.
  */
-const Tooltip = props => {
-    const {bbox, border_color, background_color, id, loading_state} = props;
+const Tooltip = ({
+    show = true,
+    targetable = false,
+    direction = 'right',
+    border_color = '#d6d6d6',
```
gvwilson, line 13: again, can we (in future?) replace magic color constants with names that are all defined in one place?

gvwilson (approved):
some questions but no blockers

@plotly/dash-generator-test-component-typescript/src/props.ts
AnnMarieW: I changed this to get tests to pass, but @T4rk1n can you confirm if this is OK?

T4rk1n: Yes that is the good type.

Issue #2329, opened 2022-11-21
[Feature Request] Allow currency format for dcc.Input of type='number'


celia-plotly:
**Describe the solution you'd like**
Allow currency format for dcc.Input while keeping the value type as number and the arrows to increase/decrease it.

**Describe alternatives you've considered**
Clientside circular callback that updates the value format based on [this JS solution](https://stackoverflow.com/questions/9372624/formatting-a-number-as-currency-using-css):
```
import dash
from dash import Input, Output, State, callback, dcc, html, ctx

app = dash.Dash(__name__)

app.layout = html.Div([
    dcc.Input(id='input-number', value=522.31, type='number')
])

app.clientside_callback(
    """
    function(value) {
        return value.toLocaleString('us-US', { style: 'currency', currency: 'USD' });
    }
    """,
    Output('input-number', 'value'),
    Input('input-number', 'value')
)

if __name__ == '__main__':
    app.run_server(debug=True)
```
It raises this error in the JS Console:
```
react-dom@16.v2_7_0m1667919873.14.0.js:1857 The specified value "-US$1.00" cannot be parsed, or is out of range.
```
If `type='number'` is not specified, the value is accepted but the arrows to increase/decrease the value disappear.

Other option is a one-cell table with html.Buttons and a callback to increase/decrease the value (it's very complex for just a simple input)

PR #2932, opened 2024-07-24
Implements #2329 : currencyFormat for Plotly Dash 

rodasarede:
Implements #2329
With this commit it is now allowed currency formatting by setting the currencyFormat property to True only when I have type='number'.
When currencyFormat is enabled, the currencySymbol property allows you to specify the currency symbol to be used.
For example, if the input value is 12345, and currencyFormat = true, currencySymbol = '€' it will be displayed as €12345.

## Contributor Checklist

- [x] I have run the tests locally and they passed. (refer to testing section in [contributing]

### DIFF
--- a/components/dash-core-components/src/components/Input.react.js
+++ b/components/dash-core-components/src/components/Input.react.js
@@ -31,6 +31,8 @@ const inputProps = [
     'size',
     'style',
     'id',
+    'currencyFormat',
+    'currencySymbol',
 ];
 
 /**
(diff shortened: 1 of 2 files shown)


components/dash-core-components/src/components/Input.react.js
```diff
@@ -90,25 +96,58 @@ export default class Input extends PureComponent {
             this.setState({value: this.props.value});
         }
     }
+    formatCurrency(value) {
+        if (isNaN(value) || value === '') {
+            return '';
+        }
+        const {currencySymbol} = this.props;
+
+        return `${currencySymbol}${Number(value).toLocaleString()}`;
```
gvwilson, line 105: are there cases where the currency symbol comes _after_ the number? Or is that just odd English typography like `15¢` that isn't used for international currencies?

gvwilson (approved):

robertclaus:
Given the age of the PR and inactivity I am going to close it. Please re-open if you are still working on it.

PR #2949, opened 2024-08-19
Fix/serve component suites FileNotFoundError 

yzcheung:
-  [x] Added a try/except block to catch FileNotFoundError when attempting to serve component suites.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 orbs:
   win: circleci/windows@5.0.0
   percy: percy/agent@0.1.3
-  browser-tools: circleci/browser-tools@1.4.6
+  browser-tools: circleci/browser-tools@1.4.8
 
 
 jobs:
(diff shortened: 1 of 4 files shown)


yzcheung:
In the development process of Dash components, .js.map files are often utilized for debugging purposes. However, these files may not be generated or included in the production releases. The absence of these files in the production environment can lead to a FileNotFoundError when the code attempts to load them.

To prevent this error from causing disruptions, I have added a try/except block to catch the FileNotFoundError when the Dash component suite attempts to serve these files. This modification ensures that the application continues to run smoothly in production environments, even if the .js.map files are missing.

This change is crucial for maintaining functionality in production environments where .js.map files are typically not included.

ndrezn (changes requested):
It looks like there are quite a few diffs in the imports list that aren't related to this PR using something like `isort`. Could you revert the changes to the ordering of import statements? This way the diffs are clear.

PR #2951, opened 2024-08-19
Update CONTRIBUTING.md to have easier to read build steps

marthacryan:
Updates the CONTRIBUTING.md file to split out a large list of bash commands separated by comments into smaller code chunks separated by text. This is easier to read because the comments blend in with the bash commands.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,37 +2,47 @@
 
 ## Getting Started
 
-Glad that you decided to make your contribution in Dash, to set up your development environment, run the following commands:
-
+To set up your development environment, start by cloning the repo and creating a virtual environment:
 ```bash
 # in your working directory
-$ git clone git@github.com:plotly/dash.git
-$ cd dash
-$ python3 -m venv .venv/dev
-# activate the virtualenv
-# on windows `.venv\dev\scripts\activate`
-# on some linux / mac environments, use `.` instead of `source`
-$ source .venv/dev/bin/activate
-# install dash and dependencies
-$ pip install -e .[ci,dev,testing,celery,diskcache]  # in some shells you need \ to escape []
-$ npm ci
-# this script will build the dash-core-components, dash-html-components, dash-table,
-# and renderer bundles; this will build all bundles from source code in their
-# respective directories. The only true source of npm version is defined
-# in package.json for each package.
-#
-$ npm run build  # runs `renderer build` and `npm build` in dcc, html, table
-# build and install components used in tests
-# on windows, the developer will need to use `npm run first-build` this performs additional first steps
-#
-# Alternatively one could run part of the build process e.g.
-$ dash-update-components "dash-core-components"
-# to only build dcc when developing dcc
-# But when you first clone check out a new branch, you must run the full build as above.
-#
-$ npm run setup-tests.py # or npm run setup-tests.R
-# you should see dash points to a local source repo
-$ pip list | grep dash
+git clone git@github.com:plotly/dash.git
+cd dash
+python3 -m venv .venv/dev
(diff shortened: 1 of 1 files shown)


marthacryan:
Oh I'm also noticing that there is no `setup-tests.py` file but that's the actual command in the package.json. Not sure if other people agree but I found that very confusing and I was looking for a file called `setup-tests.py`. Why wouldn't we call it something like `setup-python-tests`?

marthacryan:
Also: should we be including specific virtual environment instructions? I use conda right now so I haven't tested whether it actually works with venv, but I would expect people to use their preferred python env. We could link to the docs for venv and conda instead?

CONTRIBUTING.md
```diff
@@ -2,37 +2,47 @@
...
+```bash
+pip install -e ".[ci,dev,testing,celery,diskcache]"  # in some shells you need \ to escape []
+npm ci
+```
+
+This script will build the dash-core-components, dash-html-components, dash-table, and renderer bundles; this will build all bundles from source code in their respective directories. The only true source of npm version is defined in package.json for each package.
+```bash
+npm run build  # runs `renderer build` and `npm build` in dcc, html, table
+```
+To build and install components used in tests on windows, the developer will need to use `npm run first-build` this performs additional first steps.
```
emilykl, line 26: move "On windows" to the beginning of this sentence

CONTRIBUTING.md
```diff
@@ -2,37 +2,47 @@
...
+
+Alternatively one could run part of the build process e.g.
+```bash
+dash-update-components "dash-core-components"
+```
+to only build dcc when developing dcc. But when you first clone check out a new branch, you must run the full build as above.
```
emilykl, line 32: when you first clone **or** check out a new branch

CONTRIBUTING.md
```diff
@@ -2,37 +2,47 @@
...
+
+To set up tests, run:
```
emilykl, line 34: What does "setting up tests" entail? Does this need to be run each time before running tests? or each time a new branch is checked out? or only when the repo is first set up?

emilykl (approved):
Thanks for doing this @marthacryan ! 🎉

CONTRIBUTING.md
marthacryan: I have no idea! Maybe @T4rk1n could check this part out because I actually made this PR because I couldn't really figure out how to run the tests

emilykl:
> Also: should we be including specific virtual environment instructions? I use conda right now so I haven't tested whether it actually works with venv, but I would expect people to use their preferred python env. We could link to the docs for venv and conda instead?

We could say something along the lines of "instructions provided here for `venv` but you can use your preferred Python virtual environment manager (e.g. `conda`)".

I'm also OK with leaving as-is for now since the `venv` instructions were already there and you're just reformatting.

marthacryan:
I'm going to go ahead and merge this because most of the remaining discussion points / questions are unrelated to the formatting updates in this PR. They're worth discussing though!

CONTRIBUTING.md
T4rk1n: The command build the tests libraries in the `@plotly` directory, they are used to test the component generator.

T4rk1n (approved):
:woman_dancing: Looks good.

AnnMarieW:
Nice updates here! :confetti_ball: 

Any chance of helping get https://github.com/plotly/dash/pull/2893 over the finish line since you're in the neighborhood?

marthacryan:
#2893 ended up covering basically all of the changes in this PR, so I'm gonna go ahead and close this!

PR #2959, opened 2024-08-23
fix(server): handle FileNotFoundError in serve_component_suites

C0deBeez:
Added a try/except block to catch FileNotFoundError when attempting to serve component suites. This prevents the server from crashing when a requested file is not found, and instead returns a 404 status code.

In the development process of Dash components, .js.map files are often utilized for debugging purposes. However, these files may not be generated or included in the production releases. The absence of these files in the production environment can lead to a FileNotFoundError when the code attempts to load them.

To prevent this error from causing disruptions, I have added a try/except block to catch the FileNotFoundError when the Dash component suite attempts to serve these files. This modification ensures that the application continues to run smoothly in production environments, even if the .js.map files are missing.

This change is crucial for maintaining functionality in production environments where .js.map files are typically not included.


yzcheung:
@ndrezn  Sorry, there were some issues with my local environment, so I have opened a new PR here.

gvwilson (approved):

C0deBeez:
@gvwilson Is it okay to proceed with merging？

gvwilson:
@C0deBeez we're trying to get a release out the door next week but I'll have someone take a look at this as soon as that's done. thanks - @gvwilson

Issue #2590, opened 2023-07-07
[BUG] deprecation warnings for selenium when using `dash_duo`


prokie:
**Describe your context**
When using `dash_duo` and performing tests on our application we get the following deprecation warnings

```
  C:\repositories\database-gui\venv\Lib\site-packages\selenium\webdriver\remote\remote_connection.py:391: DeprecationWarning:
  HTTPResponse.getheader() is deprecated and will be removed in urllib3 v2.1.0. Instead use HTTPResponse.headers.get(name, default).
```

-  replace the result of `pip list | grep dash` below
```
dash                      2.11.1 
dash-bootstrap-components 1.4.1  
dash-chart-editor         0.0.1a4
dash-core-components      2.0.0  
dash-extensions           1.0.1  
dash-html-components      2.0.0  
dash-table                5.0.0  
dash-testing-stub         0.0.2  
```
**Describe the bug**
When running pytest with `dash_duo` the following deprecation warnings show up.
```
  C:\repositories\database-gui\venv\Lib\site-packages\selenium\webdriver\remote\remote_connection.py:391: DeprecationWarning:
  HTTPResponse.getheader() is deprecated and will be removed in urllib3 v2.1.0. Instead use HTTPResponse.headers.get(name, default).
```

**Expected behavior**

I expect to not get any deprecation warnings.

prokie:
This issue seems to be solved if we update `requires-testing.txt` to allow for selenium version 4.7.0.

```
# Dependencies necessary for utilizing Dash provided testing utilities
beautifulsoup4>=4.8.2
cryptography<3.4;python_version<"3.7"
lxml>=4.6.2
percy>=2.0.2
pytest>=6.0.2
requests[security]>=2.21.0
selenium>=3.141.0,<=4.7.0
waitress>=1.4.4
multiprocess>=0.70.12
psutil>=5.8.0
dash_testing_stub>=0.0.2
```

T4rk1n:
This is on selenium side, should be fixed in selenium 4.6.1 and above.

prokie:
Seems like the latest working selenium version is `4.9.1`. The version latest version `4.10.0` is giving me errors.

```
E       selenium.common.exceptions.InvalidArgumentException: Message: invalid argument: unrecognized capability: loggingPrefs
E       Stacktrace:
E       Backtrace:
E               GetHandleVerifier [0x00ADA813+48355]
E               (No symbol) [0x00A6C4B1]
[...]

PR #2964, opened 2024-08-25
Change selenium version to avoid depecration warnings

AugustoMagalhaes:
Some tools depending on Dash have been facing the same deprecation warning from issue #2590, which can be solved by modifying the selenium upper range version to 4.6.1, as found in https://stackoverflow.com/questions/75382508/deprecationwarning-httpresponse-getheader-is-deprecated-and-will-be-removed-i. I've tested the change in python versions 3.8.x, 3.9.x, 3.10.x, 3.11.x and 3.12.x and no problem has been found.

## Contributor Checklist

   -  [x] Change selenium version range
   -  [x] Run tests from python 3.8 to 3.12 to check if this change could break the tests

### DIFF
--- a/requirements/testing.txt
+++ b/requirements/testing.txt
@@ -5,7 +5,7 @@ lxml>=4.6.2
 percy>=2.0.2
 pytest>=6.0.2
 requests[security]>=2.21.0
-selenium>=3.141.0,<=4.2.0
+selenium>=3.141.0,<=4.6.1
 waitress>=1.4.4
 multiprocess>=0.70.12
 psutil>=5.8.0


T4rk1n:
There is a few failures that are related to our tests than the `dash[testing]` api.
Here we have a removal: `AttributeError: 'WebElement' object has no attribute 'find_element_by_tag_name'` for this test: - tests/integration/callbacks/test_multiple_callbacks.py:585 test_cbmt013_chained_callback_should_be_blocked that needs to be changed.
`test_duo001_wait_for_text_error` and `test_duo002_wait_for_text_value` test the testing api, probably needs to be adapted too.
For the failures in dcc tests, I think it's the dash_duo methods that is failing and may need to be fixed: `click_at_coord_fractions`

AugustoMagalhaes:
> There is a few failures that are related to our tests rather than the `dash[testing]` api. Here we have a removal: `AttributeError: 'WebElement' object has no attribute 'find_element_by_tag_name'` for this test: - tests/integration/callbacks/test_multiple_callbacks.py:585 test_cbmt013_chained_callback_should_be_blocked that needs to be changed. `test_duo001_wait_for_text_error` and `test_duo002_wait_for_text_value` test the testing api, probably needs to be adapted too. For the failures in dcc tests, I think it's the dash_duo methods that is failing and may need to be fixed: `click_at_coord_fractions`

Thanks for the feedback, T4rk1n. So just to confirm: I should work on the dash_duo methods for dcc failing tests or should I also work on fixing the other mentioned ones?

T4rk1n:
>  So just to confirm: I should work on the dash_duo methods for dcc failing tests or should I also work on fixing the other mentioned ones?

Yes they both need fixing.

robertclaus:
Closing due to age and the fact that this should be resolved in https://github.com/plotly/dash/pull/3336

PR #2985, opened 2024-09-05
Deprecation notices

T4rk1n:
Add deprecation notices:

- Deprecate `Dash.run_server`, use `Dash.run` instead.
- Deprecate dynamic component loader, used by old libraries (Created during early dash 0 version, before the component generator). These components are loaded directly from the metadata json and the classes are created at runtime, preventing the use of auto complete and upcoming typing for the components.


gvwilson (approved):

ndrezn (approved):
💃

PR #2987, opened 2024-09-05
Fix multioutput requiring same number of no_update.

### LINKED ISSUE #2986: [REGRESSION] Number of no_output must match number of outputs in dash==2.18.0
In `dash==2.18.0`, the number of `no_update ` returned must match the number of outputs. This is not true in earlier versions of Dash.

This can be resolved by making `no_update` match but is a regression.

Consider:
```python
import dash
from dash import html, dcc, Input, Output, no_update

# Initialize the Dash app
app = dash.Dash(__name__)

# Define the layout
app.layout = html.Div(
    [
        dcc.Input(id="input-box", type="text", value=""),
        html.Button("Submit", id="button"),
        html.Div(id="output-1", children="Output 1 will be displayed here"),
        html.Div(id="output-2", children="Output 2 will be displayed here"),
    ]
)


# Callback with two outputs
@app.callback(
    Output("output-1", "children"),
    Output("output-2", "children"),
    Input("button", "n_clicks"),
    Input("input-box", "value"),
)
def update_outputs(n_clicks, value):
    if n_clicks is None:
        return no_update
    return "Hello", "world!"


# Run the app
if __name__ == "__main__":
    app.run_server(debug=True)
```

Dash 2.18
Screenshot 2024-09-05 at 1 37 26 PM

Dash 2.17
Screenshot 2024-09-05 at 1 37 44 PM

T4rk1n:
Fixes #2986

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [UNRELEASED]
 
+## Fixed
+
+- [#2987](https://github.com/plotly/dash/pull/2987) Fix multioutput requiring same number of no_update. Fixes [#2986](https://github.com/plotly/dash/issues/2986)
+
 ## Deprecated
 
 - [#2985](https://github.com/plotly/dash/pull/2985) Deprecate dynamic component loader.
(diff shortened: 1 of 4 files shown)


gvwilson (approved):

ndrezn (approved):
Nice test 😉

PR #3011, opened 2024-09-20
Fix #3010 Assigning None to exact or shape array properties causes exception

### LINKED ISSUE #3010: [BUG] Assigning None to exact or shape array properties causes exception
For array-type properties where individual elements are of type `exact` or `shape`, and each element accepts component-type sub-properties, assigning `None` to this array property can lead to some vague error messages. For example, with `dcc.Dropdown`:

image

```python
import dash
from dash import html, dcc

app = dash.Dash(__name__)

app.layout = html.Div(
    [
        dcc.Dropdown(
            placeholder="Currently, there are no selectable options", options=None
        )
    ],
    style={"padding": 50},
)

if __name__ == "__main__":
    app.run(debug=True)

```

CNFeffery:
For array-type properties where individual elements are of type exact or shape, and each element accepts component-type sub-properties, assigning None to this array property can lead to an exception error, closes #3010

### DIFF
--- a/@plotly/dash-test-components/src/components/ArrayOfExactOrShapeWithNodePropAssignNone.js
+++ b/@plotly/dash-test-components/src/components/ArrayOfExactOrShapeWithNodePropAssignNone.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const ArrayOfExactOrShapeWithNodePropAssignNone = (props) => {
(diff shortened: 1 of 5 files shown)


CNFeffery:
run `pytest components\dash-table\tests\selenium\test_basic_operations.py` got no error in my local tests:

image

T4rk1n (approved):
:woman_dancing: Nice catch. thank you.

Issue #2804, opened 2024-03-20
The `infer_jupyter_proxy_config` method does not work due to timeout


rickymagner:
**Describe your context**
Please provide us your environment, so we can easily reproduce the issue.

-  replace the result of `pip list | grep dash` below
```
dash                 2.16.1
dash-core-components 2.0.0
dash-html-components 2.0.0
dash-table           5.0.0
jupyter-dash         0.4.2
```
-  if frontend related, tell us your Browser, Version and OS

    - OS: MacOS 14.4
    - Browser: Chrome
    - Version: 122.0.6261.129

**Describe the bug**

This use case might be a little niche, but I'm trying to use Dash inside a notebook hosted through a web service called [Terra](app.terra.bio). The notebook is hosted via a proxy, and in the past a hack (discovered by a colleague) was able to get it to work using `JupyterDash`. However, it looks like the integrated solution in Dash doesn't work here due to a timeout. In particular, running:
```
import dash

dash.jupyter_dash.infer_jupyter_proxy_config()
```
produces the error:
```
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Cell In[9], line 1
----> 1 dash.jupyter_dash.infer_jupyter_proxy_config()

File ~/.local/lib/python3.10/site-packages/dash/_jupyter.py:205, in JupyterDash.infer_jupyter_proxy_config(self)
    203     return
    204 # Assume classic notebook or JupyterLab
--> 205 _request_jupyter_config()

File ~/.local/lib/python3.10/site-packages/dash/_jupyter.py:137, in _request_jupyter_config(timeout)
    134 while True:
    135     if (time.time() - t0) > timeout:
    136         # give up
--> 137         raise EnvironmentError(
    138             "Unable to communicate with the jupyter_dash notebook or JupyterLab \n"
    139             "extension required to infer Jupyter configuration."
    140         )
    141     if _jupyter_comm_response_received():
    142         break

OSError: Unable to communicate with the jupyter_dash notebook or JupyterLab 
extension required to infer Jupyter configuration.
```

I tried cloning the repo and editing the hard-coded timeout (at 2 seconds), but couldn't get the package to install with `pip install .`. (If anyone could provide some instructions on getting this to work I could try testing this out.)

**Further Details**

Previously, the "hack" to get this to work in the past went as follows:

```
## Run these two then comment and restart the kernel
# ! pip install dash
# ! pip install jupyter_dash

import os
[...]

Issue #2998, opened 2024-09-12
[BUG] Jupyter Dash infer_jupyter_config_proxy() running forever


Khagesh16:
Thank you so much for helping improve the quality of Dash!

We do our best to catch bugs during the release process, but we rely on your help to find the ones that slip through.

**Describe your context**

We are hosting jupyterlab on eks cluster with jupyterhub integration for the analytical purposes. In our lab environment we have python kernel with all necessary visualization libs. Dash was working fine previously with separate jupyter-dash installation but now its not working.  

Based on the documentation we followed this step 

<img width="621" alt="image" src="https://github.com/user-attachments/assets/efe221ab-63ec-4060-ad84-5bb8e079f0fa">

-  replace the result of `pip list | grep dash` below
```
dash                                    2.18.0
dash-bootstrap-components               1.3.1
dash_colorscales                        0.0.4
dash-core-components                    2.0.0
dash_daq                                0.5.0
dash-html-components                    2.0.0
dash-table                              5.0.0
```

Jupyter configuration 

```
IPython          : 8.27.0
ipykernel        : 6.29.5
ipywidgets       : 8.1.5
jupyter_client   : 8.6.2
jupyter_core     : 5.7.2
jupyter_server   : 2.14.2
jupyterlab       : 4.2.5
nbclient         : 0.10.0
nbconvert        : 7.16.4
nbformat         : 5.10.4
notebook         : 7.2.2
qtconsole        : not installed
traitlets        : 5.14.3```


-  if frontend related, tell us your Browser, Version and OS

    - OS: [e.g. iOS]
    - Browser [e.g. chrome, safari]-chrome
    - Version [e.g. 22]

**Describe the bug**

Based on the investigation probably this function is causing the issue.[here](https://github.com/plotly/dash/blob/9323c2893755f7ad4532022d8c5db930d11a5c82/dash/_jupyter.py#L135). I run this code blocked yesterday and its still running in my remote sandbox. As mentioned earlier hosted on eks pod.
```
from dash import Dash, html, dcc
import plotly.express as px
import pandas as pd

from dash import jupyter_dash
jupyter_dash.infer_jupyter_proxy_config()
```


**Expected behavior**

`jupyter_dash.infer_jupyter_proxy_config()` should infer the proxy config and let the dash render the plots inline.  

**Screenshots**

<img width="569" alt="image" src="https://github.com/user-attachments/assets/5582c5c0-8d8c-463a-b90d-355a70ab1915">

[...]

PR #3028, opened 2024-10-08
Allow jupyterlab v4 in dependency constraints

mthiboust:
Fixes https://github.com/plotly/dash/issues/2804 and https://github.com/plotly/dash/issues/2998

The current `dash` extension for `jupyterlab` cannot be installed with `jupyerlab v4`. In practice, it means that the backend is not able to communicate with the juperlab frontend to retrieve the base URL when using the `jupyter_dash.infer_jupyter_proxy_config()` method. I made it work by simply changing the dependency constraints and it is now working.

### DIFF
--- a/@plotly/dash-jupyterlab/package.json
+++ b/@plotly/dash-jupyterlab/package.json
@@ -33,9 +33,9 @@
     "watch": "tsc -w"
   },
   "dependencies": {
-    "@jupyterlab/application": "^2.0.0 || ^3.0.0",
-    "@jupyterlab/notebook": "^2.0.0 || ^3.0.0",
-    "@jupyterlab/console": "^2.0.0 || ^3.0.0"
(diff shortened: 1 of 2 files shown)


gvwilson:
@marthacryan @T4rk1n does this overlap with ongoing work in Plotly.py / Plotly.js on Jupyter support?

T4rk1n:
@gvwilson No, this is for fixing `infer_jupyter_proxy_config` for hosted jupyter notebooks.

@mthiboust Look like there is some conflict with the new version and typescript, try updating typescript in the devDependencies.

mthiboust:
> @mthiboust Look like there is some conflict with the new version and typescript, try updating typescript in the devDependencies.

Should be ok now

T4rk1n (approved):
:woman_dancing: Looks good thank you.

PR #3029, opened 2024-10-08
Add hooks

T4rk1n:
Add `dash.hooks` system to enhance created apps with functionality. These hooks can be automatically added to apps without imports by adding a `dash` entry to `entry_points` in  `setup.py`.

## Hooks:

- `layout`: wraps the layout before serving. Take the layout as first argument and the return value is used as the new layout.
```python
from dash  import hooks, html

@hooks.layout
def hook_layout(layout):
  return [html.Div("hooked")] + layout
```
- `setup`: Called when a `Dash.__init__` is called, used to get a reference to the app.
- `callback`: Add a callback to all the apps. Same signature as `dash.callback`.
- `route`: Add a route to the flask server, the route will be prefixed with `routes_pathname_prefix`.
- `error`: Add a global error handler to be used with callbacks.


dash/_hooks.py
```diff
@@ -0,0 +1,93 @@
...
+    "layout": [],
+    "routes": [],
+    "error": [],
+    "callback": [],
+}
+
+
+def layout(func):
+    """
+    Run a function when serving the layout, the return value
```
gvwilson, line 14: This looks like it saves the function `func` but doesn't run it - is the comment incorrect?

dash/_hooks.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def route(name=None, methods=("GET",)):
+    """
+    Add a route to the Dash server.
+    """
+
+    def wrap(func):
+        _name = name or func.__name__
+        _ns["routes"].append((_name, func, methods))
```
gvwilson, line 36: hm - so some of the `_ns` entries are functions and some are tuples, so the structure of entries in `_ns["key"]` depends on the value of `"key"`?

dash/_hooks.py
```diff
@@ -0,0 +1,93 @@
...
+        return func
+
+    return wrap
+
+
+def error(func: _t.Callable[[Exception], _t.Any]):
```
gvwilson, line 42: you're declaring the type of `func` here but you don't declare types in the previous registration functions - declare them all for consistency?

dash/_hooks.py
```diff
@@ -0,0 +1,93 @@
...
+    return func
+
+
+def callback(*args, **kwargs):
+    """
+    Add a callback to all the apps with the hook installed.
+    """
+
+    def wrap(func):
+        _ns["callback"].append((list(args), dict(kwargs), func))
```
gvwilson, line 54: again, it looks like the structure of values stored in `_ns` depends on the key, which feels like it should be documented somewhere in this file to help the next person reading the code.

dash/_hooks.py
```diff
@@ -0,0 +1,93 @@
...
+        return func
+
+    return wrap
+
+
+class HooksManager:
+    _registered = False
```
gvwilson, line 61: please add a comment explaining what `_registered` is for - in particular, why is it a class-level variable instead of an instance variable?

dash/_hooks.py
```diff
@@ -0,0 +1,93 @@
...
+    @staticmethod
+    def get_hooks(hook: str):
+        return _ns.get(hook, []).copy()
+
+    @classmethod
+    def register_setuptools(cls):
+        if cls._registered:
+            return
+
+        import importlib.metadata  # pylint: disable=import-outside-toplevel
```
gvwilson, line 87: why not import this at the top of the file?

dash/_hooks.py
```diff
@@ -0,0 +1,93 @@
...
+
+        for dist in importlib.metadata.distributions():
+            for entry in dist.entry_points:
+                if entry.group != "dash":
```
gvwilson, line 91: hm - hard-coded string embedded in the file - define as a constant at the top of the file to make it easier to find? and a comment here explaining what this filtering is doing would be welcome - I don't really understand it.

dash/dash.py
```diff
@@ -573,6 +573,8 @@ def __init__(  # pylint: disable=too-many-statements
             for plugin in plugins:
                 plugin.plug(self)
 
+        self._setup_hooks()
+
         # tracks internally if a function already handled at least one request.
```
gvwilson, line 578: why do we need/want this?

dash/dash.py
```diff
@@ -588,6 +590,25 @@ def __init__(  # pylint: disable=too-many-statements
...
         self.setup_startup_routes()
 
+    def _setup_hooks(self):
+        # pylint: disable=import-outside-toplevel
+        from ._hooks import HooksManager
+
+        self._hooks = HooksManager
+        self._hooks.register_setuptools()
+
+        for setup in self._hooks.get_hooks("setup"):
```
gvwilson, line 600: is this where the difference in structure between the things that are stored in different `_ns` entries shows up? this loop is getting one value per list item; the loop below is getting three.

gvwilson (commented):
some questions - happy to chat

dash/_hooks.py
T4rk1n: No, it is only run `when` serving the layout, the docstring are placeholder, we can improve the wording.

dash/_hooks.py
T4rk1n: There is some issue with cyclic dependencies with the other types, I'll see what I can do.

dash/_hooks.py
T4rk1n: The `_ns` just saves the info for the different hooks, it isn't meant for public usage. For routes/callbacks it needs more arguments so it's saved in a tuple for easy unpacking.

dash/_hooks.py
T4rk1n: This just a filter on the `entry_points` of a `setup.py`, the entry_points are used to add cli and other functionalities to python packages. I usually don't define a variable if it's only used one time.

dash/dash.py
T4rk1n: Yes, it's the setup that load the plugins and add the setup/callback/error hooks.

dash/_hooks.py
```diff
@@ -0,0 +1,205 @@
...
+        hks.append(_Hook(func, priority=priority, data=data))
+        self._ns[hook] = sorted(hks, reverse=True, key=lambda h: h.priority)
+
+    def get_hooks(self, hook: str) -> _t.List[_Hook]:
+        final = self._finals.get(hook, None)
+        if final:
+            final = [final]
+        else:
+            final = []
+        return self._ns.get(hook, []) + final
```
LiamConnors, line 71: Should a user of `hooks` call these two functions directly? If so, what would be the reason to do that?

T4rk1n: No, they should not be called by the users directly.

dash/_hooks.py
```diff
@@ -0,0 +1,222 @@
+import typing as _t
```
gvwilson, line 1: just curious: why this rather than importing the specific things you need from `typing` (e.g., `from typing import TypeVar`)?

dash/_hooks.py
```diff
@@ -0,0 +1,222 @@
...
+    LayoutType = None
+    ComponentType = None
+    Dash = None
+
+
+HookDataType = _tx.TypeVar("HookDataType")
+
+
+# pylint: disable=too-few-public-methods
+class _Hook(_tx.Generic[HookDataType]):
```
gvwilson, line 28: docstrings? or do we not document internal classes like this?

dash/_hooks.py
```diff
@@ -0,0 +1,222 @@
...
+    def __init__(self) -> None:
+        self._ns = {
+            "setup": [],
+            "layout": [],
+            "routes": [],
+            "error": [],
+            "callback": [],
+            "index": [],
+        }
+        self._js_dist = []
```
gvwilson, line 49: what is `_dist` short for?

dash/_hooks.py
```diff
@@ -0,0 +1,222 @@
...
+        self._css_dist = []
+        self._clientside_callbacks: _t.List[
+            _t.Tuple[ClientsideFuncType, _t.Any, _t.Any]
+        ] = []
+        self._finals = {}
+
+    def add_hook(
+        self, hook: str, func: _t.Callable, priority=None, final=False, data=None
```
gvwilson, line 57: no types for `priority`, `final`, and `data`?

dash/_hooks.py
```diff
@@ -0,0 +1,222 @@
...
+        if not priority and len(hks):
+            priority_max = max(h.priority for h in hks)
+            priority = priority_max - 1
+        elif not priority:
+            priority = 0
+        hks.append(_Hook(func, priority=priority, data=data))
+        self._ns[hook] = sorted(hks, reverse=True, key=lambda h: h.priority)
+
+    def get_hooks(self, hook: str) -> _t.List[_Hook]:
+        final = self._finals.get(hook, None)
```
gvwilson, line 75: is this just to ensure that `final` is the last hook in this list? if so, maybe a comment to that effect?

dash/_hooks.py
```diff
@@ -0,0 +1,222 @@
...
+
+        def _wrap(func: _t.Callable[[LayoutType], LayoutType]):
+            self.add_hook("layout", func, priority=priority, final=final)
+            return func
+
+        return _wrap
+
+    def setup(self, priority: _t.Optional[int] = None, final: bool = False):
+        """
+        Can be used to get a reference to the app after it is instantiated.
```
gvwilson, line 96: if the hook's purpose is to get the app, why is the hook called `setup`?

dash/_hooks.py
```diff
@@ -0,0 +1,222 @@
...
+                priority=priority,
+                final=final,
+                data=dict(name=_name, methods=methods),
+            )
+            return func
+
+        return wrap
+
+    def error(self, priority=None, final=False):
+        """Automatically add an error handler to the dash app."""
```
gvwilson, line 130: line break after/before `"""` (see line 139-141 below)

dash/_hooks.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def script(self, distribution: _t.List[ResourceType]):
+        """Add js scripts to the page."""
+        self._js_dist.extend(distribution)
+
+    def stylesheet(self, distribution: _t.List[ResourceType]):
+        """Add stylesheets to the page."""
+        self._css_dist.extend(distribution)
+
+    def index(self, priority=None, final=False):
```
gvwilson, line 171: I don't know what "the index" means in this context?

dash/_hooks.py
gvwilson: see above

dash/dash.py
```diff
@@ -588,6 +590,38 @@ def __init__(  # pylint: disable=too-many-statements
...
+        for hook in self._hooks.get_hooks("callback"):
+            callback_args, callback_kwargs = hook.data
+            self.callback(*callback_args, **callback_kwargs)(hook.func)
+
+        for (
+            clientside_function,
+            args,
+            kwargs,
+        ) in self._hooks.hooks._clientside_callbacks:
+            _callback.register_clientside_callback(
```
gvwilson, line 612: nice and clean

dash/resources.py
```diff
@@ -2,20 +2,53 @@
...
 from .development.base_component import ComponentRegistry
 from . import exceptions
 
 
+# ResourceType has `async` key, use the init form to be able to provide it.
+ResourceType = _tx.TypedDict(
+    "ResourceType",
+    {
+        "namespace": str,
+        "async": _t.Union[bool, _t.Literal["eager", "lazy"]],
```
gvwilson, line 18: so legal values are `True`, `False`, `"eager"`, and `"lazy"`? Are these documented somewhere?

tests/integration/test_hooks.py
```diff
@@ -0,0 +1,188 @@
+from flask import jsonify
```
gvwilson, line 1: nice

gvwilson (approved):
a few questions for my own curiosity but I think this is ready to merge

dash/_hooks.py
T4rk1n: It's to setup the app, called during app setup. Can be used to automatically setup flask extensions, get a reference to the app, add flask blueprint/static routes, etc.

dash/_hooks.py
T4rk1n: The index is the raw html of the served dash index (like index.html served for '/').

dash/resources.py
T4rk1n: Yes those are the valid values, along with None/undefined for all of them properties since `total=False` set them as optional.
This is mostly for usage of `async` chunks of components libraries (dcc.Graph, dcc.Markdown, dash_table.DataTable). The usual value is `True` or undefined, eager or lazy are abstractions that are not very well documented, it is commented here:
https://github.com/plotly/dash/blob/d26c9d37c4b50b001588ff6b2862baf0c66dbe85/dash/resources.py#L66-L72

With `Dash(eager_loading=True)`, the `async: "eager"` chunks normally loaded when the component mounts, would be loaded on page load.

PR #3034, opened 2024-10-11
Remove whitespace in metadata.json

alexcjohnson:
Any reason we need `metadata.json` files to be human-readable? In some cases this is a large fraction of the total package size, eg https://github.com/snehilvj/dash-mantine-components/pull/347 (cc @AnnMarieW)

### optionals


T4rk1n (approved):
:woman_dancing:
The metadata.json is only kept as a relic of the past runtime component generator/loader.
I think the only reason to include it in the package is for tooling interactions, like if we had a layout editor it could use the data to list the components and have a named prop editor with the right types.

alexcjohnson:
And perhaps for runtime typechecking of TS components, right?

T4rk1n:
> And perhaps for runtime typechecking of TS components, right?

No I think it should be generated in a `.js` file at generation time to have less overhead. The check is basically just `window['componentlib'].component.propTypes(props, typeSpecName, componentName, ...)` so the js file can set the propType for the components on the window.

AnnMarieW:
So we should just delete the `metadata.json` file before deployment?

PR #3035, opened 2024-10-11
add prop types to tsx components

### LINKED ISSUE #3014: [BUG] No runtime type checking with TypeScript components
While working with the `dash-mantine-components`  library, I noticed that no helpful error messages are provided in the browser DevTools when an invalid prop is passed to a dmc component -- unlike with  other Dash libraries like `dash-core-components.`

The issue arises because dmc components are built using TypeScript.  They do not use  `propTypes`  which is how the runtime type checking is done in Dash.  TypeScript is not available at runtime.  

Here is an example:
```python
from dash import Dash, dcc,_dash_renderer
_dash_renderer._set_react_version("18.2.0")

app = Dash()

app.layout = dmc.MantineProvider([
    dmc.RangeSlider(min="b"),
    dcc.RangeSlider(min="b")
])

if __name__ == "__main__":
    app.run(debug=True)
```
The first very helpful error message is for the dcc component (which does not use TypeScript), the second is the dmc component:

image

After discussing with the Plotly team,  a solution could be to create a script that automatically adds `propTypes` to TypeScript components when running with `debug=True`.  This script could use information from the `metadata.json` file, which already has details about the props and their types.

T4rk1n:
Fix #3014 

We use the `propTypes` defined on the react components to provide runtime prop validation on dash components. This property is missing from typescript components as they don't define that attribute, this PR generate propTypes for the components using the extracted metadata and saves it in a `proptypes.js` in the package directory. This file should be added as a dev dependencies in the package  `__init__.py` `_js_dist` like this:

```python
_js_dist = [
...,
"dev_package_path": "proptypes.js",
"namespace": "package"
]
```


alexcjohnson:
Ah ok - so after that can we remove `metadata.json` from the packages entierly?

T4rk1n:
> Ah ok - so after that can we remove metadata.json from the packages entierly?

It hasn't been needed since the components are generated in files, think it was only kept around in the libraries after for compatibilities with older version. The runtime component loader will be removed too. There might be some usage with docs.

gvwilson (approved):

gvwilson:
@T4rk1n do you want this one in the release or save it for after?

T4rk1n:
Yes this shouldl go into dash 3.0 since it requires components dev to regenerate it would go at the same as the typing generation.

PR #3066, opened 2024-11-07
Improve performance of context components re-rendering

### LINKED ISSUE #3057: Serious performance issues related to React context
When using components associated with the `XxxProvider`, severe performance issues can arise when there is a large amount of page content. Here are some examples related to well-known component libraries in the Dash ecosystem:

- with `dmc`

In `dmc`, it is required that the application be wrapped inside the `MantineProvider`. With the React Developer Tools, you can see that any interaction with an internal component will trigger a **re-render** of all components on the current page.
Image
```python
from dash import Dash, _dash_renderer

_dash_renderer._set_react_version("18.2.0")

app = Dash(external_stylesheets=dmc.styles.ALL)

app.layout = dmc.MantineProvider([dmc.Button("test", style={"margin": 5})] * 200)

if __name__ == "__main__":
    app.run(debug=True)
```
Even placing components from `dcc` under the `MantineProvider` will cause the same issue:
Image
```python
from dash import Dash, _dash_renderer, dcc

_dash_renderer._set_react_version("18.2.0")

app = Dash(external_stylesheets=dmc.styles.ALL)

app.layout = dmc.MantineProvider([dcc.Input(style={"margin": 5})] * 200)

if __name__ == "__main__":
    app.run(debug=True)
```

- with `fac`

In [fac](https://github.com/CNFeffery/feffery-antd-components), the similar component `AntdConfigProvider` is not a must-use, but the same issue will also occur:
Image
```python
import dash
from dash import html
import feffery_antd_components as fac

app = dash.Dash(__name__)

app.layout = html.Div(
    fac.AntdConfigProvider(
        [fac.AntdButton("test", type="primary", style={"margin": 5})] * 100
    )
)

if __name__ == "__main__":
    app.run(debug=True)
```
---

However, the issue of global re-rendering does not occur with components within `html`, such as for `html.Div` (which has the functionality to update the click event to the component's `n_clicks` property):

- with `dmc`

```python
from dash import Dash, _dash_renderer, html

_dash_renderer._set_react_version("18.2.0")

app = Dash(external_stylesheets=dmc.styles.ALL)

app.layout = dmc.MantineProvider(
    [html.Div(style={"height": 25, "border": "1px solid black", "marginBottom": 5})]
    * 100
)

if __name__ == "__main__":
    app.run(debug=True)
```
- with `fac`
```python
import dash
from dash import html
import feffery_antd_components as fac

app = dash.Dash(__name__)

app.layout = html.Div(
    fac.AntdConfigProvider(
        [html.Div(style={"height": 25, "border": "1px solid black", "marginBottom": 5})]
        * 100
    )
)

if __name__ == "__main__":
    app.run(debug=True)
```
I hope to receive more help on this issue, to explore the deeper reasons and possible solutions.

T4rk1n:
Fix #3057

- Refactor setProps reducer to partially updates the layout instead of returning a new layout object.
- New rendering component, DashWrapper replace TreeContainer.
- Remove DashContext, the state is now handled in react-redux selectors.

Gif showing only the clicked button is re-rendered:
single-update


T4rk1n:
The test_redraw shows failure with 5 redraw instead of 2.

There is two cases of additional redraw that need to be fixed in this pr:
- the `loading_state` selector get triggered. Adding +2 redraw.
- While the props is the same when returning, the equality check fails for props since it's a new object. Adding +1 redraw.
  - Checking all the props all the time is too expensive.
  - Shallow check will fail for objects that might be simple or complex.

gvwilson (approved):

dash/dash-renderer/src/APIController.react.js
```diff
@@ -97,46 +82,26 @@ const UnconnectedContainer = props => {
...
         content = (
-            <DashContext.Provider value={provider.current}>
+            <>
                 {Array.isArray(layout) ? (
                     layout.map((c, i) =>
                         isSimpleComponent(c) ? (
                             c
                         ) : (
-                            <TreeContainer
+                            <DashWrapper
```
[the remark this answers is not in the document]
T4rk1n: No, it shouldn't as this is an internal component.

dash/dash-renderer/src/config.ts
```diff
@@ -1,4 +1,4 @@
-type Config = {
+export type DashConfig = {
```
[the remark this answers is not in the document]
T4rk1n: It is used in the new DashWrapper.tsx.

T4rk1n:
I think for the additional redraw might be coming from the new path, before it was `JSON.stringify(path)` and had a memo on that. Think it might need to be back like that or with an additional equalityFn on the memo for the path where it does JSON.stringify to compare the paths.

NOTMEE12:
Hi. Are there any updates on this?

AnnMarieW:
@T4rk1n 

I ran the dmc-docs using this branch, and it's much faster!  Thanks so much for doing this PR. This will make large apps using DMC perform much better.

I don't know if it's related to changes in this PR, but it seems like the DashIconify library is no longer compatible.  Do you know what's causing this error?  (It works fine with the current Dash version )

```python
from dash import Dash, _dash_renderer
_dash_renderer._set_react_version("18.2.0")
from dash_iconify import DashIconify
app = Dash(external_stylesheets=dmc.styles.ALL)



app.layout = dmc.MantineProvider(
    dmc.Checkbox(
            label="Custom checked icon",
            checked=True,
            icon=DashIconify(icon="ion:bag-check-sharp"),
            size="lg",
        )
)

if __name__ == "__main__":
    app.run(debug=True)


```

Here's one of the error messages (There are a few more as well)

```
Uncaught TypeError: Cannot read properties of undefined (reading 'props')
    at n.47474.s.default.createElement.r.icon (dash_mantine_components.v0_15_1m1733503561.js:2:1076461)
    at renderWithHooks (react-dom@18.v2_18_2m1733760928.2.0.js:16315:20)
    at mountIndeterminateComponent (react-dom@18.v2_18_2m1733760928.2.0.js:20084:15)
    at beginWork (react-dom@18.v2_18_2m1733760928.2.0.js:21597:18)
    at HTMLUnknownElement.callCallback (react-dom@18.v2_18_2m1733760928.2.0.js:4151:16)
    at Object.invokeGuardedCallbackDev (react-dom@18.v2_18_2m1733760928.2.0.js:4200:18)
    at invokeGuardedCallback (react-dom@18.v2_18_2m1733760928.2.0.js:4264:33)
    at beginWork$1 (react-dom@18.v2_18_2m1733760928.2.0.js:27461:9)
    at performUnitOfWork (react-dom@18.v2_18_2m1733760928.2.0.js:26567:14)
    at workLoopSync (react-dom@18.v2_18_2m1733760928.2.0.js:26476:7)

```

T4rk1n:
@AnnMarieW I removed the `_dashprivate_layout` from the given props, it's a "private" props and shouldn't have been used outside of the renderer.

https://github.com/snehilvj/dash-mantine-components/blob/529231820bc2ca25db396cb35d637618b6df62cc/src/ts/components/core/checkbox/Checkbox.tsx#L66

AnnMarieW:
@T4rk1n Thanks for your speedy response.  I understand that we shouldn't have been using `_dashprivate` anything.  Do you have any suggestions for a workaround?  

Dash Bootstrap Components uses this too, and I would like to give Tom a headsup and/or do a PR to fix it there too.

alexcjohnson:
There's one other user of `_dashprivate_layout` in this repo (`dcc.ConfirmDialogProvider`), and it's also used by both DDK and DBE https://github.com/search?q=org%3Aplotly+_dashprivate_layout&type=code, in addition to [DBC](https://github.com/search?q=repo%3Afacultyai%2Fdash-bootstrap-components%20_dashprivate_layout&type=code) and as @AnnMarieW points out DMC. @T4rk1n is removing this necessary to the performance improvements? If not, it seems like we really shouldn't tamper with it. But if it is necessary, we'll need to come up with an alternate pattern that will work for our own uses of it, and then give these other libraries time and assistance to adopt the pattern as well.

AnnMarieW:
@T4rk1n 

Thanks for the suggestion of using context instead.  We are using the `dashprivate` props in multiple components in DMC, so this will take a while to update.  

You make a good point here:
>  contains the whole layout starting from the component, duplicated for each components the memory footprint is non-negligible for just a few usecases.

Given that this PR substantially increases performance, would you consider keeping the `dashprivate` props for now, and removing  in a future update?

T4rk1n:
> Given that this PR substantially increases performance, would you consider keeping the dashprivate props for now, and removing in a future update?

Yes, I'll put it back since it's used everywhere. There could be an alternative `get_props` on the clientside that would return the raw props in a future update before deprecating and removing this dashprivate api.

AnnMarieW:
This would be awesome:
> There could be an alternative get_props on the clientside that would return the raw props in a future update before deprecating and removing this dashprivate api.

 :heart:  :heart:

T4rk1n:
Not sure I can add back those props in the way it was, there was another component in between the TreeContainer and the library component that is not longer the case. Might need to refactor the new Wrapper to have a middle component but that changes the order of renders and is not as optimal as a single component since it had overhead.

T4rk1n:
And I think now with this to access the new props the new path would be just without a level and the `_dashprivate_layout.props`.
So instead of `child.props._dashprivate_layout.props` you would use `child.props` directly.

AnnMarieW:
> so the old hack cannot work anymore with this solution.

Is there a "new hack" that might be easier than using context?

T4rk1n:
> > so the old hack cannot work anymore with this solution.
> 
> Is there a "new hack" that might be easier than using context?

A Context solution might only be valid in some cases, for the mantine checkbox case 0it want to add extra props to the component.

I see a couple solution for those cases:
1. Could add `extraProps` to DashWrapper props that would be merged with the component props from the store, this would make `React.cloneElement(child, {extraProps: {selected: true}})` to work.
2. change the clientside `set_props` to accept a path instead of an id. Then you can use `children[0]._dashprivate_path` to add new props to the components.

For accessing the props (dcc.Tabs cases), a new `get_props(path)` could be added or refactor the components to use a context.

T4rk1n:
@AnnMarieW I added all three solutions from the last post.

1. Added extras prop to `DashWrapper` component, think this one could replace dmc checkbox:
```js
const iconFunc = ({ indeterminate, ...others }) => {
    const selected: any = indeterminate ? indeterminateIcon : icon;
    return React.cloneElement(selected, {extras: others});
};
```
2. `set_props` can take a dash path instead of id.
3. Added `get_props(pathOrId)`, to get the props of a component. The path is available in `children[0].props.componentPath`.

Let me know if that work.

AnnMarieW:
Thanks for the update @T4rk1n   I'll try to figure out how to use these new features.

Will it handle this pattern too?  https://github.com/snehilvj/dash-mantine-components/blob/83a1cc12e5e6b210b7a1e27c7a83068abe2830c5/src/ts/components/core/timeline/Timeline.tsx#L46

T4rk1n:
> Thanks for the update @T4rk1n I'll try to figure out how to use these new features.
> 
> Will it handle this pattern too? https://github.com/snehilvj/dash-mantine-components/blob/83a1cc12e5e6b210b7a1e27c7a83068abe2830c5/src/ts/components/core/timeline/Timeline.tsx#L46

For that one can use `window.dash_clientside.get_props(child.props.componentProps)`

AnnMarieW:
@T4rk1n 

Do you know how  can get the component type as used here in in this [Stepper component](https://github.com/snehilvj/dash-mantine-components/blob/master/src/ts/components/core/stepper/Stepper.tsx#L69)

```
 const childType = child.props._dashprivate_layout.type;
 if (childType === "StepperCompleted") {
      ....
                   
```

BSd3v:
@T4rk1n,

Hmm... would it be possible to change it from `get_props` to `get_attributes`? Where it could return `type, namespace, props` and if you wanted to you, you could target `props` just like how you are targeting the `value` in the one component?

T4rk1n:
I changed `get_props` to `get_layout` to match the previous `_dashprivate_layout` it returns type, namespace, props. Can also concat the path to get props values, eg: `[...child.props.componentPath, 'props', 'value']`

AnnMarieW:
@T4rk1n 

It looks like `dcc.Loading` is not working with this update .

Here's the first example from the docs - loading spinners don't show up

```python


from dash import Dash, dcc, html, Input, Output, callback
import time

app = Dash()

app.layout = html.Div([
    html.H3("Edit text input to see loading state"),
    html.Div("Input triggers local spinner"),
    dcc.Input(id="loading-input-1"),
    dcc.Loading(
        id="loading-1",
        type="default",
        children=html.Div(id="loading-output-1")
    ),
    html.Div([
        html.Div('Input triggers nested spinner'),
        dcc.Input(id="loading-input-2"),
        dcc.Loading(
            id="loading-2",
            children=[html.Div([html.Div(id="loading-output-2")])],
            type="circle",
        )
    ]),
])


@callback(Output("loading-output-1", "children"), Input("loading-input-1", "value"))
def input_triggers_spinner(value):
    time.sleep(1)
    return value


@callback(Output("loading-output-2", "children"), Input("loading-input-2", "value"))
def input_triggers_nested(value):
    time.sleep(1)
    return value


if __name__ == "__main__":
    app.run(debug=False)


```

emilhe:
In some cases, it is necessary to desireable (necessary?) to render _other Dash components_ beyond what is passed to the `children` property. The simple exaple is `icon` propety in dmc, where a content is to be rendered (including callback bindings) similar to the `children` property. As I see it, we need two pieces of functionality,

1) To be able to lookup a component type
2) To be able to render a component (given type and props), including callback bindings

It seems that the proposed `get_layout` function satisfies (1). I wrote `[renderDashComponents](https://github.com/emilhe/dash-extensions-js/blob/8b07ce3b61bbd3150cbf8dd6c5fb6d0204e74adf/index.js#L337)` to provide (2), but I am not sure if there are any other utility functions available providing this functionality?

As a last resort, I guess I could re-write the `renderDashComponents` function in `dash-extensions-js` to adopt the new syntax, but I would of course prefer an official, non-private API based solution if possible :)

T4rk1n:
> It looks like dcc.Loading is not working with this update .

Yes, the loading state stuff incurred two extra renders so I had merged the loading_state selector with the props but the equality function still need to be adapted or something, working on that.

T4rk1n:
> To be able to render a component (given type and props), including callback bindings

I don't think this is a good pattern, the components are already hydrated, they only need to be mounted for the rendering. Can now add props `set_props(pathOrId, payload)` or `React.cloneElement(dashComponent, {extras: {style: {background: 'red'}})`. The path would now be available at `component.props.componentPath`.

The `renderDashComponent` does not return a Wrapper or previously TreeContainer and may be disconnected from the store.

AnnMarieW:
@T4rk1n 

In [DMC PR # 458](https://github.com/snehilvj/dash-mantine-components/pull/458) I've used the new features you added to eliminate all the `_dashprivate_layout` from DMC.  Works great - Thanks! :tada:   

DMC is using Emil's `renderDashComponets` function in these [4 components](https://github.com/search?q=repo%3Asnehilvj%2Fdash-mantine-components+renderDash&type=code) in case you wanted to see a use-case.

Is it OK to continue to use `_dashprivate_pushstate` ?  We use it in [dmc.Anchor](https://github.com/snehilvj/dash-mantine-components/blob/83a1cc12e5e6b210b7a1e27c7a83068abe2830c5/src/ts/utils/anchor.ts#L59) so it works like  `dcc.Link`.      It's also used in  [these components](https://github.com/search?q=repo%3Afacultyai%2Fdash-bootstrap-components%20dashprivate_pushstate&type=code) in the Dash Bootstrap Components library.

T4rk1n:
> Is it OK to continue to use _dashprivate_pushstate ? 

@AnnMarieW Yea that is just for the location component, it shouldn't have been prefixed with `_dashprivate` and a better name but it's unlikely to change for now.

AnnMarieW:
@T4rk1n 
Does this mean that every component that adds `data-dash-is-loading` needs to be updated too?
https://github.com/plotly/dash/blob/dev/components/dash-core-components/src/fragments/Graph.react.js#L523

T4rk1n:
> @T4rk1n Does this mean that every component that adds `data-dash-is-loading` needs to be updated too? https://github.com/plotly/dash/blob/dev/components/dash-core-components/src/fragments/Graph.react.js#L523

Yes, if it's really needed, there will be `ctx.isLoading` and `ctx.useLoading()` as replacement.

AnnMarieW:
> Yes, if it's really needed, there will be ctx.isLoading and ctx.useLoading() as replacement.

Could you show an example of how to update the dcc.Graph component?  It would be good to still enabling styling a component that's loading  via CSS as described on at the end of this page:  https://dash.plotly.com/loading-states

T4rk1n:
@AnnMarieW For class component we can replace the div with `data-dash-is-loading` with a component like this one:

```
import React from 'react';

export default function LoadingDiv({children, loading_state, ...props}) {
    let loading;
    if (window.dash_component_api) {
        const ctx = window.dash_component_api.useDashContext();
        loading = ctx.useLoading();
    } else {
        loading = loading_state?.is_loading;
    }
    
    return (
        <div {...props} data-dash-is-loading={loading || undefined}>
            {children}
        </div>
    );
}
```

This would also be backward compatible with the old loading_state.

AnnMarieW:
Hi @T4rk1n 

I see the new `LoadingElement`  - that looks great!  Will it work with elements like a Mantine component rather than a standard jsx component?  It causes some problems if a Mantine component is wrapped in an extra `div` as described in [ issue  463](https://github.com/snehilvj/dash-mantine-components/issues/463)

T4rk1n:
> Hi @T4rk1n
> 
> I see the new `LoadingElement` - that looks great! Will it work with elements like a Mantine component rather than a standard jsx component? It causes some problems if a Mantine component is wrapped in an extra `div` as described in [ issue 463](https://github.com/snehilvj/dash-mantine-components/issues/463)

If the component accept `data-dash-is-loading` and pass it to it's dom node can use `const  loading = ctx.useLoading()` in function components.

insistence:
> @AnnMarieW I added all three solutions from the last post.
> 
> 1. Added extras prop to `DashWrapper` component, think this one could replace dmc checkbox:
> 
> ```js
> const iconFunc = ({ indeterminate, ...others }) => {
>     const selected: any = indeterminate ? indeterminateIcon : icon;
>     return React.cloneElement(selected, {extras: others});
> };
> ```
> 
> 2. `set_props` can take a dash path instead of id.
> 3. Added `get_props(pathOrId)`, to get the props of a component. The path is available in `children[0].props.componentPath`.
> 
> Let me know if that work.

I have a question about the `Added extras prop to DashWrapper component`. If the third-party React component itself uses React.cloneElement but does not pass the extras property, will the current change not be able to pass additional parameters to child components? Previous related PR #2819 #2831 .

T4rk1n:
@insistence 
> I have a question about the Added extras prop to DashWrapper component. 

I added the `extras` as an alternative to `dashprivate` api's, I refactored the test you added a couple months ago yesterday to use the new `extras` instead and it felt kinda wrong. I think I can do `...restProps` instead of the `extras` prop that would be better so I'll change that.

T4rk1n:
@AnnMarieW This will be ready for merge, I'd like to get the api right before and I am thinking of one last minute change. Putting `window.dash_clientside.get_layout(path)` on `window.dash_component_api.get_layout()` instead. This way would be more separated between what should be used in clientside callbacks and inside dash components. What do you think?

AnnMarieW:
> Putting window.dash_clientside.get_layout(path) on window.dash_component_api.get_layout() instead. This way would be more separated between what should be used in clientside callbacks and inside dash components. 

I agree - it would be more consistent.  Is there any use-case for the `get_layout` being included in `dash_clientside` ?  Would people use in in clientside callbacks?

components/dash-core-components/src/components/Textarea.react.js
```diff
@@ -73,6 +73,11 @@ const Textarea = ({
     );
 };
 
+Textarea.dashPersistence = {
```
AnnMarieW, line 76: Is this the recommended way of setting default for persistence?  
It will only work in Dash >=3 ?

T4rk1n: Yes, only for dash>=3 for support of the persistence props that was previously in `defaultProps` in functional components and react 18.

T4rk1n:
> Would people use it in clientside callbacks?

I don't think so, it requires the componentPath which is not something the clientside callbacks would really know about.

components/dash-core-components/src/components/Textarea.react.js
AnnMarieW: Does it work in dash< 3 to have the default set in the function signature?  If not, we should check if Dash 3.0 is being use and if not use `defaultProps`?  Is the best way to check for Dash 3.0 by seeing if `window.dash_component_api` exists?

I also I get an error when trying to set default props in the function signature when the prop is a required prop by Mantine.  This might be a separate issue, but do you know how to handle that case?

AnnMarieW:
Hey Philippe,

When doing a build based on the latest version here, I get an Error message:

>  File "/home/amw/PycharmProjects/dash/dash/development/_py_prop_typing.py", line 115, in <listcomp>
    _get_literal_value(json.loads(v["value"].replace("'", '"')))
AttributeError: 'int' object has no attribute 'replace'

There are some types defined in Mantine and imported.  An example, this `order` prop, which is an integer with values between 2-6
https://github.com/snehilvj/dash-mantine-components/blob/master/src/ts/components/core/accordion/Accordion.tsx#L33

idling-mind:
hi, need some help!
I had used the `_dashprivate_layout` in one of my components like this
[Link to repo](https://github.com/idling-mind/dash_snap_grid/blob/9d0001055257c9d133874ae43b515e5adb30593f/src/lib/components/Grid.react.js#L50)

```js
const dom = this.props.children.map((child) => {
            if (
                !child.props._dashprivate_layout ||
                !child.props._dashprivate_layout.props.id
            ) {
                throw new Error(
                    'All children of Grid must have a unique id prop.'
                );
            }
            return (
                <div key={child.props._dashprivate_layout.props.id}>
                    {child}
                </div>
            );
        });
```
What would have been a better alternative? With the new release, how should I change this?

AnnMarieW:
Hi @idling-mind 

Check out the new docs for more info. 
https://dash.plotly.com/dash-3-for-component-developers

 You can get the props using the new `dash_components_api.getLayout` function.

```
 window.dash_component_api.getLayout(child.props.componentPath).props;
```

PR #3072, opened 2024-11-11
Fallback on long callback function names if source cannot be found

### LINKED ISSUE #1885: [BUG] long_callback fails when bundling python code

corebit-nl:
Additional patch to #2922 to fully fix #1885 
- Adds patch made in #2922 to similar code, to make long callbacks fully work in compiled binaries (eg. Pyinstaller).


T4rk1n:
This has been reimplemented from another PR.

PR #3077, opened 2024-11-15
Add new parameter assets_path_ignore for dash.Dash()

### LINKED ISSUE #3076: [Feature Request] Support for ignoring specific folders under the assets directory
The existing parameter `assets_ignore` can ignore files under the assets folder that match a specified regular expression pattern, but it cannot ignore specific folders under the assets folder as a whole, which would be useful in many scenarios.

CNFeffery:
Add new parameter `assets_path_ignore` to `dash.Dash()`, for the static resource directory, set several folders that need to be ignored during the initial automatic loading of static resources.

closes #3076

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -246,6 +246,12 @@ class Dash:
         to sensitive files.
     :type assets_ignore: string
 
+    :param assets_path_ignore: A list of regex, each regex as a string to pass to ``re.compile``, for
+        assets path to omit from immediate loading. The files in these ignored paths will still be
+        served if specifically requested. You cannot use this to prevent access
+        to sensitive files.
+    :type assets_path_ignore: list of strings
+
     :param assets_external_path: an absolute URL from which to load assets.
         Use with ``serve_locally=False``. assets_external_path is joined
         with assets_url_path to determine the absolute url to the
@@ -391,6 +397,7 @@ class Dash:
         use_pages: Optional[bool] = None,
         assets_url_path: str = "assets",
         assets_ignore: str = "",
+        assets_path_ignore: List[str] = None,
         assets_external_path: Optional[str] = None,
         eager_loading: bool = False,
         include_assets_files: bool = True,
@@ -451,6 +458,7 @@ class Dash:
             ),  # type: ignore
             assets_url_path=assets_url_path,
             assets_ignore=assets_ignore,
+            assets_path_ignore=assets_path_ignore,
             assets_external_path=get_combined_config(
                 "assets_external_path", assets_external_path, ""
             ),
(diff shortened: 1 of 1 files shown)


CNFeffery:
@T4rk1n Should I resubmit a PR for this feature to adapt to the latest version of `Dash`? This feature is very useful for those who want to quickly load static resources such as `js` on demand in a `Dash` application. For example, when building an online spreadsheet feature based on [Univer](https://github.com/dream-num/univer) in `Dash`, there are requirements for the loading order of relevant `js` dependencies. By adding the `assets_path_ignore` parameter, it is convenient to ensure that the `js` in the relevant static file directory will not be automatically loaded in a disorderly manner in advance during application initialization.

T4rk1n:
> @T4rk1n Should I resubmit a PR for this feature to adapt to the latest version of `Dash`? This feature is very useful for those who want to quickly load static resources such as `js` on demand in a `Dash` application. For example, when building an online spreadsheet feature based on [Univer](https://github.com/dream-num/univer) in `Dash`, there are requirements for the loading order of relevant `js` dependencies. By adding the `assets_path_ignore` parameter, it is convenient to ensure that the `js` in the relevant static file directory will not be automatically loaded in a disorderly manner in advance during application initialization.

Yes, I want to include this in next release 3.1, up to you if you want to resolve the conflicts in this PR or a new one.

CNFeffery:
> > > Look like the format check failed:
> > > > ERROR: "private::lint.black" exited with 1.
> > > 
> > > 
> > > I changed the pre-commit hook to only lint staged files and should runs almost instantly for these cases now.
> > 
> > 
> > I can't figure out exactly what caused the code format check to fail 😢, I've used `ruff` to format related `.py` files.
> 
> That is because we use black not ruff 😆

Yes, I used to think that `ruff` was 100% compatible with the `black` standard 😣. After executing `npm run format`, the relevant lint tests have passed, but I don't know how to deal with the remaining failed test `test-312` 😢

T4rk1n:
> > > > Look like the format check failed:
> > > > > ERROR: "private::lint.black" exited with 1.
> > > > 
> > > > 
> > > > I changed the pre-commit hook to only lint staged files and should runs almost instantly for these cases now.
> > > 
> > > 
> > > I can't figure out exactly what caused the code format check to fail 😢, I've used `ruff` to format related `.py` files.
> > 
> > 
> > That is because we use black not ruff 😆
> 
> Yes, I used to think that `ruff` was 100% compatible with the `black` standard 😣. After executing `npm run format`, the relevant lint tests have passed, but I don't know how to deal with the remaining failed test `test-312` 😢

No worry, the `test-312` circleci container has been instable since the update to their v2 container. I started moving tests to github actions.

PR #3078, opened 2024-11-15
Deprecation removals

T4rk1n:


T4rk1n:
Ready for review!

emilykl:
Woohoo!! 🙌 

It looks like `app.run_server` might still be used in a handful of app templates -- @LiamConnors can you confirm? Or are these templates which are not used anymore?

https://github.com/search?q=repo%3Aplotly%2Fddk-dash-docs+run_server+language%3APython&type=code&l=Python

emilykl:
@T4rk1n What's the backstory regarding `dcc.LogoutButton`? I'm not up to date

dash/_callback.py
```diff
@@ -342,7 +344,7 @@ def register_callback(
         flat_state,
         inputs_state_indices,
         prevent_initial_call,
-        long=long,
```
emilykl, line 345: @T4rk1n What about this argument called `long=` here, is there a reason it needs to stay?

T4rk1n:
> What's the backstory regarding dcc.LogoutButton?

The button was made for DE4 but doesn't work with DE5.

emilykl (approved):
@T4rk1n You've addressed all my comments -- looks good to me. 🚀

PR #3089, opened 2024-11-21
adding support for async callbacks and page layouts

BSd3v:
This PR is attempting to allow for layouts and callbacks to be performed in a way that will allow for async functions natively.


ndrezn:
@BSd3v is there a feature ticket for this change anywhere?

PR #3092, opened 2024-11-22
Allow `npm run lint` to work through and display all errors

BSd3v:
allowing `npm run lint` to go through all linting steps providing all the errors that need to be fixed, not just the first encountered failing test


ndrezn (approved):
Chatted with Brian about this change offline -- this looks good to me. If there is a linting error it will still show as failed in CI.

💃 

@T4rk1n will give you final call

T4rk1n (approved):
:woman_dancing:

PR #3095, opened 2024-11-27
Remove werkzeug upperbound

marcstern14:
`werkzeug>=3.1` is compatible within the set flask boundaries. Removing the boundaries for `werkzeug` would allow for more flexible installations on deployed apps without triggering vulnerabilities.

### DIFF
--- a/requirements/install.txt
+++ b/requirements/install.txt
@@ -1,5 +1,5 @@
 Flask>=1.0.4,<3.1
-Werkzeug
+Werkzeug>=3.0.6
 plotly>=5.0.0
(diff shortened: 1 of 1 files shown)


ndrezn:
We've had a few issues in the past with backwards incompatibilities in werkzeug causing issues in Dash (https://github.com/plotly/dash/issues/1992) which led to the pin.

Could you open a ticket requesting that we release the upper bound for discussion? Want to make sure we get this right if we decide to change the dependency on werkzeug.

marcstern14:
Yes will do! Thanks @ndrezn

Issue #2924, opened 2024-07-18
[BUG] Layout as list of components does not work with Dash Pages page_container


antonymilne:
**Describe your context**

```
dash                       2.17.1
dash_ag_grid               31.2.0
dash-bootstrap-components  1.6.0
dash-core-components       2.0.0
dash-html-components       2.0.0
dash-mantine-components    0.12.1
dash-table                 5.0.0
dash-testing-stub          0.0.2
```

**Describe the bug**

https://github.com/plotly/dash/pull/2795 enabled you to pass a list of components to `app.layout`. However, this does not work with `page_container`.

```
import dash

from dash import Dash, html, page_container

app = Dash(__name__, use_pages=True, pages_folder="")
dash.register_page("/", layout=html.H1("Hello"))
app.layout = [page_container, html.H2("Stuff")]
# These work ok:
# app.layout = page_container
# app.layout = html.Div([page_container, html.H2("Stuff")])

app.run()
```
The first request will give the following exception:
```
Exception: `dash.page_container` not found in the layout
```
Subsequent page refreshes are ok because this check only occurs on the first request.

The problem is that the following check is failing: https://github.com/plotly/dash/blob/bbd013cd9d97d8c41700b240dd95c32e6b875998/dash/dash.py#L2257 

The reason for that is as follows:
```
assert "a" in html.Div(html.Div(id="a")) # passes
assert "a" in html.Div([html.Div(id="a")]) #passes
assert "a" in html.Div([[html.Div(id="a")]]) # FAILS - this is the relevant case here
```

As with https://github.com/plotly/dash/issues/2905 probably the easiest solution is to wrap the layout in `html.Div` in this case so that the check passes:
```
assert "a" in html.Div([html.Div([html.Div(id="a")])]) # passes
```

Note that while #2905 is very similar it looks like the proposed fix #2915 will not solve this case. Given these two very similar issues, maybe there's a better elsewhere that would fix both (and any other similar undiscovered issues) simultaneously?

ProfessionalCaddie:
Hello, I am new to open source. May I be assigned this?

kenshima4:
Hi dash team, I've opened a pull request for this. Can someone take a look please and let me know if anything needs to be changed please? #3108

PR #3108, opened 2024-12-17
changed typecheck for list value to use isinstance instead of type

kenshima4:
Solves: Layout as list of components does not work with Dash Pages page_container #2924

I've checked if the values passed to layout setter isinstance list, if so, set the values to html.Div (dash Component)

## Contributor Checklist

   -  [x] Set values passed as list to html.Div
   - [x] Changed typecheck to use isintance instead of type
- [ ] I have run the tests locally and they passed (based on automated tests). (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))

### optionals


gvwilson:
@T4rk1n this one looks good to me, but can you please check it before we merge?

kenshima4:
Hi guys, will fix the percyu, testing etc this weekend. Will revert back.

.gitignore
```diff
@@ -92,3 +92,5 @@ VERSION.txt
 !components/dash-core-components/tests/integration/upload/upload-assets/upft001.csv
 !components/dash-table/tests/assets/*.csv
 !components/dash-table/tests/selenium/assets/*.csv
+app.py
+launch.json
```
ndrezn, line 96: I'd recommend just avoiding committing these files rather than changing `.gitignore`; while unlikely we may need these file names in the future.

kenshima4:
Hi @T4rk1n, I made the changes you recommended, please check. I'd just like to ask if all the integration tests need to remain unchanged or would I need to change them based on my changes? Because I see even on dev branch the integration tests are failing (unless I'm wrong). Can check with `pytest -k test_cbcx005_grouped_clicks` from dash directory. Is this right or can I go ahead and modify these failing integration tests?

.gitignore
kenshima4: Hi @ndrezn will remove them from gitignore and make sure not to commit. Thank you.

T4rk1n (approved):
:woman_dancing: Looks good.

PR #3113, opened 2024-12-30
Stripping Input Data from Background Callback Polling

### LINKED ISSUE #3111: Background callback polling request sends all callback inputs, states, outputs each time
**Describe your context**
As far as I can tell this is a general issue I have reproduced it in Jupyterlab with Diskcache and also in a separately served app with Gunicorn and Celery.

```
dash                      2.18.2
dash_ag_grid              31.2.0
```

**Describe the bug**

When a dash background callback is executed there is one initial request which sends all of the related inputs and states to the server and then a series of polling requests that contain the caheKey identifying the background callback until it has completed. Looking at the request being sent, I think it is sending all of the callback data with each polling request instead of just sending it once with the first request and then using only the cacheKey on subsequent requests.

The impact of this behaviour is that if there is a lot of input data the polling is very slow (for example uploading with dcc.Upload and then running a longer processing and saving in a background callback). Since the polling interval gets long due to all the uploading, progress and set_props updates can be missed entirely.

**Expected behavior**

It would be preferable to make the polling as fast as possible by dropping all of the callback inputs/states and just using the cacheKey to identify it.

BSd3v:
Attempting to make polling requests not eat bandwidth

This is to fix #3111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [unreleased]
+
+## Changed
+- [#3113](https://github.com/plotly/dash/pull/3113) Adjusted background polling requests to strip the data from the request, this allows for context to flow as normal. This addresses issue [#3111](https://github.com/plotly/dash/pull/3111)
+
+
 ## [3.0.1] - 2025-03-24
 
 ## Fixed
@@ -160,10 +166,6 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3190](https://github.com/plotly/dash/pull/3190) Fix issue with cache key generation by adding option to include triggered inputs. Fixes [#3189](https://github.com/plotly/dash/issues/3189)
 - [#3130](https://github.com/plotly/dash/pull/3130) Fix HOST variable when using conda.
 - [#3066](https://github.com/plotly/dash/pull/3066) Improve performance of context components re-rendering.
-
-## Changed
-- [#3113](https://github.com/plotly/dash/pull/3113) Adjusted background polling requests to strip the data from the request, this allows for context to flow as normal. This addresses issue [#3111](https://github.com/plotly/dash/pull/3111)
-
 
 ## [2.18.2] - 2024-11-04
 


gvwilson:
@T4rk1n can you please review? thx

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -439,6 +439,7 @@ function handleServerside(
     const fetchCallback = () => {
         const headers = getCSRFHeader() as any;
         let url = `${urlBase(config)}_dash-update-component`;
+        let new_body = body;
```
T4rk1n, line 442: We use camelCase for variable that are strictly js based, usually if there is a snake_case variable in the js code it's because it's transfered back and forth with the backend or part of the clientside api.

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -447,11 +448,19 @@ function handleServerside(
...
-            addArg('cacheKey', cacheKey);
-        }
-        if (job) {
-            addArg('job', job);
+        if (cacheKey || job) {
+            if (cacheKey) addArg('cacheKey', cacheKey);
+            if (job) addArg('job', job);
+
+            // clear inputs as background callback doesnt need inputs, just verify for context
+            const tmp_body = JSON.parse(new_body);
```
T4rk1n, line 456: nitpick: I think it would be more efficient to keep body as is and instead call the stringify only in the merge on line 476.

T4rk1n (commented):
Looks good overall. I just have a question if we could just do `body = "{}"` if we have the cache key instead of setting the inputs to null, think that might require some work on the backend callback but wouldn't it be better?

BSd3v:
@T4rk1n,

This would indeed take much more effort to unwind how the background callbacks function to completely strip the need for the input structure, as this is used to verify the callback structure, as well as pull things like context (I'm assuming for using things like progress updates)

dash/dash-renderer/src/actions/callbacks.ts
BSd3v: This would make the process always trigger a parse and stringify for each call, this way makes it only necessary in the lifecycle of a background callback without adjusting regular callbacks.

T4rk1n (approved):

T4rk1n:
`table-node` failed with unrelated npm connection error, merging.

PR #3121, opened 2025-01-13
Add very basic update notifications

marthacryan:
**Warning: Very much in-progress**

This PR adds the ability to notify dash developers of new versions of dash through the developer tools. It mostly updates the errors section of the dash developer tools frontend. The only change to python is adding an additional config passed from the python side to send the dash python version to the frontend.

The notification prompts the developer to upgrade their version of dash and has 3 options for actions:
* Don't show again -> sets a boolean flag in local storage if the user clicks. 
* Remind me later -> sets a field in local storage to the time that this button is clicked. The frontend then uses that to determine whether to show the notification again.
* Skip this version -> sets a field in local storage to the version that was received from the server. The frontend then uses that to determine whether to show the notification again. 

Todos:
- [ ] Add style to UI (There is currently no style applied to the buttons / very little applied to the notification)
- [ ] Update dev tools menu options - create separate button for toggling notifications? 
- [ ] Code cleanup - clean up handling of button clicks, clean up names of directories and files that assume notifications are errors, etc. 
- [ ] Swap in actual server information for notifications

<details> <summary>A couple of sample files to help test </summary>
Server: 

```
from flask import Flask, request
from flask_cors import CORS

app = Flask(__name__)
CORS(app)
@app.route('/sample', methods=['POST'])
def index():
    # Get contents of post request
    data = request.json
    if data['dash_version'] == current_version:
        return []
    else: 
        return [{
            'message': 'A new version of dash is available! Run `pip install dash --upgrade` to upgrade.',
            'buttons': ['dont_show_again', 'remind_me_later', 'skip_this_version'],
            'version': current_version
        }]
   

if __name__ == '__main__':
    # on port 8080
    app.run(debug=True, port=8080)
```

Dash app that adds errors to show how errors fit with these notifications:

```
## generic dash app
import dash

app = dash.Dash(__name__)

app.layout = html.Div([
    dcc.Input(id='input', value='initial value', type='text'),
    html.Div(id='output')
])

@app.callback(
    Output(component_id='output', component_property='children'),
    [Input(component_id='input', component_property='value')]
)
def update_value(input_data):
    # Throw a sample error
    raise ValueError('This is a sample error')
    return 'Input: "{}"'.format(input_data)

if __name__ == '__main__':
    app.run_server(debug=True)
```
</details>

### DIFF
--- a/dash/dash-renderer/src/components/error/FrontEnd/FrontEndErrorContainer.react.js
+++ b/dash/dash-renderer/src/components/error/FrontEnd/FrontEndErrorContainer.react.js
@@ -3,15 +3,60 @@ import './FrontEndError.css';
 import PropTypes from 'prop-types';
 import {FrontEndError} from './FrontEndError.react';
 
+const DAY_IN_MS = 86400000;
+
+function compareVersions(v1, v2) {
+    const v1Parts = v1.split('.').map(Number);
+    const v2Parts = v2.split('.').map(Number);
+
+    for (let i = 0; i < Math.max(v1Parts.length, v2Parts.length); i++) {
+        const part1 = v1Parts[i] || 0;
+        const part2 = v2Parts[i] || 0;
+
+        if (part1 > part2) return 1;
+        if (part1 < part2) return -1;
+    }
+
+    return 0;
+}
+
 class FrontEndErrorContainer extends Component {
     constructor(props) {
         super(props);
+
+        const {config} = props;
+
+        if (!localStorage.getItem('noNotifications')) {
+            fetch('http://127.0.0.1:8080/sample', {
+                method: 'POST',
+                body: JSON.stringify({
+                    dash_version: config.dash_version
+                }),
+                headers: {
+                    'Content-Type': 'application/json'
+                }
+            }).then(response => {
+                response.json().then(body => {
+                    this.setState({...this.state, notificationInfo: body});
+                });
+            });
+        }
+
+        this.state = {
+            dismissed: [false * this.props.errors.length],
+            notificationInfo: []
+        };
     }
 
     render() {
-        const {errors, connected, errorsOpened, clickHandler} = this.props;
+        const {errors, errorsOpened} = this.props;
+        const {dismissed, notificationInfo} = this.state;
         const errorsLength = errors.length;
-        if (errorsLength === 0 || !errorsOpened) {
+
+        if (
+            (errorsLength === 0 || !errorsOpened) &&
+            notificationInfo.length === 0
+        ) {
             return null;
         }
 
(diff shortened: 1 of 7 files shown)


dash/dash-renderer/src/components/error/menu/DebugMenu.react.js
```diff
@@ -1,156 +1,243 @@
...
-                    className={classes('dash-debug-menu', openVariant)}
-                    onClick={() => {
-                        this.setState({opened: !opened});
-                    }}
-                >
-                    <DebugIcon
-                        className={classes('dash-debug-menu__icon', 'debug')}
-                    />
-                </div>
-                <GlobalErrorOverlay
```
T4rk1n, line 146: Is this comoponent not used anymore?

T4rk1n (commented):
Would be best to refactor the new version functionality in it's component file.
Also should add a flag in the config to disable the system entirely in an opt out value.

dash/dash-renderer/src/components/error/menu/DebugMenu.react.js
marthacryan: Ah yeah I did remove that. It had very little logic in it so I just moved it into the `DebugMenu.react.js`. The only thing I see it doing is concatenating the frontend and the backend errors, which I now do on line 148. Here's the current file: https://github.com/plotly/dash/blob/7c0318749da90424f7de83badeb66b523f6a60a5/dash/dash-renderer/src/components/error/GlobalErrorOverlay.react.js#L18

dash/dash-renderer/src/components/error/menu/DebugMenu.react.js
```diff
@@ -1,156 +1,243 @@
...
+                        newDashVersion
+                    ) ? (
+                        <button
+                            className='dash-debug-menu__upgrade-button'
+                            onClick={toggleShowUpgradeTooltip}
+                        >
+                            Dash update available - v{newDashVersion}
+                        </button>
+                    ) : null}
+                </div>
```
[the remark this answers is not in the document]
marthacryan: Good point! I moved the upgrade info (current version and if necessary the upgrade button / tooltip) into `VersionInfo.react.js`, let me know if that looks ok. 

For the other buttons, they each have a bit of custom logic to the point that I'm not sure it will save any lines of code for them to be separate. But let me know if you think otherwise.

dash/dash-renderer/src/components/error/menu/DebugMenu.react.js
```diff
@@ -1,156 +1,243 @@
...
+        ) < 0
+    ) {
+        return true;
+    } else {
+        return !lastDismissed && !lastDismissedVersion;
+    }
+}
+
+async function requestDashVersionInfo(currentDashVersion) {
+    return fetch('https://dash-version.plotly.com:8080/sample', {
```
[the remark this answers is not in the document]
marthacryan: Done!

dash/dash-renderer/src/components/error/menu/DebugMenu.react.js
```diff
@@ -1,156 +1,243 @@
...
         this.state = {
             opened: false,
-            callbackGraphOpened: false,
-            errorsOpened: true
+            popup: 'errors',
+            upgradeInfo: []
         };
+
+        // Close the upgrade tooltip if the user clicks outside of it
+        document.addEventListener('click', e => {
```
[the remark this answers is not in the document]
marthacryan: Done! Does that look ok?

dash/dash-renderer/src/components/error/menu/DebugMenu.react.js
```diff
@@ -1,156 +1,243 @@
...
+        }
+    }).then(response => response.json());
+}
 
 class DebugMenu extends Component {
     constructor(props) {
         super(props);
+        const {config} = props;
+
+        requestDashVersionInfo(config.dash_version).then(body => {
```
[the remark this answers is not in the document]
marthacryan: Done! Does that look ok? I use react state rather than redux.

marthacryan:
> Would be best to refactor the new version functionality in it's component file. Also should add a flag in the config to disable the system entirely in an opt out value.

We're still deciding how opt outs should happen so I'm going to wait on this!

dash/dash-renderer/src/components/error/menu/VersionInfo.react.js
```diff
@@ -0,0 +1,160 @@
...
+
+    return (
+        <div className='dash-debug-menu__version'>
+            {upgradeTooltipOpened ? (
+                <div className='dash-debug-menu__upgrade-tooltip'>
+                    <a
+                        target='_blank'
+                        href='https://dash.plotly.com/installation'
+                    >
+                        Read details
```
LiamConnors, line 130: Is this intended to link to detail on how to install it or details about the features in the version?

marthacryan: Yeah! I didn't have anywhere to point to yet but I should definitely add that to the TODO's.

BSd3v:
Would be nice if you guys also were able to dismiss error messages... I know that this becomes annoying with the window open and the messages showing.

Believe I have an issue opened for this...

dash/dash.py
```diff
@@ -77,6 +77,22 @@
 from ._jupyter import jupyter_dash, JupyterDisplayMode
 from .types import RendererHooks
 
+dash_design_kit_version = None
+try:
+    import dash_design_kit
+
+    dash_design_kit_version = dash_design_kit.__version__
```
emilykl, line 84: Maybe use `importlib.metadata` instead, so it'll still work if DDK changes the name of the `__version__` property (and also we don't have to load DDK into memory)

```suggestion
    dash_design_kit_version = importlib.metadata.version("dash_design_kit")
```

marthacryan: Done!

marthacryan: Ended up being better anyway because then you don't have to import the package

dash/dash-renderer/src/components/error/menu/VersionInfo.react.js
```diff
@@ -0,0 +1,182 @@
...
+            });
+    }
+}
+
+function shouldShowUpgradeNotification(currentDashVersion, newDashVersion) {
+    const showNotifications = localStorage.getItem('showNotifications');
+    const lastDismissed = localStorage.getItem('lastDismissed');
+    const lastDismissedVersion = localStorage.getItem('lastDismissedVersion');
+    if (
+        currentDashVersion == newDashVersion ||
```
emilykl, line 75: I could imagine edge cases where `currentDashVersion` is "greater" than `newDashVersion` -- maybe they are running a prerelease or custom build. 

Are the versions in a format where we can do greater-than/less-than comparisons?

```suggestion
        currentDashVersion >= newDashVersion ||
```

dash/dash-renderer/src/components/error/menu/VersionInfo.react.js
```diff
@@ -0,0 +1,182 @@
+import React, {useEffect, useState} from 'react';
+
+import './VersionInfo.css';
+
+const DAY_IN_MS = 86400000;
+
+function compareVersions(v1, v2) {
```
emilykl, line 7: Speaking of edge cases / prerelease versions, I think this will not work with alphanumeric version parts (e.g. `rc0`)... [Python guidelines](https://packaging.python.org/en/latest/specifications/version-specifiers/) don't require a dot between the version number and the prerelease spec either.

marthacryan: Ahh good point! I'll work on adding some logic for that

ndrezn: You can also do:
```python
>>> # pip install packaging
>>> from packaging.version import Version
>>> Version("2.3.1") < Version("10.1.2")
True
>>> Version("1.3.a4") < Version("10.1.2")
True
```

I'm not sure if `packaging` is a built-in so this isn't an option if we need to add a dependency but since our versioning scheme is pretty simple all things considered (we don't use `post` etc), this could be a good option.

ndrezn: If you want to steal implementation: https://github.com/pypa/packaging/blob/9b4922dd3c26c8522d716bec79d7e0ed408631c1/src/packaging/version.py#L161

marthacryan: I added some logic to strip off anything that isn't a digit or dot, and everything after that. So assuming that the versions will always start with `x.y.z`, it'll just compare using that version number.

marthacryan: > You can also do:
> 
> ```python
> >>> # pip install packaging
> >>> from packaging.version import Version
> >>> Version("2.3.1") < Version("10.1.2")
> True
> >>> Version("1.3.a4") < Version("10.1.2")
> True
> ```
> 
> I'm not sure if `packaging` is a built-in so this isn't an option if we need to add a dependency but since our versioning scheme is pretty simple all things considered (we don't use `post` etc), this could be a good option.

Oh sorry just seeing this! That's in python, and this logic will all have to be in javascript.

dash/dash-renderer/src/components/error/FrontEnd/FrontEndErrorContainer.react.js
```diff
@@ -28,11 +24,8 @@ class FrontEndErrorContainer extends Component {
             <div className={cardClasses}>
                 <div className='dash-error-card__topbar'>
                     <div className='dash-error-card__message'>
-                        🛑 Errors (
-                        <strong className='test-devtools-error-count'>
```
T4rk1n, line 32: This className is required for the tests and need to be put back or the tests adapted.

marthacryan: Oh ok I can add that back

marthacryan: Ok ended up just adding the class to where the new error count is. Waiting on the CI now

T4rk1n (approved):
:woman_dancing: Looks good.

PR #3122, opened 2025-01-15
Fixed bug in that user specified host/port/proxy were being overwritten

trbedwards:
Now the environment variables HOST, PORT and DASH_PROXY are fallbacks rather than overwriting the user specified arguments for host, port, proxy.


gvwilson:
@T4rk1n if this looks good to you, can it go into the upcoming release?

T4rk1n (approved):
:woman_dancing: Looks good, thanks.

Issue #3090, opened 2024-11-22
dash.no_update is not serializable by json is raised if other, not serializiable type is passed as output


smoky1337:
Hi everyone, 

thanks for providing such a wonderful package. I had an issue and I have been cracking my head at this long enough to belief this is a bug with `dash.no_update` not being serializiable. While writing this issue, I found that `dash.no_update` is not the issue but obfusicates the real error. 

Situation: Given a user picked date range, I gather data and display it. However, I want to cover the case when there simply is no data to present. 
For this, I aim to not update the already presented data and instead show an alert, notifiying the user of the absence of data. 

See this line (context below):
```python

return no_update, no_update, True, df
```
`df` is an `AssertionError`. Apparently, those are not serializiable by `json`. However, see error log at bottom, dash reports that `dash.no_update `seems to be the culprit.

The error below is resolved by parsing df into a string via ´str(df)´ 

```python

@callback(
    Output(component_id='figcontainer', component_property='figure'),
     Output(component_id='table', component_property='data'),
     Output(component_id="alert", component_property='is_open'),
     Output(component_id="alert",component_property='children'),
    Input(component_id='date_picker', component_property='start_date'),
    Input(component_id='date_picker', component_property='end_date'),
)
def update_graph(start, end):
    df = get_bookings_in_interval(start, end)
    # if there is no data, keep previous states and use alert
    if type(df) is AssertionError:
        return no_update, no_update, True, df

    #else simply display the information and dont update the alert
    all_appointments, best_days_with_numbers, cleaning_schedule = create_dfs(df)
    
    fig = create_figure(start, end, all_appointments, cleaning_schedule)

    
    return fig.to_dict(), best_days_with_numbers.to_dict(orient="records"), no_update, no_update


#either returns a df or an AssertionError
def get_bookings_in_interval(start, end):
    df = None
    try:
        data = get(#do something)
        assert data.status_code == 200 , "Failed to fetch bookings"
        parsed_data = dict(data.json())
        assert len(parsed_data["bookings"]) > 0, "No items in Response")
        #do something
            
    except AssertionError as e:
        print(e)
        return e

    return data
``` 

Environment used:
```

dash                      2.18.2
dash-bootstrap-components 1.6.0
dash-core-components      2.0.0
dash-html-components      2.0.0
dash-table                5.0.0
```

Error Log:

```
[...]

PR #3127, opened 2025-01-21
 is a Valid Prop

2Ryan09:
Solves #3090.

I was able to reproduce the error using a minimal example inferred by code provided by the issue author:

```python
from types import SimpleNamespace

from dash import html, Dash, dcc, Input, Output, no_update, callback, dash_table


app = Dash()

app.layout = html.Div(
        [
            dbc.Alert(id="alert", is_open=False, duration=4000),
            dcc.DatePickerRange(
                id="date_picker",
                start_date="2021-01-01",
                end_date="2021-01-31",
            ),
            dcc.Graph(id="figcontainer"),
            dash_table.DataTable(id="table"),
        ]
    )


@callback(
    Output(component_id="figcontainer", component_property="figure"),
    Output(component_id="table", component_property="data"),
    Output(component_id="alert", component_property="is_open"),
    Output(component_id="alert", component_property="children"),
    Input(component_id="date_picker", component_property="start_date"),
    Input(component_id="date_picker", component_property="end_date"),
)
def update_graph(start, end):
    df = get_bookings_in_interval(start, end)
    # if there is no data, keep previous states and use alert
    if type(df) is AssertionError:
        return no_update, no_update, True, df

    fig = go.Figure()

    return (
        fig.to_dict(),
        {},
        no_update,
        no_update,
    )

mock_response = SimpleNamespace(
    status_code=404,
)

# either returns a df or an AssertionError
def get_bookings_in_interval(start, end):
    df = None
    try:
        data = mock_response
        assert data.status_code == 200, "Failed to fetch bookings"
        parsed_data = dict(data.json())
        assert len(parsed_data["bookings"]) > 0, "No items in Response"
        # do something

    except AssertionError as e:
        print(e)
        return e

    return data


if __name__ == '__main__':
    app.run(debug=True)
```

raises

```
  File "/Users/raw/dash/dash/_callback.py", line 562, in add_context
    jsonResponse = to_json(response)
                   ^^^^^^^^^^^^^^^^^
  File "/Users/raw/dash/dash/_utils.py", line 26, in to_json
    return to_json_plotly(value)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/raw/miniconda3/envs/dash/lib/python3.12/site-packages/plotly/io/_json.py", line 171, in to_json_plotly
    return _safe(orjson.dumps(cleaned, option=opts).decode("utf8"), _swap_orjson)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Type is not JSON serializable: AssertionError

During handling of the above exception, another exception occurred:

dash.exceptions.InvalidCallbackReturnValue: The callback for `[<Output `figcontainer.figure`>, <Output `table.data`>, <Output `alert.is_open`>, <Output `alert.children`>]`
                returned a value having type `NoUpdate`     <--- Incorrectly attributes the error to `NoUpdate`


The value in question is either the only value returned,
or is in the top level of the returned list,

                and has string representation
                `<dash._no_update.NoUpdate object at 0x10a3da870>`

                or lists of those.
```

The apparent error was `NoUpdate` was being blamed for not being JSON serializable, even though the `AssertionError` output was to blame. The root cause ended up being `NoUpdate` was not deemed to be a valid prop by both `valid_props` and `valid_children` within `_validate.fail_callback_output`.

If the error-causing output was placed before the `NoUpdate` outputs in the return, the correct error message would have been displayed.

E.g.

```python
# ... existing code
    if type(df) is AssertionError:
        return df, no_update, no_update, True # df is `AssertionError`
```

raises

```
  File "/Users/raw/dash/dash/_callback.py", line 562, in add_context
    jsonResponse = to_json(response)
                   ^^^^^^^^^^^^^^^^^
  File "/Users/raw/dash/dash/_utils.py", line 26, in to_json
    return to_json_plotly(value)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/raw/miniconda3/envs/dash/lib/python3.12/site-packages/plotly/io/_json.py", line 171, in to_json_plotly
    return _safe(orjson.dumps(cleaned, option=opts).decode("utf8"), _swap_orjson)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Type is not JSON serializable: AssertionError

During handling of the above exception, another exception occurred:

dash.exceptions.InvalidCallbackReturnValue: The callback for `[<Output `figcontainer.figure`>, <Output `table.data`>, <Output `alert.is_open`>, <Output `alert.children`>]`
                returned a value having type `AssertionError`    <------ Here `AssertionError` is being properly blamed


The value in question is either the only value returned,
or is in the top level of the returned list,

                and has string representation
                `Failed to fetch bookings`

                or lists of those.
```

This PR adds `NoUpdate` to `valid_props` and `valid_children`, which results in the expected error.

One open question is the meaningful location for the `NoUpdate` definition. I made a new `_no_update` module, which seems like overkill, but I did not see a better place and importing `_callback` into `_validate` resorts in a circular dependency.

Side note: hello `NoUpdate` again, my [old friend](https://github.com/plotly/dash/pull/2800) :)

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -37,21 +37,11 @@ from ._utils import (
 from . import _validate
 from .long_callback.managers import BaseLongCallbackManager
 from ._callback_context import context_value
+from ._no_update import NoUpdate
 
 
 def _invoke_callback(func, *args, **kwargs):  # used to mark the frame for the debugger
     return func(*args, **kwargs)  # %% callback invoked %%
-
-
-class NoUpdate:
-    def to_plotly_json(self):  # pylint: disable=no-self-use
-        return {"_dash_no_update": "_dash_no_update"}
-
-    @staticmethod
-    def is_no_update(obj):
-        return isinstance(obj, NoUpdate) or (
-            isinstance(obj, dict) and obj == {"_dash_no_update": "_dash_no_update"}
-        )
 
 
 GLOBAL_CALLBACK_LIST = []
--- a/dash/_no_update.py
+++ b/dash/_no_update.py
@@ -1,0 +1,9 @@
+class NoUpdate:
+    def to_plotly_json(self):  # pylint: disable=no-self-use
+        return {"_dash_no_update": "_dash_no_update"}
+
+    @staticmethod
+    def is_no_update(obj):
+        return isinstance(obj, NoUpdate) or (
+            isinstance(obj, dict) and obj == {"_dash_no_update": "_dash_no_update"}
+        )
--- a/dash/_validate.py
+++ b/dash/_validate.py
@@ -215,8 +215,6 @@ def fail_callback_output(output_value, output):
     valid_children = (str, int, float, type(None), Component, NoUpdate)
     valid_props = (str, int, float, type(None), tuple, MutableSequence, NoUpdate)
 
-    print("================================")
-
     def _raise_invalid(bad_val, outer_val, path, index=None, toplevel=False):
         bad_type = type(bad_val).__name__
         outer_id = f"(id={outer_val.id:s})" if getattr(outer_val, "id", False) else ""
@@ -264,7 +262,6 @@ def fail_callback_output(output_value, output):
         return isinstance(val, valid_props)
 
     def _can_serialize(val):
-        print("checking ability to serialize")
         if not (_valid_child(val) or _valid_prop(val)):
             return False
         try:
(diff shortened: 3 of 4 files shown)


2Ryan09:
I can confirm this issue is still present in Dash 3.0.4.

gvwilson:
@2Ryan09 we're working through issues when we can, but we are a small team with a large backlog - I apologize for not being able to move faster.

PR #3137, opened 2025-01-29
Modernizing Logout Handling replacing doc.Logoutbutton with a React-Based solution

SHRISHTISHUKLA-0:
here removed dcc.LogoutButton and Implement a standard react <button> for logout
use fetch API for logout request here improved UX and error handling and added an onLogout callback.

correct me where I am wrong and help me to know more about issue.


ndrezn:
We're planning to remove `dcc.LogoutButton`, see: #2894. I don't think it is worth updating the button to work differently -- I would err on letting developers use their own strategies here, or leverage Dash Enterprise Auth in a business context.  @T4rk1n might have a differing opinion though.

T4rk1n:
The logout button has already been removed in dash-3.0 branch.

PR #3157, opened 2025-02-07
Fix hook priority

### LINKED ISSUE #3150: Are hooks correctly saving assigned priority?
From `_hooks.py` we see that hook decorators, e.g. `@hooks.layout`, are actually just wrappers around `self.add_hook`. Looking at `self.add_hook` and how it handles input parameter `priority`, how is the value given to `priority` passed down to the hook? 

It is only used in the check `if not priority`, which fails if a non-zero value is given and is otherwise not used. The value given to `_Hook`, is `priority=p`, and `p=0` is set. 

Is this a bug or am I misunderstanding how hooks are later used/called?

See: https://github.com/plotly/dash/blob/v3.0.0rc1/dash/_hooks.py#L58-L81

T4rk1n:
Fix #3150

### DIFF
--- a/dash/_hooks.py
+++ b/dash/_hooks.py
@@ -71,10 +71,10 @@ class _Hooks:
             return
         hks = self._ns.get(hook, [])
 
-        p = 0
+        p = priority or 0
         if not priority and len(hks):
-            priority_max = max(h.priority for h in hks)
-            p = priority_max - 1
+            priority_min = min(h.priority for h in hks)
+            p = priority_min - 1
 
         hks.append(_Hook(func, priority=p, data=data))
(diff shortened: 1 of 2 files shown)


dash/_hooks.py
```diff
@@ -71,10 +71,10 @@ def add_hook(
             return
         hks = self._ns.get(hook, [])
 
-        p = 0
+        p = priority or 0
         if not priority and len(hks):
-            priority_max = max(h.priority for h in hks)
-            p = priority_max - 1
+            priority_min = min(h.priority for h in hks)
```
gvwilson, line 76: worth a comment in the code so the next person doesn't flip it back?

gvwilson (approved):

PR #3161, opened 2025-02-12
Add generated caller filename hash as clientside callback namespace.

T4rk1n:
Fix for https://github.com/plotly/dash-embedded/issues/87

Generate hash from the `clientside_callback` caller filename to use as namespace, this will make it unique instead of modifying the namespace in dash-embedded.

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -1,6 +1,7 @@
 import collections
 import hashlib
 from functools import wraps
+import importlib
 from typing import Callable, Optional, Any
 
 import flask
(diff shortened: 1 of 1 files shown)


gvwilson (approved):

T4rk1n:
Repeating from dash-embedded thread, the inline scripts are added to the scripts tag on the index but that is not recommended and doesn't allow caching. We need to bring the functionality on embedded that created a new route for the inline scripts and added it to the scripts to load.

PR #3170, opened 2025-02-17
Add ExternalWrapper to dash_component_api

T4rk1n:
Add new exposed component `ExternalWrapper` to `window.dash_component_api`.

It has the props:

- `componentPath` path to the component in the store. Should use the parent path to create a new path, `[...parentPath, "external"]`
- `componentType` string type of the component to hydrate.
- `componentNamespace` string of the namespace where the component is located. example `dash_core_components`, `dash_table`. The namespace must have been loaded before use.
- The rest of the props are given as extra props to the wrapper, allowing the component author to control the props of the component.

Intended usage in libraries to reuse dash components without them being present in the original layout.

### DIFF
--- a/@plotly/dash-test-components/src/components/ExternalComponent.js
+++ b/@plotly/dash-test-components/src/components/ExternalComponent.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+const ExternalComponent = ({ id, text, input_id }) => {
+    const ctx = window.dash_component_api.useDashContext();
+    const ExternalWrapper = window.dash_component_api.ExternalWrapper;
+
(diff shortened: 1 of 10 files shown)


gvwilson:
I'm happy to approve this so that it can be merged but will be doing so on faith - I don't understand the Dash code well enough to do a proper review.

AnnMarieW:
@T4rk1n 

Can you say more about this use case for this?

What if the component is not a simple component but also has components in `children`?

Is this a replacement for the `renderDashComponents` function used in dmc?

T4rk1n:
> Is this a replacement for the renderDashComponents function used in dmc?

Yes this is intended for this purpose, rendering components out of the layout tree.

PR #3173, opened 2025-02-20
Enable storing the callback outputs in the persistence storage with the flag

petar-qb:
This PR is similar to https://github.com/plotly/dash/pull/3144. It solves the issue but with an optional callback flag called `enable_persistence`.
This PR behaves more like a `feature`, and the other one more like a `bugfix`.

Add enable_persistence that when True enables storing the callback outputs in the persistence storage.

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -71,6 +71,7 @@ def callback(
     manager=None,
     cache_args_to_ignore=None,
     on_error: Optional[Callable[[Exception], Any]] = None,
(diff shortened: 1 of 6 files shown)


gvwilson:
cc @T4rk1n

ndrezn:
My understanding is that this PR is parallel with #3144, we either merge this or #3144 depending on whether we want to have this as a flag or a default behaviour.

PR #3182, opened 2025-02-21
Fixes set_props to merge instead of overwriting props

### LINKED ISSUE #3104: [BUG] set_props only updating last output in background callbacks
**Describe your context**
Mac OS, occurring in using Safari & Chrome
```
dash                 2.18.2
dash-core-components 2.0.0
dash-html-components 2.0.0
dash-table           5.0.0
```

**Describe the bug**

When updating two different components via set_props, in a background callback with running, **only the last output is updated**. No console errors encountered.

If I remove the background, and running, then both outputs are updated as expected. 

See reproducible example below:
```
from dash import Dash, Output, Input, html, set_props, DiskcacheManager
import diskcache
import dash

app = Dash()

app.layout = [
    html.Button("start", id="start"),
    html.Div("initial", id="output"),
    html.Div("initial 2", id="output-2"),
]

@app.callback(
    Input("start", "n_clicks"),
    background=True,
    running=[
        (Output("start", "disabled"), True, False)
    ],
    manager=background_callback_manager,
)
def on_click(_):
    set_props("output", {"children": "changed"})
    set_props("output-2", {"style": {"background": "red"}})

if __name__ == "__main__":
    app.run(debug=True)
```

**Expected behavior**

All outputs should update, not just the last one.

Raahul-Singh:

Fixes #3104 

## Contributor Checklist

   - [x] **Fix `_set_props` inside `_make_job_fn` to accumulate updates instead of overwriting.**
   - [x] **Ensure all `set_props` calls persist across multiple updates.**
   - [x] **Test multiple `set_props` calls in a background callback to confirm the fix.**

### optionals

- [ ] If this PR needs a follow-up in **dash docs**, **community thread**, I have mentio

ned the relevant URLS as follows

This change gets the expected behavior in @faulty13' snippet

https://github.com/user-attachments/assets/1e6d0eb9-6982-4526-85a9-67ba06f4c02c

### DIFF
--- a/dash/long_callback/managers/diskcache_manager.py
+++ b/dash/long_callback/managers/diskcache_manager.py
@@ -179,7 +179,15 @@ def _make_job_fn(fn, cache, progress):
         maybe_progress = [_set_progress] if progress else []
 
         def _set_props(_id, props):
-            cache.set(f"{result_key}-set_props", {_id: props})
+            set_props_key = f"{result_key}-set_props"
+            existing_props = cache.get(set_props_key, {})
+
+            if _id in existing_props:
+                existing_props[_id].update(props)  # Merge instead of overwrite
+            else:
+                existing_props[_id] = props  # First-time update
+
+            cache.set(set_props_key, existing_props)
 
         ctx = copy_context()
 


T4rk1n (commented):
This is only for diskcache, we need the same behavior in regular callbacks, diskcache and celery.

Also need to add tests for the merging.

BSd3v:
Would be interesting to see if this could take on a behaviour of `Patch`. Sometimes there may be a reason to append multiple children throughout the span of the entire callback.

PR #3190, opened 2025-02-25
fix cache key issue

### LINKED ISSUE #3189: [BUG]
```
dash                 2.18.2
dash-core-components 2.0.0
dash-html-components 2.0.0
dash-table           5.0.0
```
**Describe the bug**
When a background callback resets one of its own outputs, you cannot set the value of that output back to the previous value, because the basis for the cache key would be the same, and the cached result would be returned. This is an edge case, but it happens frequently enough to warrant a change.  

**Expected behavior**

A callback invoked with the same input values, but different triggered inputs should not retrieve the cached result.

minimal example:
Select "World", and wait for the callback to finish. Press "clear page", and wait for the callback to finish. If you now try to select "World", you cannot do so.
`
import dash
from dash import html, dcc, Input, Output, no_update
import diskcache
from dash import DiskcacheManager
from uuid import uuid4
import time

launch_uid = uuid4()

background_callback_manager = DiskcacheManager(
    cache, cache_by=[lambda: launch_uid], expire=60
)

app = dash.Dash(__name__, background_callback_manager=background_callback_manager)

app.layout = html.Div(
    [
        dcc.RadioItems(["Hello", "World"], id="input", value="Hello"),
        html.Div(id="output", children="Hello is selected"),
        html.Button("clear page", id="button"),
    ]
)

@app.callback(
    Output("output", "children"),
    Output("input", "value"),
    Input("input", "value"),
    Input("button", "n_clicks"),
    prevent_initial_call=True,
    background=True
)
def update_output(value, n_clicks):
    triggered_id = dash.callback_context.triggered_id
    print(triggered_id)
    time.sleep(5)
    match triggered_id:
        case "input":
            return value + " is selected", no_update
        case "button":
            return "Hello is selected", "Hello"
        case _:
            return no_update, no_update

if __name__ == "__main__":
    app.run_server()
`

MartinSA04:
Proposed fix for issue #3189 
Closes #3189 

Adds triggered_inputs as additional basis for cache_key generation.

## Contributor Checklist

   -  [x] fix issue
   -  [x] add as optional parameter
- [ ] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR (unsure if this is needed?)

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 - [#3080](https://github.com/plotly/dash/pull/3080) Fix docstring generation for components using single-line or nonstandard-indent leading comments
 - [#3103](https://github.com/plotly/dash/pull/3103) Fix Graph component becomes unresponsive if an invalid figure is passed
+- [#3190](https://github.com/plotly/dash/pull/3190) Fix issue with cache key generation by adding option to include triggered inputs. Fixes [#3189](https://github.com/plotly/dash/issues/3189)
 
 ## [2.18.2] - 2024-11-04
 
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -70,6 +70,7 @@ def callback(
     cancel=None,
     manager=None,
     cache_args_to_ignore=None,
+    cache_ignore_triggered=True,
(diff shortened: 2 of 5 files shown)


T4rk1n (approved):

PR #3195, opened 2025-02-26
Change to use GET request when requesting dash version from server

marthacryan:
Changes from using a POST request to retrieve the upgrade information from the server to using a GET request for more flexibility in the future.

### DIFF
--- a/dash/dash-renderer/src/components/error/menu/VersionInfo.react.js
+++ b/dash/dash-renderer/src/components/error/menu/VersionInfo.react.js
@@ -32,9 +32,6 @@ async function requestDashVersionInfo(config) {
         ddk_version: ddkVersion,
         plotly_version: plotlyVersion
     } = config;
-    if (!shouldRequestDashVersion(config)) {
-        return {};
-    }
     const cachedVersionInfo = localStorage.getItem('cachedNewDashVersion');
     const cachedNewDashVersionLink = localStorage.getItem(
         'cachedNewDashVersionLink'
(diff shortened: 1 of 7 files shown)


gvwilson (approved):

dash/testing/application_runners.py
```diff
@@ -162,6 +162,7 @@ def run():
             app.css.config.serve_locally = True
 
             options = kwargs.copy()
+            options["disable_version_check"] = True
```
T4rk1n, line 165: Change to `dev_tools_disable_version_check`

dash/dash-renderer/src/components/error/menu/VersionInfo.react.js
```diff
@@ -46,25 +46,20 @@ async function requestDashVersionInfo(config) {
...
+    } else if (shouldRequestDashVersion(config)) {
+        const queryParams = new URLSearchParams({
+            dash_version: currentDashVersion,
+            python_version: pythonVersion,
+            ddk_version: ddkVersion,
+            plotly_version: plotlyVersion
+        }).toString();
+        return fetch(dashVersionUrl + '?' + queryParams, {mode: 'cors'})
             .then(response => response.json())
             .catch(() => {
```
T4rk1n, line 58: I think the catch should be moved to the end of the tail. Since it's CORS, we can't get the regular error and it will throw on the `.json()` call and go into the last then and throw again. The catch should handle all the cases.

T4rk1n: Also could refactor those then to `await` the fetch with a `try catch` since the function is marked as `async`

T4rk1n (approved):
:woman_dancing:

PR #3219, opened 2025-03-14
reworking dash debug tools interface

BSd3v:
A couple of things to note here:
- I removed color declarations in order to allow app definitions to take precedence over the colors.
- I adjusted the color on the toolbar itself to have some alpha in order to allow the app background to get added to
- the toolbar is now toggle-able, when hidden error count goes to the toggle button
- the toolbar is now a footer
- when the toolbar is shown, the app is scaled dynamically based upon the screen size and given 40px for the toolbar to display
- when the toolbar is not shown, the app is full size, and the toggle button is placed on top with a zIndex of 10000
- the toolbar no longer needs zIndex as it has been given space in the page

The main thing that needs to get worked on here is the toggle button:
- obviously the button needs to be styled
- when the screen size is narrow, the toolbar and button collide, need to implement some sort of way to implement a selection for the tool to display and the other info to display (dash version)

@marthacryan @ndrezn

### DIFF
--- a/dash/dash-renderer/src/components/error/GlobalErrorContainer.react.js
+++ b/dash/dash-renderer/src/components/error/GlobalErrorContainer.react.js
@@ -11,14 +11,16 @@ class UnconnectedGlobalErrorContainer extends Component {
     render() {
         const {config, error, children} = this.props;
         return (
-            <div id='_dash-global-error-container'>
+            <div id='_dash-global-error-container' style={{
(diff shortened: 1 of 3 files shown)


BSd3v:
image
image
image
image
image
image

gvwilson:
thanks very much for this @BSd3v - we are going to ship Dash 3.0.0 with the existing dev tools UI and will then come back immediately to look at improving it (either via this PR or something similar).

PR #3224, opened 2025-03-17
reworking dash debug tools interface

BSd3v:
See https://github.com/plotly/dash/pull/3219 for comments on the PR.

Reopening to target dev branch vs the deleted 3.0 branch.


BSd3v:
@gvwilson, I think for now, we are alright with how the minimized tools are working out. 🙂

PR #3232, opened 2025-03-20
Add error handling for when localStorage is disabled

marthacryan:


gvwilson (approved):
And there are even meaningful comments! thank you

dash/dash-renderer/src/components/error/menu/VersionInfo.react.js
```diff
@@ -75,12 +88,19 @@ async function requestDashVersionInfo(config) {
...
-    const showNotifications = config.disable_version_check
-        ? false
-        : showNotificationsLocalStorage !== 'false';
-    const lastFetched = localStorage.getItem('lastFetched');
+    let showNotificationsLocalStorage, showNotifications, lastFetched;
+    try {
+        showNotificationsLocalStorage =
+            localStorage.getItem('showNotifications');
+
+        showNotifications = config.disable_version_check
```
T4rk1n, line 96: Can we move that check to the DebugMenu render like: `{!config.disable_version_check && <VersionInfo />}` ?

marthacryan: The `VersionInfo` component also shows the current version of dash, which we want to be shown either way

T4rk1n (approved):
:woman_dancing:

PR #3239, opened 2025-03-24
Remove stringcase

### LINKED ISSUE #3238: The stringcase package can no longer be installed with setuptools version >=78.
The [https://pypi.org/project/stringcase/](stringcase) package, that is hard dependacy of Dash since this PR can no longer be installed using setuptools >=78, because of the following PR [https://github.com/pypa/setuptools/pull/4870](#4870). As a result, Dash can no longer be installed with setuptools version >=78.

_Originally posted by @ivan-mingolov-blue-technologies in https://github.com/plotly/dash/issues/3220#issuecomment-2748637195_

T4rk1n:
Fix #3238

### DIFF
--- a/dash/development/_py_prop_typing.py
+++ b/dash/development/_py_prop_typing.py
@@ -3,7 +3,8 @@ import json
 import string
 import textwrap
 import importlib
-import re
+
+from .._utils import pascal_case
 
 
 shapes = {}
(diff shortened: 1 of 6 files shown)


heng-m-xr (approved):

ndrezn (approved):

dash/development/_py_prop_typing.py
```diff
@@ -52,9 +51,15 @@ def generate_any(*_):
     return "typing.Any"
 
 
+def pascal_case(name: str):
+    return name[0].upper() + re.sub(
+        r"[\-_\.\s]([a-z])", lambda match: match.group(1).upper(), name[1:]
+    )
+
+
 def generate_shape(type_info, component_name: str, prop_name: str):
```
emilykl, line 60: @T4rk1n Is this function covered by tests? Will anything break in a very deep way if this function returns a slightly different output following this change? (The regex looks solid but just in case)

T4rk1n: It would just be the name of the generated type, I don't think it would break. It's also almost the same as what stringcase was doing.

Issue #2337, opened 2022-11-27
[Feature Request] Individual Dismissible Error Messages in Debug=True


BSd3v:
**Is your feature request related to a problem? Please describe.**
When trying to use the application in debug=True mode, not dealing with all errors one by one. If I have a components that I need to interact with on the right side of the screen, they become hidden by error messages, no longer being able to use without turning all errors off. This is not welcome as new error messages do not automatically show.

**Describe the solution you'd like**
If we could hide/dismiss individual or "known" issues from the error detail on the right side of the screen, to where when all messages have been dismissed, that the error messages placeholder would also disappear. This would allow for continued use of the site without hinderance.

Then, if you want to see all the errors again, you interact through the menu on the bottom

**Describe alternatives you've considered**
I have considered writing my own javascript code to dismiss the individual elements of the by adding buttons to each message.

englucas-data:
Someone help me with logout this group. Unfortunately I don't have access
to my oldest git, and because of this I can't get out of this group.

marthacryan:
Agreed

PR #3241, opened 2025-03-24
Add collapse button to devtools

marthacryan:
## User facing changes
* Adds a collapse / expand button that uses local storage for collapse state. It has the same little rotating animation as the old dev tools!
devtools-collapse

## Code changes
* Changes the `DebugMenu` from a class component to functional component (it made using state easier and we're trying to move components to that anyway).


gvwilson (approved):

BSd3v:
What happens if there are errors when it is collapsed?

marthacryan:
@BSd3v Good question! Currently what happens is the list of errors pop up either way, and it doesn't affect the collapse state of the devtools menu. So if it's collapsed and there's an error, you see this at the bottom right:
image
@ndrezn Any thoughts on this? I know we discussed having a little red error count over the collapse button but I didn't remember if we had a decision there. 

This did actually help me catch an edge case though where if you dismiss the error dialog and the devtools are collapsed, nothing happens when you get an error.

BSd3v:
This is where this would be handy. ;)

https://github.com/plotly/dash/issues/2337

ndrezn:
Good ideas, but let's not let scope creep here! We came up with plenty of others during the devtools redesign 😉 When there is an error/version update but the devtools are minimized it will show a red notification dot with no number. @jeromevaldez is on design for that.

dash/dash-renderer/src/components/error/menu/DebugMenu.css
```diff
@@ -137,10 +137,10 @@
     border: 1px solid rgba(0, 24, 102, 0.1);
 }
 .dash-debug-menu__outer.dash-debug-menu__outer--collapsed {
-    width: 50px;
+    max-width: 50px;
 }
 .dash-debug-menu__outer.dash-debug-menu__outer--expanded {
-    width: 682px;
+    max-width: 682px;
```
BSd3v, line 143: Did you try `fit-content`? It might work as long as the internal components use a width pre-defined

T4rk1n (approved):
Looks good, nice refactor to a functional component :woman_dancing:

dash/dash-renderer/src/components/error/menu/DebugMenu.css
marthacryan: Just tried that and it didn't work. It's tricky to add the sliding behavior! Looks like there's this [experimental css feature](https://developer.mozilla.org/en-US/docs/Web/CSS/calc-size) that could be good once it has higher adoption, but from what I've seen there aren't amazing options other than the max-width hard-coding for now.

Issue #3243, opened 2025-03-25
Prop tests in Debug=True slow when manipulating lots of children


BSd3v:
Good day,

You can find a discussion here for some examples: https://github.com/RenaudLN/dash-pydantic-form/issues/86

Anyways, if you run the below example with:
```
dash>=3.0.0
dash-mantine-component==1.1.0
```

```

import dash_mantine_components as dmc
from dash import Output, Input, html, callback, Dash, ctx, _dash_renderer
from dash.exceptions import PreventUpdate
from dash_iconify import DashIconify

# this is optional when using dash >3.0.0
_dash_renderer._set_react_version("18.2.0")

app = Dash(__name__, external_stylesheets=dmc.styles.ALL)

app.layout = dmc.MantineProvider(
    [
        dmc.Accordion(
            [dmc.AccordionItem([
                dmc.AccordionControl(f'{x}'),
                dmc.AccordionPanel([
                    html.Div([dmc.Anchor(f'{y}', href=f'{y}'), dmc.Text(), dmc.TextInput()]) for y in range(100)
                ])
            ], value =f'{x}') for x in range(10)]
        )
    ]
)


if __name__ == "__main__":
    app.run(debug=True)
```

If you keep as is, it takes a good amount of time to render an update to the accordion. If, however, you disable debug mode, the results are what we would expect less than half a second to open.

This seems to be related to the check props while in debug mode. This needs to be more efficient, or possibly even controlled by the app configuration.

BSd3v:
With the way this renders, especially in debug mode, yes. 

The code above takes half the time in dash 2 compared to dash 3 when in debug mode. This can help it perform better by memoizing the response. 

It doesn't solve the issue of numerous components being rendered at the same time though.

PR #3248, opened 2025-03-26
Adjusting memoization on the `DashWrapper` to be utilized

BSd3v:
This is a PR to test the adjustment of memoization on the `DashWrapper`

you can see info here: https://github.com/plotly/dash/issues/3243


BSd3v:
The initial test failed due to not being able to `redraw` components if the parent is adjusted.

The current test for this hash is `startsWith` I have updated it to render the first descendent of the `updatedPath`, however, this will not work if the desired outcome is to render and adjustment to every upstream component.

If the above is the desired outcome, I recommend having devs add a new prop to components that want to redraw on upstream adjustments and using that to switch to the older function which uses `startsWith` without any limit. Applying this to all descendants does not seem efficient.

BSd3v:
Another thing on the last comment, I adjusted the way the memoizing works, it only holds onto the last rendered time. This helps with rerenders.

dash/dash-renderer/src/wrapper/DashWrapper.tsx
```diff
@@ -41,24 +46,61 @@ type DashWrapperProps = {
      */
     componentPath: DashLayoutPath;
     _dashprivate_error?: any;
+    _passedComponent?: any;
+    _newRender?: any;
```
T4rk1n, line 50: Maybe add a comment describing those new props.

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -357,10 +359,14 @@ function updateComponent(component_id: any, props: any, cb: ICallbackPayload) {
...
         }
+        const component = getComponentLayout(componentPath, _state);
         dispatch(
             updateProps({
                 props,
-                itempath: componentPath
+                itempath: componentPath,
+                component,
+                config,
+                renderType: 'callback'
```
T4rk1n, line 369: We could avoid passing the component and config in all the calls by using an action creator for the `updateProps` function.
Rename this action to `onPropChange` https://github.com/plotly/dash/blob/f397f8a1997f1db8f61ef68cdddf96b8a8259a96/dash/dash-renderer/src/actions/index.js#L20

Then add:
```js
export const updateProps = payload => (dispatch, getState) => {
    const {itempath} = payload;
    const state = getState();
    const component = getComponentLayout(itempath, state);
    dispatch(onPropChange({...payload, component, config: state.config }));
}
```

T4rk1n (commented):
Nice job, just a few minor comments.

dash/dash-renderer/src/actions/callbacks.ts
BSd3v: The config and component arent used anymore, it was if we needed to crawl which I opted to instead force rerendering of the components as props if the changeProp was the prop of the component.

dash/dash-renderer/src/reducers/reducer.js
```diff
@@ -27,7 +27,23 @@ export const apiRequests = [
...
+        strPath,
+        {
+            hash: prev + 1,
+            changedProps: action.payload.props,
+            renderType: action.payload.renderType
+        },
+        state
+    );
+    return state;
+}
```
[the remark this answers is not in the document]
BSd3v: I had pulled it out in order to make recursive, we can go back to inside the same function.

dash/dash-renderer/src/wrapper/DashWrapper.tsx
```diff
@@ -41,24 +46,61 @@ type DashWrapperProps = {
...
-    const [component, componentProps] = useSelector(
+    // eslint-disable-next-line prefer-const
+    let [component, componentProps, h, changedProps, renderType] = useSelector(
         selectDashProps(componentPath),
         selectDashPropsEqualityFn
     );
 
+    /* eslint-disable @typescript-eslint/ban-ts-comment, @typescript-eslint/no-unused-vars */
+    // @ts-ignore
+    const newlyRendered = useMemo(() => {
```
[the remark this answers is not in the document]
BSd3v: I'll check with the `@ts-ignore` the unused vars needs to be ignored.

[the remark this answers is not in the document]
BSd3v: Still errors out in the build.

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -336,7 +336,8 @@ async function handleClientside(
 
 function updateComponent(component_id: any, props: any, cb: ICallbackPayload) {
     return function (dispatch: any, getState: any) {
-        const {paths, config} = getState();
+        const _state = getState();
+        const {paths, config} = _state;
```
T4rk1n, line 340: combine those back since _state is not need anymore.

dash/dash-renderer/src/observers/executedCallbacks.ts
```diff
@@ -45,7 +45,8 @@ const observer: IStoreObserverDefinition<IStoreState> = {
         } = getState();
 
         function applyProps(id: any, updatedProps: any) {
-            const {layout, paths} = getState();
+            const _state = getState();
+            const {layout, paths} = _state;
```
T4rk1n, line 49: combine those back since _state is not need anymore.

dash/dash-renderer/src/utils/clientsideFunctions.ts
```diff
@@ -16,16 +16,18 @@ function set_props(
     for (let y = 0; y < ds.length; y++) {
         const {dispatch, getState} = ds[y];
         let componentPath;
+        const _state = getState();
+        const {paths} = _state;
```
T4rk1n: combine those back since _state is not need anymore.

dash/dash-renderer/src/wrapper/DashWrapper.tsx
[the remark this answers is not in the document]
T4rk1n: Oh, I don't think `useMemo` is right in this case, should be `useEffect`.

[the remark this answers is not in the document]
BSd3v: `useEffect` is not sync, while `useMemo` is sync. For the `current` to be used within the same sync it has to be `useMemo`.

T4rk1n (approved):
:woman_dancing:

PR #3249, opened 2025-03-26
Move the dash.layout.setter so that type checkers can find it easier

peter-gerlagh-kyos:
This rearranges two methods of the Dash class.
Specifically you now have `@property def layout // @layout.setter def layout // def _layout_value` instead of `@property def layout // def _layout_value // @layout.setter def layout`.
The point is to make it easier for type checkers (such as mypy) to connect the `@property layout` with the `layout.setter`, so that people who use this library with a type checker need fewer `#type: ignore` statements.
This is a very minimal change, and I don't see how it could break anything, or make anything worse. I don't think a note in the changelog is necessary.

## Contributor Checklist

   -  [x] change order of methods
   -  [x] confirm mypy is happy
   -  [ ] run tests

### optionals


peter-gerlagh-kyos:
I've not been able to run the test, and I don't understand the problem. dash_test_components seems to be missing (?)

T4rk1n:
> I've not been able to run the test, and I don't understand the problem. dash_test_components seems to be missing (?)

For the you need to run `setup-tests.py`, but it might takes pretty long, can just run the tests that fails the ci.

T4rk1n (approved):
Look good, if that helps with linting doesn't hurt to change the order. :woman_dancing:

Issue #3226, opened 2025-03-18
[BUG] Dash 3.0 fails mypy


gothicVI:
With the release of dash 3.0 our CI/CD fails for stuff that used to work.
Here's a minimum working example:
```python
from dash import Dash, dcc, html
from dash.dependencies import Input, Output
from typing import Callable

app = Dash(__name__)

def create_layout() -> html.Div:
    return html.Div([
        dcc.Input(id='input-text', type='text', value='', placeholder='Enter text'),
        html.Div(id='output-text')
    ])  

app.layout = create_layout

# Explicitly cast Input and Output to satisfy mypy
input_component: Input = Input('input-text', 'value')
output_component: Output = Output('output-text', 'children')

@app.callback(output_component, input_component)
def update_output(value: str) -> str:
    return f'You entered: {value}'

if __name__ == '__main__':
    app.run(debug=True, port=9000)
```
running mypy on this file results in:
```bash
$ mypy t.py --strict
t.py:1: error: Skipping analyzing "dash": module is installed, but missing library stubs or py.typed marker  [import-untyped]
t.py:2: error: Skipping analyzing "dash.dependencies": module is installed, but missing library stubs or py.typed marker  [import-untyped]
t.py:2: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
t.py:19: error: Untyped decorator makes function "update_output" untyped  [misc]
Found 3 errors in 1 file (checked 1 source file)
```
which we used to solve by adding `# type: ignore[misc]` to every `callback` call and by adding
```toml
[[tool.mypy.overrides]]
module = [
  "dash.*",
  "dash_ag_grid",
  "dash_bootstrap_components.*",
  "plotly.*",
]
ignore_missing_imports = true
```
to our `pyproject.toml`.

However, when updating to version 3, we get:
```bash
$ mypy --strict t.py 
t.py:8: error: Returning Any from function declared to return "Div"  [no-any-return]
t.py:13: error: Property "layout" defined in "Dash" is read-only  [misc]
t.py:16: error: Call to untyped function "Input" in typed context  [no-untyped-call]
t.py:17: error: Call to untyped function "Output" in typed context  [no-untyped-call]
t.py:19: error: Call to untyped function "callback" in typed context  [no-untyped-call]
t.py:19: note: Error code "no-untyped-call" not covered by "type: ignore" comment
Found 5 errors in 1 file (checked 1 source file)
```
without changing anything else.

[...]

PR #3254, opened 2025-04-01
feat: adding more type annotations

gvwilson:
This PR will address #3226. Using `pyright --pythonversion 3.9 ./dash` to check.


gvwilson:
The warning in `dash/table/__init__.py` can be fixed by changing the imports at the top from:

```
from ._imports_ import *  # noqa: E402, F401, F403
from ._imports_ import __all__ as _components
from . import Format  # noqa: F401, E402
from . import FormatTemplate  # noqa: F401, E402
```

to:

```
from . import DataTable  # noqa: F401, E402
from . import Format  # noqa: F401, E402
from . import FormatTemplate  # noqa: F401, E402
```

However, this file is generated during `npm run build`, so we need to figure out where and how to modify the generator.

dash/dash.py
```diff
@@ -18,10 +18,11 @@
 import base64
 import traceback
 from urllib.parse import urlparse
-from typing import Any, Callable, Dict, Optional, Union, Sequence
+from typing import Any, Callable, Dict, Optional, Union, Sequence, Literal
 
 import flask
 
+from flask.typing import RouteCallable
```
mayankg90, line 25: This is not present in flask 1.1.1 and [requirements](https://github.com/plotly/dash/blob/a4bdc7d8c586b23cc7a1b9df9f0528e9ab472168/requirements/install.txt#L1) are set to 1.0.4 

Flask only has typing [after 2.0.0](https://github.com/pallets/flask/tree/2.0.0/src/flask)

anurag1paul: RouteCallable in flask.typing was added by PR https://github.com/pallets/flask/pull/4676 which was released in flask 2.2.0 https://github.com/pallets/flask/commits/2.2.0?after=b17bb9ed563ab2857c0db9a07ec4e6407404c7be+34, so probably need to update requirements to >= 2.2.0

PR #3263, opened 2025-04-07
Add note on params that only work with background callbacks

LiamConnors:
@T4rk1n do you know if there are any other params that only apply to background callbacks.

This PR relates to this issue https://github.com/plotly/ddk-dash-docs/issues/3318
which is specifically related to the `cancel` param, but I think it's the same for the rest of these. 

---

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -118,12 +118,14 @@ def callback(
             the app layout. The second element is the value that the property
             should be set to while the callback is running, and the third element
             is the value the property should be set to when the callback completes.
+            This parameter only applies to background callbacks (`background=True`).
         :param cancel:
             A list of `Input` dependency objects that reference a property of a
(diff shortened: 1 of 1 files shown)


dash/_callback.py
```diff
@@ -118,12 +118,14 @@ def callback(
             the app layout. The second element is the value that the property
             should be set to while the callback is running, and the third element
             is the value the property should be set to when the callback completes.
+            This parameter only applies to background callbacks (`background=True`).
```
LiamConnors:
```suggestion
```

gvwilson (approved):
The prose looks good, but I'll defer to @T4rk1n on correctness.

T4rk1n (approved):
:woman_dancing:

PR #3264, opened 2025-04-08
Updating `componentPath` from the `DashWrapper` `setProps`

BSd3v:
allows for the componentPath to be updated when moving around in the children and triggering `setProps`


T4rk1n (approved):
:woman_dancing: Looks good

PR #3265, opened 2025-04-08
localizing the scope of the resize handler

BSd3v:
localizing the scope of the resize handler

### DIFF
--- a/components/dash-core-components/tests/integration/graph/test_graph_responsive.py
+++ b/components/dash-core-components/tests/integration/graph/test_graph_responsive.py
@@ -206,6 +206,6 @@ def test_grrs002_graph(dash_dcc):
     dash_dcc.wait_for_text_to_equal("#generate-btn", "Generate Figures")
     dash_dcc.find_element("#generate-btn").click()
     dash_dcc.wait_for_text_to_equal("#bounding-output", "loaded")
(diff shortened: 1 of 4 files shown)


components/dash-core-components/tests/integration/graph/test_graph_responsive.py
```diff
@@ -205,5 +206,6 @@ def update_figures(n_clicks):
     dash_dcc.wait_for_text_to_equal("#generate-btn", "Generate Figures")
     dash_dcc.find_element("#generate-btn").click()
     dash_dcc.wait_for_text_to_equal("#bounding-output", "loaded")
+    time.sleep(.1) ## must wait for the full render
```
ndrezn: Always a bit skeptical of adding artificial sleeps to load times. Is there any way we can actually await the full render?

BSd3v: The components are there immediately, the debounce has a 50ms delay. I'm not sure there is a way to handle this.

I'm open to figuring out how to delay and wait for something on the screen to demonstrate the proper sizing. XD

ndrezn: cc @T4rk1n

T4rk1n (approved):
:woman_dancing:

Issue #3252, opened 2025-03-31
[BUG] Dash 3 dcc.Tab cannot be updated in a callback


AnnMarieW:
It looks like the component is not being re-rendered when a prop in the dcc.Tab component is updated in a callback.

As reported on the [forum](https://community.plotly.com/t/dynamically-changing-disable-state-of-dbc-tab/91527) This app works in dash 2.18.2 but not in dash 3.0.1

```python
import dash
from dash import dcc, html, Input, Output

app = dash.Dash()

app.layout = html.Div(
    [
        html.Button("Enable Tabs", id="button", n_clicks=0),
        dcc.Store(id="store-data", data=None),
        dcc.Tabs(
            [
                dcc.Tab(label="Tab A", value="tab-a", id="tab-a", disabled=True),
                dcc.Tab(label="Tab B", value="tab-b", id="tab-b", disabled=True),
            ],
            id="tabs",
            value="tab-a",
        ),
    ]
)


@app.callback(Output("store-data", "data"), Input("button", "n_clicks"))
def update_store_data(clicks):
    if clicks > 0:
        return {"data": "available"}
    return None


@app.callback(
    Output("tab-a", "disabled"), Output("tab-b", "disabled"),
    Input("store-data", "data"),
)
def toggle_tabs(store_data):    
    if store_data is not None and "data" in store_data:       
        return False, False
    return True, True


if __name__ == "__main__":
    app.run(debug=True)



```

kummerer94:
This is a really annoying bug that we also noticed. It makes disabling / enabling tabs impossible. Only when you manually change the tab, is the tab really disabled.

Another example to reproduce:

```python
from dash import Dash, Input, Output, callback, dcc, html

external_stylesheets = ["https://codepen.io/chriddyp/pen/bWLwgP.css"]

app = Dash(__name__, external_stylesheets=external_stylesheets)

app.layout = html.Div(
    [
        dcc.Checklist(["Yes"], ["Yes"], id="checklist"),
        dcc.Tabs(
            [
                dcc.Tab(label="Tab one", id="tab1", value="tab1", children=["FIRST"]),
                dcc.Tab(
                    label="Tab two",
                    id="tab2",
                    value="tab2",
                    children=["SECOND"],
                ),
                dcc.Tab(
                    label="Tab three",
                    id="tab3",
                    value="tab3",
                    children=["THIRD"],
                ),
            ],
            id="tabs",
        ),
    ],
)


[...]

PR #3268, opened 2025-04-10
allows for parents to listen to descendent updates

BSd3v:
In Dash 3, it has become apparent that some components need to listen to updates in descendent components, and to rerender again. This PR adds this ability for component devs to subscribe to updates in descendents by placing `dashChildrenUpdate = true` on their component.

Currently, there is no limit to the scope that this will rerender for.

This fixes: https://github.com/plotly/dash/issues/3252


T4rk1n (approved):
:woman_dancing: Nice job, thank you.

Issue #3269, opened 2025-04-10
[BUG] dcc.Tooltip() incorect placement: "direction" and/or "bbox" params are not respected


embar-:
**Context**
After update from Dash 2.18.2 to 3.0.2, dcc.Tooltip is placed in wrong position and partially hide target objects.

```
dash 3.0.2
dash-bootstrap-components 2.0.0
```
    - OS: Linux, Windows
    - Browsers: Firefox, Edge, Chrome

**Describe the bug**
dcc.Tooltip left top corner is always placed at same position (relative to "bbox") regardless of "direction" parameter, and/or ignores some "bbox" values (presumably  "x1" and "y1")

**Expected behavior**
Tooltip should be placed in direction according "direction" parameter relative to "bbox" rectangle area.

**Screenshots**
In Firefox 128.8.0esr:
![Image](https://github.com/user-attachments/assets/cfdd9b19-6eff-406c-892e-7996e558ce16)

In Chrome 134:
![Image](https://github.com/user-attachments/assets/5d8c902e-c427-43e1-8362-38b76c63fd7a)

embar-:
Sample code to reproduce bug and provided images:

```
import dash
from dash import dcc, html

app = dash.Dash(__name__)

app.layout = html.Div([
    html.Table(
        style={
            "margin": "50px",
            "background": "red",
            "minWidth": "50px",
            "minHeight": "50px",
        }
    ),
    dcc.Tooltip(
        "top",
        show=True,
        bbox={"x0": 50, "y0": 50, "x1": 100, "y1": 100},
        direction='top'
    ),
    html.Table(
        style={
            "margin": "50px",
            "background": "yellow",
            "minWidth": "50px",
            "minHeight": "50px",
        }
    ),
    dcc.Tooltip(
        "bottom",
        show=True,
        bbox={"x0": 50, "y0": 150, "x1": 100, "y1": 200},
        direction='bottom'
    ),
    html.Table(
        style={
            "margin": "50px",
            "background": "green",
            "minWidth": "50px",
            "minHeight": "50px",
        }
    ),
    dcc.Tooltip(
        "left",
        show=True,
        bbox={"x0": 50, "y0": 250, "x1": 100, "y1": 300},
        direction="left"
    ),
    html.Table(
        style={
            "margin": "50px",
            "background": "blue",
            "minWidth": "50px",
            "minHeight": "50px",
        }
    ),
    dcc.Tooltip(
        "right",
        show=True,
[...]

PR #3271, opened 2025-04-11
fixing issue with tooltip styling

BSd3v:
fixes https://github.com/plotly/dash/issues/3269


gvwilson (approved):
Why does this fix it?

BSd3v:
The issue was that the `LoadingElement` wasnt able to receive the `jsx` generated classname, therefore the styling was never applied.

Reverting to a simple `div` allows for the jsx classname to be passed once again. The `isLoading` is already utilized in the component, so just used the attribute from it instead of applying the `LoadingElement`.

This still needs a real test made. 🙂

PR #3273, opened 2025-04-14
Fix dash_hooks plugin entry point

### LINKED ISSUE #3272: [BUG] "dash-hooks" is an invalid group name in python and so plugins using it don't install
Initially reported in this community post https://community.plotly.com/t/plugin-cannot-be-installed-with-pip/91307

**Describe the bug**

follow the[ plugin guide in dash docs ](https://dash.plotly.com/dash-plugins-using-hooks?_gl=1*9cbi6o*_gcl_au*MTgwNjA2MjYzOS4xNzQxODYyODM5*_ga*MTYyMTAwMDc3Ny4xNzQxODYyODM5*_ga_6G7EE0JNSC*MTc0MjI3ODY2MS4xNi4xLjE3NDIyODA5MTMuMTEuMC4w#configure-the-dash-entrypoint-in-setup.py) and set `entry_points={"dash-hooks": ["callback_error_plugin = callback_error_plugin"]},` in the `setup.py` file so that the hooks gets added to dash apps automatically.

Trying to pip install the resulting package whl results in `ValueError: (‘Invalid group name’, ‘dash-hooks’)`  because hyphens aren't allowed from https://packaging.python.org/en/latest/specifications/entry-points/

> Group names must be one or more groups of letters, numbers and underscores, separated by dots (regex ^\w+(\.\w+)*$).

replacing the hyphen in this check in dash https://github.com/plotly/dash/blob/700b70690e37fd5cd9af1bbd08996cfd71f296d8/dash/_hooks.py#L252-L253 with an underscore and then modifying the setup.py file of my plugin accordingly fixed the issue for me though

cc @T4rk1n @LiamConnors

T4rk1n:
Fixes #3272

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## Fixed
 - [#3264](https://github.com/plotly/dash/pull/3264) Fixed an issue where moving components inside of children would not update the `setProps` path, leading to hashes being incorrect
 - [#3265](https://github.com/plotly/dash/pull/3265) Fixed issue where the resize of graphs was cancelling others
+- [#3273](https://github.com/plotly/dash/pull/3273) Fix hooks entry point, renamed from invalid hyphen `dash-hooks` to underscored `dash_hooks`. Fix [#3272](https://github.com/plotly/dash/issues/3272)
 
 ## Added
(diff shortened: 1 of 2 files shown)


ndrezn (approved):
@LiamConnors this will require docs changes FYI

PR #3279, opened 2025-04-16
Store callback Output value in persistence storage

### LINKED ISSUE #2678: [BUG]Component properties set through dynamic callbacks cannot be persisted.

petar-qb:
Fixes #2678

This PR:
- Solves #2678 
- Solves the [question](https://community.plotly.com/t/dash-callback-output-value-clears-the-persistence-of-the-dcc-components/88314) from the Dash Plotly forum.

TLDR:
This PR ensures that the callback output value is written (instead of being pruned) within the browser's persistence storage for its `dcc` components that has `persistence=True` set.

---

For the following code (also posted in the Dash Plotly forum question that's linked above):
```py

app = Dash(__name__)

app.layout = html.Div([
    html.Button("Select All", id="select_all_button"),
    dcc.Checklist(
        id='checkbox',
        options=[1, 2, 3],
        value=[],
        persistence=True,
        persistence_type='session'
    ),
])


@app.callback(
    Output('checkbox', 'value'),
    Input('select_all_button', 'n_clicks'),
    prevent_initial_call=True,
)
def select_all_options(n_clicks):
    return [1, 2, 3]


if __name__ == '__main__':
    app.run(debug=True)
```

Here's how it works from the `plotly/dash::dev`:

https://github.com/user-attachments/assets/065d3a46-e8a4-4585-b36a-2a4e95c087ba

Here's how it works from the `petar-qb/dash::feature/store-output-value-in-persistence-storage`:

https://github.com/user-attachments/assets/9d3020de-035a-4187-9858-63d16bcad9de

---

## Contributor Checklist

### optionals

### DIFF
--- a/dash/dash-renderer/src/actions/index.js
+++ b/dash/dash-renderer/src/actions/index.js
@@ -1,4 +1,4 @@
-import {once} from 'ramda';
+import {once, path} from 'ramda';
 import {createAction} from 'redux-actions';
 import {addRequestedCallbacks} from './callbacks';
 import {getAppState} from '../reducers/constants';
@@ -7,6 +7,7 @@ import cookie from 'cookie';
 import {validateCallbacksToLayout} from './dependencies';
 import {includeObservers, getLayoutCallbacks} from './dependencies_ts';
 import {computePaths, getPath} from './paths';
+import {recordUiEdit} from '../persistence';
 
 export const onError = createAction(getAction('ON_ERROR'));
 export const setAppLifecycle = createAction(getAction('SET_APP_LIFECYCLE'));
@@ -17,9 +18,19 @@ export const setHooks = createAction(getAction('SET_HOOKS'));
 export const setLayout = createAction(getAction('SET_LAYOUT'));
 export const setPaths = createAction(getAction('SET_PATHS'));
 export const setRequestQueue = createAction(getAction('SET_REQUEST_QUEUE'));
-export const updateProps = createAction(getAction('ON_PROP_CHANGE'));
 export const insertComponent = createAction(getAction('INSERT_COMPONENT'));
 export const removeComponent = createAction(getAction('REMOVE_COMPONENT'));
+
+// Change the variable name of the action
+export const onPropChange = createAction(getAction('ON_PROP_CHANGE'));
+
+export function updateProps(payload) {
+    return (dispatch, getState) => {
+        const component = path(payload.itempath, getState().layout);
+        recordUiEdit(component, payload.props, dispatch);
+        dispatch(onPropChange(payload));
+    };
+}
 
 export const addComponentToLayout = payload => (dispatch, getState) => {
     const {paths} = getState();
--- a/dash/dash-renderer/src/persistence.js
+++ b/dash/dash-renderer/src/persistence.js
@@ -501,46 +501,21 @@ export function prunePersistence(layout, newProps, dispatch) {
         depersistedProps = mergeRight(props, update);
     }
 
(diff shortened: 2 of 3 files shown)


dash/dash-renderer/src/persistence.js
```diff
@@ -501,46 +501,21 @@ export function prunePersistence(layout, newProps, dispatch) {
...
-                            `${propName}.${propPart}`,
-                            finalPersistence
-                        )
-                    );
-                }
-            } else {
-                finalStorage.removeItem(
-                    getValsKey(id, propName, finalPersistence)
-                );
-            }
```
T4rk1n, line 542: What did this do? It's inclusion/removal doesn't seem to have an effect.

dash/dash-renderer/src/actions/index.js
```diff
@@ -17,10 +18,20 @@ export const setHooks = createAction(getAction('SET_HOOKS'));
 export const setLayout = createAction(getAction('SET_LAYOUT'));
 export const setPaths = createAction(getAction('SET_PATHS'));
 export const setRequestQueue = createAction(getAction('SET_REQUEST_QUEUE'));
-export const updateProps = createAction(getAction('ON_PROP_CHANGE'));
 export const insertComponent = createAction(getAction('INSERT_COMPONENT'));
 export const removeComponent = createAction(getAction('REMOVE_COMPONENT'));
 
+// Change the variable name of the action
```
T4rk1n, line 24: :hocho: This comment will lose context once this PR is merged. `onPropChange` is the proper name for this variable anyway.

T4rk1n (commented):
Looks good, but there is a failing test rdps011. I am not sure if it's because the bahavior changed to persist value on callbacks or if it's a regression.

Maybe @alexcjohnson can provide more insights on this test?

dash/dash-renderer/src/actions/index.js
petar-qb: Removed 👍

dash/dash-renderer/src/persistence.js
petar-qb: The issue with this code is that it clears the persisted value before the new selection is written via `recordUiEdit`, once the callback is triggered. For example, if the storage initially contains `[[1, 2], []]` and `callback_select_all` (which selects `[1, 2, 3]`) is triggered:
1. The removed code clears the persistence storage prematurely which causes the original value (`originalVal`) to be lost, which is problematic.
2. Then `recordUiEdit` writes `[[1, 2, 3], [1, 2]]` into storage.

This is incorrect because what should be persisted is `[[1, 2, 3], []]`. 

I might need to handle this differently. What do you think?

petar-qb:
@T4rk1n the changelog is added https://github.com/plotly/dash/pull/3279/commits/fc5eea6268b71cbfe2cef36dc162429cd7e8393b. Happy to see this fix merged and released soon 🥳 

Thanks for all your support!

petar-qb:
P.S. the `cd components/dash-table && npm ci` failed in the CI tests due to "network connectivity" issues. These tests passed before I pushed the minor changelog changes so it's flaky.

T4rk1n (approved):
:woman_dancing:

PR #3281, opened 2025-04-16
Use routes_pathname_prefix for internal Alive URL in Dash app

Chiragasourabh:
When running a Dash app behind a proxy, the internal alive check (used during app startup) fails if the app is initialized with only `requests_pathname_prefix`, or if `requests_pathname_prefix` differs from `routes_pathname_prefix`.

This results in a misleading error:
```
OSError: Address 'http://127.0.0.1:9090' already in use.
Try passing a different port to run.
```

This is due to the alive check attempting to construct the URL using `requests_pathname_prefix`, which may not route correctly internally.

###  Solution

This PR updates the internal alive URL construction logic to use `routes_pathname_prefix`, which is the correct prefix for server-side routing and internal endpoint resolution.

### This

- Prevents startup failures when `requests_pathname_prefix` != `routes_pathname_prefix`
- Ensures internal health check (alive check) functions correctly
- Unblocks deployment in more complex reverse-proxy setups

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals

### DIFF
--- a/dash/_jupyter.py
+++ b/dash/_jupyter.py
@@ -294,6 +294,8 @@ class JupyterDash:
         else:
             requests_pathname_prefix = "/"
 
+        routes_pathname_prefix = app.config.get("routes_pathname_prefix", "/")
+
         # FIXME Move config initialization to main dash __init__
         # low-level setter to circumvent Dash's config locking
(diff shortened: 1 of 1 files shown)


T4rk1n (approved):

PR #3289, opened 2025-04-24
allow_duplicates to work with Loading

### LINKED ISSUE #3288: [BUG] Allow dcc.Loading's target_components to work with allow_duplicate=True Outputs
Hi,

I observed that the target_components dictionary available in the dcc.Loading component does not work with components, which use the keyword argument allow_duplicate=True in their respective Output(...) definition.

Using debug=True in my example below indicates, that this stems from the fact, that allow_duplicate=True adds an @<HASH_VALUE> to the property name and hence the property referenced in the target_components dictionary cannot be found.

A pre-discussion of the issue can be found in the Plotly forum: [https://community.plotly.com/t/using-dcc-loading-spinner-with-target-components-and-an-allow-duplicate-true-output/91998?u=robster](url)

I have tested this behaviour with Dash 2.18.2 und Dash 3.0.3. Moreover for my example below I tested with dbc 1.7.1 and 2.0.2 respectively.

Best

Robert

--

```
from time import sleep

from dash import Dash, html, Input, Output
from dash.dcc import Loading
from dash_bootstrap_components import Input as dbcInput

app = Dash(__name__)

id_str = 'text-input'

app.layout = html.Div(children=html.Table(children=[
    html.Tr(children=html.Td(children=Loading(
            id='text-input-spinner',
            debug=True,
            target_components={id_str: 'value'},
            type='default',
            children=html.Div(children=dbcInput(
                    size='sm',
                    id=id_str,
                    type='text')
            )))),
    html.Tr(children=html.Td(html.Button(children='Click Me',
                                         id='example-button'))),
    html.Tr(children=html.Td(html.Label(children='Dummy',
                                        id='le-label')))]))

@app.callback(output=[Output(id_str, 'value'),
                      Output('le-label', 'children')],
              inputs=[Input('example-button', 'n_clicks')],
              prevent_initial_call=True)
def update_output(_: int):
    sleep(2)
    return 'dummy_value', 'Check'

if __name__ == '__main__':
    app.run(debug=True)
```

BSd3v:
fixes #3288


components/dash-core-components/src/components/Loading.react.js
```diff
@@ -45,9 +45,9 @@ const loadingSelector = (componentPath, targetComponents) => state => {
                             return false;
                         }
                         if (Array.isArray(target)) {
-                            return includes(l.property, target);
+                            return includes(l.property.split('@')[0], target);
```
T4rk1n, line 48: Think it might be better to set that in the reducer or the action.

There is a `// duplicate outputs` comment that I think may have been a todo here: https://github.com/plotly/dash/blob/9cc96271b4574c4587836467031366f7cab18b8e/dash/dash-renderer/src/reducers/loading.ts#L36

To change on the callback would be here: https://github.com/plotly/dash/blob/3f87c9b92a02022416631823a9da292d7976d1c0/dash/dash-renderer/src/actions/callbacks.ts#L751

BSd3v: I was curious about that, didnt want to mess around with the hash in the callback. Wasnt sure about side-effects of that change.

T4rk1n (approved):
:woman_dancing:

Issue #3067, opened 2024-11-08
allow optional callback inputs


RenaudLN:
## Problem
In a number of cases, I have callbacks that have Inputs/States that are only conditionally present on the page, which causes console errors (ReferenceError), and prevents the callback to run.

Examples where this could happen:
* A global store at the app level being updated from inputs on a given page -> this will raise the ReferenceError log when I am on a different page
* Several objects that are conditionally present on a page and update another element

## Expected solution
I would like to be able to mark inputs as optional, in which case they would default to None in the callback when the element is not present.

## Current workaround
In some cases I can use a pattern matching callback with `ALL` to make the input optional (no input will give an empty list), however this feels hackish and does not always work with existing pattern-matching ids.

PR #3294, opened 2025-04-29
adds `allow_optional` to State and Input to place no value as the placeholders

BSd3v:
Sometimes there is a need to have conditional inputs/states in callbacks, this allows this to happen by passing the conditions to the clientside to be parsed and null values passed back.

This is accessible by `allow_optional=True` on the Input/State.

fixes: https://github.com/plotly/dash/issues/3067

### DIFF
--- a/dash/dependencies.py
+++ b/dash/dependencies.py
@@ -35,6 +35,7 @@ class DashDependency:  # pylint: disable=too-few-public-methods
     allow_duplicate: bool
     component_property: str
     allowed_wildcards: Sequence[_Wildcard]
+    allow_optional: bool
 
     def __init__(self, component_id: ComponentIdType, component_property: str):
 
@@ -45,6 +46,7 @@ class DashDependency:  # pylint: disable=too-few-public-methods
 
         self.component_property = component_property
         self.allow_duplicate = False
+        self.allow_optional = False
 
     def __str__(self):
         return f"{self.component_id_str()}.{self.component_property}"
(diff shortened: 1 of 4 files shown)


dash/dependencies.py
```diff
@@ -134,13 +137,31 @@ def __init__(
...
-    allowed_wildcards = (MATCH, ALL, ALLSMALLER)
+    def __init__(
+        self,
+        component_id: ComponentIdType,
+        component_property: str,
+        allow_optional: bool = False,
+    ):
+        super().__init__(component_id, component_property)
+        self.allow_optional = allow_optional
+        self.allowed_wildcards = (MATCH, ALL, ALLSMALLER)
```
T4rk1n, line 162: That was a class attribute before, it doesn't change on the object and I mistyped the typing on L37 it should finish in `= ()` to be a class attribute.

Searching for reference usage, there is none, I think it could actually be removed. The actual check happens on the frontend and they are defined here: https://github.com/plotly/dash/blob/3488bbecacd50676cd512b1c7cdc3b4e4764acb6/dash/dash-renderer/src/actions/dependencies.js#L56-L60

dash/dependencies.py
```diff
@@ -56,7 +58,8 @@ def component_id_str(self) -> str:
         return stringify_id(self.component_id)
 
     def to_dict(self) -> dict:
-        return {"id": self.component_id_str(), "property": self.component_property}
+        specs = {"id": self.component_id_str(), "property": self.component_property}
+        return {**specs, "allow_optional": True} if self.allow_optional else specs
```
T4rk1n, line 62: This is kinda weird to read, should be:
```python
if self.allow_optional:
    specs["allow_optional"] = True
return specs
```

T4rk1n (approved):
:woman_dancing: Look good, just a minor code style comment.

PR #3305, opened 2025-05-26
Fix CI

T4rk1n:
I've identified two different issue in the recent CI failures:
- setuptools >=80.0.0 changed something with the path resolution of local installed package, breaking the typing tests. I also migrated the tests to run on github workflow instead.
- Migrated the background callback tests to run on github workflow instead of circleci, the tests now passes. There is a new v2 container on circleci that seems to be the culprit.

Also ignore the new html attribute from #3295


gvwilson:
:+1: well done

PR #3323, opened 2025-06-04
Make dash wsgi compliant

### LINKED ISSUE #3131: Remove server=app.server requirement for Gunicorn
Our usual recommendation for deployment is `gunicorn` as the production server, and you set up a `Procfile` or similar with:
```console
gunicorn app:server --workers 4
```

But wait... what's `server`? Well, you need to add another line to your `app.py` (that's the module `app` referred to before the colon):
```python
app = Dash()

...

server = app.server
```

Ugh, a code change! Ripe for problems. Instead we could add a method to Dash, e.g. `wsgi_app` which returns the flask app, and this way the `Procfile` can be written as:
```console
gunicorn app:app --workers 4
```

And the `app` object already in your Dash app is picked up. Simpler, one fewer code change. nice. Vizro does this: https://vizro.readthedocs.io/en/stable/pages/user-guides/run/#jupyter .

T4rk1n:
Resolve #3131

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -2395,3 +2395,11 @@ class Dash(ObsoleteChecker):
                 Output(_ID_DUMMY, "children"),
                 Input(_ID_STORE, "data"),
             )
+
+    def __call__(self, environ, start_response):
+        """
+        This method makes instances of Dash WSGI-compliant callables.
+        It delegates the actual WSGI handling to the internal Flask app's
+        __call__ method.
+        """
+        return self.server(environ, start_response)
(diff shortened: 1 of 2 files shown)


dash/dash.py
```diff
@@ -2395,3 +2395,11 @@ def update(pathname_, search_, **states):
                 Output(_ID_DUMMY, "children"),
                 Input(_ID_STORE, "data"),
             )
+
+    def __call__(self, environ, start_response):
```
syl-oh, line 2399: nice 👌 no need to add `server = app.server` anymore

emilykl:
This is huge -- thanks @T4rk1n @ndrezn !

PR #3329, opened 2025-06-09
Add plotly run command

T4rk1n:
Add a command to run dash app from the terminal.

The command can be run two methods:
- `plotly run app:app --debug`
- `python -m dash run app:app --debug`

The arguments follow `Dash.run`, run help:
```
$ plotly run --help
usage: plotly run [-h] [--host HOST] [--port PORT] [--proxy PROXY] [--debug | --no-debug | -d] [--dev-tools-ui | --no-dev-tools-ui] [--dev-tools-props-check | --no-dev-tools-props-check]
                  [--dev-tools-serve-dev-bundles | --no-dev-tools-serve-dev-bundles] [--dev-tools-hot-reload | --no-dev-tools-hot-reload] [--dev-tools-hot-reload-interval DEV_TOOLS_HOT_RELOAD_INTERVAL]
                  [--dev-tools-hot-reload-watch-interval DEV_TOOLS_HOT_RELOAD_WATCH_INTERVAL] [--dev-tools-hot-reload-max-retry DEV_TOOLS_HOT_RELOAD_MAX_RETRY]
                  [--dev-tools-silence-routes-logging | --no-dev-tools-silence-routes-logging] [--dev-tools-disable-version-check | --no-dev-tools-disable-version-check] [--dev-tools-prune-errors | --no-dev-tools-prune-errors]
                  app

Run a local development server for a Dash app.

positional arguments:
  app                   The Dash app to run, in the format "module:variable".

options:
  -h, --help            show this help message and exit
  --host HOST           Host IP used to serve the application (Default: "127.0.0.1").
  --port PORT, -p PORT  Port used to serve the application (Default: "8050").
  --proxy PROXY         Proxy configuration string, e.g., "http://0.0.0.0:8050::https://my.domain.com".
  --debug, --no-debug, -d
                        Enable/disable Flask debug mode and dev tools.

dev tools options:
  --dev-tools-ui, --no-dev-tools-ui
                        Enable/disable the dev tools UI.
  --dev-tools-props-check, --no-dev-tools-props-check
                        Enable/disable component prop validation.
  --dev-tools-serve-dev-bundles, --no-dev-tools-serve-dev-bundles
                        Enable/disable serving of dev bundles.
  --dev-tools-hot-reload, --no-dev-tools-hot-reload
                        Enable/disable hot reloading.
  --dev-tools-hot-reload-interval DEV_TOOLS_HOT_RELOAD_INTERVAL
                        Interval in seconds for hot reload polling (Default: 3).
  --dev-tools-hot-reload-watch-interval DEV_TOOLS_HOT_RELOAD_WATCH_INTERVAL
                        Interval in seconds for server-side file watch polling (Default: 0.5).
  --dev-tools-hot-reload-max-retry DEV_TOOLS_HOT_RELOAD_MAX_RETRY
                        Max number of failed hot reload requests before failing (Default: 8).
  --dev-tools-silence-routes-logging, --no-dev-tools-silence-routes-logging
                        Enable/disable silencing of Werkzeug's route logging.
  --dev-tools-disable-version-check, --no-dev-tools-disable-version-check
                        Enable/disable the Dash version upgrade check.
  --dev-tools-prune-errors, --no-dev-tools-prune-errors
                        Enable/disable pruning of tracebacks to user code only.

```

### DIFF
--- a/dash/_cli.py
+++ b/dash/_cli.py
@@ -1,0 +1,172 @@
+import argparse
+import importlib
+import sys
+from typing import Any, Dict
+
+from dash import Dash
+
+
+def load_app(app_path: str) -> Dash:
+    """
+    Load a Dash app instance from a string like "module:variable".
+
+    :param app_path: The import path to the Dash app instance.
+    :return: The loaded Dash app instance.
+    """
+    if ":" not in app_path:
+        raise ValueError(
+            f"Invalid app path: '{app_path}'. "
+            'The path must be in the format "module:variable".'
+        )
+
+    module_str, app_str = app_path.split(":", 1)
+
+    if not module_str or not app_str:
+        raise ValueError(
+            f"Invalid app path: '{app_path}'. "
+            'Both module and variable names are required in "module:variable".'
+        )
+
+    try:
+        module = importlib.import_module(module_str)
+    except ImportError as e:
+        raise ImportError(f"Could not import module '{module_str}'.") from e
+
+    try:
+        app_instance = getattr(module, app_str)
+    except AttributeError as e:
+        raise AttributeError(
+            f"Could not find variable '{app_str}' in module '{module_str}'."
+        ) from e
+
+    if not isinstance(app_instance, Dash):
+        raise TypeError(f"'{app_path}' did not resolve to a Dash app instance.")
+
+    return app_instance
+
+
+def create_parser() -> argparse.ArgumentParser:
+    """Create the argument parser for the Plotly CLI."""
+    parser = argparse.ArgumentParser(
+        description="A command line interface for Plotly Dash."
+    )
+    subparsers = parser.add_subparsers(dest="command", required=True)
+
+    # --- `run` command ---
+    run_parser = subparsers.add_parser(
+        "run",
+        help="Run a Dash app.",
+        description="Run a local development server for a Dash app.",
+    )
+
+    run_parser.add_argument(
+        "app", help='The Dash app to run, in the format "module:variable".'
+    )
+
+    # Server options
+    run_parser.add_argument(
+        "--host",
+        type=str,
+        help='Host IP used to serve the application (Default: "127.0.0.1").',
+    )
+    run_parser.add_argument(
+        "--port",
+        "-p",
+        type=int,
+        help='Port used to serve the application (Default: "8050").',
+    )
+    run_parser.add_argument(
+        "--proxy",
+        type=str,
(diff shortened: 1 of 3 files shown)


dash/_cli.py
syl-oh, line 1: Do we want to use that in the CI for some test to avoid regression?

syl-oh (commented):
Nice, that'll make remembering all the debug flags easier

T4rk1n:
> @T4rk1n What do you think about using `app` as the default variable name if no variable name is provided? So e.g. if you had a file named `my_dash_app.py` containing an app variable called `app`, you could just run
> 
> `plotly run app --debug`

Yes that might be more convenient :+1: 



> @T4rk1n For the flags, is there a way to make the helptext clearer to the user about which option is the default? I.e. if I can pass `--debug` or `--no-debug`, how do I know what is the default behavior if I pass neither?
> 
> For that matter (since I do know that no-debug is the default), what is the use case for the `--no-debug` flag?

Oh, the `--no-<option>` is added by `argparse.BooleanOptionalAction`, I didn't notice them before but they don't make much sense since those are flags, I'll see how to remove them.

dash/_cli.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def load_app(app_path: str) -> Dash:
+    """
+    Load a Dash app instance from a string like "module:variable".
+
+    :param app_path: The import path to the Dash app instance.
+    :return: The loaded Dash app instance.
+    """
+    app_split = app_path.split(":")
```
gvwilson, line 16: check that `len(app_split) == 2` here?

gvwilson: OK, I see that further down you're allowing users to provide just the module name, then using the first `Dash` you find in that module.

dash/_cli.py
```diff
@@ -0,0 +1,171 @@
...
+            raise AttributeError(
+                f"Could not find variable '{app_str}' in module '{module_str}'."
+            ) from e
+    else:
+        for module_var in vars(module).values():
+            if isinstance(module_var, Dash):
+                app_instance = module_var
+                break
+
+    if not isinstance(app_instance, Dash):
```
gvwilson, line 41: if I run with just the module name, but the module doesn't define a `Dash`, I think `app_instance` will be uninitialized at the point of the `isinstance()` check.

robertclaus:
@T4rk1n is this still relevant?

PR #2991, opened 2024-09-09
Added support for URL decoding of the search parameter for pages

yzcheung:
Here’s a structured PR description following the format you provided:

---

## Description of this PR

This PR adds support for URL decoding of the `search` parameter in pages mode. The changes ensure that URL-encoded characters in the `search` parameter are properly decoded, improving the handling of query strings in pages mode.

## Contributor Checklist

- [x] I have broken down my PR scope into the following TODO tasks
   - [x] Implement URL decoding for the `search` parameter in pages mode
- [ ] I have run the tests locally and they passed. (refer to testing section in [contributing](https://github.com/plotly/dash/blob/master/CONTRIBUTING.md))
- [ ] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR

### optionals

- [ ] I have added entry in the `CHANGELOG.md`
- [ ] If this PR needs a follow-up in **dash docs**, **community thread**, I have mentioned the relevant URLs as follows
   - [ ] this GitHub [#PR number]() updates the dash docs
   - [ ] here is the show and tell thread in Plotly Dash community

This structure aligns with the given format and provides clear documentation of the changes.

PR #3341, opened 2025-06-25
Fix query string parsing regression introduced in 2.18.2 #3106

### LINKED ISSUE #3106: Search param removes values after ampersand, introduced in 2.18.2
I pass to one of my pages a search string, like:
?param1=something&param2=something2

Accessing it uding:
def layout(**kwargs):

In 2.18.1, this works for values that included ampersand. For example:
?param1=something&bla&param2=something2
would result in 
kwargs[param1]='something&bla'

With 2.18.2 I get just:
kwargs[param1]='something'
with anything after the ampersand removed.

I would guess this is related to #2991 .

To be clear, I specifically downgraded dash to 2.18.1 and the issue went away.

C0deBeez:
# Fix query string parsing regression introduced in 2.18.2

This PR fixes a regression introduced in Dash 2.18.2 where query string values containing unencoded `&` characters were being truncated at the first `&`. 

**Problem**: In 2.18.2, `?param1=something&bla&param2=something2` would result in `param1='something'` instead of the expected `param1='something&bla'` from 2.18.1.

**Root Cause**: The automatic URL decoding feature from #2991 caused `unquote()` to decode `%26` to `&`, which `parse_qs()` then misinterpreted as parameter delimiters.

**Solution**: Remove the automatic `unquote()` call to restore 2.18.1 behavior while maintaining support for properly encoded URLs.

Fixes #3106

## Contributor Checklist

## Changes Made

```python
# Before (2.18.2) - BROKEN
def _parse_query_string(search):
    search = unquote(search)   # REMOVED
    if search and len(search) > 0 and search[0] == "?":
        search = search[1:]
    else:
        parsed_qs = {}
    for (k, v) in parse_qs(search).items():
        v = v[0] if len(v) == 1 else v
        parsed_qs[k] = v
    return parsed_qs

# After (This Fix) - WORKING
def _parse_query_string(search):
    if not search or not search.startswith("?"):
        return {}

    query_string = search[1:]
    
    parsed_qs = parse_qs(query_string, keep_blank_values=True)
    
    return {
        k: v[0] if len(v) == 1 else v
        for k, v in parsed_qs.items()
    }

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [UNRELEASED]
 
 ## Fixed
+- [#3341](https://github.com/plotly/dash/pull/3341) Fixed query string parsing regression introduced in 2.18.2 where values containing unencoded `&` characters were being truncated. [#3106](https://github.com/plotly/dash/issues/3106)
 - [#3279](https://github.com/plotly/dash/pull/3279) Fix an issue where persisted values were incorrectly pruned when updated via callback. Now, callback returned values are correctly stored in the persistence storage. Fix [#2678](https://github.com/plotly/dash/issues/2678)
 - [#3298](https://github.com/plotly/dash/pull/3298) Fix dev_only resources filtering.
 - [#3315](https://github.com/plotly/dash/pull/3315) Fix pages module is package check.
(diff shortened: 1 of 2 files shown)


gvwilson:
@T4rk1n please review

T4rk1n (approved):

PR #3347, opened 2025-06-27
Api exposure callbacks

BSd3v:
This PR strives to allow callbacks to be exposed by the underlying server as an API by simply passing the callback a `api_endpoint`. The api will accept a body with the functional arguments provided as `kwargs`.

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -72,6 +72,7 @@ class NoUpdate:
 GLOBAL_CALLBACK_LIST = []
 GLOBAL_CALLBACK_MAP = {}
 GLOBAL_INLINE_SCRIPTS = []
+GLOBAL_API_PATHS = {}
 
 
 # pylint: disable=too-many-locals
@@ -87,6 +88,7 @@ def callback(
     cache_args_to_ignore: Optional[list] = None,
     cache_ignore_triggered=True,
     on_error: Optional[Callable[[Exception], Any]] = None,
+    api_path: Optional[str] = None,
     **_kwargs,
 ) -> Callable[..., Any]:
     """
@@ -178,6 +180,7 @@ def callback(
     )
     callback_map = _kwargs.pop("callback_map", GLOBAL_CALLBACK_MAP)
     callback_list = _kwargs.pop("callback_list", GLOBAL_CALLBACK_LIST)
+    callback_api_paths = _kwargs.pop("callback_api_paths", GLOBAL_API_PATHS)
 
     if background:
         background_spec: Any = {
@@ -217,12 +220,14 @@ def callback(
         callback_list,
         callback_map,
         config_prevent_initial_callbacks,
+        callback_api_paths,
         *_args,
         **_kwargs,
         background=background_spec,
         manager=manager,
         running=running,
         on_error=on_error,
+        api_path=api_path,
     )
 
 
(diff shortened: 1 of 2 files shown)


BSd3v:
Here is an example of how to spin up the app and how to perform the api calls:

app.py
```
import time
import dash
from dash import html, dcc, Input, Output, ctx, callback
from flask import jsonify
import asyncio
import json

app = dash.Dash(__name__, use_async=True)
app.layout = html.Div([
    html.Button("Slow Callback", id="slow-btn"),
    html.Div(id="slow-output"),
    html.Button("Fast Callback", id="fast-btn"),
    html.Div(id="fast-output"),
])

async def get_async_data(n_clicks):
    # Simulate an async data fetch
    await asyncio.sleep(1)
    return f"Data fetched - {n_clicks}"

def get_data(n_clicks):
    # Simulate an async data fetch
    time.sleep(1)
    return f"Data fetched - {n_clicks}"

@app.callback(
    Output("slow-output", "children"),
    Input("slow-btn", "n_clicks"),
    prevent_initial_call=True,
    api_endpoint='/api/slow_callback',  # Example API path for the slow callback
)
def slow_callback(n_clicks):
    start = time.time()
    data = {}
    for i in range(5):
        data[f'step_{i}'] = get_data(n_clicks)
    ret = f"{json.dumps(data)} Time taken: {time.time() - start:.2f} seconds"
    if ctx:
        return ret
    return jsonify(ret)

@app.callback(
    Output("fast-output", "children"),
    Input("fast-btn", "n_clicks"),
    prevent_initial_call=True,
    api_endpoint='/api/fast_callback',  # Example API path for the fast callback
)
async def fast_callback(n_clicks):
    start = time.time()
    coros = [get_async_data(n_clicks) for _ in range(5)]
    results = await asyncio.gather(*coros)
    data = {f'step_{i}_async': result for i, result in enumerate(results)}
    ret = f"{json.dumps(data)} Time taken: {time.time() - start:.2f} seconds"
    if ctx:
        return ret
    return jsonify(ret)

app.setup_apis()

if __name__ == "__main__":
    app.run(debug=True)
```

api calls

```
import requests

resp = requests.post('http://127.0.0.1:8050/api/slow_callback',
                     json={"n_clicks": 1},
                    headers={"Content-Type": "application/json"})

print(resp.json())

resp = requests.post('http://127.0.0.1:8050/api/fast_callback',
                     json={"n_clicks": 1},
                    headers={"Content-Type": "application/json"})

print(resp.json())
```

BSd3v:
@T4rk1n 

Is there another way to setup the apis rather than calling `app.setup_apis()`

antonymilne:
This seems very cool as a feature 🙌 But I can't think of where I would actually use it... Please could you give some examples of where you think this would be useful for inspiration?

BSd3v:
@antonymilne 

This would be useful for a few things:
- allowing someone to pull info from your system pertaining to their account
- allowing a company to add data to your db
- allowing someone to trigger something that you already have an interface setup to handle

In an api first landscape, having this built-in allows for better flexibility when it pertains to process flows, specifically when it comes to B2B. This also decreases developer load, you dont have to have a separate server to cater apis. As such, the process flow for the dash app and the api are essentially identical.

You can do things like require specific headers to be present, eg: `apiKey` before performing the task if there is no callback context available.

PR #3350, opened 2025-06-27
throw an error if there is more than 5000 slider marks

### LINKED ISSUE #3331: [BUG] Limit total tick marks on sliders
Slider components, when set with many tick marks, will freeze the UI. 

While developers may notice this and fix it, in AI systems setting too many tick marks causes problems in rendering really easily, and is hard to debug.

We can plausibly just restrict the total tick marks to ~5k or so as a maximum (most displays are 5k pixels wide, so 5k tick marks is more than enough).

T4rk1n:
resolve #3331

### DIFF
--- a/components/dash-core-components/src/fragments/Slider.react.js
+++ b/components/dash-core-components/src/fragments/Slider.react.js
@@ -17,6 +17,8 @@ import {
 } from '../utils/formatSliderTooltip';
 import LoadingElement from '../utils/LoadingElement';
 
+const MAX_SLIDER_MARKS = 5000;
+
 const sliderProps = [
     'min',
     'max',
@@ -75,6 +77,15 @@ export default class Slider extends Component {
             verticalHeight,
         } = this.props;
         const value = this.state.value;
+
+        // Validate marks count
+        if (marks && Object.keys(marks).length > MAX_SLIDER_MARKS) {
+            throw new Error(
+                'Slider marks cannot exceed 5000 items. Received ' +
+                    Object.keys(marks).length +
+                    ' marks.'
+            );
+        }
 
         let tipProps, tipFormatter;
         if (tooltip) {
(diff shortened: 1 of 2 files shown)


gvwilson (approved):
Looks good to me - @ndrezn please have a look as well.

ndrezn:
@chriddyp / @cleaaum is an error message here sufficient? I ask because in theory we expect an agent to respond to error messages and address them, so it will require an additional loop.

The other options here are:
1. Throw a warning, and/or
2. Truncate automatically to 5k or so ticks... but how do we decide how to downsample? Is 5k the right upper bound in this case? Usually we would want far fewer.

cleaaum:
I think using the default settings could be useful (~5 ticks) with a warning message that its done that - makes it a little easier to use

gvwilson:
@cleaaum would you want generated code to _always_ use the default, or do you want the library to switch to the 5-tick default if the requested number of ticks is too large?

emilykl:
If I understand correctly, the main issue we want to resolve is that if the user supplies too many ticks to `dcc.Slider` in their code, the Dash app becomes unresponsive in the browser — is that correct?

So we want to avoid the scenario where the Dash app is unresponsive (due to too many ticks). 

If that's correct, I think it would be reasonable for `dcc.Slider` to ignore the `marks` value if it's above some threshold and log a warning saying it's done so. So basically I agree with @cleaaum .

cleaaum:
@gvwilson the latter: would want the library to switch to the 5-tick default if the requested number of ticks is too large.

ndrezn:
Let's go with that route and start with 500 ticks, which feels like a reasonable starting place.

So:
1. If a user supplies > 500 ticks,
2. Default to standard behaviour (~5 ticks),
3. Raise a warning in the Javascript console (_not_ the Python console) that the tick count was reduced, as the overwrite happens in the frontend

Sound good?

T4rk1n:
closing in favor of #3365

PR #3352, opened 2025-06-27
run async test in separate step & flaky cbsc001

T4rk1n:
The async_tests fails or timeout often. Run them in a separate step and mark cbsc001 flaky.

### DIFF
--- a/tests/async_tests/test_async_callbacks.py
+++ b/tests/async_tests/test_async_callbacks.py
@@ -1,6 +1,8 @@
 import json
 import time
 
+import flaky
+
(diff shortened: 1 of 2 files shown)


tests/async_tests/test_async_callbacks.py
```diff
@@ -27,6 +29,7 @@
 from tests.utils import is_dash_async
 
 
+@flaky.flaky(max_runs=3)
```
gvwilson, line 32: I didn't know `flaky` existed - nice catch.

gvwilson (approved):

PR #3359, opened 2025-07-01
Make #3351 acceptable only for sync version (async version has self-owned if not config.suppress_callback_exceptions ... block)

### LINKED ISSUE #3358: [BUG] #3351 is too global
**Describe your context**

```
dash                 3.1.1
```

**Describe the bug**

#3351 has an effect on both the synchronous and asynchronous versions, whereas should only affect the synchronous

Erraen:
Fix #3358

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -2444,8 +2444,8 @@ class Dash(ObsoleteChecker):
                             else self.layout
                         ]
                     )
-                if _ID_CONTENT not in self.validation_layout:
-                    raise Exception("`dash.page_container` not found in the layout")
+                    if _ID_CONTENT not in self.validation_layout:
+                        raise Exception("`dash.page_container` not found in the layout")
             else:
 
                 @self.callback(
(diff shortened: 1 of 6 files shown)


gvwilson:
@T4rk1n please have a look

dash/dash.py
```diff
@@ -2491,27 +2491,27 @@ def update(pathname_, search_, **states):
...
+                        self.validation_layout = html.Div(
+                            [
+                                page["layout"]()
+                                if callable(page["layout"])
+                                else page["layout"]
+                                for page in _pages.PAGE_REGISTRY.values()
+                            ]
+                            + layout
+                        )
+                    if _ID_CONTENT not in self.validation_layout:
```
T4rk1n, line 2513: I think this one need to be indented one more level too.

Erraen: If you mean another additional indented for checking "if _ID_CONTENT ... " - then as far as I understand, no. Here, only the self.validation_layout property is redefined if necessary. It can also be set in the layout() method/property, and the checking should occur in both cases.

PR #3364, opened 2025-07-02
Add filepath to extract-meta errors

### LINKED ISSUE #3362: Expose file path in errors for dash-generate-components
Today, `dash-generate-components` can fail with confusing errors like:
```
ERROR: "__@iterator@3570" matches reserved word pattern: /^_.*$/

or

ERROR: "__html" matches reserved word pattern: /^_.*$/
```

and no context on where the failing component comes from.  We should surface the path to the failing component in the error message.

T4rk1n:
Fixes #3362


gvwilson (approved):

ndrezn:
@T4rk1n , is it possible to generate a tarball with a prerelease with these changes? Would facilitate testing early.

AnnMarieW:
This PR looks great to me!   

I actually ran into an error like the one reported in issue 3362 when I was working on DMC recently.  It failed during the build with this message and I was stuck. 

> ERROR: "__@iterator@3570" matches reserved word pattern: /^_.*$/


To test this PR, I checked out that failing commit in DMC and tried it with this solution.  There was no error message like the one above during the build, so I could take the next step and run the tests.  I was able to see this helpful error message.  This PR would have saved me lots of time!

```
        args = {k: _locals[k] for k in _explicit_args}
    
        for k in ['fontsize', 'length', 'localeCompare', 'match', 'normalize', 'search', 'split']:
            if k not in args:
>               raise TypeError(
                    'Required argument `' + k + '` was not specified.')
E               TypeError: Required argument `fontsize` was not specified.



```
@ndrezn do you need any other scenarios tested?

ndrezn:
I don't recall actually why I wanted early testing on this 🫠

T4rk1n:
There is extra changes in this PR that needs removed because I was trying to fix the CI.

While it skips: `symbol properties (e.g., __@iterator@3570, __@asyncIterator@3571, etc.)`, some of those are actually valid props that get their name mangled in some cases of wrapping external libraries props.

It should emit a warning but no error, think we can get the parent and the previous prop name to help identify the invalid prop.

T4rk1n:
reimplemented in #3895

PR #3371, opened 2025-07-16
Add devtool hook

T4rk1n:
- Add a new hook to add react/dash components to the devtool bar.
- Fix allow_optional triggering a warning for not found input.
- ExternalWrapper components can now target a base property out of the regular layout.

### DIFF
--- a/dash/dash-renderer/src/actions/dependencies.js
+++ b/dash/dash-renderer/src/actions/dependencies.js
@@ -5,6 +5,7 @@ import {
     any,
     ap,
     assoc,
+    concat,
     difference,
     equals,
     evolve,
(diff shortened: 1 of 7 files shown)


dash/dash-renderer/src/actions/dependencies.js
```diff
@@ -568,10 +569,20 @@ export function validateCallbacksToLayout(state_, dispatchError) {
     function validateMap(map, cls, doState) {
         for (const id in map) {
             const idProps = map[id];
+            const fcb = flatten(values(idProps));
```
gvwilson, line 572: I'm trying to guess what `fcb` stands for…

gvwilson (approved):

dash/dash-renderer/src/actions/dependencies.js
syl-oh: flatten callbacks?

gvwilson: aahh... thanks

CNFeffery:
> @CNFeffery you may find this one interesting...

At present, there is no relevant introduction in https://dash.plotly.com/dash-plugins-using-hooks. Let's first understand the operation mechanism and usage form through the test cases involved in this PR🚀.

PR #3415, opened 2025-08-26
Fix the client-side callback with multiple Outputs return single no_update

### LINKED ISSUE #3366: [BUG] Error caused by returning a single no_update for multiple Outputs in the client-side callback
```
dash                 3.1.1
```
For `client-side` callbacks, if for multiple `Outputs`, a single `no_update` is returned uniformly as in the previous version, an `Invalid array length` error will be triggered. The following is a case to reproduce the problem:

```python
import dash
from dash import html

app = dash.Dash(__name__)

app.layout = html.Div(
    [
        html.Button('test', id='test-input'),
        html.Div('block1', id='test-output1'),
        html.Div('block2', id='test-output2'),
    ],
    style={'padding': 50},
)

app.clientside_callback(
    """(n_clicks) => {
        return window.dash_clientside.no_update;
    }""",
    Output('test-output1', 'children'),
    Output('test-output2', 'children'),
    Input('test-input', 'n_clicks'),
)

if __name__ == '__main__':
    app.run(debug=True)
```

Image

CNFeffery:
Fix the error triggered when only a single no_update is returned for client-side callback functions with multiple Outputs, Closes #3366 .

### DIFF
--- a/dash/dash-renderer/src/actions/callbacks.ts
+++ b/dash/dash-renderer/src/actions/callbacks.ts
@@ -226,8 +226,16 @@ function refErr(errors: any, paths: any) {
 const getVals = (input: any) =>
     Array.isArray(input) ? pluck('value', input) : input.value;
 
-const zipIfArray = (a: any, b: any) =>
-    Array.isArray(a) ? zip(a, b) : [[a, b]];
+const zipIfArray = (a: any, b: any) => {
+    if (Array.isArray(a)) {
+        // For client-side callbacks with multiple Outputs, only return a single dash_clientside.no_update
+        if (b?.description === 'Return to prevent updating an Output.') {
+            return zip(a, [b]);
+        }
+        return zip(a, b);
+    }
+    return [[a, b]];
+};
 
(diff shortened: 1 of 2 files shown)


dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -226,8 +226,16 @@ function refErr(errors: any, paths: any) {
 const getVals = (input: any) =>
     Array.isArray(input) ? pluck('value', input) : input.value;
 
-const zipIfArray = (a: any, b: any) =>
-    Array.isArray(a) ? zip(a, b) : [[a, b]];
+const zipIfArray = (a: any, b: any) => {
+    if (Array.isArray(a)) {
+        // For client-side callbacks with multiple Outputs, only return a single dash_clientside.no_update
+        if (b?.description === 'Return to prevent updating an Output.') {
```
T4rk1n, line 232: Wouldn't `b === window.dash_clientside.no_update` be the same comparison since it's "singleton-like" variable?

PR #3424, opened 2025-09-01
brings patch to side-updates and allows for use in clientside

### LINKED ISSUE #3101: set_props does not work with Patch()
**Describe the bug**

`set_props` does not work with `Patch`. I guess, it replaces the Figure object with a Patch object instead of merging them. 

As a result, in order to update a figure partially, one has to use old-style `Output` in server-side callbacks.

I haven't seen anything in the docs that would suggest that this is expected behaviour.

**Expected behavior**

`set_props` works with `Patch` or `Figure` equally well when updating `Graph` component's `figure` property.

**Reproduce**

<details>

<summary>Code to reproduce</summary>

```python

import numpy as np
import plotly
from plotly import express as px, graph_objects as go, figure_factory as ff
from dash import Dash
from dash import html, dcc, callback, Input, Output, Patch, set_props

def make_figure():
    x = np.arange(6)
    y = np.random.randint(6, size=(10))
    return go.Figure(go.Scatter(x=x, y=y))


fig1 = make_figure()
button1 = html.Button("Output()", id="button1", n_clicks=0)
button2 = html.Button("set_props", id="button2", n_clicks=0)
children = [button1, button2, dcc.Graph(id="fig1", figure=fig1)]


@callback(
    Output("fig1", "figure", allow_duplicate=True),
    Input("button1", "n_clicks"),
    prevent_initial_call=True,
)
def regenerate_1(n_clicks):
    fig1 = make_figure()
    patch = Patch()
    patch["data"] = fig1["data"]
    fig1 = patch  # comment/uncomment this
    return fig1


@callback(
    Input("button2", "n_clicks"),
    prevent_initial_call=True,
)
def regenerate_2(n_clicks):
    fig1 = make_figure()
    patch = Patch()
    patch["data"] = fig1["data"]
    fig1 = patch  # comment/uncomment this
    set_props("fig1", {"figure": fig1})


app = Dash()
app.layout = html.Div(children)
app.run(jupyter_mode="external", port=8051)
```
</details>

<details>

<summary>`pip list | grep dash` below</summary>

```
dash                      2.18.1
dash-bootstrap-components 1.6.0
dash-mantine-components   0.12.1
```

</details>

BSd3v:
This brings patch functionality to:
- `set_props` both clientside and server side
- exposed a `PatchBuilder` class at 'Patch' on the clientside to build the user friendly version of the patch

fixes #3101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3395](https://github.com/plotly/dash/pull/3396) Add position argument to hooks.devtool
 - [#3403](https://github.com/plotly/dash/pull/3403) Add app_context to get_app, allowing to get the current app in routes.
 - [#3407](https://github.com/plotly/dash/pull/3407) Add `hidden` to callback arguments, hiding the callback from appearing in the devtool callback graph.
+- [#3424](https://github.com/plotly/dash/pull/3424) Adds support for `Patch` on clientside callbacks class `dash_clientside.Patch`, as well as supporting side updates, eg: (Running, SetProps).
 
 ## Fixed
 - [#3395](https://github.com/plotly/dash/pull/3395) Fix Components added through set_props() cannot trigger related callback functions. Fix [#3316](https://github.com/plotly/dash/issues/3316)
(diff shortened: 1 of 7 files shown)


T4rk1n (commented):
Looks good, just need a changelog entry and tests similar to the one we have for Python side.

PR #3427, opened 2025-09-02
issue with windows build lerna

BSd3v:
With the upgrade of lerna, windows ceased to build:
- fixes the Windows build
- reimplements the windows build test

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -91,6 +91,43 @@ jobs:
           path: packages/ # Upload the contents of the dist directory
           retention-days: 1 # Keep artifact for 1 day (adjust as needed)
 
+  build-windows:
+    name: Build Dash Package - Windows
+    runs-on: windows-latest
+    timeout-minutes: 30
+    outputs:
+      artifact_name: dash-packages
+    steps:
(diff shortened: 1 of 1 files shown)


gvwilson (approved):

robertclaus:
@T4rk1n could you take a look and see if this is still good to merge?

dash/development/update_components.py
```diff
@@ -43,7 +43,7 @@ def bootstrap_components(components_source, concurrency, install_type):
         else "{dash-core-components,dash-html-components,dash-table}"
     )
 
-    cmdstr = f"npx lerna exec --concurrency {concurrency} --scope='{source_glob}' -- npm {install_type}"
+    cmdstr = f"npx lerna exec --concurrency {concurrency} --scope={source_glob} -- npm {install_type}"
```
T4rk1n: We needed those `''` for spaces that might be in the directory path, while `'` is not windows compatible, `"` is and work on both os.

BSd3v: Just tried it, and it did not work.

BSd3v: It was an issue with the `posix`, needed to be passed `True`

Issue #1571, opened 2021-03-11
[Feature Request] FastAPI / other ASGI support


3tilley:
Hello!

We love Dash, it's been perfect for our team of Python developers who want to provide rich web frontends.

We're coming up against the issue that it's difficult to host alongside our other web services though, primarily because we use FastAPI. FastAPI is the most popular of the ASGI frontends but by no means the only one, if a solution could be found to support ASGI as well as WSGI, I think the other webservers could just slot in as well.

I've read up on various topics where this is mentioned, and aside from the non-trivial work to make it generic, it seems like the major blocker is Python 2.7 support, and Python 3.6 needed for async/await. I have every sympathy with having to support Enterprise Python, and I'm sure you're well aware of this, but Python 2.7 has been end of life for 15 months, and 3.5 has been end of life for six months.

Is there a roadmap for Dash 2.0? Would that be the time to support multiple backends and draw a line in the sand on versions?
---------------------
_References_

https://community.plotly.com/t/websockets-devices-performance-gains/39669
https://github.com/plotly/dash/issues/1559

alexcjohnson:
We are indeed working on a Dash 2.0 right now, but we're not quite ready to make the plans public yet - stay tuned, we'll be ready to share sometime this spring 🌸 

But I'm happy to say we will drop Py2 support with Dash 2.0, and I believe Py3.6 is a reasonable new minimum. According to https://pypistats.org/packages/dash we still have a couple percent of users installing on 3.5 (and another couple percent on 2.7 🙈 ) but hopefully this change will help encourage them to upgrade!

So if you're interested in contributing a feature that requires Py3.6, in a couple of months we'll happily accept it into the core dash repo 😄

3tilley:
Thanks!

Do you think that there would be appetite for a switchable backend at that point, instead of just Flask?

alexcjohnson:
As long as it's done in a way that's backward compatible for folks using the Flask server and interfacing directly with it via `app.server`, I think it should be fine to generalize to other back ends.

oTree-org:
I'm potentially interested in contributing/testing ASGI support. (I'm using Starlette/Uvicorn.)

3tilley:
Hello, I'm just circling back to this to ask if there is more clarity on Dash 2.0. I can see the announcements have been made, was it decided what versions of Python will be supported? Python 3.6 is EoL at the end of the year, so I wonder if even making 3.7 the minimum is worth considering, especially if the release will be in 2022.

[...]

PR #3430, opened 2025-09-09
Building support for decoupling Dash from Flask and supporting Quart / FastAPI servers

BSd3v:
This is an open PR draft, to contribute please target my forked branch.

The goal of this PR is to modularize the Dash setup to be independent of Flask (will fallback to Flask) and allow devs to configure their own backend.

- [x] decouple callback context from Flask Request
- [x] decouple index, assets, etc from build
- [x] build Quart server support
- [x] build FastAPI server support
- [x] allow for custom server build
- [x] add tests for servers

fixes https://github.com/plotly/dash/issues/1571

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@lerna/filter-options": "^6.4.1",
         "@types/jest": "^30.0.0",
         "husky": "8.0.3",
-        "lerna": "^7.4.2",
+        "lerna": "^9.0.4",
         "lint-staged": "^16.1.0"
       }
     },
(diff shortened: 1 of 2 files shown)


dash/_hooks.py
```diff
@@ -127,7 +126,7 @@ def route(
         Add a route to the Dash server.
         """
 
-        def wrap(func: _t.Callable[[], _f.Response]):
+        def wrap(func: _t.Callable[[], _t.Any]):
```
T4rk1n, line 129: This can be a future pain point to update existing hooks, maybe we can try adding a flask.Response converter in the adapter?

dash/backends/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+backend: BaseDashServer
+
+
+_backend_imports = {
+    "flask": ("dash.backends._flask", "FlaskDashServer"),
+    "fastapi": ("dash.backends._fastapi", "FastAPIDashServer"),
+    "quart": ("dash.backends._quart", "QuartDashServer"),
+}
```
T4rk1n, line 14: I think we should have a way for this to be customizable, maybe add a hook for that. Then the backend library can have the hook inside it's init.

dash/backends/_utils.py
```diff
@@ -0,0 +1,108 @@
...
+          </div>
+          <h2 class=\"traceback\">Traceback <em>(most recent call last)</em></h2>
+          {cards_html}
+          <blockquote>{error_type}: {error_msg}</blockquote>
+          <div class=\"plain\">
+            <p>This is the Copy/Paste friendly version of the traceback.</p>
+            <textarea readonly>{formatted_tb}</textarea>
+          </div>
+          <div class=\"explanation\">
+            The debugger caught an exception in your ASGI application. You can now
```
T4rk1n:
```suggestion
            The debugger caught an exception in your Dash application. You can now
```

dash/backends/base_server.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @property  # pragma: no cover - interface
+    @abstractmethod
+    def cookies(self):
+        raise NotImplementedError()
+
+    @property  # pragma: no cover - interface
+    @abstractmethod
+    def headers(self):
+        raise NotImplementedError()
```
T4rk1n, line 81: Maybe we can type those as `dict`, then make sure the backends returns a dict copied from all the cookies/headers.

In the same way, maybe we can add a single `get_cookie(...)` and `get_header(...)` for single usage.

dash/backends/base_server.py
```diff
@@ -0,0 +1,133 @@
...
+    def run(
+        self, dash_app, host: str, port: int, debug: bool, **kwargs
+    ) -> None:  # pragma: no cover - interface
+        pass
+
+    @abstractmethod
+    def make_response(
+        self, data, mimetype=None, content_type=None
+    ) -> Any:  # pragma: no cover - interface
+        pass
```
T4rk1n: The usual usecase for the response in the dash context is to set cookies/headers, maybe we should a global `set_cookie(...)` and `set_header` akin to `set_props` that can be used across the different backends with ease.

dash/backends/_fastapi.py
```diff
@@ -0,0 +1,537 @@
...
+    JSONResponse = None
+    RedirectResponse = None
+    StaticFiles = None
+    StarletteResponse = None
+    MutableHeaders = None
+    ASGIApp = None
+    Scope = None
+    Receive = None
+    Send = None
+    uvicorn = None
```
T4rk1n: Importing like this set the type to None creates types error in the editor. I don't think we need to set them to None, but instead raise back the import error  with a message to install `dash[fastapi]` if this file is imported without the proper libraries installed.

dash/backends/base_server.py
```diff
@@ -0,0 +1,133 @@
...
+        raise NotImplementedError()
+
+    @property  # pragma: no cover - interface
+    @abstractmethod
+    def path(self) -> str:
+        raise NotImplementedError()
+
+
+class BaseDashServer(ABC):
+    server_type: str
```
T4rk1n, line 167: Maybe `server_type` should be a class attribute, set to `"base"` and then it could be used without an instance to match in the auto resolution methods.

dash/backends/_fastapi.py
```diff
@@ -0,0 +1,537 @@
...
+    except (json.JSONDecodeError, OSError):
+        pass  # ignore errors
+    return resp
+
+
+def _remove_config():
+    try:
+        os.remove(_CONFIG_PATH)
+    except FileNotFoundError:
+        pass
```
T4rk1n: It's not really clear to me why we need that config file, maybe add some more context in comments.

dash/backends/_fastapi.py
```diff
@@ -0,0 +1,537 @@
...
+        request = Request(scope, receive=receive)
+        token = set_current_request(request)
+        try:
+            await self.app(scope, receive, send)
+        finally:
+            reset_current_request(token)
+
+
+# Internal config helpers (local to this file)
+_CONFIG_PATH = os.path.join(os.path.dirname(__file__), "dash_config.json")
```
T4rk1n: Since this is the same file all the time, this might cause issues if multiple servers are running are at the same time.

dash/backends/_fastapi.py
```diff
@@ -0,0 +1,537 @@
...
+            **{f"dev_tools_{k}": v for k, v in dev_tools.items()},
+        )
+        _save_config(config)
+        if debug:
+            if kwargs.get("reload") is None:
+                kwargs["reload"] = True
+        if kwargs.get("reload"):
+            # Dynamically determine the module name from the file path
+            file_path = frame.filename
+            spec = spec_from_file_location("app", file_path)
```
T4rk1n: We need the real file here, when I try with an app not named app with `debug=True` I get:
```
INFO:     Uvicorn running on http://127.0.0.1:8050 (Press CTRL+C to quit)
INFO:     Started reloader process [80425] using StatReload
ERROR:    Error loading ASGI app. Could not import module "app".
WARNING:  StatReload detected changes in 'quart_app.py'. Reloading...
ERROR:    Error loading ASGI app. Could not import module "app".
```

dash/backends/_fastapi.py
```diff
@@ -0,0 +1,537 @@
...
+    def add_redirect_rule(self, app, fullname, path):
+        self.server.add_api_route(
+            fullname,
+            self._create_redirect_function(app.get_relative_path(path)),
+            methods=["GET"],
+            name=fullname,
+            include_in_schema=False,
+        )
+
+    def dispatch(self, dash_app: Dash):
```
T4rk1n: Might be a good time to rename this to a better name like `serve_callback`.

dash/backends/_quart.py
```diff
@@ -0,0 +1,407 @@
...
+                return sync_view_func
+
+            view_func = _make_view_func(handler)
+            self.server.add_url_rule(
+                route, endpoint=endpoint, view_func=view_func, methods=methods
+            )
+
+    def _serve_default_favicon(self):
+        if Response is None:
+            raise RuntimeError("Quart not installed; cannot generate Response")
```
T4rk1n: This should not be happening, we should raise back the import error with a custom message to install `dash[quart]`.

dash/backends/_quart.py
```diff
@@ -0,0 +1,407 @@
...
+        return Response(
+            pkgutil.get_data("dash", "favicon.ico"), content_type="image/x-icon"
+        )
```
T4rk1n: This is the same for all the backends, maybe it be in the base and just call the proper make_response to handle that?

dash/backends/_quart.py
```diff
@@ -0,0 +1,407 @@
...
+            package.__path__,
+        )
+        data = pkgutil.get_data(package_name, path_in_pkg)
+        headers = {}
+        if has_fingerprint:
+            headers["Cache-Control"] = "public, max-age=31536000"
+
+        if Response is None:
+            raise RuntimeError("Quart not installed; cannot generate Response")
+        return Response(data, content_type=mimetype, headers=headers)
```
T4rk1n, line 279: This function will mostly be the same across backends with a difference in the Response type, in this case we are missing the etag stuff that might need to be adapted.

I think maybe we could have a Response type that is dash specific and then we have a ResponseAdapter like we do for the RequestAdapter.

dash/_hooks.py
BSd3v: We have the `self.backend.make_response` which uses the server adapters.

dash/backends/__init__.py
BSd3v: You can bring your own server by passing it in the Dash spin up, this is just to import them from the pre-configured backends.

dash/backends/_fastapi.py
BSd3v: This is needed due to a condition where the app looses its initial state of the settings, since this is managed not inside the specific thread. So, any arguments passed to it is lost as now the hot_reload is listening to file changes, but the main is never ran again.

dash/backends/_fastapi.py
BSd3v: I adjusted it to where if you run in `debug=True`, you need to have `server` exposed as the variable. This should make it more consistent.

dash/backends/_quart.py
BSd3v: I'm for changing this, but there are quite a few differences between the Responses, etc. 

I like your idea for Response information, I'm just not sure that its necessary for this as there may be more things that popup inside of this process for different servers that need to be handled inside of this function.

dash/backends/__init__.py
T4rk1n: I meant if an external lib wants to add a new backend and have it available as a string. eg: `app = Dash(backend="django")` it could add it via a hook.

Something like: `hooks.register_backend("django", "django_dash", "DjangoDashBackend")`

dash/_hooks.py
T4rk1n: It's more like I have a few route hooks that does flask.jsonify because it was typed before, they would need then to be updated for the different backends, but that is not a blocker. I think if we can add an abstraction for a specific dash Response that might be the better way to handle these cases.

dash/backends/_quart.py
```diff
@@ -0,0 +1,397 @@
...
+        request,
+        Blueprint,
+        g as quart_g,
+        has_request_context,
+        redirect,
+    )
+except ImportError:
+    raise ImportError(
+        "All dependencies not installed. Please install it with `dash[quart]` to use the Quart backend."
+    ) from None
```
T4rk1n:
```suggestion
except ImportError as _err:
    raise ImportError(
        "All dependencies not installed. Please install it with `dash[quart]` to use the Quart backend."
    ) from _err
```

dash/backends/_fastapi.py
```diff
@@ -0,0 +1,524 @@
...
+        self.request_adapter = FastAPIRequestAdapter
+        self._before_request_funcs = []
+
+        fname = inspect.stack()[2]
+        file_path = fname.filename
+        rel_path = os.path.relpath(file_path, os.getcwd())
+        module_name = os.path.splitext(rel_path)[0].replace(os.sep, ".")
+        # Internal config helpers (local to this file)
+        # This is used to persist dev tools config between reloads, since uvicorn runs a new process
+        self._CONFIG_PATH = os.path.join(os.path.dirname(file_path), f"_{module_name}_dash_config.json")
```
T4rk1n: But this create the file in the project folder. Let's make it a temp file in `~/.local/share/plotly-dash-configs/<hash.json>` with `appdirs.user_data_dir`.

Where the hash is generated from the filepath of the app file.

BSd3v: I swapped uvicorn run command for using a subprocess, it seems to be under a little more control over the config file.

maxschulz-COL:
Just came here to comment that this is very exciting :)

dash/dash.py
```diff
@@ -250,6 +244,12 @@ class Dash(ObsoleteChecker):
         ``flask.Flask``: use this pre-existing Flask server.
     :type server: boolean or flask.Flask
 
+    :param backend: The backend to use for the Dash app. Can be a string
+        (name of the backend) or a backend class. Default is None, which
+        selects the Flask backend. Currently, "flask", "fastapi", and "quart" backends
```
LiamConnors, line 249: Does the `server` param docstring above need an update too? 

Question - For the difference between the too - is it you'd use `server` if you want Dash to use an existing instance of the server? vs `backend` if you want Dash to create the instance?

BSd3v: Yes, I believe so.

You should be able to pass a preconfigured server of any of the supported types.

T4rk1n:
@BSd3v This is looking good, I have made some significant changes to the fastapi implementation. The config is now transfered via environment variable and I merged all the middlewares into one.

Could you review those changes, then we can merge and release a rc version :rocket:

BSd3v:
@T4rk1n 

I dont know that I necessarily agree with keeping the logic of `set_header` here, this appends to an already defined key, where as `set` makes the dev believe they are overriding. This function was done this way since `headers.add` in Flask is handled this way.

I think we should explicitly make it to where it is set, and if people want to append, then we can possibly expose a secondary headers `append_header`?

T4rk1n:
> @T4rk1n
> 
> I dont know that I necessarily agree with keeping the logic of `set_header` here, this appends to an already defined key, where as `set` makes the dev believe they are overriding. This function was done this way since `headers.add` in Flask is handled this way.
> 
> I think we should explicitly make it to where it is set, and if people want to append, then we can possibly expose a secondary headers `append_header`?

Yea I don't think we want to append but only set for the set_headers and add append method for clarity.

Issue #3442, opened 2025-09-17
[Feature Request] dcc.Slider - Enable user to define slider direction


areleu:
I would like to know if there is any interest on adding the feature requested by:

https://github.com/plotly/dash-core-components/issues/801#event-3304866528 

> I have an application where I am like to use sliders to crop an image (heatmap graph). The image's (0,0) is defined as the top left of the image. I'd like the y slider to start with 0 at the top, and end with the height of the image at the bottom. Currently, I cannot find a method to invert the slider to allow for the slider to go {top: low, bottom: high}, instead of the default {top: high, bottom: low}
> 
> Describe the solution you'd like
> A slider where the direction from minimum to maximum could be swapped.
> 
> Describe alternatives you've considered
> I tried setting the min to be higher than the max, did not work.
> I've tried flipping the slider with CSS and the result is... erratic

Unfortunately, it has been an issue for a while: https://community.plotly.com/t/slider-order-reversible/46656

gvwilson:
@areleu We'd be happy to add it, but don't have anyone we can put on it now (or in the foreseeable future). We'd be happy to prioritize review of a community PR if someone would like to create one. Thanks - @gvwilson

PR #3445, opened 2025-09-21
Added API to reverse direction of slider

enioarz:
Solves #3442 
## Contributor Checklist

   -  [ ] Extended the Slider API to allow reverse rendering. 

### optionals


KoolADE85:
> Looks okay, @KoolADE85 where would this fit with the redesign of the slider ?

It will need a similar prop passed to the new slider component, but that will be trivial to implement! Looks good to me too.

T4rk1n (approved):
:woman_dancing:

PR #3453, opened 2025-09-26
dcc redesign: checklist

KoolADE85:
This PR implements `dcc.Checklist` and `dcc.RadioItems` by extracting common functionality out of the Dropdown component into a reusable `OptionsList` component.

All 3 of these components now use the same code for displaying their lists of options.

Some of the type information in `types.ts` is also de-duplicated by extracting common attributes (like `id`, `setProps`) into a `BaseComponentProps` type.


components/dash-core-components/src/utils/optionRendering.tsx
```diff
@@ -0,0 +1,157 @@
...
+    optionClassName,
+    optionStyle,
+    inputType = 'checkbox',
+    inputClassName,
+    inputStyle,
+    labelClassName,
+    labelStyle,
+    index,
+    id,
+}) => {
```
T4rk1n, line 65: There is two style of function component, maybe just use one for consistency.

KoolADE85: done!

PR #3468, opened 2025-10-03
dcc redesign: TextArea and Tooltip

KoolADE85:
This PR is a small refactor of `Textarea` and `Tooltip` components to match the current designs and be Typescript components.


gvwilson:
@T4rk1n please review

components/dash-core-components/src/components/Textarea.tsx
```diff
@@ -0,0 +1,99 @@
...
+        if (['true', 'TRUE', 'True'].includes(value)) {
+            return true;
+        }
+        if (['false', 'FALSE', 'False'].includes(value)) {
+            return false;
+        }
+        return undefined;
+    }
+    return value;
+};
```
T4rk1n, line 41: I don't think those conversion are necessary, the props are transfered in json and bool and number. The prop type should be bool or number but the string interpolation makes it harder on the python type check.

PR #3488, opened 2025-10-29
Fix pkgutil.loader removal

### LINKED ISSUE #3139: Fix pkgutil.find_loader deprecation warning
Users might encounter:
```
'pkgutil.find_loader' is deprecated and slated for removal in Python 3.14; use importlib.util.find_spec() instead.
```

in Dash. To ensure forwards compatibility we should update to the new recommended method.

T4rk1n:
Recreation of #3476 with further fixes

Fix #3139


ndrezn (approved):

dash/development/base_component.py
camdecoster: Do you have a convention for using `import typing` vs. `from typing import ...`? If so, could you switch to that convention across all these type annotations?

camdecoster (approved):
Looks good. I made a few comments, but nothing stopping merge.

Issue #3451, opened 2025-09-26
[BUG] dcc.DatePickerRange class removed for first date after max_date_allowed


rzvcs:
**Describe your context**
I use `dcc.DatePickerRange` and `max_date_allowed` with CSS override of class `CalendarDay__blocked_out_of_range` to set a background for the dates beyond maximum. This helps our UX well. However, using the calendar to select a new range removes this class for the next date after `max_date_allowed`.

-  replace the result of `pip list | grep dash` below
```
dash                 3.2.0
```
-  if frontend related, tell us your Browser, Version and OS

    - OS: Windows 11 24H2 26100.6584
    - Browser: Chrome
    - Version: 140.0.7339.128

**Describe the bug**

Using dcc.DatePickerRange with `max_date_allowed` sets css class `CalendarDay__blocked_out_of_range` for days beyond the `max_date_allowed`. When clicking on a new start or end date in the calendar, the class is removed for date == max date allowed. Closing the calendar and opening it again repeats the whole behaviour.

Use sample code in `main.py`:
```python
from datetime import date
from dash import Dash, html, dcc

app = Dash(
    assets_folder="./assets"
)

STRFORMAT = "%Y-%m-%d"
app.layout = [
    html.Div(
        children=[
            'Repro', 
            dcc.DatePickerRange(
                start_date=date(2025,9,21).strftime(STRFORMAT),
                end_date=date(2025,9,25).strftime(STRFORMAT),
                max_date_allowed=date(2025,9,26).strftime(STRFORMAT),
                display_format='YYYY-MM-DD',
                id="my-date-range-picker",
                updatemode="bothdates"
            )
        ]
    )
]

if __name__ == '__main__':
    app.run(debug=True)
```

Create a folder `assets` at `main.py` location and add `style.css`:
```css
.CalendarDay__blocked_out_of_range {
    background-color: rgba(240, 240, 240, 1) !important;
}
```

**Expected behavior**

CSS Class should still be enforced.

**Screenshots**
Open the calendar picker and the 27th has the class applied:

<img width="925" height="393" alt="Image" src="https://github.com/user-attachments/assets/1c735f7c-f4d9-4c37-8864-5607c5dfe046" />

Click on a new date and the 27th doesn't have the class applied:

<img width="928" height="386" alt="Image" src="https://github.com/user-attachments/assets/c371733f-ecfe-4a02-b04a-f48310e28bc2" />

Farkites:
[...]

Issue #3425, opened 2025-09-02
[BUG] DatePickerRange updatemode not respected


rzvcs:
**Describe your context**
Please provide us your environment, so we can easily reproduce the issue.

-  replace the result of `pip list | grep dash` below
```
dash                          3.2.0
dash_ag_grid                  31.3.1
dash-bootstrap-components     1.7.1
dash-table                    5.0.0
```
-  if frontend related, tell us your Browser, Version and OS

    - OS: Windows 11
    - Browser: Chrome
    - Version: 139.0.7258.155

**Describe the bug**

When using dcc.DateRangePicker with updatemode="bothdates" from the UI, selecting the start date waits for the end date selection. However, selecting the end date triggers linked callbacks.

**Expected behavior**

In this updatemode, selecting the end date should wait for selecting the start date.

Farkites:
This issue is fixed with `dash==4.0.0rc3`

pierovera:
I am still seeing this issue with dash 4.1.0. The issue is exactly as described in the original report. I can provide additional info if needed.

elklaassen:
I also still see the issue with dash 4.2.0. I would like to add that after selecting start date, end date is (set to) None and as described before, callback is triggered before selecting end date. This causes my code to crash as I cannot have end date = None. Workarounds like using dcc.Store to preserve the end date value do not work for me.

PR #3495, opened 2025-11-07
dcc redesign: `DatePickerSingle` and `DatePickerRange`

### LINKED ISSUE #3391: update dcc date pickers
Tracking work to update the DCC date pickers.

KoolADE85:
At long last, here is a PR to modernize the date pickers.

You can play with them here:
https://dcc-redesign.plotly.app/datepickers-single
https://dcc-redesign.plotly.app/datepickers-range

Some improvements:
* Consistent with `dcc.Dropdown`'s look and feel
* Keyboard accessible with focus carefully managed
* Support for mouse drag & wheel (and click of course)
* DatePickerRange supports selecting backwards as well as forwards
* A range of tests added for both new and existing features.

Closes #3391 #3451 #3425
Fixes https://github.com/plotly/dash-design-kit/issues/1504

### DIFF
--- a/components/dash-core-components/src/fragments/DatePickerRange.tsx
+++ b/components/dash-core-components/src/fragments/DatePickerRange.tsx
@@ -143,7 +143,7 @@ const DatePickerRange = ({
             // Only end changed - send just that one
             setProps({end_date: dateAsStr(internalEndDate)});
         }
-    }, [start_date, internalStartDate, end_date, internalEndDate, updatemode]);
+    }, [internalStartDate, internalEndDate, updatemode]);
 
     const sendStartInputAsDate = useCallback(
         (focusCalendar = false) => {
--- a/components/dash-core-components/src/fragments/DatePickerSingle.tsx
+++ b/components/dash-core-components/src/fragments/DatePickerSingle.tsx
@@ -76,7 +76,7 @@ const DatePickerSingle = ({
         if (dateChanged) {
             setProps({date: dateAsStr(internalDate)});
         }
-    }, [date, internalDate, setProps]);
+    }, [internalDate]);
 
     const parseUserInput = useCallback(
         (focusCalendar = false) => {
--- a/components/dash-core-components/tests/integration/calendar/test_date_picker_range.py
+++ b/components/dash-core-components/tests/integration/calendar/test_date_picker_range.py
@@ -1,6 +1,7 @@
 from datetime import datetime
+import time
 
-from dash import Dash, html, dcc
+from dash import Dash, Input, Output, html, dcc
 from selenium.common.exceptions import (
     ElementClickInterceptedException,
     TimeoutException,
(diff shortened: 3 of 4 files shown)


KoolADE85:
Thanks @AnnMarieW ! And wow, I did not realize moment was fading away like that! I will take a look at replacing moment - it shouldn't be too much effort there.

> Wow! First look at this is amazing 🤩
> 
> I love how you can easily skip months and years - that's been a pain point in the old date pickers for a long time.
> 
> This is an incredible amount of work given that the Radix components does not have date pickers and you had to build all of this from scratch.
> 
> I see you are still using moment.js for the date formatting. Since they are no longer recommending moment.js to be used in new project, would you consider using one of their recommended alternatives? It might make it easier to add support for locale at some point https://momentjs.com/docs/#/-project-status/

components/dash-core-components/src/components/DatePickerRange.tsx
```diff
@@ -0,0 +1,70 @@
+import React, {lazy, Suspense} from 'react';
+import datePickerRange from '../utils/LazyLoader/datePickerRange';
+import transformDate from '../utils/DatePickerPersistence';
+import {DatePickerRangeProps, PersistedProps, PersistenceTypes} from '../types';
+
+const RealDatePickerRange = lazy(datePickerRange);
+
+/**
+ * DatePickerRange is a tailor made component designed for selecting
+ * timespan across multiple days off of a calendar.
```
T4rk1n: I'd remove the `tailor made` from the description.

components/dash-core-components/src/fragments/DatePickerRange.tsx
```diff
@@ -0,0 +1,447 @@
...
+import {
+    dateAsStr,
+    strAsDate,
+    formatDate,
+    isDateDisabled,
+    isSameDay,
+} from '../utils/calendar/helpers';
+import '../components/css/datepickers.css';
+import uuid from 'uniqid';
+import moment from 'moment';
```
T4rk1n: These imports should be grouped/sorted, third parties first then local file.

T4rk1n (commented):
Looks good.

T4rk1n:
> I see you are still using moment.js for the date formatting. Since they are no longer recommending moment.js

I would also recommend we change to something else, I don't think we use that much moment.js. Luxon seems like a good alternative, I also had a good experience with `date-fns` before.

KoolADE85:
> > I see you are still using moment.js for the date formatting. Since they are no longer recommending moment.js
> 
> I would also recommend we change to something else, I don't think we use that much moment.js. Luxon seems like a good alternative, I also had a good experience with `date-fns` before.

Oh nice, `date-fns` looks super lightweight, I'll give it a try!
Just want to call out... we document `moment`-style formatting for props like `display_format` and `month_format`. I guess in the interest of backwards-compatibility, I will write something that maps `moment`-style formats to `date-fns`-style (😒)

components/dash-core-components/src/fragments/DatePickerRange.tsx
KoolADE85: Good call, done!

AnnMarieW:
Sorry to comment on this after you already moved to `date-fns`, but just wanted to check if it's  possible to add locales from a dash app with `date-fns`?  

For example with `day.js` you can add different locales with a script.  I didn't see an option to do that with `date-fns` without bundling the locales in the component.
```
scripts = [
    "https://cdnjs.cloudflare.com/ajax/libs/dayjs/1.10.8/dayjs.min.js",      # dayjs  
    "https://cdnjs.cloudflare.com/ajax/libs/dayjs/1.10.8/locale/fr.min.js",  # french locale
]

app = Dash(external_scripts=scripts)

```

Also, the `day.js` has a syntax that is more like `moment` so you may not have to write scripts for those format props.

KoolADE85:
> Sorry to comment on this after you already moved to `date-fns`, but just wanted to check if it's possible to add locales from a dash app with `date-fns`?
> 
> For example with `day.js` you can add different locales with a script. I didn't see an option to do that with `date-fns` without bundling the locales in the component.
> 
> ```
> scripts = [
>     "https://cdnjs.cloudflare.com/ajax/libs/dayjs/1.10.8/dayjs.min.js",      # dayjs  
>     "https://cdnjs.cloudflare.com/ajax/libs/dayjs/1.10.8/locale/fr.min.js",  # french locale
> ]
> 
> app = Dash(external_scripts=scripts)
> ```
> 
> Also, the `day.js` has a syntax that is more like `moment` so you may not have to write scripts for those format props.

Excellent call @AnnMarieW ! It does support adding locales like that (or from the `assets` folder). However, I neglected to add the code that would make use of it! This is now added, along with a test.

@ndrezn 
I think this qualifies as a "API change" for the DatePicker, so just want to run it past you. With this approach, if users want a localized calendar, they will need to provide an external script that provides the translations (e.g. https://cdn.jsdelivr.net/npm/date-fns@3.6.0/locale/fr/cdn.min.js for French) or to put the script in their assets folder.
I don't think it's a big deal, but want to run it past you since non-English apps will need to perform this additional step when upgrading to the new components.

AnnMarieW:
Wow - that's so awesome that you included the locale now.  I just wanted to make sure it was possible to add at some point since this has been a frequently requested feature. 

I think it's OK to document how to add the extra script, since this is a new feature.

KoolADE85:
> Wow - that's so awesome that you included the locale now. I just wanted to make sure it was possible to add at some point since this has been a frequently requested feature.
> 
> I think it's OK to document how to add the extra script, since this is a new feature.

oh that's great then! For some reason, I had thought that moment.js had translations bundled in and handled them automatically. But if this actually adds a new translation feature, then even better!

PR #3497, opened 2025-11-07
Fix incorrect rendering in nested children

### LINKED ISSUE #3330: Disappearing and Swapping Content in Dash v3
**Describe your context**

Dash package versions:
```
dash                      3.0.4
dash_ag_grid              31.3.1
dash-bootstrap-components 2.0.3
```
Browsers used:
    - OS: Windows
    - Browser: Chrome and Edge
    - Version: latest

**Describe the bug**

I have a dash app which has multiple tabs and dropdowns which control the majority of the content on the page.

Using v3 of dash and highlighting text or clicking on interactive (e.g. buttons) and non-interactive (e.g. text) content on the page causes elements from previous tabs to appear and replace those on the current page.
Examples being:
- clicking on a dropdown on the currently viewed tab and it gets replaced by a title element of another tab
- highlighting some text from a H3 element causes the whole element to be replaced by a dash table from another tab

The issue doesn't seem to occur after downgrading the installed version of dash to version 2.18.2.

Watching the network tool within the browser's developer tools when the issue occurs would suggest that when the content changes on the page, there is no communication between the browser and the dash app and it's just local javascript running.

Not sure if it's related but also received this error while in debug mode:
```
Error: undefined was not found.
   at Object.resolve (http://server:8040/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v3_0_3m1749492493.dev.js:83970:11)
   at hydrateFunc (http://server:8040/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v3_0_3m1749492493.dev.js:84756:69)
   at DashWrapper (http://server:8040/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v3_0_3m1749492493.dev.js:84774:16)
   at renderWithHooks (http://server:8040/_dash-component-suites/dash/deps/react-dom@18.v3_0_3m1749492493.3.1.js:15496:20)
   at mountIndeterminateComponent (http://server:8040/_dash-component-suites/dash/deps/react-dom@18.v3_0_3m1749492493.3.1.js:20113:15)
   at beginWork (http://server:8040/_dash-component-suites/dash/deps/react-dom@18.v3_0_3m1749492493.3.1.js:21636:18)
   at beginWork$1 (http://server:8040/_dash-component-suites/dash/deps/react-dom@18.v3_0_3m1749492493.3.1.js:27475:16)
   at performUnitOfWork (http://server:8040/_dash-component-suites/dash/deps/react-dom@18.v3_0_3m1749492493.3.1.js:26609:14)
   at workLoopSync (http://server:8040/_dash-component-suites/dash/deps/react-dom@18.v3_0_3m1749492493.3.1.js:26515:7)
   at renderRootSync (http://server:8040/_dash-component-suites/dash/deps/react-dom@18.v3_0_3m1749492493.3.1.js:26483:9)
```

**Expected behavior**

No changes to page when clicking etc.

AnnMarieW:
closes #3330 

Note -this solution was provided by a Dash Enterprise customer. :tada: 

In sample dash app below,  a callback updates the children component with a layout of nested components.  If there is a change in one of the nested components there can be unexpected results.  

For example, if  you select a date in the date picker,  then "refresh" the layout in a callback, it re-renders the old state when you interact with the date picker again.  This error happens in Dash 3 but not in Dash 2.

A workaround is to give the layout's parent component an id that changes with each update.  This forces the component to re-render correctly.

This PR makes that workaround unnecessary.  It  updates the reducer in the dash renderer so it clears out the child path hashes and prevents stale values from triggering incorrect renders

Potential issues:
- This may impact performance when the children component has many nested components
- May cause issues when using partial property updates.

------------

```python

import dash
from dash import Input, Output, State, dcc, html

app = dash.Dash(__name__, prevent_initial_callbacks=True)


def layout(refresh_count=0):
    return html.Div(
       # id=f"container-{refresh_count}",  # this works
        id="container",                    # this doesn't work
        children=[
            html.H1(f"Layout refreshed {refresh_count} times"),
            html.P(
                html.Ul(
                    [
                        html.Li("Pick a date in the date picker"),
                        html.Li("Click on the 'Refresh layout' button"),
                        html.Li(
                            "Open date picker. It will change its value and text on the right will change too"
                        ),
                    ]
                )
            ),
            html.Div(
                html.Div(
                    html.Div(
                        id="nested-container",
                        children=[
                            dcc.DatePickerSingle(id="date-picker", date=None),
                            html.Span(
                                f"Nothing picked, {refresh_count} refreshes",
                                id="text",
                                style={
                                    "marginLeft": "10px",
                                    "backgroundColor": "#d0d0d0",
                                    "padding": "5px",
                                },
                            ),
                        ],
                    ),
                ),
            ),
        ],
    )


app.layout = html.Div(
    [
        html.H1("Dash bug showcase"),
        html.Button("Refresh layout", id="refresh-button"),
        html.P(
            id="layout",
            style={"padding": "20px", "border": "2px dashed red"},
            children=layout(),
        ),
    ]
)


@app.callback(
    Output("layout", "children"),
    Input("refresh-button", "n_clicks"),
)
def update_output(n_clicks):
    return layout(n_clicks)


@app.callback(
    Output("text", "children"),
    Input("date-picker", "date"),
    State("refresh-button", "n_clicks"),
)
def update_text(date, n_clicks):
    return f"{date}, {n_clicks} refreshes"


if __name__ == "__main__":
    app.run(debug=True)


```


ndrezn:
@T4rk1n @AnnMarieW I see performance risks:
> This may impact performance when the children component has many nested components
> May cause issues when using partial property updates.

Can we mitigate these issues or does this need to be an optional rendering config?

AnnMarieW:
Here's an interesting bit of data.  The example above works fine in Dash 4 ( using the new dcc date picker), but does not work in Dash 3.  

One more clue:  In Dash 3, if the children are wrapped in a div instead of just being a list, it updates correctly:
```python
                    children=html.Div([     # This line changed
                            dcc.DatePickerSingle(id="date-picker", date=None),
                            html.Span(
                                f"Nothing picked, {refresh_count} refreshes",
                                id="text",
                                style={
                                    "marginLeft": "10px",
                                    "backgroundColor": "#d0d0d0",
                                    "padding": "5px",
                                },
                            ),
                        ]),
```

AnnMarieW:
I expect this PR should be closed in favor of https://github.com/plotly/dash/pull/3570 ?

AnnMarieW:
closed in favor of #3570

PR #3511, opened 2025-11-13
Add issue checker

### LINKED ISSUE #3508: Require that every open PR be associated with an issue
Often, PRs are opened and it's not clear why it was opened or what the priority of the solution is.

We should require that all PRs be associated with an issue before they can be merged.

Maybe configuration in Github settings.

ndrezn:
Closes #3508

Primarily to help the core team review incoming community PRs and keep track of changes.

GHA can be found in: https://github.com/plotly/pr-issue-checker

### DIFF
--- a/.github/workflows/pr-issue-check.yml
+++ b/.github/workflows/pr-issue-check.yml
@@ -12,6 +12,6 @@ jobs:
   check-linked-issue:
     runs-on: ubuntu-latest
     steps:
-      - uses: plotly/pr-issue-checker
+      - uses: plotly/pr-issue-checker@main
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr-issue-check.yml
+++ b/.github/workflows/pr-issue-check.yml
@@ -16,6 +16,6 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           exempt-labels: "no-issue-needed"
-          request-issue-message: "Could you please link an issue to this PR? This helps us track the context and purpose of changes."
-          thank-you-message: "Thank you for your contribution! 🎉"
+          request-issue-message: "Could you please link an issue to this PR? This helps us track the context and purpose of changes. If an issue doesn't exist yet, create one before linking it."
+          thank-you-message: "Thank you for your contribution to Dash! 🎉"
           issue-added-message: "Thanks for adding an issue! ✅"
--- a/.github/workflows/pr-issue-check.yml
+++ b/.github/workflows/pr-issue-check.yml
@@ -18,4 +18,4 @@ jobs:
           exempt-labels: "no-issue-needed"
           request-issue-message: "Could you please link an issue to this PR? This helps us track the context and purpose of changes. If an issue doesn't exist yet, create one before linking it."
           thank-you-message: "Thank you for your contribution to Dash! 🎉"
-          issue-added-message: "Thanks for adding an issue! ✅"
+          issue-added-message: ""
--- a/.github/workflows/pr-issue-check.yml
+++ b/.github/workflows/pr-issue-check.yml
@@ -15,3 +15,7 @@ jobs:
       - uses: plotly/pr-issue-checker@v1
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          exempt-labels: "no-issue-needed"
+          request-issue-message: "Could you please link an issue to this PR? This helps us track the context and purpose of changes."
+          thank-you-message: "Thank you for your contribution! 🎉"
+          issue-added-message: "Thanks for adding an issue! ✅"
--- a/.github/workflows/pr-issue-check.yml
+++ b/.github/workflows/pr-issue-check.yml
@@ -2,7 +2,7 @@ name: PR Issue Check
 
 on:
   pull_request:
-    types: [opened, edited, synchronize, labeled, unlabeled]
+    types: [opened, edited, labeled, unlabeled]
 
 permissions:
   pull-requests: write
--- a/.github/workflows/pr-issue-check.yml
+++ b/.github/workflows/pr-issue-check.yml
@@ -12,6 +12,6 @@ jobs:
   check-linked-issue:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-issue-check.yml
```diff
@@ -0,0 +1,17 @@
...
+
+permissions:
+  pull-requests: write
+  issues: read
+
+jobs:
+  check-linked-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: plotly/pr-issue-checker@main
```
emilykl: should reference a commit or tag, not `main` (security)

ndrezn: I'll issue a release of `pr-issue-checker` (e.g. `v1`)

.github/workflows/pr-issue-check.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  check-linked-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: plotly/pr-issue-checker@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          exempt-labels: "no-issue-needed"
+          request-issue-message: "Could you please link an issue to this PR? This helps us track the context and purpose of changes. If an issue doesn't exist yet, create one before linking it."
+          thank-you-message: "Thank you for your contribution to Dash! 🎉"
```
emilykl, line 20: imo message should be something along the lines of "Thank you for your contribution to Dash! We will review this PR as soon as possible."

.github/workflows/pr-issue-check.yml
```diff
@@ -0,0 +1,21 @@
...
+          issue-added-message: "Thanks for adding an issue! ✅"
```
emilykl: maybe don't even need an additional comment in this case, could just be ""

.github/workflows/pr-issue-check.yml
```diff
@@ -0,0 +1,21 @@
+name: PR Issue Check
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, labeled, unlabeled]
```
emilykl: is `synchronize` needed here?

AnnMarieW:
What does the error message say if there is no linked issue?  Lots of people don’t know you have to use a magic word before the issue number like `fixes` or `closes`.  And it needs to be in the first issue message block, otherwise GH doesn’t automatically close the issue when it’s merged.

.github/workflows/pr-issue-check.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  check-linked-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: plotly/pr-issue-checker@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          exempt-labels: "no-issue-needed"
+          request-issue-message: "Could you please link an issue to this PR? This helps us track the context and purpose of changes. If an issue doesn't exist yet, create one before linking it."
```
camdecoster, line 19: Could you add a link to help people understand how to link the issue? This [action page section](https://github.com/plotly/pr-issue-checker#how-to-link-issues) is a good option.

.github/workflows/pr-issue-check.yml
camdecoster: It does seem to be [redundant](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request):
> A pull request's head branch was updated. For example, the head branch was updated from the base branch or new commits were pushed to the head branch.

.github/workflows/pr-issue-check.yml
camdecoster: I'd avoid adding any language that implies a timeline for a review.

.github/workflows/pr-issue-check.yml
camdecoster: I agree. The extra comment seems unnecessary.

.github/workflows/pr-issue-check.yml
emilykl: alternative idea, don't leave a comment at all unless the PR is missing a link to an issue

ndrezn: Why not? It's always nice to thank contributors. I can add copy like, "We have many contributions and will do our best to review our PR in a timely manner" or similar

ndrezn:
@AnnMarieW the failure message looks like this:
image

.github/workflows/pr-issue-check.yml
ndrezn: See latest comment with screenshot

.github/workflows/pr-issue-check.yml
camdecoster: I understand your point, but it also seems like it would be noise. If there's no problem, then why comment? We can add a comment (if necessary) during our triage meetings. Ultimately, it's not a blocker for me.

camdecoster (approved):
There might be some kinks to work out, but it seems like we should turn it on and see how it goes.

.github/workflows/pr-issue-check.yml
emilykl: I agree it's nice to thank contributors but a thank you feels better coming from a human than an automation. If we have genuinely useful information to share we can put it in an auto comment, otherwise it's just noise

ndrezn: Let's leave for now so comments aren't being created/deleted all the time

ndrezn:
@T4rk1n could I get a ✅ ?

.github/workflows/pr-issue-check.yml
emilykl: @ndrezn my vote would be:

- no comment if PR has linked issue on creation
- leave comment if PR
- if author links issue, edit comment to say "Thanks for linking an issue!" (but don't leave that comment initially)

but if you want to leave as-is for now that's fine

ndrezn: Leave as-is for simplicity, adjust later as we get some real-world feedback on the flow?

T4rk1n (approved):

PR #3523, opened 2025-11-22
Fallback on background callback function names if source cannot be found

### LINKED ISSUE #1885: [BUG] long_callback fails when bundling python code

corebit-nl:
Additional patch to #2922 to fully fix #1885 . Re-submitted version of #3072 .

This MR adds a patch made in #2922 to similar code, to make background callbacks fully work in compiled binaries (eg. PyInstaller).

## Contributor Checklist

### optionals


T4rk1n (approved):
:woman_dancing:

corebit-nl:
Hi @T4rk1n @ndrezn @emilykl ! Are there any remaining actions I should perform to get this PR merged? Thanks!

camdecoster:
@corebit-nl could you please add a changelog entry? With that, we can get this merged.

corebit-nl:
> @corebit-nl could you please add a changelog entry? With that, we can get this merged.

I did not find the time until now to do so, but I see you already did add a message. Thanks for your effort on getting this merged!

PR #3536, opened 2025-12-04
Feature: `dcc.Button`

KoolADE85:
This PR adds a `dcc.Button`

It is the same as `html.Button` except for the default styles, which match the look-and-feel of the other core components.
image


T4rk1n (approved):
:woman_dancing:

emilykl:
Awesome @KoolADE85! I've wanted this for ages.

PR #3540, opened 2025-12-12
Expose more types for better static typing options

tlauli:
Hi, I have two small improvements to static typing. This is my first contribution, so hopefully I did it right.

## Contributor Checklist

   -  [x] Introduce `ComponentSingleType`, representing a singular renderable entity -> it is possible to render a list of these
   -  [x] Make `_Wildcard` type public by renaming it to `Wildcard`, and making it an `Enum`, which I feel better reflects the intent behind it, plays nice with static type-checkers and prevents users from doing funky stuff with it since it is now public (subclassing or instantiating anything other than the tree variants defined by dash)
- [x] I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR.

### optionals


dash/development/base_component.py
```diff
@@ -454,14 +454,11 @@ def _validate_deprecation(self):
...
 # Renderable node type.
 ComponentType = typing.Union[
-    str,
-    int,
-    float,
-    Component,
-    None,
-    typing.Sequence[typing.Union[str, int, float, Component, None]],
+    ComponentSingleType,
+    typing.Sequence[ComponentSingleType],
```
T4rk1n, line 461: This type is also duplicated on generate components for backward compatibility. The template is here: https://github.com/plotly/dash/blob/1163158047d424d6521829ec56bada99d2c2d52c/dash/development/_py_components_generation.py#L28-L35

tlauli: I replaced the duplicated code with import from the `base_component`, is that ok?

T4rk1n: No, it's for backward compatibility of generated components in external components libraries, if the library is built with the new imports but used with a previous version of dash it would not work.

tlauli: Oh, I see. I removed the import and used the new type.

T4rk1n (approved):
:woman_dancing:

dash/development/base_component.py
tlauli: Would it be ok to replace the duplicated code with the import for 4.0?

T4rk1n: No, it's really just building the components need to be backward compatible as we don't want to force components libraries to depend on a particular version of dash.

PR #3543, opened 2025-12-15
Bugfix: DCC redesign feedback from `rc4`

KoolADE85:
This PR addresses the feedback collected from `rc4`:

General:
- Bolder (thicker) outline when components are focused
- Small adjustments to default colours so that WCAG colour contrast requirements are met
- Font family & font size consistently inherit from overall page font/size

Dropdowns:
- Hide radio buttons in single-select dropdowns
- Allow `Backspace` and `Delete` keys to clear selection

Input:
- Fix `type=number` increment/decrement buttons could appear disabled even though they are not


T4rk1n (approved):
:woman_dancing:

PR #3544, opened 2025-12-15
Bugfix/datepicker feedback

KoolADE85:
This PR fixes issues found & reported in the new DatePicker:

* Month labels could be cut-off in some languages; so the dropdown width is now dynamic
* Enter key now activates the calendar popup when focused on the dropdown
* Smarter determination of how to handle `numberOfMonthsShown` & _which_ months are shown, based on how the user interacts
    * when opening a calendar for the first time, the current month will be "centered" in the list of months
    * when navigating "forward" in the calendar, the current month will now be displayed last in the list of months
    * when navigating "backward" in the calendar, the current month will be displayed first
    * in all cases, when navigating to a month that is already visible (`numberOfMonthsShown > 1`), do not re-calculate or re-position which months are shown: just keep the current view.

### DIFF
--- a/components/dash-core-components/src/components/css/calendar.css
+++ b/components/dash-core-components/src/components/css/calendar.css
@@ -16,7 +16,6 @@
     font-weight: normal;
     color: var(--Dash-Text-Weak);
     text-align: center;
-    border-radius: 4px;
     cursor: default;
     user-select: none;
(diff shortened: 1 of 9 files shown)


T4rk1n (approved):
:woman_dancing:

Issue #3552, opened 2025-12-18
[BUG] - dash 4.0.0rc5 - step on input doesn't increment correctly


LiamConnors:
With the following code:

```
from dash import Dash, dcc

app = Dash(__name__)

app.layout = [dcc.Input(type="number", value=0, step=0.1)]

if __name__ == "__main__":
    app.run(debug=True, port=8063)
```

<img width="328" height="104" alt="Image" src="https://github.com/user-attachments/assets/9bdd049a-ea7e-45be-bec3-1e4d7ac378d6" />

The step doesn't increment correctly with Dash 4.0.0rc5

<img width="1493" height="182" alt="Image" src="https://github.com/user-attachments/assets/f7694f33-f097-485e-a400-8b83dc7c42e1" />

In an app with a callback that shows the value as `None`:

<img width="478" height="169" alt="Image" src="https://github.com/user-attachments/assets/b8cbb1fd-6ae9-4beb-9ad0-8ac1336fdf4e" />

vs

<img width="482" height="169" alt="Image" src="https://github.com/user-attachments/assets/53fe4661-da8b-4e4c-9210-55211ec18e87" />

camdecoster:
@KoolADE85 was this fixed by #3554? If so, could you please close this issue?

KoolADE85:
Fixed in rc6.

Issue #3550, opened 2025-12-18
[BUG] - dash 4.0.0rc5 - can't see calendar when selecting datepicker


LiamConnors:
-  replace the result of `pip list | grep dash` below
```
dash                4.0.0rc5
```

**Describe the bug**

Sometimes the datepickers appear can't be seen when you select them. For example, in this app, when selecting either the datepickersingle or range:

<img width="970" height="318" alt="Image" src="https://github.com/user-attachments/assets/d9bc88a1-256e-4af3-8ebd-39280977690a" />

```
import dash
from dash import Dash, dcc, html
from datetime import date

app = Dash()

app.layout = html.Div([
    html.H3(f"Dash version: {dash.__version__}"),
    dcc.DatePickerSingle(date=date(2024, 1, 15)),
    dcc.DatePickerRange(
        start_date=date(2024, 1, 1),
        end_date=date(2024, 1, 15)
    )
])

if __name__ == '__main__':
    app.run(debug=True)
```

KoolADE85:
Fixed in rc6.

PR #3554, opened 2025-12-30
Bugfix: dcc redesign rc5 feedback

KoolADE85:
This PR addresses feedback from the latest `rc5` release.

Notable issues fixed:
* #3550  Remove css that attempted to manage popover placement: these rules have been causing more problems than they solved. The core change is that popovers will be `position: fixed` again (rather than absolute). While these rules were  necessary for compatibility with Dash Design Kit, they have been negatively affecting OSS apps. I will move these rules into DDK where they can be narrowly constrained.
* #3552 - floating point precision bug when incrementing/decrementing number inputs


T4rk1n (approved):
:woman_dancing:

PR #3562, opened 2026-01-07
Remove space in css variable (which broke the CSS variable link)

emilhe:
Remove space in css variable (which broke the CSS variable link)


KoolADE85 (approved):
Thanks for this @emilhe ! 💃

T4rk1n (approved):

Issue #3525, opened 2025-11-24
[Feat/bug] Make `hidden` work for clientside callbacks


antonymilne:
I love the new `hidden=True` feature introduced in https://github.com/plotly/dash/pull/3407 to hide callbacks from the devtools graph, but unfortunately it doesn't do anything for clientside callbacks. In practice this unfortunately makes it completely useless for our purposes since all the callbacks we want to hide happen to currently be clientside 😅 

Is there any way `hidden` could be implemented for clientside callbacks too, pretty please? 🙏

From reading the PR and docs it wasn't clear to me whether this was a bug or just an un-implemented feature, so not quite sure how to classify the PR. Either way I would be hugely grateful if this could be added. Happy to raise the PR myself if it sounds like a good idea.

robertclaus:
Good idea! We likely won't get to this in the next few months, but would be thrilled to review a PR for it!

antonymilne:
Closed by #3563

PR #3563, opened 2026-01-07
Add hidden to clientside callbacks as configurable parameter

l0uden:
Added `hidden` to clientside callbacks as configurable parameter.
Now it can be configured from `cilentside_callback` function.

Solves this issue: https://github.com/plotly/dash/issues/3525

## Contributor Checklist

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -854,6 +854,7 @@ def register_clientside_callback(
         None,
         prevent_initial_call,
         no_output=no_output,
+        hidden=True,
     )
 
     # If JS source is explicitly given, create a namespace and function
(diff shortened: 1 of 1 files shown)


T4rk1n (approved):
:woman_dancing:

dash/_callback.py
```diff
@@ -854,6 +854,7 @@ def register_clientside_callback(
         None,
         prevent_initial_call,
         no_output=no_output,
+        hidden=True,
```
CNFeffery: I don't quite understand why it is mandatory to set it to `True` here.

T4rk1n: Oh right, it should be taken from the args, not hide all the clientside callbacks

T4rk1n (changes requested):
Should take the hidden parameter from the arguments.

dash/_callback.py
l0uden: Thanks!

T4rk1n (approved):
:woman_dancing:

Issue #3493, opened 2025-11-05
[Feature Request] Allow only displaying some specified callback function relationships in the callback graph


CNFeffery:
In the current debug tools, the callback graph displays all callback function relationships by default (unless the callback functions are hidden based on https://github.com/plotly/dash/commit/36c71715f9b0f809faa2d1e455bb8be3f08265c4).

<img width="1943" height="1414" alt="Image" src="https://github.com/user-attachments/assets/2a9a6375-5b5c-480f-aadf-904f763a7f24" />

However, in many cases, the application functions that can be developed using `Dash` are becoming increasingly complex, making it difficult to identify effective information from the displayed callback relationship graph. Adding a new parameter to `dash.Dash()` to set all callback functions to be hidden by default, and adding a new parameter to `callback()` to allow the current callback function to be displayed in the callback graph, would facilitate callback debugging during the development of more complex applications.

Formally, it is similar to `prevent_initial_callbacks` in `dash.Dash()` and `prevent_initial_call` in `callback()`.

BSd3v:
@CNFeffery,

I am fairly certain that this was added in 3.3.0, `hidden=True` can you confirm?

https://github.com/plotly/dash/pull/3407

CNFeffery:
@BSd3v However, if my application involves 100 callback functions, and I only want to debug a few of them, using the `hidden` parameter would require manually setting it for most of the callback functions, which is very troublesome.

BSd3v:
You could iterate through the callback map and add the hidden to all but the ones you want to display?

CNFeffery:
> You could iterate through the callback map and add the hidden to all but the ones you want to display?

That would be a bit more troublesome but still doable. However, it would be very convenient if there were a parameter like `hide_all_callbacks` in dash.Dash(). Many of my friends who use Dash often find, after debugging a non-functioning callback, that the cause is something simple like a misspelled ID. If they could first use the callback graph to attempt to reverse and display only the target callback function for debugging, such low-level issues would be much easier and quicker to identify.

Aaron-Wrote-This:
I appreciate the ability to hide them at all, but it would be nice to have a hidden by default option display only the callbacks I'm focused on.

CNFeffery:
@Aaron-Wrote-This as mentioned in `3.4.0`, use `hide_all_callbacks` to focus on specific callbacks:

https://github.com/plotly/dash/releases/tag/v3.4.0

PR #3564, opened 2026-01-08
run() add new parameter hide_all_callbacks

CNFeffery:
For the scenario corresponding to https://github.com/plotly/dash/issues/3493, add a new parameter `hide_all_callbacks` to `Dash()`, which is used to change the default application value of the `hidden` parameter in `app.callback()`, improving the convenience of callback debugging for complex applications. Example:

image

```python
import dash
from dash import html, Input, Output, State

app = dash.Dash(__name__, hide_all_callbacks=True)

app.layout = html.Div(
    [
        html.Div(
            [
                html.Button("Button 1", id="button1"),
                html.Button("Button 2", id="button2"),
                html.Button("Button 3", id="button3"),
            ]
        ),
        html.Div(0, id="output"),
    ],
    style={"padding": 50},
)


@app.callback(
    Output("output", "children"),
    Input("button1", "n_clicks"),
    State("output", "children"),
    prevent_initial_call=True,
)
def button1_update_output(n_clicks, children):
    return children + 1


@app.callback(
    Output("output", "children", allow_duplicate=True),
    Input("button2", "n_clicks"),
    State("output", "children"),
    prevent_initial_call=True,
)
def button2_update_output(n_clicks, children):
    return children + 1


@app.callback(
    Output("output", "children", allow_duplicate=True),
    Input("button3", "n_clicks"),
    State("output", "children"),
    prevent_initial_call=True,
    hidden=False,
)
def button3_update_output(n_clicks, children):
    return children + 1


if __name__ == "__main__":
    app.run(debug=True)
```

### DIFF
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -79,7 +79,7 @@ def callback(
     on_error: Optional[Callable[[Exception], Any]] = None,
     api_endpoint: Optional[str] = None,
     optional: Optional[bool] = False,
-    hidden: Optional[bool] = False,
+    hidden: Optional[bool] = None,
     **_kwargs,
 ) -> Callable[..., Any]:
     """
@@ -181,6 +181,7 @@ def callback(
     config_prevent_initial_callbacks = _kwargs.pop(
         "config_prevent_initial_callbacks", False
     )
+    config_hide_all_callbacks = _kwargs.pop("config_hide_all_callbacks", False)
     callback_map = _kwargs.pop("callback_map", GLOBAL_CALLBACK_MAP)
     callback_list = _kwargs.pop("callback_list", GLOBAL_CALLBACK_LIST)
 
@@ -222,6 +223,7 @@ def callback(
         callback_list,
         callback_map,
         config_prevent_initial_callbacks,
+        config_hide_all_callbacks,
         *_args,
         **_kwargs,
         background=background_spec,
@@ -265,6 +267,7 @@ def insert_callback(
     callback_list,
     callback_map,
     config_prevent_initial_callbacks,
+    config_hide_all_callbacks,
     output,
     outputs_indices,
     inputs,
@@ -277,10 +280,13 @@ def insert_callback(
     dynamic_creator: Optional[bool] = False,
     no_output=False,
     optional=False,
-    hidden=False,
+    hidden=None,
 ):
     if prevent_initial_call is None:
         prevent_initial_call = config_prevent_initial_callbacks
+
+    if hidden is None:
+        hidden = config_hide_all_callbacks
 
     _validate.validate_duplicate_output(
         output, prevent_initial_call, config_prevent_initial_callbacks
@@ -600,6 +606,7 @@ def register_callback(
     callback_list,
     callback_map,
     config_prevent_initial_callbacks,
+    config_hide_all_callbacks,
     *_args,
     **_kwargs,
 ):
@@ -639,6 +646,7 @@ def register_callback(
         callback_list,
         callback_map,
         config_prevent_initial_callbacks,
+        config_hide_all_callbacks,
         insert_output,
         output_indices,
         flat_inputs,
@@ -651,7 +659,7 @@ def register_callback(
         running=running,
         no_output=not has_output,
         optional=_kwargs.get("optional", False),
-        hidden=_kwargs.get("hidden", False),
+        hidden=_kwargs.get("hidden"),
     )
 
     # pylint: disable=too-many-locals
@@ -836,6 +844,7 @@ def register_clientside_callback(
     callback_list,
     callback_map,
     config_prevent_initial_callbacks,
+    config_hide_all_callbacks,
     inline_scripts,
     clientside_function: ClientsideFuncType,
     *args,
@@ -847,6 +856,7 @@ def register_clientside_callback(
         callback_list,
         callback_map,
         config_prevent_initial_callbacks,
+        config_hide_all_callbacks,
         output,
         None,
         inputs,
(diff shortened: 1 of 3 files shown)


dash/_callback.py
```diff
@@ -181,6 +181,7 @@ def callback(
     config_prevent_initial_callbacks = _kwargs.pop(
         "config_prevent_initial_callbacks", False
     )
+    config_hide_all_callbacks = _kwargs.pop("config_hide_all_callbacks", False)
```
T4rk1n: How are those actually insert when using `@callback` directly? I think we need a better system for those since we want the same behavior for both `@app.callback` and `@callback`.

Think you can refactor these config into a context and use the context instead of passing them around?

CNFeffery: > How are those actually insert when using `@callback` directly? I think we need a better system for those since we want the same behavior for both `@app.callback` and `@callback`.
> 
> Think you can refactor these config into a context and use the context instead of passing them around?

Yes, similarly, for `prevent_initial_callbacks`, the callback functions defined by `dash.callback()` cannot obtain the setting values from `dash.Dash()` too.

CNFeffery:
Taking the following application as an example, both the `prevent_initial_callbacks` and `hide_all_callbacks` parameters can now work properly in the callback functions defined by `app.callback()` and `dash.callback()`:

image

```python
import dash
from dash import html, Input, Output, State

app = dash.Dash(__name__, hide_all_callbacks=True, prevent_initial_callbacks=True)

app.layout = html.Div(
    [
        html.Div(
            [
                html.Button("Button 1", id="button1"),
                html.Button("Button 2", id="button2"),
                html.Button("Button 3", id="button3"),
            ]
        ),
        html.Div(0, id="output"),
    ],
    style={"padding": 50},
)


@app.callback(
    Output("output", "children"),
    Input("button1", "n_clicks"),
    State("output", "children"),
    prevent_initial_call=True,
)
def button1_update_output(n_clicks, children):
    return children + 1


@dash.callback(
    Output("output", "children", allow_duplicate=True),
    Input("button2", "n_clicks"),
    State("output", "children"),
)
def button2_update_output(n_clicks, children):
    return children + 1


@app.callback(
    Output("output", "children", allow_duplicate=True),
    Input("button3", "n_clicks"),
    State("output", "children"),
    prevent_initial_call=True,
    hidden=False,
)
def button3_update_output(n_clicks, children):
    return children + 1


if __name__ == "__main__":
    app.run(debug=True)
```

dash/_callback.py
```diff
@@ -178,9 +179,13 @@ def callback(
 
     background_spec: Any = None
 
-    config_prevent_initial_callbacks = _kwargs.pop(
-        "config_prevent_initial_callbacks", False
+    # Get prevent_initial_callbacks and hide_all_callbacks from get_app().config
+    current_app_config = get_app().config
```
T4rk1n: Nice usage of get_app, but this will not work if the app hasn't been instanced yet which we can't guarantee with `dash.callback`.

We can defer it all and map the configs values in the _setup_server after this line: https://github.com/plotly/dash/blob/a9a173f7e27ab70e9a89b2c2afc13f8d107e3fe1/dash/dash.py#L1648

CNFeffery:
After refactoring, the current `hide_all_callbacks` parameter can take effect on callback functions defined in various types:

- Callback functions defined based on `dash.callback()` before the `app` is instantiated
- Callback functions defined based on `dash.callback()` after the `app` is instantiated
- Callback functions defined based on `app.callback()`

Example application:

image

```python
import dash
from dash import html, Input, Output, State


@dash.callback(
    Output("output", "children", allow_duplicate=True),
    Input("button2", "n_clicks"),
    State("output", "children"),
    prevent_initial_call=True,
)
def button2_update_output(n_clicks, children):
    return children + 1


app = dash.Dash(__name__, hide_all_callbacks=True)

app.layout = html.Div(
    [
        html.Div(
            [
                html.Button("Button 1", id="button1"),
                html.Button("Button 2", id="button2"),
                html.Button("Button 3", id="button3"),
            ]
        ),
        html.Div(0, id="output"),
    ],
    style={"padding": 50},
)


@app.callback(
    Output("output", "children"),
    Input("button1", "n_clicks"),
    State("output", "children"),
    prevent_initial_call=True,
)
def button1_update_output(n_clicks, children):
    return children + 1


@dash.callback(
    Output("output", "children", allow_duplicate=True),
    Input("button3", "n_clicks"),
    State("output", "children"),
    prevent_initial_call=True,
    hidden=False,
)
def button3_update_output(n_clicks, children):
    return children + 1


if __name__ == "__main__":
    app.run(debug=True)

```

dash/dash.py
```diff
@@ -1645,7 +1659,18 @@ def _setup_server(self):
 
             self.callback_map[k] = _callback.GLOBAL_CALLBACK_MAP.pop(k)
 
+        # Get config of current app instance
+        current_app_config = get_app().config
```
T4rk1n: Can use self.config in this case.

T4rk1n (approved):
:woman_dancing:

ndrezn:
Because this is primarily used for debugging, would it make more sense to put this into the `run` command rather than the `Dash` instantiation? I guess it gets a bit messy since you need to manually specify the callbacks to show in callback definitions but want to throw the idea out there.

CNFeffery:
> Because this is primarily used for debugging, would it make more sense to put this into the `run` command rather than the `Dash` instantiation? I guess it gets a bit messy since you need to manually specify the callbacks to show in callback definitions but want to throw the idea out there.

This is indeed more in line with the purpose of development and debugging. I'll look into how to refactor it.

T4rk1n (approved):

PR #3566, opened 2026-01-11
fix to loadingOptions.rawPath type

LudwigAJ:
loadingOptions.rawPath was incorrectly set to boolean. Following show docs example (https://dash.plotly.com/dash-3-for-component-developers) so it aligns with example.


LudwigAJ:
simple change so ts linter stops complaining

T4rk1n (approved):
:woman_dancing:

PR #3568, opened 2026-01-12
update Clipboard

### LINKED ISSUE #3565: [Feature Request] Allow customisation of dcc.Clipboard icon/children
Similar to dmc.CopyButton: https://www.dash-mantine-components.com/components/copybutton#

**Current functionality:**

https://github.com/user-attachments/assets/e79d5160-e4c4-404c-a122-3e08d2f9c576

- The copy icon is always shown.
- It can be hidden with css properties, but then a workaround with a callback needs to be implemented to trigger the copy functionality.
- There are python libraries that allow triggering copy-to-clipboard like `pyperclip` (https://pypi.org/project/pyperclip/) but (1) they still require a callback (2) they don't work on Dash Enterprise because "on Linux, this module makes use of the xclip or xsel commands, which should come with the os" and their installation as apt-packages doesn't work properly.

**Desired functionality:**

https://github.com/user-attachments/assets/53b7d972-e14a-487f-b9a4-f55c7279cb12

This functionality was achieved with the code included in the "Current workarounds" section.

Suggestion for how it would look code-wise: **dcc.Clipboard as a wrapper for the component that triggers the copy functionality (similar to html.A). No `id` or callbacks needed.**
```
app.layout = html.Div([
    html.H3("Click the button to copy the text to the clipboard"),
    ### Relevant code ###
    html.Div([
            dcc.Clipboard(
                children=html.Button(children=symbol), 
                style={"margin-right":"10px", "font-size":20}
                ),
         ], style={"margin":"10px", "display":"inline-flex"}) for symbol in symbols
    ### End of relevant code ###
    ], style={"display":"inline-flex"}),
    html.H3("Try to paste the copied symbol here"),
    dcc.Textarea(id="textarea", style={'width': '100%', 'height': 300})
], style={"margin":"30px"})
```

## Current workarounds

This code is for the app that is run in the "Desired functionality" video.
It uses a hidden dcc.Clipboard and a callback that triggers its `n_clicks` property when a corresponding html.Button is clicked:
```
import dash
from dash import dcc, html, callback, Input, Output, State, MATCH

app = dash.Dash(__name__)

symbols = ["√", "π", "∑", "∆", "γ"]

app.layout = html.Div([
    html.H3("Click the button to copy the text to the clipboard"),
    html.Div([
        html.Div([
            html.Button(children=symbol, id={"type":"symbol", "index":str(i)}),
            dcc.Clipboard(
                target_id={"type":"symbol", "index":str(i)},
                id={"type":"copy-symbol", "index":str(i)},
                style={"display":"none"})
         ], style={"margin":"10px"}) for i, symbol in enumerate(symbols)
    ], style={"display":"inline-flex"}),
    html.H3("Try to paste the copied symbol here"),
    dcc.Textarea(id="textarea", style={'width': '100%', 'height': 300})
], style={"margin":"30px"})

@callback(
    Output({"type":"copy-symbol", "index":MATCH}, "n_clicks"),
    Input({"type":"symbol", "index":MATCH}, "n_clicks"),
    prevent_initial_call=True
)
def copy_symbol(n_clicks):
    return n_clicks
    
if __name__ == "__main__":
    app.run(debug=True)
```

## Sample code for current functionality
```
import dash
from dash import dcc, html, callback, Input, Output, State, MATCH

app = dash.Dash(__name__)

symbols = ["√", "π", "∑", "∆", "γ"]

app.layout = html.Div([
    html.H3("Click the button to copy the text to the clipboard"),
    html.Div([
        html.Div([
            html.Div(
                children=symbol, 
                id={"type":"symbol", "index":str(i)}, 
                style={"margin-right":"10px", "font-size":20}
                ),
            dcc.Clipboard(
                target_id={"type":"symbol", "index":str(i)},
                id={"type":"copy-symbol", "index":str(i)},
                )
         ], style={"margin":"10px", "display":"inline-flex"}) for i, symbol in enumerate(symbols)
    ], style={"display":"inline-flex"}),
    html.H3("Try to paste the copied symbol here"),
    dcc.Textarea(id="textarea", style={'width': '100%', 'height': 300})
], style={"margin":"30px"})
    
if __name__ == "__main__":
    app.run(debug=True)
```

AnnMarieW:
Closes #3565

See issue for details and examples

### DIFF
--- a/components/dash-core-components/src/components/Clipboard.react.js
+++ b/components/dash-core-components/src/components/Clipboard.react.js
@@ -128,20 +128,12 @@ export default class Clipboard extends React.Component {
     }
 
     render() {
-        const {
-            id,
-            title,
-            className,
-            style,
-            copied_className,
-            copied_style,
-            children,
-            copied_children,
-        } = this.props;
+        const {id, title, className, style, children, copied_children} =
+            this.props;
 
         const isCopied = this.state.copied;
 
-        const content = isCopied
+        const button_children = isCopied
             ? copied_children ?? <FontAwesomeIcon icon={faCheckCircle} />
             : children ?? <FontAwesomeIcon icon={faCopy} />;
 
@@ -149,11 +141,11 @@ export default class Clipboard extends React.Component {
             <LoadingElement
                 id={id}
                 title={title}
-                style={isCopied ? copied_style ?? style : style}
-                className={isCopied ? copied_className ?? className : className}
+                style={style}
+                className={className}
                 onClick={this.onClickHandler}
             >
-                {content}
+                {button_children}
             </LoadingElement>
         ) : null;
     }
@@ -173,12 +165,12 @@ Clipboard.propTypes = {
     id: PropTypes.string,
 
     /**
-     * The children of this component.  By default copy icon
+     * Children rendered inside the Clipboard button before copying. By default, a copy icon.
      */
     children: PropTypes.node,
 
     /**
-     * The children of this component displayed while copying.  By default checked icon
+     * Children rendered inside the Clipboard button after the value has been copied. By default, a check icon.
      */
     copied_children: PropTypes.node,
 
@@ -195,7 +187,7 @@ Clipboard.propTypes = {
     content: PropTypes.string,
 
     /**
-     * The number of times copy button was clicked
+     * The number of times Clipboard button was clicked
      */
     n_clicks: PropTypes.number,
 
@@ -218,15 +210,6 @@ Clipboard.propTypes = {
      * The class  name of the icon element
      */
     className: PropTypes.string,
-    /**
-     * The icon's styles while copying
-     */
-    copied_style: PropTypes.object,
-
-    /**
-     * The class  name of the icon element while copying
-     */
-    copied_className: PropTypes.string,
 
     /**
      * Dash-assigned callback that gets fired when the value changes.
--- a/components/dash-core-components/src/components/Clipboard.react.js
+++ b/components/dash-core-components/src/components/Clipboard.react.js
@@ -128,20 +128,33 @@ export default class Clipboard extends React.Component {
     }
 
     render() {
-        const {id, title, className, style} = this.props;
-        const copyIcon = <FontAwesomeIcon icon={faCopy} />;
-        const copiedIcon = <FontAwesomeIcon icon={faCheckCircle} />;
-        const btnIcon = this.state.copied ? copiedIcon : copyIcon;
+        const {
+            id,
+            title,
+            className,
+            style,
+            copied_className,
+            copied_style,
+            children,
+            copied_children,
+        } = this.props;
+
+        const isCopied = this.state.copied;
+
+        const content = isCopied
+            ? copied_children ?? <FontAwesomeIcon icon={faCheckCircle} />
+            : children ?? <FontAwesomeIcon icon={faCopy} />;
 
         return clipboardAPI ? (
             <LoadingElement
+                elementType={'button'}
                 id={id}
                 title={title}
-                style={style}
-                className={className}
+                style={isCopied ? copied_style ?? style : style}
+                className={isCopied ? copied_className ?? className : className}
                 onClick={this.onClickHandler}
             >
-                <i> {btnIcon}</i>
+                {content}
             </LoadingElement>
         ) : null;
     }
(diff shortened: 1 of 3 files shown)


AnnMarieW:
Changing the `elementType` to `Button` would be better for accessibility, but it would change the layout, so might be considered a breaking change.  This can be done in V4

components/dash-core-components/src/components/Clipboard.react.js
```diff
@@ -160,6 +172,16 @@ Clipboard.propTypes = {
      */
     id: PropTypes.string,
 
+    /**
+     * The children of this component.  By default copy icon
+     */
+    children: PropTypes.node,
```
KoolADE85, line 170: I'm a big fan of replacing the hard-coded `<i>` and giving users control of the child elements!
More than just the icon, I think that the default element should include the wrapper button/div itself. For example, a user might prefer to use an anchor instead.

components/dash-core-components/src/components/Clipboard.react.js
```diff
@@ -160,6 +172,16 @@ Clipboard.propTypes = {
...
+
+    /**
+     * The children of this component displayed while copying.  By default checked icon
```
KoolADE85:
```suggestion
     * The children of this component displayed after copying.  By default checked icon
```

components/dash-core-components/src/components/Clipboard.react.js
```diff
@@ -196,6 +218,15 @@ Clipboard.propTypes = {
...
     className: PropTypes.string,
+    /**
+     * The icon's styles while copying
+     */
+    copied_style: PropTypes.object,
+
+    /**
+     * The class  name of the icon element while copying
+     */
+    copied_className: PropTypes.string,
```
KoolADE85: Hmm, this seems like we're adding 2 or 3 ways of solving this problem at once. If we are allowing custom `children` as above, is there really any need to have these 2 style props? Users could apply their own styling to the child components and achieve the same effect with less complexity on our end.
(also, there might be good reasons here, I'm just not seeing it immediately!)

components/dash-core-components/src/components/Clipboard.react.js
AnnMarieW: That seems like a good idea, however, the Clipboard is essentially a Button, (and the base element should be updated to be a Button rather than a Div).  It’s invalid HTML to have a Button or other interactive element  as children of a Button.    The purpose of updating the children is to simply provide feedback on the copy state.

components/dash-core-components/src/components/Clipboard.react.js
AnnMarieW: I may have misunderstood you first comment above.   Can you clarify?  Did you mean that the children and copied_children could be buttons or anchors?  Or were you referring to only the root element of the Clipboard?

But if children and copied_children is a div with styling, I agree that you would then not need the coied_style and copied_className

KoolADE85: I was thinking the base element of the clipboard could remain a `<div>` and the default `children` could be
```
<button>
  <Icon />
</button>
```

As you note, if we eventually make the base element a button, then there are certain child components that become invalid (and this would not be obvious to Python users).
If, instead, we make the default child a button, then users can safely replace that with any Dash component, including a customized button.

AnnMarieW: The issue is that the click handler for the copy is in the base element, making it the interactive element -- even if it's a div.    

Internally, we could make the children a button as you propose and move the click handler for the copy there.  But  how would that work if someone provided their own button as children?

AnnMarieW:
Updated to remove `copied_style` and `copied_className`.


It will be necessary to document that `children` and `copied_children`   are rendered as the Clipboard button’s children and must not include interactive elements such as buttons or links.   


Here are some sample app to show how to style the children like a button

```python

from dash import Dash, html, dcc

app = Dash(external_stylesheets=[dbc.themes.BOOTSTRAP])

app.layout = html.Div([
    dcc.Clipboard(
        children=html.Div("∑", className="btn btn-primary"),
        copied_children = html.Div("∑", className="btn btn-secondary"),
        content="∑",
        id="clipoard",
        title="Copy me!"
    ),
    html.H5("Paste the copied symbol here"),
    dcc.Textarea(id="textarea"),

], style={"padding": 24})

if __name__ == "__main__":
    app.run(debug=True)
```

```python

from dash import Dash, html, dcc

app = Dash(external_stylesheets=[dbc.themes.BOOTSTRAP])

app.layout = html.Div([
    dcc.Clipboard(
        children="Copy URL",
        copied_children="Copied!",
        className="btn btn-primary",
        content="https://dash.plotly.com/"
    ),
    html.H5("Paste the copied content here"),
    dcc.Textarea(id="textarea")
], style={"padding": 24})

if __name__ == "__main__":
    app.run(debug=True)

```

KoolADE85 (approved):
Looks great to me!

T4rk1n (approved):

PR #3591, opened 2026-01-27
DCC Redesign: v4 release

KoolADE85:
This PR removes the `rc` tags from the Dash version, in preparation for the `v4` release.
(I won't be merging this until we are ready to do the actual v4 release).


T4rk1n (approved):
💃

PR #3597, opened 2026-02-03
Fix typo in docstring of Dash.callback function

Dimon9lul:
## This is a Quick-fix:

**The typo can can hint towards using a nonexistent paramter,** although application doesn't crash.
But it can still be confusing, when modern IDEs display the docstring.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -1450,7 +1450,7 @@ class Dash(ObsoleteChecker):
 
         The last, optional argument `prevent_initial_call` causes the callback
         not to fire when its outputs are first added to the page. Defaults to
-        `False` unless `prevent_initial_callbacks=True` at the app level.
+        `False` unless `prevent_initial_call=True` at the app level.
 
 
         """


dash/dash.py
```diff
@@ -1450,7 +1450,7 @@ def callback(self, *_args, **_kwargs) -> Callable[..., Any]:
 
         The last, optional argument `prevent_initial_call` causes the callback
         not to fire when its outputs are first added to the page. Defaults to
-        `False` unless `prevent_initial_callbacks=True` at the app level.
+        `False` unless `prevent_initial_call=True` at the app level.
```
T4rk1n, line 1453: The docstring was right, it's `prevent_initial_callbacks` at the app level and `prevent_initial_call` for individual callbacks.

https://github.com/plotly/dash/blob/3b42e03b4e101ecff70058cca8e026a250fb5a2a/dash/dash.py#L459

Dimon9lul: Oh, sorry. It threw me off when I was making my own app, but I wasn't at app level.
But thanks for clarifying!

PR #3621, opened 2026-02-23
[Bug] Component generation needs double quote `extract_path`

antonymilne:
Currently when the path to `extract-meta.js` contains a space, the `node` command run during component generation crashes with an error like this:
```
> dash-generate-components ./src/ts/components vizro_dash_components -p package-info.json  --ignore \.test\.

node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module '/Users/antony_milne/Library/Application'
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Error generating metadata in vizro_dash_components (status=1)
```

Note the path should be "Application Support/..." but it got cut off. Now with the quote marks it should all work nicely  🎉


T4rk1n:
But now it fails on windows, it seems like the extract path is collated with the previous argument:
`'C:\Users\circleci\dash\components\dash-html-components\"C:\Users\circleci\dash\dash\extract-meta.js"'`

Maybe we could not use the shlex.split and pass the command argument like `["node", extract_path, ignore, reserved_pattern, component_source]`.

antonymilne:
Ah yes, good spot @T4rk1n, thanks for that! Let's see if [2f90b0e](https://github.com/plotly/dash/pull/3621/commits/2f90b0e0365b56ecc8d9ca74378243405c3ad344) fixes it.

T4rk1n (approved):
💃

PR #3627, opened 2026-02-24
Make focused dropdowns searchable without first opening

### LINKED ISSUE #3572: [BUG] Dash 4.0.0rc5 dcc.Dropdowns
Hi
We're getting feedback from our users that the new dropdowns in 4.0.0 are deprecating a really nice behavior in the old dcc.Dropdowns.

**Describe your context**

```
dash                                     4.0.0rc5
dash-ag-grid                             31.3.1
dash-bootstrap-components                2.0.4
dash-bootstrap-templates                 1.3.0
dash-extensions                          1.0.20rc4
```

**Describe the bug**

When "tab"'ing around. In the old dcc.Dropdown you would be directed into search if you started searching when highlighted:

It's not possible to search on highlight anymore:
Image

Another nice thing about the old search, was that it had a "top 1" select on Enter behavior due to how the focus worked. I would expect in the new, that if I am in the search and hits Enter, it would select the "top 1" element. Here Enter does nothing anymore.

Image

Going from the old flow to the new creates the user flow:

_before_
"tab" -> "write" -> "enter" (done)

_after_
"tab" -> arrow-down / enter -> "write" -> "arrow-down" -> "enter" -> "esc" (done) 

**Expected behavior**

In the old dcc.Dropdown version it was default that I could search when starting to write (when highlighted) and select the first option ("top 1" select) on Enter behavior.

Image

**Screenshots**
In another project, we have a similar dropdown as 4.0, but it behaves more like the old, as the focus is auto-shifted.

Image

KoolADE85:
This PR updates dropdowns to be searchable without first opening:

* When focused, you can now start typing and the dropdown will automatically open with the search field focused.
* Pressing enter while searching will select/deselect the first available option.

Fixes #3572 

## Contributor Checklist

### optionals

### DIFF
--- a/components/dash-core-components/src/fragments/Dropdown.tsx
+++ b/components/dash-core-components/src/fragments/Dropdown.tsx
@@ -1,4 +1,4 @@
-import {isNil, without, isEmpty} from 'ramda';
+import {isNil, without, append, isEmpty} from 'ramda';
 import React, {
     useState,
     useCallback,
@@ -48,6 +48,7 @@ const Dropdown = (props: DropdownProps) => {
         document.createElement('div')
     );
     const searchInputRef = useRef<HTMLInputElement>(null);
+    const pendingSearchRef = useRef('');
 
     const ctx = window.dash_component_api.useDashContext();
     const loading = ctx.useLoading();
@@ -80,6 +81,8 @@ const Dropdown = (props: DropdownProps) => {
         (selection: OptionValue[]) => {
             if (closeOnSelect !== false) {
                 setIsOpen(false);
+                setProps({search_value: undefined});
+                pendingSearchRef.current = '';
             }
 
             if (multi) {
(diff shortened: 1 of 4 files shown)


components/dash-core-components/src/fragments/Dropdown.tsx
```diff
@@ -475,6 +495,22 @@ const Dropdown = (props: DropdownProps) => {
...
+                                    ) {
+                                        return;
+                                    }
+                                    const firstVal = displayOptions[0].value;
+                                    const isSelected =
+                                        sanitizedValues.includes(firstVal);
+                                    const newSelection = isSelected
+                                        ? without([firstVal], sanitizedValues)
+                                        : append(firstVal, sanitizedValues);
+                                    updateSelection(newSelection);
```
T4rk1n, line 554: I think this would read better with a standard if else.

T4rk1n (approved):
💃 Looks good.

PR #3630, opened 2026-02-26
fix: replace bare except clause with except Exception

haosenwang1018:
Replace bare except clause with except Exception.


AnnMarieW:
This looks fine, but it’s not a big deal since it’s not production code, it’s just a test.
@haosenwang1018  looks like a bot?

image

haosenwang1018:
CI failures require manual investigation.

PR #3633, opened 2026-02-27
fix: replace 1 bare except clause with except Exception

haosenwang1018:
Replace bare `except:` clauses with `except Exception:` for PEP 8 compliance.


emilykl:
On its own this is a pretty small change and has no effect since it's only in the tests, we'd prefer to look at our use of `except` across the whole codebase.

PR #3637, opened 2026-03-02
add debounce to Dropdown

### LINKED ISSUE #3620: [Feature Request] debounce for dcc.Dropdown(multi=True)
When using the dcc.Dropdown with multi=True, each selection/unselection of an item trigger a change of the value prop. 
It would be great to have a "debounce" feature implemented where the value is changed only when the dropdown is closed (or the x / deselect all shortcut is clicked) and the selection has changed.

AnnMarieW:
closes #3620 

This PR adds  an optional `debounce` (boolean) prop to the `Dropdown`

When `debounce=True`, changes to selected values will be sent back to the Dash server only when dropdown menu closes.

```python
from dash import Dash, dcc, html, Input, Output,callback

app = Dash()
app.layout = html.Div([
    html.Div(id='dd-output-container'),
    dcc.Dropdown(
        ['NYC', 'MTL', 'SF'],
        'NYC', id='demo-dropdown',
        debounce=True,
        closeOnSelect=False,
        multi=True),
])


@callback(
    Output('dd-output-container', 'children'),
    Input('demo-dropdown', 'value')
)
def update_output(value):
    return f'You have selected {value}'


if __name__ == '__main__':
    app.run(debug=True)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [UNRELEASED]
 
+## Added
+- [#3637](https://github.com/plotly/dash/pull/3637)  Added `debounce` prop to `Dropdown`.
+
 ## Fixed
 - [#3629](https://github.com/plotly/dash/pull/3629) Fix date pickers not showing date when initially rendered in a hidden container.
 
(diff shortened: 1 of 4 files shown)


AnnMarieW:
The two failed tests in dcc-312 are for date pickers which is unrelated to this PR. Flakey tests?

T4rk1n (approved):
:dancer:

PR #3641, opened 2026-03-04
fix: flatten outputs before building loadingOutputs

### LINKED ISSUE #3619: [BUG] dcc.Loading spinner does not trigger when callback Output uses ALL wildcard
**Describe your context**

```
dash                               4.0.0
dash-bootstrap-components          2.0.4
dash-table                         5.0.0
```

- OS: Windows 11
- MS Edge
- Version 145

**Describe the bug**

`dcc.Loading` never displays the loading spinner when a callback's Output uses the ALL wildcard in a pattern-matching ID, e.g.:

```python
Output({"type": "my-graph", "id": ALL, "name": MATCH}, "figure")
````

The same callback with MATCH or a fully-specified ID works correctly — the spinner appears during the callback execution.

**Root cause**

In [dash-renderer/src/actions/callbacks.ts](https://github.com/plotly/dash/blob/dev/dash/dash-renderer/src/actions/callbacks.ts#L794-L798), loadingOutputs is constructed as:

```python
loadingOutputs = outputs.map(function (out) {
    return {
        path: getPath(paths, out.id),
        property: out.property?.split('@')[0],
        id: stringifyId(out.id)
    };
});
```

When the output spec contains ALL, unwrapIfNotMulti returns the resolved outputs as an array (because isMultiValued returns true). This array is pushed into outputs as a single element, producing a nested structure like [[[{id, property}, ...]]](http://vscodecontentref/11). When .map() iterates over this, out is the inner array, so out.id is undefined → getPath returns undefined → the loading reducer receives a bogus path → `dcc.Loading` never matches it → no spinner.

Notably, the same function already uses flatten(outputs) correctly on [L829](https://github.com/plotly/dash/blob/dev/dash/dash-renderer/src/actions/callbacks.ts#L829) and [L912](https://github.com/plotly/dash/blob/dev/dash/dash-renderer/src/actions/callbacks.ts#L912) when processing the actual response data — the flattening was simply missed for the loading dispatch.

**Fix**

Change `outputs.map(…)` to `outputs.flat().map(…)` on [L794](https://github.com/plotly/dash/blob/dev/dash/dash-renderer/src/actions/callbacks.ts#L794) so the nested arrays from multi-valued outputs are flattened before resolving paths:

```python
loadingOutputs = outputs.flat().map(function (out) {
    // ...same body
});
```

**Expected behavior**

`dcc.Loading` should display the loading spinner for callbacks with ALL in the Output, the same as it does for MATCH or fully-specified outputs.

**Minimal reproducible example**

```python
"""Minimal Dash app to test dcc.Loading with ALL vs specific callback outputs."""

import time

from dash import ALL, MATCH, Dash, Input, Output, dcc, html

app = Dash(__name__, external_stylesheets=[dbc.themes.BOOTSTRAP])

FIGURE_A = {"data": [{"x": [1, 2, 3], "y": [1, 3, 2]}], "layout": {"title": "Option A"}}
FIGURE_B = {"data": [{"x": [1, 2, 3], "y": [3, 1, 2]}], "layout": {"title": "Option B"}}

RADIO_OPTIONS = [
    {"label": "Option A", "value": "a"},
    {"label": "Option B", "value": "b"},
]

graph_id_all = {"type": "graph-all", "id": "component", "name": "test"}
graph_id_match = {"type": "graph-match", "id": "component", "name": "test"}

app.layout = html.Div([
    html.H3("dcc.Loading test: ALL vs MATCH in Output"),
    dbc.Row([
        # ALL output — spinner BROKEN
        dbc.Col([
            html.H5("ALL in Output (broken)"),
            dbc.RadioItems(options=RADIO_OPTIONS, value="a",
                id={"type": "radio-all", "id": "component", "name": "test"}, inline=True),
            dcc.Loading(
                dcc.Graph(id=graph_id_all, figure=FIGURE_A),
                type="dot",
                overlay_style={"visibility": "visible", "opacity": 0.5, "filter": "blur(2px)"},
            ),
        ], width=6),
        # MATCH output — spinner WORKS
        dbc.Col([
            html.H5("MATCH in Output (works)"),
            dbc.RadioItems(options=RADIO_OPTIONS, value="a",
                id={"type": "radio-match", "id": "component", "name": "test"}, inline=True),
            dcc.Loading(
                dcc.Graph(id=graph_id_match, figure=FIGURE_A),
                type="dot",
                overlay_style={"visibility": "visible", "opacity": 0.5, "filter": "blur(2px)"},
            ),
        ], width=6),
    ]),
], className="p-3")


@app.callback(
    Output({"type": "graph-all", "id": ALL, "name": MATCH}, "figure"),
    Input({"type": "radio-all", "id": ALL, "name": MATCH}, "value"),
    prevent_initial_call=True,
)
def update_all(values):
    time.sleep(2)
    return [FIGURE_A if values[0] == "a" else FIGURE_B]


@app.callback(
    Output({"type": "graph-match", "id": MATCH, "name": MATCH}, "figure"),
    Input({"type": "radio-match", "id": MATCH, "name": MATCH}, "value"),
    prevent_initial_call=True,
)
def update_match(value):
    time.sleep(2)
    return FIGURE_A if value == "a" else FIGURE_B


if __name__ == "__main__":
    app.run(debug=True, port=8051)
```

i-murray:
## Summary

Fix `dcc.Loading` spinner not triggering when a callback `Output` uses the `ALL` wildcard.

## Problem

In `dash-renderer/src/actions/callbacks.ts`, `loadingOutputs` was built with `outputs.map(...)`. When the output spec contains `ALL`, `unwrapIfNotMulti` returns resolved outputs as a nested array. Iterating with `.map()` over this nested structure meant `out.id` was `undefined`, so `getPath` returned `undefined` and the loading reducer never matched the component path.

## Fix

Changed `outputs.map(...)` to `flatten(outputs).map(...)` on L794, consistent with how outputs are already flattened on L829 and L912 in the same function.

Fixes #3619


AnnMarieW:
Thanks for the fix @i-murray 

I tried your sample app after the fix and verified that it works.  Here is a simplified version that can be added as a test in this file:  https://github.com/plotly/dash/blob/dev/components/dash-core-components/tests/integration/loading/test_loading_component.py

```python

def test_ldcp019_loading_component_pattern_matching(dash_dcc):
    lock = Lock()

    app = Dash(__name__)

    app.layout = html.Div(
        [dcc.Loading([html.Div(id={"type": "div-1", "index": 1, "name": "test"}, className="div-1")], className="loading")],
        id={"type": "root", "index": 1, "name": "test"},
        className="root"
    )

    @app.callback(Output({"type": "div-1", "index": ALL, "name": MATCH}, "children"), Input({"type": "root", "index": ALL, "name": MATCH}, "n_clicks"))
    def updateDiv(n_clicks):
        if n_clicks ==  [1]:
            time.sleep(.1)
            return ["changed"]
        return ["content"]

    with lock:
        dash_dcc.start_server(app)
        dash_dcc.wait_for_text_to_equal(".div-1", "content")

        dash_dcc.find_element(".root").click()

        dash_dcc.find_element(".loading .dash-spinner")
        # mounted but hidden, so looks like no text
        dash_dcc.wait_for_text_to_equal(".div-1", "")

    dash_dcc.wait_for_text_to_equal(".div-1", "changed")

```

AnnMarieW:
@T4rk1n This is ready for review - I ran the new test locally, but I don't think I can add it to the PR since I'm not a maintainer.

i-murray:
@AnnMarieW, thank you for writing the test for this fix. I've added it.

T4rk1n (approved):
💃 Looks good, just need to run `npm run format` to pass the linting.

ndrezn:
Thanks for the contribution @i-murray !

i-murray:
@T4rk1n linting should pass now. There was also a duplicate test function name I fixed as well.

PR #3646, opened 2026-03-11
support react 19

### LINKED ISSUE #3231: implement support for React 19
We should start exploring React 19 support. React 19 has many changes which may affect how Dash works.

We know that component libraries will be affected as `defaultProps` is removed in React 19.

From @T4rk1n :

> The `defaultProps` was mostly handled with the upgrade to react 18 and most components libs now.
`propTypes` removal doesn't really affect us because we still have the runtime check on our own, we also have our own typescript prop type support altho that might be difficult to update in the future if typescript indeed goes with the go rewrite that been in the news recently.
> `contextType` removal is a new one I didn't see before, might have to update some components for it to work.
string refs removals was also part of the 18 update along with ReactDOM.render -> ReactDOM.createRoot
> `findDOMNode` removal was part of react 18 upgrade.
new one is JSX namespace is changed https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript might need some adaptation to the types.
> The rest of the changes I don't think would affect us that much, I don't see anything that would prevent us to update. The initial fear with react 19 was with their new compiler but I think that was dropped.

T4rk1n:
- Remove react 16 support.
- Add react 19 experimental support, use with: 
```python
from dash._dash_renderer import _set_react_version

_set_react_version("19.2.0")
```


AnnMarieW:
Thanks for supporting React 19 :tada:

I see all the tests pass, but when I run apps setting react to 19.2.0 some components work (like dcc.Markdown and dcc.Input) but others do not, such as dcc.RangeSlider and the date picker components.  The error is:

> Cannot read properties of undefined (reading 'ReactCurrentOwner')

I see the same error if I set react 19 in the tests for those components.

AnnMarieW:
I tried this again, and here's some more feedback:

The new React 19 shim seems to fix the ReactCurrentOwner error with DCC components.  

I also tried a simple quick start app in a  few other libraries:

-  dash-ag-grid 35.2 worked fine.
-  dash-daq 0.6.0 worked fine.
-  dash-bootstrap-components 2.0.4 had the errors shown below.   DBC v2 is based on react-bootstrap v2 which does not support React.19, so that wasn't surprising.
- dash-mantine-components 2.6.1 had the same errors as DBC.  However,  DMC v2 is based on Mantine v8 which is compatible with React 19, so that was a little unexpected.




------------
Errors in both DMC and DBC if a component is defined in the app:

> Uncaught TypeError: Cannot read properties of undefined (reading 'ReactCurrentOwner')
    at Array.<anonymous> (dash_bootstrap_components.v2_0_4m1759328733.min.js:2:12535)
    at n (dash_bootstrap_components.v2_0_4m1759328733.min.js:2:13035)
    at Array.<anonymous> (dash_bootstrap_components.v2_0_4m1759328733.min.js:2:12343)
    at n (dash_bootstrap_components.v2_0_4m1759328733.min.js:2:13035)
    at dash_bootstrap_components.v2_0_4m1759328733.min.js:2:17559
    at dash_bootstrap_components.v2_0_4m1759328733.min.js:2:227635
    at dash_bootstrap_components.v2_0_4m1759328733.min.js:2:227674


Also, if the component is added to the layout, the app doesn't render, and there is also the following errors:

> Uncaught Error: dash_bootstrap_components was not found.
    at Object.resolve (registry.js:15:15)
    at getProps (persistence.js:293:30)
    at persistenceMods (persistence.js:404:9)
    at persistence.js:437:29
    at Array.forEach (<anonymous>)
    at persistenceMods (persistence.js:435:18)
    at applyPersistence (persistence.js:370:12)
    at APIController.react.js:161:53
    at exports.unstable_batchedUpdates (react-dom.development.js:411:14)
    at storeEffect (APIController.react.js:150:10)Understand this error


> installHook.js:1 An error occurred in the <UnconnectedContainer> component.
Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://react.dev/link/error-boundaries to learn more about error boundaries.
overrideMethod @ installHook.js:1Understand this warning

T4rk1n:
Superseded by #3880

PR #3656, opened 2026-03-11
dcc.Dropdown: Use virtualized lists for large collections of options

### LINKED ISSUE #3616: [BUG] dcc.Dropdown in dash 4.0.0 is considerably slower than before
I have a dash app with dcc.Dropdown with thousands of options (around 8000). In dash 3.4.0 the dropdown works amazingly (e.g. beating dmc.Select by a margin in speed). In dash 4.0.0, the performance degraded considerably.

Here is a minimal app to simulate the issue:
```
import random
import string

import dash

# Generate 8000 random options
def random_string(length=8):
    return ''.join(random.choices(string.ascii_letters + string.digits, k=length))

options_list = [
    {"label": f"Option {i} - {random_string()}", "value": f"value_{i}"}
    for i in range(8000)
]

app = dash.Dash(__name__)

app.layout = dash.dcc.Dropdown(
    id="my-dropdown",
    options=options_list,
    placeholder="Select an option",
    searchable=True
)

if __name__ == "__main__":
    app.run(debug=True, host="127.0.0.1", port="8050")
```
For the two versions, try typing "78" to obtain filtered options and then deleting the search string to obtain the full list of options and compare the timing.

I understand that this many options may be stretching the limits a bit (and I know it would be better to only show a subset in most cases). Nevertheless, it would be nice to keep the previously achieved performance :-)

Finally, let me appreciate the new styling (of the dropdown) in dash 4.0.0!

KoolADE85:
This PR virtualizes lists when rendering lists of options (using `react-window` library).

Also, replaced a few `options.map()` calls with a Set for faster lookups.

This allows the dropdown component to render huge amounts of options without any perceivable performance impact.

Fixes #3616 

## Contributor Checklist

### optionals


T4rk1n (approved):
💃 Looks good

PR #3660, opened 2026-03-12
Fix single-date selection in DatePickerRange

### LINKED ISSUE #3645: [BUG] Dash 4.0.0 Issue DatePickerRange: unable to select same start/end date and end_date silently retained after new start_date selection (minimum_nights=0)
**Describe your context**

```
dash==4.0.0
dash-bootstrap-components==2.0.4
dash-iconify==0.1.2
dash-mantine-components==2.5.1
```

    - Windows
    - Edge
    - Version 145.0.3800.70 (Official build) (64-bit)

**Describe the bug**

Dash 4.0.0 has made some changes to DateRangePicker. I assume it is a bug, but we're unable to select the same date for start and end date. Meaning it is no longer possible to select a single date as it was before, even if we keep `minimum_nights=0`.

At the same time, it is possible to induce a "hidden end date". If you select a range, and select a single date afterwards, the function uses the previously selected end date. In dash \<4.0.0 this was visible in the UI, but in 4.0.0 this left over end date is now hidden. 

You can use the following script:

```

app = Dash(__name__)

app.layout = html.Div([
    dcc.DatePickerRange(
        id="date-picker",
        minimum_nights=0,
    ),
    html.Div(id="output"),
])

@app.callback(
    Output("output", "children"),
    Input("date-picker", "start_date"),
    Input("date-picker", "end_date"),
)
def display_dates(start_date, end_date):
    print(f"Selected dates: {start_date} to {end_date}")
    return f"Selected: {start_date} to {end_date}"

if __name__ == "__main__":
    app.run(host="0.0.0.0", port=8080, debug=True)
```

**Issue with being unable to select a single date**

Try to select the same date twice. You won't be able to, but you could do this with dash <4.0.0. 

**Issue with hidden end date**

Now select a range, e.g. 2026-02-02 to 2026-02-11

Image

Then select 2026-02-04

Image

You can now see that the end date is the same as before but hidden. This caused a lot of confusion among our users. 

**Expected behavior**

1. Single date selection: When minimum_nights=0, selecting the same date for start and end should work (start_date == end_date).
2. No hidden end date: When a new start date is selected, the end date should be cleared/reset, not silently retained from a previous selection.

KoolADE85:
Allows the same date to be used for both start_date and end_date when selecting dates.

Fixes #3645 

## Contributor Checklist

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## Fixed
 - [#3629](https://github.com/plotly/dash/pull/3629) Fix date pickers not showing date when initially rendered in a hidden container.
 - [#3627][(](https://github.com/plotly/dash/pull/3627)) Make dropdowns searchable wheen focused, without requiring to open them first
+- [#3660][(](https://github.com/plotly/dash/pull/3660)) Allow same date to be selected for both start and end in DatePickerRange components
 
 
 
--- a/components/dash-core-components/src/fragments/DatePickerRange.tsx
+++ b/components/dash-core-components/src/fragments/DatePickerRange.tsx
@@ -174,8 +174,10 @@ const DatePickerRange = ({
                 end_date: null,
             });
         } else if (updatemode === 'singledate' && internalStartDate) {
+            // Only start changed - send just that one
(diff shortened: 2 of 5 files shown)


T4rk1n (approved):
💃

PR #3668, opened 2026-03-17
fixing issue where FastAPI would not allow new paths to be added

### LINKED ISSUE #3667: [BUG] v4.1 FastAPI Paths
In v4.1, FastAPI order matters for when adding url rules.

Subsequently, `/{path:path}` was registered on app init, this meant all routes after this were never used by the server.

BSd3v:
fixes #3667 

This uses the lifespan to load the catchall when the server starts, allowing all other urls to be added properly.


dash/backends/_fastapi.py
```diff
@@ -179,6 +184,12 @@ async def _handle_error(
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         # Handle lifespan events (startup/shutdown)
         if scope["type"] == "lifespan":
+            try:
+                dash_app = get_app()
+                dash_app.backend._setup_catchall()
+            except:
+                print("Error during catch-all setup:")
+                print(traceback.format_exc())
```
T4rk1n: `traceback.print_exc()` should be enough.

dash/backends/_fastapi.py
```diff
@@ -275,11 +286,24 @@ async def index(_request: Request):
...
+        and timing to work correctly on all routes. FastAPI will match this catch-all route
+        for any path that isn't matched by a more specific route, allowing the middleware to
+        process the request and then return the appropriate response (e.g., 404 if no Dash route matches).'''
 
-        # pylint: disable=protected-access
-        dash_app._add_url("{path:path}", catchall, methods=["GET"])
+
+    def _setup_catchall(self):
+        try:
+            print("Setting up catch-all route for unmatched paths")
```
T4rk1n: Remove print or replace with a logger.debug call.

dash/backends/_fastapi.py
```diff
@@ -289,6 +313,7 @@ def add_url_rule(
         methods: list[str] | None = None,
         include_in_schema: bool = False,
     ):
+        print(f"Adding URL rule: {rule} -> {view_func} (endpoint: {endpoint}, methods: {methods})")
```
T4rk1n: logger.debug

PR #3672, opened 2026-03-19
adjustments for computeGraphs 

### LINKED ISSUE #3661: [BUG] dash-renderer 4.0: O(n²) dependency graph resolution causes 8-9 second freeze in Firefox with ~400 callbacks
---

**Describe your context**

```
dash                      4.0.0
dash_ag_grid              32.3.2
dash-bootstrap-components 2.0.4
dash-bootstrap-templates  2.1.0
dash_daq                  0.6.0
dash-extensions           2.0.4
dash-iconify              0.1.2
dash-leaflet              1.1.3
dash-loading-spinners     1.0.3
dash_mantine_components   2.3.0
dash-svg                  0.0.12
```

- OS: Windows 11
- Browser: Firefox (latest stable), also tested in Chrome
- Chrome handles the same app without noticeable delay

**Describe the bug**

Our production app registers ~400 callbacks (~150 using pattern-matching with `ALL`/`MATCH`). On page load, the `_dash-dependencies` response is 332kB. All network requests complete within 1.3 seconds, but there's an 8-9 second freeze before the first `_dash-update-component` POST fires. The entire delay is client-side JavaScript execution in the dash-renderer.

A Firefox profiler capture shows the content thread burning **36.68 seconds of CPU** during this window. The hot functions are all in the dependency graph builder:

| Samples | Function |
|---|---|
| 2071 (27.8%) | `forEach` |
| 1857 (24.9%) | `Tn` |
| 1826 (24.5%) | `Ln/<` |
| 1799 (24.2%) | `some` |
| 1788 (24.0%) | `Tn/<` |
| 1768 (23.8%) | `map` |
| 79 (1.1%) | `addDependency` |

The `forEach`/`some`/`map` nesting pattern and the presence of `addDependency` suggest O(n²) behavior in the callback dependency graph resolution — for each callback, it appears to check against all others.

Chrome's V8 engine handles the same workload fine, which suggests Firefox's SpiderMonkey doesn't optimize these particular loop patterns as aggressively. That said, 400 callbacks shouldn't require 37 seconds of CPU in any browser.

**Expected behavior**

The dependency graph for ~400 callbacks should resolve in under 1 second. The current implementation appears to scale quadratically — this will become a bigger problem as apps grow.

**Steps to reproduce**

Any Dash 4 app with ~400+ callbacks, especially with heavy use of pattern-matching IDs (`ALL`/`MATCH`), loaded in Firefox. Chrome masks the issue due to engine-level optimizations, but the underlying algorithmic complexity is the same.

**Additional context**

HAR file confirms the timeline: all assets and `_dash-dependencies` load by 1.3s, then nothing happens until 9.3s when the first callback POST fires. The gap is pure JS execution. Happy to share the Firefox profiler capture and HAR if helpful.

BSd3v:
fixes #3661 
adjustments for computeGraphs to not have validation when in debug is False or `dev_tools_validate_callbacks` is False

### DIFF
--- a/tests/integration/renderer/test_benchmarking.py
+++ b/tests/integration/renderer/test_benchmarking.py
@@ -111,12 +111,13 @@ def test_compute_graph_timing(dash_duo, dev_tools, store):
     dash_duo.start_server(app, **dev_tools)
     times = []
     for _ in range(10):
+        dash_duo.wait_for_element("#input")
         wait.until(
             lambda: dash_duo.find_element("#output").text.strip() != "", timeout=4
         )
         graph_compute_time = float(
             dash_duo.driver.execute_script(
-                "return window.dash_clientside.callbackGraphTime"
+                "return window.dash_component_api.callbackGraphTime"
             )
         )
         times.append(graph_compute_time)
--- a/tests/integration/renderer/test_benchmarking.py
+++ b/tests/integration/renderer/test_benchmarking.py
@@ -4,7 +4,7 @@ import time
 import pytest
 
 
-def make_app(num_groups=500, items_per_group=10):
+def make_app(num_groups=500, items_per_group=20):
     app = Dash(__name__)
 
     NUM_GROUPS = num_groups
@@ -131,5 +131,5 @@ def test_compute_graph_timing(dash_duo, dev_tools, store):
     if store == "disabled":
         print(f"Average time with store disabled: {avg_time:.2f} ms")
         assert (
-            avg_time < 500
-        ), "Expected average time to be under 1/2 seconds with circular callback check disabled"
+            avg_time < 100
+        ), "Expected average time to be under 100 ms with circular callback check disabled"
--- a/tests/integration/renderer/test_benchmarking.py
+++ b/tests/integration/renderer/test_benchmarking.py
@@ -4,7 +4,7 @@ import time
 import pytest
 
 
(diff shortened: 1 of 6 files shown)


tests/integration/renderer/test_benchmarking.py
```diff
@@ -0,0 +1,136 @@
...
+    check_timing[store] = avg_time
+    if store == "enabled":
+        print(f"Average time with store enabled: {avg_time:.2f} ms")
+        assert (
+            check_timing["disabled"] < avg_time
+        ), "Expected faster performance with circular callback check disabled"
+    if store == "disabled":
+        print(f"Average time with store disabled: {avg_time:.2f} ms")
+        assert (
+            avg_time < 100
```
ndrezn, line 135: Having a fixed time test could lead to flakiness if our runner types change or have downtime/performance issues. @T4rk1n do you have any ideas on how to add performance regression testing for this change?

T4rk1n: Not sure there are better ways, the threshold should reflect what we think is acceptable.

BSd3v: If you want to benchmark outside of this, on specific devices, it might be worth running it on something like lambdatest, browser stack or something similar.

T4rk1n (approved):
💃 Looks good.

PR #3690, opened 2026-03-24
Fix input when min or max is None

### LINKED ISSUE #3689: [BUG] Dash 4.0.0 number dcc.Input increment doesn't work when max or min = None
When setting the min or max option in a `dcc.Input` the increment button stops working. Clicking it erases the current value in the input box. Try with the simple app added below. The increment buttons don't work for the top input field, but they do work for the bottom input field.

I used Dash 4.0.0 for testing.

```python
import dash
from dash import html, dcc

app = dash.Dash(__name__)


app.layout = html.Div(
    [
        html.H1("Hello Dash"),
        dcc.Input(min=0, max=None, type="number"),
        dcc.Input(min=0, type="number"),
    ],
    style={"width": "20%"},
)

app.run(debug=True)
```

AnnMarieW:
Closes #3689 

Fixes a bug in `Input` when `min` or `max` is set to `None`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [UNRELEASED]
+
+## Fixed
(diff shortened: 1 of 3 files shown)


T4rk1n (approved):
💃

PR #3711, opened 2026-04-01
MCP Server Part 2: Represent callback-related data with a `CallbackAdapter`

KoolADE85:
## Summary

- Add `CallbackAdapter`: wraps a Dash callback and exposes MCP-related metadata (computed lazily)
  * It combines Dash's own callback map, layout, and user docstrings/types into 1 interface that the MCP server will use to present a complete picture of the app to LLMs.
- Add `CallbackAdapterCollection` for working with collections of adapters
- Add `mcp_enabled` parameter to `@app.callback()` for opting callbacks out of MCP
- Add `mcp>=1.0.0` to dependencies

## Manual testing

* You can inspect what kind of data the `CallbackAdapter` exposes
* you can run callbacks directly with `run_callback` (which is what MCP will do)

```python

app = Dash(__name__)
app.layout = html.Div([
    dcc.Input(id="name", value="World"),
    html.Div(id="greeting"),
])
@app.callback(Output("greeting", "children"), Input("name", "value"))
def greet(name):
    """Greet the user by name."""
    return f"Hello, {name}!"

    app_context.set(app)
    from dash.mcp.primitives.tools.callback_utils import run_callback
    for adapter in collection:
        print(f"Tool: {adapter.tool_name}")
        print(f"  Docstring: {adapter._docstring}")
        print(f"  Outputs: {[o['id_and_prop'] for o in adapter.outputs]}")
        print(f"  Inputs:  {[i['name'] for i in adapter.inputs]}")

    # Execute the callback through the adapter
    adapter = collection[0]
    result = run_callback(adapter, {"name": "Alice"})
    print(f"Callback result: {result}")

### DIFF
--- a/dash/mcp/primitives/tools/callback_adapter.py
+++ b/dash/mcp/primitives/tools/callback_adapter.py
@@ -1,0 +1,461 @@
+"""Adapter: Dash callback → MCP tool interface.
+
+Wraps a raw ``callback_map`` entry and exposes MCP-facing
+properties (tool name, params, outputs) lazily.
+"""
+
+from __future__ import annotations
+
+import inspect
+import json
+import typing
+from functools import cached_property
+from typing import Any
+
+from mcp.types import Tool
+
+from dash import get_app
+from dash.layout import (
+    _WILDCARD_VALUES,
+    find_component,
+    find_matching_components,
+    parse_wildcard_id,
+)
+from dash.mcp.types import is_nullable
+from dash._grouping import flatten_grouping
+from dash._utils import clean_property_name, split_callback_id
+from dash.mcp.types import MCPInput, MCPOutput
+from .callback_utils import run_callback
+from .descriptions import build_tool_description
+from .input_schemas import get_input_schema
+from .output_schemas import get_output_schema
+
+
+class CallbackAdapter:
+    """Adapts a single Dash callback_map entry to the MCP tool interface."""
+
+    def __init__(self, callback_output_id: str):
+        self._output_id = callback_output_id
+
+    # -------------------------------------------------------------------
+    # Projections
+    # -------------------------------------------------------------------
+
+    @cached_property
+    def as_mcp_tool(self) -> Tool:
+        """Stub — will be implemented in a future PR."""
+        raise NotImplementedError("as_mcp_tool will be implemented in a future PR.")
+
+    def as_callback_body(self, kwargs: dict[str, Any]) -> dict[str, Any]:
+        """Transforms the given kwargs to a dict suitable for calling this callback.
+
+        Mirrors how the Dash renderer assembles the callback payload —
+        see ``fillVals()`` in ``dash-renderer/src/actions/callbacks.ts``.
+
+        For pattern-matching callbacks, wildcard deps are expanded into
+        nested arrays with concrete component IDs.
+        """
+        coerced = {k: _coerce_value(v) for k, v in kwargs.items()}
+
+        raw_inputs = self._cb_info.get("inputs", [])
+        raw_state = self._cb_info.get("state", [])
+        n_deps = len(raw_inputs) + len(raw_state)
+
+        flat_values = [None] * n_deps
+        for i, name in enumerate(self._param_names):
+            if i < n_deps and name in coerced:
+                flat_values[i] = coerced[name]
+
+        inputs_with_values = [
+            _expand_dep(dep, flat_values[i]) for i, dep in enumerate(raw_inputs)
+        ]
+        state_with_values = [
+            _expand_dep(dep, flat_values[len(raw_inputs) + i])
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: dash/mcp/primitives/tools/callback_adapter.py [hunk 1]
```diff
@@ -0,0 +1,461 @@
...
+    parse_wildcard_id,
+)
+from dash.mcp.types import is_nullable
+from dash._grouping import flatten_grouping
+from dash._utils import clean_property_name, split_callback_id
+from dash.mcp.types import MCPInput, MCPOutput
+from .callback_utils import run_callback
+from .descriptions import build_tool_description
+from .input_schemas import get_input_schema
+from .output_schemas import get_output_schema
```
T4rk1n: dash.mcp.types appears two times and there is a mix of absolute vs relative imports.

dash/mcp/primitives/tools/callback_adapter.py
```diff
@@ -0,0 +1,461 @@
...
+        if self._func_signature is None:
+            return [None] * len(self._dep_param_map)
+        try:
+            hints = typing.get_type_hints(self._original_func)
+        except Exception:
+            hints = getattr(self._original_func, "__annotations__", {})
+        return [hints.get(func_name) for func_name, _ in self._dep_param_map]
+
+
+def _expand_dep(dep: dict, value: Any) -> Any:
```
T4rk1n: Add a type to this return value.

dash/mcp/primitives/tools/callback_adapter.py
```diff
@@ -0,0 +1,461 @@
...
+
+    # LLM provides browser-like format
+    if isinstance(value, list):
+        return value
+    if isinstance(value, dict) and "id" in value:
+        return value
+    return {**dep, "value": value}
+
+
+def _expand_output_spec(output_id: str, cb_info: dict, resolved_inputs: list) -> Any:
```
T4rk1n: Add a type to this return value.

dash/mcp/primitives/tools/callback_adapter.py
```diff
@@ -0,0 +1,461 @@
...
+            if len(expanded) == 1:
+                results.append(expanded[0])
+            else:
+                results.append(expanded)
+        else:
+            results.append({"id": pid, "property": prop})
+
+    if len(results) == 1:
+        return results[0]
+    return results
```
T4rk1n: I think returning the array and handling the results length at call site would be better since this makes the return value an union.

dash/mcp/primitives/tools/callback_adapter.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+def _derive_output_ids(
+    output_pattern: dict, resolved_inputs: list
+) -> list[dict] | None:
```
T4rk1n: Add type definition to the dict if possible.

dash/mcp/primitives/tools/callback_adapter.py
```diff
@@ -0,0 +1,461 @@
...
+        # MATCH: single {id, property, value} dict
+        elif isinstance(entry, dict) and isinstance(entry.get("id"), dict):
+            out = _substitute(entry["id"])
+            if out:
+                return [out]
+
+    return None
+
+
+def _coerce_value(value: Any) -> Any:
```
T4rk1n: Could add a generic here `[TInput, TOutput]`

dash/mcp/types/exceptions.py
```diff
@@ -0,0 +1,30 @@
...
+class InvalidParamsError(MCPError):
+    """Invalid or missing parameters for a tool call."""
+
+    code = -32602
+
+
+class CallbackExecutionError(MCPError):
+    """Callback raised an exception during execution."""
+
+    code = -32603
```
T4rk1n: What are these code? This one has the same code as the base MCPError

dash/mcp/primitives/tools/callback_adapter.py
KoolADE85: Good call, this is now always returning a list.

dash/mcp/primitives/tools/callback_adapter.py
KoolADE85: Even better: this function wasn't really needed anymore as most usages had been refactored away. I just refactored its last remaining user and this function is now removed.

dash/mcp/types/exceptions.py
KoolADE85: These are standard JSON-RPC error codes, as per the MCP spec on [error handling](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling).
I removed the redundant one.

T4rk1n (approved):
💃

PR #3712, opened 2026-04-01
MCP Server Part 3: Dash App `Resources`

KoolADE85:
## Summary

- Add read-only MCP resources:
    * `dash://layout`: retrieves the layout in the same json format as the `_dash-layout` endpoint
    * `dash://components`: retrieves a flat list of components which have IDs, along with their types
    * `dash://pages`: retrieves all pages from the page registry
    * `dash://page-layout/{path}`: a template for retrieving the layout of a specific page
    * `dash://clientside-callbacks`: describes all clientside callbacks for LLM discovery

- Resources may be conditionally listed (e.g. `dash://pages` only appears for multi-page apps)

## Manual testing

```python
import json

app = Dash(__name__)
app.layout = html.Div([
    dcc.Dropdown(id="city", options=["NYC", "LA", "Chicago"], value="NYC"),
    dcc.Graph(id="chart"),
])
@app.callback(Output("chart", "figure"), Input("city", "value"))
def update_chart(city):
    return {"data": [{"x": [1, 2], "y": [3, 4]}], "layout": {"title": city}}

    app_context.set(app)
    from dash.mcp.primitives.resources import list_resources, read_resource
    result = list_resources()
    for r in result.resources:
        print(f"Resource: {r.uri} — {r.name}")
    result = read_resource("dash://components")
    print(json.loads(result.contents[0].text))

### DIFF
--- a/dash/mcp/primitives/resources/resource_clientside_callbacks.py
+++ b/dash/mcp/primitives/resources/resource_clientside_callbacks.py
@@ -1,0 +1,95 @@
+"""Clientside callbacks resource."""
+
+from __future__ import annotations
+
+import json
+from typing import Any
+
+from mcp.types import (
+    ReadResourceResult,
+    Resource,
+    ResourceTemplate,
+    TextResourceContents,
+)
+
+from dash import get_app
+from dash._utils import clean_property_name, split_callback_id
+
+URI = "dash://clientside-callbacks"
+
+
+def get_resource() -> Resource | None:
+    if not _get_clientside_callbacks():
+        return None
+    return Resource(
+        uri=URI,
+        name="dash_clientside_callbacks",
+        description=(
+            "Actions the user can take manually in the browser "
+            "to affect clientside state. Inputs describe the "
+            "components that can be changed to trigger an effect. "
+            "Outputs describe the components that will change "
+            "in response."
+        ),
+        mimeType="application/json",
+    )
+
+
+def get_template() -> ResourceTemplate | None:
+    return None
+
+
+def read_resource(uri: str = "") -> ReadResourceResult:
+    data = {
+        "description": (
+            "These are actions that the user can take manually in the "
+            "browser to affect the clientside state. Inputs describe "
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: dash/mcp/primitives/resources/resource_clientside_callbacks.py [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+            "components that can be changed to trigger an effect. "
+            "Outputs describe the components that will change "
+            "in response."
+        ),
+        mimeType="application/json",
+    )
+
+
+def get_template() -> ResourceTemplate | None:
+    return None
```
T4rk1n: Why is this necessary if it only returns None? Seems like the modules are being used as a class, maybe refactor to a base class instead.

dash/mcp/primitives/resources/resource_page_layout.py
```diff
@@ -0,0 +1,77 @@
...
+
+def read_resource(uri: str) -> ReadResourceResult:
+    path = uri[len(URI) :]
+    if not path.startswith("/"):
+        path = "/" + path
+
+    try:
+        from dash._pages import PAGE_REGISTRY
+    except ImportError:
+        raise ValueError("Dash Pages is not available.")
```
T4rk1n: Why would this import be not available?

T4rk1n (commented):
Looks good overall, just wondering why the pages import wouldn't work since that is a pretty standard module.

dash/mcp/primitives/resources/resource_clientside_callbacks.py
KoolADE85: That is a great idea! I added a base class and removed the no-op methods from each type of resource.

dash/mcp/primitives/resources/resource_page_layout.py
KoolADE85: Good point - I guess it would always be there. I removed the defensive import and put it at the top.

T4rk1n (approved):
Looks good 💃

PR #3723, opened 2026-04-02
Fix misaligned slider marks with empty labels

### LINKED ISSUE #3715: [Bad behaviour] dcc.Slider marks misaligned with empty label in Dash 4
When creating custom marks for a dcc.Slider in Dash 4, the marks/ticks become misaligned if the label is empty. See this example slider:

Image

The marks/ticks without label sink to the bottom of the mark+label box. In this example I used an empty string for an empty label. If I add some spaces there, the result is the same. 

Use Dash 4.1.0 and this small app to replicate:
```
import dash
from dash import dcc, html

app = dash.Dash(__name__)

empty_label = ""
# empty_label = "\xa0"
marks = {val: (val if val % 10 == 0 else empty_label) for val in range(0, 151, 5)}

app.layout = html.Div(
    [
        html.H2("Dash 4 Slider tick misalignment"),
        dcc.Slider(
            id="slider",
            min=0,
            max=150,
            step=None,
            value=0,
            marks=marks,
        ),
    ],
    style={"width": "40%"}
)

if __name__ == "__main__":
    app.run(debug=True)
```

I did find a (temporary) workaround. By taking an invisible unicode character as the empty label, for example: `\xa0`, it is possible to get the marks aligned again, see:

Image

Though it would be nicer if we wouldn't have to rely on a hack like that.

terminalchai:
## Summary

- add an explicit `height: 12px` to `.dash-slider-mark` so empty labels keep the same mark box height as labeled marks
- add an integration regression test covering the empty-string marks case from `#3715`
- add the corresponding unreleased changelog entry

## Contributor Checklist

   - [x] add the slider mark height fix
   - [x] add a regression test for empty mark labels
   - [x] add the changelog entry

### optionals

Fixes #3715

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Fixed
 - [#3690](https://github.com/plotly/dash/pull/3690) Fixes Input when min or max is set to None
+- [#3723](https://github.com/plotly/dash/pull/3723) Fix misaligned `dcc.Slider` marks when some labels are empty strings
 
 ## [4.1.0] - 2026-03-23
 
--- a/components/dash-core-components/src/components/css/sliders.css
+++ b/components/dash-core-components/src/components/css/sliders.css
@@ -92,6 +92,7 @@
 .dash-slider-mark {
     position: absolute;
     font-size: 12px;
+    height: 12px;
     line-height: 12px;
     color: var(--Dash-Text-Strong);
     white-space: nowrap;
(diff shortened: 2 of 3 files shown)


AnnMarieW:
Thanks!  Looks good.

AnnMarieW:
Just needs a changelog entry

terminalchai:
Added the unreleased changelog entry and updated the PR checklist accordingly.

KoolADE85 (approved):
Thanks for the fix, it's looking good to me.

PR #3731, opened 2026-04-08
MCP Server Part 4: Expose callbacks as tools

KoolADE85:
## Summary

This PR turns Dash callbacks into fully-described MCP tools. Each tool has four main parts:

### Tool description (`descriptions/`)
A human-readable summary of what the callback does. Built from pluggable sources that each contribute lines:
- `description_outputs.py` — semantic summary of what the callback returns (e.g. "Returns chart/visualization data" for a `Graph.figure` output)
- `description_docstring.py` — the callback function's Python docstring

Each source extracts what it needs from a given `CallbackAdapter` in order to produce text descriptions. New sources can be added by appending to the `_SOURCES` list.

### Input schema (`input_schemas/`)
JSON Schema that describes valid inputs for each callback parameter. Each schema is derived from:
- Python type annotations on the callback function (e.g. `def callback(metric: str):` → `{"type": "string"}`)
- Automatic introspection of the callback's components (e.g. `Input("dropdown", "value")` → `{"type": { "anyOf": ["string", "number", "boolean", null]}}`)
- Manually curated overrides for components where introspection is insufficient (e.g. the "date" format of date pickers; dropdown values are an array only when `multi=True`)

### Input description (`input_descriptions/`)
Each parameter also gets a text description assembled from various sources:
- Text content of an `html.Label` that is associated with the input
- Per-parameter text from the callback's docstring
- A list of props that are used in the app  (`options`, `min`/`max`, `value`, etc) and any chained callbacks that set values for this input

### Output schema (`output_schemas/`)
JSON Schema for the tool's `outputSchema` field
- A generic schema derived from the dict that Dash callbacks return to `dash-renderer`.

## Manual testing

You can see the JSON that will be presented to LLMs by testing manually:

```python
import json

app = Dash(__name__)
app.layout = html.Div([
    html.Label("Choose a metric", htmlFor="metric"),
    dcc.Dropdown(id="metric", options=["revenue", "users"], value="revenue"),
    html.Label(["Threshold value", dcc.Input(id="threshold", type="number", value=100)]),
    dcc.Dropdown(id="sub-metric"),
    html.Div(id="result"),
])

@app.callback(Output("sub-metric", "options"), Input("metric", "value"))
def update_sub_metrics(metric):
    """Update available sub-metrics based on the selected metric."""
    if metric == "revenue": return ["gross", "net", "recurring"]
    return ["daily", "monthly"]

@app.callback(Output("result", "children"), Input("metric", "value"),
              Input("threshold", "value"), Input("sub-metric", "value"))
def filter_data(metric: str, threshold: float, sub_metric: str):
    """Filter data by metric and threshold."""
    return f"{metric}/{sub_metric} > {threshold}"

    app_context.set(app)
    for adapter in collection:
        tool = adapter.as_mcp_tool
        print(tool)

### CODE REVIEW

File: .github/workflows/testing.yml [hunk 1]
```diff
@@ -105,6 +105,8 @@ jobs:
           echo "DISPLAY=:99" >> $GITHUB_ENV
 
       - name: Run lint
+        env:
+          PYLINT_EXTRA_ARGS: ${{ matrix.python-version == '3.8' && '--ignored-modules=mcp' || '' }}
```
T4rk1n, line 109: Maybe out of scope of this PR, but are we really running 3.8? min should be 3.9

tests/unit/mcp/tools/input_schemas/input_descriptions/test_descriptions.py
```diff
@@ -0,0 +1,424 @@
...
+                    dcc.Input(id="inp"),
+                    html.Div(id="out"),
+                ]
+            ),
+            ("inp", "value"),
+        )
+        tool = _user_tool(_tools_list(app))
+        assert "Your Name" in _desc_for(tool)
+
+    def test_html_for_not_adjacent(self):
```
T4rk1n: These tests need a code and number like the rest of the codebase. It makes the test easier to find and you can also filter easily which tests runs.

tests/unit/mcp/tools/input_schemas/input_descriptions/test_descriptions.py
```diff
@@ -0,0 +1,424 @@
...
+        desc = _desc_for(tool)
+        assert "Other Field" not in (desc or "")
+
+
+# ---------------------------------------------------------------------------
+# Component-specific: date pickers
+# ---------------------------------------------------------------------------
+
+
+class TestDatePickerDescriptions:
```
T4rk1n: I'd rather see different files and simple function for the tests instead of classes as sections.

More files == more evenly splitting in the ci.

T4rk1n (changes requested):
Looks good, but need to make the tests in style of the rest of the repo.

- Single function, no class unless necessary.
- code for the tests eg: test_mscp00x_* would be for tests in mcp schema component proptypes.
- More tests files for separation and splitting.

.github/workflows/testing.yml
KoolADE85: yeah, all the CI tests are in 3.8 still.
We should definitely consider bumping the min version. The `mcp` package requires `>=3.10` so if we can make that the min version, we would be able to run all tests without any conditions.
I also considered that out of scope for this PR, but maybe worth revisiting?

T4rk1n (approved):
💃

PR #3732, opened 2026-04-09
Initialize ai agents documentation

T4rk1n:
Add claude/ai agents documentation for the project.


AnnMarieW:
Who is the audience for the docs?  Is it MCP/LLM runtime docs?

The COMMANDS.md  TESTING.md and RENDERER.md seems like it's more relevant to a dash contributor.

How was the file generated?  The TROUBLESHOOTING.md looks shallow and some of it is incorrect, for example:

```
"Circular dependency detected"
Cause: Callback output is also its own input (directly or indirectly).

Solution: Restructure callbacks to break the cycle. Use State instead of Input where possible, or split into multiple callbacks.

```

T4rk1n:
> Who is the audience for the docs? Is it MCP/LLM runtime docs?

It's AI generated and for AI agents like claude. It is meant to describe the project context. The TESTING describe how dash_duo works since it's part of the framework. If it's useful to a dash contributor then I think it's doing a good job.

The TROUBLESHOOTING part I was not really sure, it was a suggestion from claude to add it. If you notice incorrect info please add in a review.

AnnMarieW:
OK - I thought this was docs for the MCP server.  For dash contributors, this looks good.  Not sure how useful the troubleshooing part is for this use-case.

.ai/TESTING.md
```diff
@@ -0,0 +1,486 @@
...
+
+# Run integration tests
+pytest tests/integration/
+
+# Run specific test
+pytest tests/integration/callbacks/test_basic.py::test_name
+
+# Headless mode (CI)
+pytest --headless tests/integration/
+```
```
KoolADE85: I have some local rules on this topic that I give to CLAUDE, maybe it makes sense as general-purpose instructions?


````suggestion
## Running tests — CRITICAL RULES

You must activate the virtual environment before running any tests.

### Unit tests

```
source .venv/bin/activate && python -m pytest "tests/unit/" 2>&1
```

### Integration tests

Always work with a specific test or test file. NEVER run all integration tests at once.

```
source .venv/bin/activate && python -m pytest tests/integration/callbacks/test_basic_callback.py --headless -xvs 2>&1
```

**NEVER truncate test output so aggressively that you cannot see failures.**
`tail -5` or `tail -10` is NEVER acceptable — you will miss the error details and waste time re-running.
Always use `tail -50` at minimum, or omit the tail entirely for short test runs.

**NEVER use `grep "FAILED"` to filter test output** — it hides the actual error messages.

When a test fails, you MUST be able to see the assertion error, traceback, and context in a SINGLE run. If your command truncates this, your command is wrong.
````

File: AGENTS.md [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+
+# Test
+pytest tests/unit                                # Unit tests
+pytest tests/integration                         # Integration tests
+pytest -k test_name                              # Specific test
+
+# Lint
+npm run lint                                     # All linters
+npm run private::format.black                    # Auto-format Python
+```
```
KoolADE85: I don't think that it's actually useful to provide this section. LLMs are pretty good at taking in all the docs, and it's likely that this section will become out-of-sync with the other example commands when we update them.

.ai/TESTING.md
T4rk1n: Good suggestion!

.ai/COMMANDS.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Run tests matching pattern
+pytest -k cbcx  # runs all tests with "cbcx" in name
+
+# Renderer unit tests (Jest)
+cd dash/dash-renderer && npm run test
+
+# Setup test components before running integration tests
+npm run setup-tests.py
+```
```
KoolADE85, line 58: similar to the "Quick Reference" I don't think we should have duplicate examples.

KoolADE85 (approved):
This is looking great!

Issue #3116, opened 2025-01-03
Add Filesystem as a backend for CeleryManager in background callbacks


celia-lm:
**How is this feature useful?**
- When the value that the background callback returns is bigger than 512MB (formatted as a string), an error will be raised: `celery.exceptions.BackendStoreError: value too large for Redis backend`
- This error occurs because the CeleryManager temporarily stores the output value of the background callback.
- A workaround for this could be using a FilesystemBackend (for example, the [DE5 Persistent Filesystem](https://dash.plotly.com/dash-enterprise/persistent-filesystem)), like this:
```
celery_app = Celery(__name__, broker=os.environ['REDIS_URL'], backend='../mount')
background_callback_manager = CeleryManager(celery_app)
```

**What's the current state?**
- In the CeleryManager for background callbacks, the values are stored as strings (ref code: [link](https://github.com/plotly/dash/blob/dev/dash/long_callback/managers/celery_manager.py#L151)). 
- When the backend is a filesystem, celery (as a dependency of the CeleryManager) expects the format to be bytes (ref: [link](https://github.com/celery/celery/blob/main/celery/backends/filesystem.py#L87)). 
- The current version of dash CeleryManager for background callbacks doesn't contemplate the case where the backend is a filesystem; it assumes Redis will be used as a backend (ref:[ link](https://github.com/celery/celery/blob/main/celery/backends/redis.py#L187)). 
- Celery does contemplate the option os using a FilesystemBackend (ref [link](https://github.com/celery/celery/blob/main/celery/backends/filesystem.py#L25), but the CeleryManager doesn't know how to handle that situation.

**What changes would be necessary?**
- In places where `cache.set(progress_key, json.dumps(progress_value, cls=PlotlyJSONEncoder))` is used directly add a conditional logic to support both RedisBackend and FilesystemBackend like:
```
from kombu.utils.encoding import str_to_bytes [[source]](https://docs.celeryq.dev/projects/kombu/en/latest/_modules/kombu/utils/encoding.html#str_to_bytes)
...

if isinstance(celery_app.backend, RedisBackend):
    cache.set(progress_key, json.dumps(progress_value, cls=PlotlyJSONEncoder))
elif isinstance(celery_app.backend, FilesystemBackend):
    celery_app.backend.set(progress_key, str_to_bytes(json.dumps(progress_value(progress_value)))
    # ref: https://github.com/celery/celery/blob/main/celery/backends/filesystem.py#L87
```
[...]

PR #3733, opened 2026-04-10
Enable Celery Background Manager for additional Backends

### LINKED ISSUE #3729: [BUG] Celery Background Callback Manager fails with Non-Key/Value-Backends
**Describe your context**

This applies to background callbacks using the Celery Background Callback Manager. This should be reproducible using the latest version of dash as well.

```
dash                                     3.3.0
dash-ag-grid                             33.3.3
dash-iconify                             0.1.2
dash-mantine-components                  2.5.1
dash-testing-stub                        0.0.2
```

**Describe the bug**

Celery supports multiple backends that are not a `BaseKeyValueStoreBackend`:

- `CassandraBackend`
- `DatabaseBackend`
- `MongoBackend`
- `RPCBackend`

However, Dash's `CeleryManager` [assumes](https://github.com/plotly/dash/blob/c897b2b094543e930a862fe51d48abfb78057df7/dash/long_callback/managers/celery_manager.py#L99) the backend to be a `BaseKeyValueStoreBackend` having the methods `set` and `get`.

Configuring any of the backends above will result in an error similar to `AttributeError: 'DatabaseBackend' object has no attribute 'get'` when triggering a background callback.

Backends that that implement `BaseKeyValueStoreBackend` might also fail, when the parameters passed to `set` and `get` do not comply with the backend's requirements (e.g. the `FilesystemBackend` expects a byte-like object for the key resulting in an error similar to `TypeError: sequence item 1: expected a bytes-like object, str found`, also see https://github.com/plotly/dash/issues/3116).

**Expected behavior**

The `CeleryManager` should either

- support all Celery backends by using `get_result` and `store_result` from `BaseBackend` instead of `get` and `set` and using byte-like objects for keys or
- make the restriction more explicit by additional documentation and erroring [early on](https://github.com/plotly/dash/blob/c897b2b094543e930a862fe51d48abfb78057df7/dash/long_callback/managers/celery_manager.py#L52) when the configured backend does not implement `BaseKeyValueStoreBackend` or is otherwise not supported.

Especially the `DatabaseBackend` and `FilesystemBackend` would be accessible options without the need for additional infrastructure to execute background callbacks.

datenzauberai:
This is an implementation to fix https://github.com/plotly/dash/issues/3729 and https://github.com/plotly/dash/issues/3116.

## Contributor Checklist

   -  [x] Error early on when the configured backend is not a `BaseKeyValueStoreBackend`
   -  [x] Use proper byte-like objects as keys for `BaseKeyValueStoreBackend`
   -  [x] Harden implementation to support celery filesystem backend
   -  [ ] Enhance implementation to support arbitrary backends via `get_result` and `store_result`

### optionals


camdecoster:
Thanks for creating this PR! Is it in a state where we can review it?

datenzauberai:
> Thanks for creating this PR! Is it in a state where we can review it?

I'm on it, but this still needs work! The manager itself works locally, but I want to have good tests first so that this can be maintained in the future.

datenzauberai:
@T4rk1n Maybe you could have a quick look. This should work now with the filesystem backend in Celery.

Enabling the other backends would need a more substantial rework, because we would need to use get_result/store_result instead of get/set/delete/expire in BaseKeyValueStoreBackend.

I wonder if it's really worth the effort to rework this, because I guess that many simple cases can now be handled using WebSocket Callbacks. What do you think?

PR #3738, opened 2026-04-11
Add missing stacklevel=2 to 9 warnings.warn() calls

stef41:
## Description

Without `stacklevel`, `warnings.warn()` reports the location of the `warn()` call itself rather than the caller's location. This makes it harder for users to identify where the warning originates in their own code.

This adds `stacklevel=2` to all 9 `warnings.warn()` calls across 7 files that were missing it.

### Files changed

- `dash/_callback_context.py` — deprecated `outputs_list`, `inputs_list`, `states_list` properties
- `dash/dash.py` — JupyterDash deprecation warning
- `dash/resources.py` — `serve_locally` config warning
- `dash/development/base_component.py` — component deprecation
- `dash/development/_jl_components_generation.py` — Julia reserved keyword warning
- `dash/development/_r_components_generation.py` — R reserved keyword warning
- `dash/testing/browser.py` — non-Chrome log warning

### Why this matters

When a user accesses a deprecated property like `callback_context.outputs_list`, the warning currently points to `dash/_callback_context.py:177` instead of the user's own code. With `stacklevel=2`, warnings correctly point to the caller's location, making it much easier to find and fix the issue.


T4rk1n (commented):
Looks good, just need to run `npm run format` to fix the linting and add a changelog.

stef41:
Thanks @T4rk1n! Done — I ran `black` on the 7 changed files and added a changelog entry under `## Fixed` in the `[UNRELEASED]` section.

T4rk1n (approved):
This should be a good improvement to our warnings, thank you 💃

PR #3745, opened 2026-04-15
feat: make CSRF cookie and header names configurable

### LINKED ISSUE #729: Name of CSRF cookie is hard-coded.
Hey folks, 

I've run into a problem with CSRF while using Dash with Django. The gist of it is that the route handler in our setup used for processing the `_dash-update-component` call requires CSRF (not something we can turn off). This is all fine, however, `api.js` in dash-renderer
hardcodes the name of the cookie to read the token from
```
'X-CSRFToken': cookie.parse(document.cookie)._csrf_token
```
This means that we have to either change our cookie names across the project, or simply not use dash. I'd rather use it cause it's a great product, but I can't change our cookie name across the board. 

Would it be possible to make this value configurable? 

Thanks!

yosofbadr:
## Summary

- Add `csrf_token_name` and `csrf_header_name` config options to the `Dash` constructor
- Allows users to match the CSRF cookie/header names used by their server framework (e.g. Django uses `csrftoken` instead of `_csrf_token`)
- Values flow through Dash's standard config pipeline: Python `app.config` -> HTML `_dash-config` JSON -> frontend Redux store
- `getCSRFHeader()` now reads cookie/header names from config with backward-compatible defaults
- When the configured CSRF cookie is absent, no header is sent (previously sent `undefined` value)

### Usage

```python
# Django users
app = Dash(__name__, csrf_token_name="csrftoken")

# Custom CSRF setup
app = Dash(__name__, csrf_token_name="my_csrf", csrf_header_name="X-My-CSRF")

# Default behavior unchanged
app = Dash(__name__)  # uses '_csrf_token' / 'X-CSRFToken'
```

### Motivation

Dash's frontend hardcodes the CSRF cookie name as `_csrf_token`, which conflicts with Django's default `csrftoken`. Users embedding Dash in Django apps with CSRF middleware enabled cannot use Dash without project-wide Django config changes. See #729 for full history.

This follows the architecture [recommended by @chriddyp in #729](https://github.com/plotly/dash/issues/729): config variables should flow through Dash's standard config pipeline, not through hooks.

### Prior art

PR #990 attempted this via the hooks system but was closed without review. This PR instead uses the config pipeline as maintainers requested.

### Files changed

| File | Change |
|------|--------|
| `dash/dash.py` | Added params, config, validation, read-only, docstrings |
| `dash-renderer/src/actions/index.js` | `getCSRFHeader(config)` with dynamic cookie/header names |
| `dash-renderer/src/actions/api.js` | Pass config through GET/POST |
| `dash-renderer/src/actions/callbacks.ts` | Pass config to `getCSRFHeader` |
| `dash-renderer/src/config.ts` | Added optional fields to `DashConfig` type |
| `tests/unit/test_configs.py` | 7 unit tests (defaults, custom, HTML output, validation, read-only) |
| `CHANGELOG.md` | Added entry under UNRELEASED |

## Test plan

- [x] Unit tests for default config values
- [x] Unit tests for custom config values
- [x] Unit tests verifying config appears in generated HTML
- [x] Unit tests for input validation (empty/whitespace rejected)
- [x] Unit tests for read-only enforcement
- [x] Validated with Django 6.0 + `CsrfViewMiddleware` active
- [x] Renderer builds successfully
- [x] ESLint + Prettier pass on all JS/TS changes
- [x] Black + flake8 pass on all Python changes
- [x] CI integration tests

Fixes #729

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Added
 - [#3680](https://github.com/plotly/dash/pull/3680) Added `search_order` prop to `Dropdown` to allow users to preserve original option order during search 
+- Added `csrf_token_name` and `csrf_header_name` config options to allow configuring the CSRF cookie and header names. Fixes [#729](https://github.com/plotly/dash/issues/729)
 
 ## Added
 - [#3523](https://github.com/plotly/dash/pull/3523) Fall back to background callback function names if source cannot be found
--- a/dash/dash-renderer/src/actions/api.js
+++ b/dash/dash-renderer/src/actions/api.js
@@ -7,22 +7,22 @@ import {JWT_EXPIRED_MESSAGE, STATUS} from '../constants/constants';
 /* eslint-disable-next-line no-console */
 const logWarningOnce = once(console.warn);
 
-function GET(path, fetchConfig) {
+function GET(path, fetchConfig, _body, config) {
     return fetch(
         path,
         mergeDeepRight(fetchConfig, {
             method: 'GET',
-            headers: getCSRFHeader()
+            headers: getCSRFHeader(config)
         })
     );
 }
 
-function POST(path, fetchConfig, body = {}) {
+function POST(path, fetchConfig, body = {}, config) {
     return fetch(
         path,
         mergeDeepRight(fetchConfig, {
             method: 'POST',
-            headers: getCSRFHeader(),
+            headers: getCSRFHeader(config),
             body: body ? JSON.stringify(body) : null
         })
     );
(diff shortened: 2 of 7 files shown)


yosofbadr:
Regarding the 2 remaining Percy visual diffs:

1. **Error dialog** — same content, slightly different scroll position (no functional change)
2. **Dropdown search** — appears to show a different filter state at the time of Percy snapshot capture. The dcc.Dropdown filtering is client-side React code, which this PR does not modify. Our changes are limited to how CSRF headers are attached to HTTP requests.

All integration and unit tests pass. These appear to be Percy snapshot timing differences, not functional regressions.

T4rk1n (approved):
Looks good, thank you. 💃

PR #3747, opened 2026-04-16
MCP Server Part 5: Pattern-matching callback schemas

KoolADE85:
## Summary

- Add pattern-matching callback support (ALL, MATCH, ALLSMALLER) to input schemas and descriptions
- `PatternMatchingSchema` generates array schemas for ALL/ALLSMALLER and object schemas for MATCH, with value types inferred from matching layout components
- `PatternMatchingDescription` explains the wildcard pattern and expected input format to the LLM

Manual verification:

```python
from dash import Dash, html, Input, Output, ALL
import json

app = Dash(__name__)
app.layout = html.Div([
    html.Div(id={"type": "item", "index": 0}, children="A"),
    html.Div(id={"type": "item", "index": 1}, children="B"),
    html.Div(id="result"),
])

@app.callback(
    Output("result", "children"),
    Input({"type": "item", "index": ALL}, "children"),
)
def combine(values):
    """Combine all item values."""
    return ", ".join(values)

    for adapter in collection:
        tool = adapter.as_mcp_tool
        print(f"Tool: {tool.name}")
        print(f"Description:\n{tool.description}\n")
        print(f"Input schema:\n{json.dumps(tool.inputSchema, indent=2)}\n")
        # Schema should show type: "array" with items containing {id, property, value}
        # Description should mention "Pattern-matching input (ALL)" and type="item"


T4rk1n (approved):
💃

PR #3748, opened 2026-04-16
MCP Server Part 6: Format callback results for LLM consumption

KoolADE85:
## Summary

- Add formatters that enrich callback responses with LLM-friendly content (images, markdown tables) for certain data types
- `PlotlyFigureResult` — renders `Graph.figure` outputs as PNG images via kaleido, returned as `ImageContent`
- `DataFrameResult` — renders `DataTable.data` and `AgGrid.rowData` outputs as markdown tables, returned as `TextContent`

## Manual verification:

```python
import plotly.express as px
from dash import Dash, html, dcc, Input, Output, dash_table

df = px.data.gapminder().query("year == 2007 and continent == 'Europe'")

app = Dash(__name__)
app.layout = html.Div([
    dcc.Dropdown(id="metric", options=["gdpPercap", "lifeExp", "pop"], value="gdpPercap"),
    dcc.Graph(id="chart"),
    dash_table.DataTable(id="table"),
])

@app.callback(Output("chart", "figure"), Input("metric", "value"))
def update_chart(metric):
    """Bar chart of the selected metric across European countries."""
    return px.bar(df, x="country", y=metric, title=f"{metric} in Europe (2007)")

@app.callback(Output("table", "data"), Input("metric", "value"))
def update_table(metric):
    """Top 5 countries for the selected metric."""
    return df.nlargest(5, metric)[["country", metric]].to_dict("records")

    from dash.mcp.primitives.tools.callback_utils import run_callback
    from dash.mcp.primitives.tools.results import format_callback_response

    for adapter in collection:
        response = run_callback(adapter, {"metric": "lifeExp"})
        result = format_callback_response(response, adapter)
        for item in result.content:
            if item.type == "text":
                print(f"[text] {item.text[:200]}...")
            elif item.type == "image":
                print(f"[image] {item.mimeType}, {len(item.data)} chars base64")
        # update_chart: [text] JSON + [image] PNG
        # update_table: [text] JSON + [text] markdown table

### CODE REVIEW

File: dash/mcp/primitives/tools/results/__init__.py [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+    response: CallbackExecutionResponse,
+    callback: CallbackAdapter,
+) -> CallToolResult:
+    """Format a callback response as a CallToolResult.
+
+    The response is always returned as structuredContent. Result
+    formatters are called per output property and may add additional
+    content items (images, markdown, etc.).
+    """
+    content: list[Any] = [
```
camdecoster, line 37: Would using a dict instead of a list make more sense here? Then you wouldn't have to iterate through the list down below (assuming you could figure out which formatter to use before calling it).

dash/mcp/primitives/tools/results/result_dataframe.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @classmethod
+    def format(
+        cls, output: MCPOutput, returned_output_value: Any
+    ) -> list[TextContent | ImageContent]:
+        if not TABULAR.matches(output.get("component_type"), output["property"]):
+            return []
+        if (
+            not isinstance(returned_output_value, list)
+            or not returned_output_value
```
camdecoster:
```suggestion
            not returned_output_value
            or not isinstance(returned_output_value, list)
```

dash/mcp/primitives/tools/results/result_dataframe.py
```diff
@@ -0,0 +1,63 @@
...
+from ..prop_roles import TABULAR
+from .base import ResultFormatter
+
+MAX_ROWS = 50
+
+
+def _to_markdown_table(rows: list[dict], max_rows: int = MAX_ROWS) -> str:
+    """Render a list of row dicts as a markdown table."""
+    columns = list(rows[0].keys())
+    total = len(rows)
```
camdecoster:
```suggestion
    total_rows = len(rows)
```

dash/mcp/primitives/tools/results/result_dataframe.py
```diff
@@ -0,0 +1,63 @@
...
+
+    lines: list[str] = []
+    lines.append(f"*{total} rows \u00d7 {len(columns)} columns*")
+    lines.append("")
+    lines.append("| " + " | ".join(columns) + " |")
```
camdecoster: If wouldn't save you much, but you could avoid adding the outer |s and still comply with the markdown spec. But every token counts, right?

dash/mcp/primitives/tools/prop_roles.py
```diff
@@ -100,7 +100,7 @@ def _compute_dropdown_value_schema(param: MCPInput) -> dict[str, Any]:
     description="Returns formatted text",
 )
 
-GENERIC_FIGURE = PropRole(
+PLOTLY_FIGURE = PropRole(
```
camdecoster, line 103: Why is this change necessary?

dash/mcp/primitives/tools/results/result_dataframe.py
```diff
@@ -0,0 +1,63 @@
...
+    return "\n".join(lines)
+
+
+class DataFrameResult(ResultFormatter):
+    """Produce a markdown table for tabular component output values."""
+
+    @classmethod
+    def format(
+        cls, output: MCPOutput, returned_output_value: Any
+    ) -> list[TextContent | ImageContent]:
```
camdecoster: Since only `TextContent` is added below, could you remove `ImageContent` here?

dash/mcp/primitives/tools/results/result_plotly_figure.py
```diff
@@ -0,0 +1,56 @@
...
+class PlotlyFigureResult(ResultFormatter):
+    """Produce a rendered PNG for Graph.figure output values."""
+
+    @classmethod
+    def format(
+        cls, output: MCPOutput, returned_output_value: Any
+    ) -> list[TextContent | ImageContent]:
+        if not PLOTLY_FIGURE.matches(output.get("component_type"), output["property"]):
+            return []
+        if not isinstance(returned_output_value, dict):
```
camdecoster:
```suggestion
        if (
            not returned_output_value
            or not isinstance(returned_output_value, dict)
        ):
```

dash/mcp/primitives/tools/results/result_plotly_figure.py
```diff
@@ -0,0 +1,56 @@
...
+from .base import ResultFormatter
+
+logger = logging.getLogger(__name__)
+
+IMAGE_WIDTH = 700
+IMAGE_HEIGHT = 450
+
+
+def _render_image(figure: Any) -> ImageContent | None:
+    """Render the figure as a base64 PNG ImageContent.
```
camdecoster: Typical for this PR:
```suggestion
    """
    Render the figure as a base64 PNG ImageContent.
```

camdecoster (approved):
Looks good. I add a few minor suggestions.

dash/mcp/primitives/tools/prop_roles.py
KoolADE85: just a style choice: I wanted it to read more clearly when it's subsequently used in the result formatter, that we are actually concerned with formatting Plotly figures, not more "generic" figures.

dash/mcp/primitives/tools/results/__init__.py
KoolADE85: Thinking about this further, I'd like to keep the current structure for two reasons:
1. consistency: the pattern of "iterate providers, each decides if it applies" is already used for input schemas, tool descriptions, and resource providers; I think it's valuable to keep a consistent pattern here too.
2. flexibility: iterating lets each formatter decide dynamically whether it applies. While it's fairly static today, we are leaving room for multiple formatters to enrich the same output without restructuring later.

dash/mcp/primitives/tools/results/result_dataframe.py
KoolADE85: good idea! Done.

PR #3749, opened 2026-04-16
MCP Server Part 7: `get_dash_component` tool and callback tool execution

KoolADE85:
## Summary

- Wires up the full tool pipeline following the same pattern as `resources`:
    * `list_tools()` builds the tool list from all registered tool providers
    * `call_tool(name, args)` routes LLM calls to the appropriate callback function

- Adds a static tool `get_dash_component` which allows LLMs to query a specific prop for a given ID.

## Manual verification:

```python
import json

app = Dash(__name__)
app.layout = html.Div([
    dcc.Dropdown(id="city", options=["NYC", "LA"], value="NYC"),
    dcc.Graph(id="chart"),
    html.Div(id="summary"),
])

@app.callback(Output("chart", "figure"), Input("city", "value"))
def update_chart(city):
    return {"data": [], "layout": {"title": city}}

@app.callback(Output("summary", "children"), Input("city", "value"))
def update_summary(city):
    return f"Showing data for {city}"

    from dash.mcp.primitives.tools import list_tools, call_tool

    # List all tools — should include update_chart, update_summary, get_dash_component
    tools = list_tools()
    for tool in tools.tools:
        print(f"- {tool.name}")

    # Inspect a component — shows it's input to both callbacks
    result = call_tool("get_dash_component", {"component_id": "city"})
    print(json.dumps(json.loads(result.content[0].text), indent=2))
    # "value" property should list both `update_chart` and `update_summary` in input_to_tool

    # Execute a callback via its tool
    result = call_tool("update_summary", {"city": "LA"})
    print(result.content[0].text)  # Shows "Showing data for LA"

### CODE REVIEW

File: dash/mcp/primitives/tools/tool_get_dash_component.py [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+            )
+
+        callback_map = get_app().mcp_callback_map
+
+        properties: dict[str, ComponentPropertyInfo] = {}
+        for prop_name in getattr(component, "_prop_names", []):
+            if prop_filter and prop_name != prop_filter:
+                continue
+
+            value = callback_map.get_initial_value(f"{comp_id}.{prop_name}")
```
camdecoster:
```suggestion
            id_and_prop = f"{comp_id}.{prop_name}"
            value = callback_map.get_initial_value(id_and_prop)
```

dash/mcp/primitives/tools/tool_get_dash_component.py
```diff
@@ -0,0 +1,131 @@
...
+            if value is None:
+                value = getattr(component, prop_name, None)
+            if value is None:
+                continue
+
+            modified_by: list[str] = []
+            input_to: list[str] = []
+            id_and_prop = f"{comp_id}.{prop_name}"
```
camdecoster:
```suggestion
```

dash/mcp/primitives/tools/tool_get_dash_component.py
```diff
@@ -0,0 +1,131 @@
...
+                inputSchema=_INPUT_SCHEMA,
+                outputSchema=_OUTPUT_SCHEMA,
+            )
+        ]
+
+    @classmethod
+    def call_tool(cls, tool_name: str, arguments: dict[str, Any]) -> CallToolResult:
+        comp_id = arguments.get("component_id", "")
+        if not comp_id:
+            raise ValueError("component_id is required")
```
camdecoster:
```suggestion
            return CallToolResult(
                content=[TextContent(type="text", text="component_id is required")],
                isError=True,
            )
```

dash/mcp/primitives/tools/tool_get_dash_component.py
```diff
@@ -0,0 +1,131 @@
...
+            msg = f"Component '{comp_id}' not found in static layout."
+            if rendering_tools:
+                msg += (
+                    f" However, the following tools would modify it: {rendering_tools}."
+                )
+            msg += " Use the dash://components resource to see statically available component IDs."
+            return CallToolResult(
+                content=[TextContent(type="text", text=msg)],
+                isError=True,
+            )
```
camdecoster, line 88: Should this return `ToolNotFoundError`?

dash/mcp/primitives/tools/tool_get_dash_component.py
```diff
@@ -0,0 +1,131 @@
...
+    def call_tool(cls, tool_name: str, arguments: dict[str, Any]) -> CallToolResult:
+        comp_id = arguments.get("component_id", "")
+        if not comp_id:
+            raise ValueError("component_id is required")
+
+        prop_filter = arguments.get("property", "")
+        component = find_component(comp_id)
+
+        if component is None:
+            callback_map = get_app().mcp_callback_map
```
camdecoster:
```suggestion
        callback_map = get_app().mcp_callback_map
        if component is None:
```

dash/mcp/primitives/tools/tool_get_dash_component.py
```diff
@@ -0,0 +1,131 @@
...
+                msg += (
+                    f" However, the following tools would modify it: {rendering_tools}."
+                )
+            msg += " Use the dash://components resource to see statically available component IDs."
+            return CallToolResult(
+                content=[TextContent(type="text", text=msg)],
+                isError=True,
+            )
+
+        callback_map = get_app().mcp_callback_map
```
camdecoster:
```suggestion
```

dash/mcp/primitives/tools/tool_get_dash_component.py
```diff
@@ -0,0 +1,131 @@
...
+            modified_by: list[str] = []
+            input_to: list[str] = []
+            id_and_prop = f"{comp_id}.{prop_name}"
+            for cb in callback_map:
+                for out in cb.outputs:
+                    if out["id_and_prop"] == id_and_prop:
+                        modified_by.append(cb.tool_name)
+                for inp in cb.inputs:
+                    if inp["id_and_prop"] == id_and_prop:
+                        input_to.append(cb.tool_name)
```
camdecoster: This could get expensive for apps with a lot of callbacks, inputs, outputs. Is there a way to create and use an index for looking this up?

camdecoster (commented):
I made a few suggestions. The one to pay attention to is the loop going through callbacks/inputs/outputs. Let me know what you think.

dash/mcp/primitives/tools/tool_get_dash_component.py
KoolADE85: Great point! The dash renderer actually implements such a mapping (but in JS obviously). I'll do the same on the python side.

dash/mcp/primitives/tools/tool_get_dash_component.py
KoolADE85: No, that would be semantically wrong since `ToolNotFoundError` is for when LLMs call a non-existent or missing tool.
Whereas this is about an error produced by a tool that does exist.

camdecoster (approved):

PR #3750, opened 2026-04-16
MCP Server Part 8: MCP server routes

KoolADE85:
## Summary

This PR pulls together everything from the previous PRs into a working MCP server you can connect to from any MCP-compatible LLM agent. Example:
```
claude mcp add dash-app --transport http http://localhost:8050/_mcp
```

- Routes are added per MCP spec
- Dash integration: control the feature with `enable_mcp` and `mcp_path` constructor params, (or `DASH_MCP_ENABLED` / `DASH_MCP_PATH` env vars)
- E2E test suite: all the previous PRs have E2E coverage via the actual MCP server endpoints.
- A number of one-liner type fixes since the mypy tests are now running in this PR.


chgiesse:
Hi Plotly team! 

First of all, thanks for integrating MCP into Dash - this looks really exciting! 
But I would like to address/discuss some concerns: 
- this PR looks like its not compatible with the new `bring your own Backend` architecture by directly importing components like Response from Flask and not the response adapter - thus the mcp doesn't work with FastApi or Quart
- the same seems to apply for the SSE streaming. Which I think should be integrated into the server adapter classes and be provided for the other backends too
- and regarding the SSE implementation - I think the this version has some serious security implications - for synchronous Flask. With the current endless loop, session wise handling and Gunicorns default synchronous workers, the life time per session per sse stream blocks a complete worker! For example, google Chrome doesn't really provide a tab limit, so I could open a 1000 tabs and block a 1000 workers in your application. Although I think there is not a direct solutions to this (besides not providing endless stream capabilities to a synchronous backend) - a way to handle this would be to provide a `DashQueueAdapter` where a user can use a [Gevent Queue](https://www.gevent.org/api/gevent.queue.html) instead of the default one. With this, the problem can be smoothed out in production. I also created a blog post ones - comparing server performance with SSE streams between Flask and Quart. https://community.plotly.com/t/flash-1-2-0-streaming-ui-updates-via-server-sent-events/93890/2 

Let me know what you think - if you see it the same way or if I overlooked something! Would also love to contribute these changes :)

Best regards,
Christian

KoolADE85:
Hey @chgiesse 

Thanks for the comments, and happy to see the excitement around the feature!

* I'll be adjusting this to be compatible with the "bring your own backend" initiative. As you can see, the two projects have been happening independently so I'll make the appropriate code adjustments here.

* As for SSE streaming, I'm thinking to remove that for the initial release. Early on, I thought it would be useful to have some server-driven messaging, and this would represent the start of that. But since the core functionality is all achieved with basic, well-described tools/resources, it's probably wise to remove SSE until we think through the security and performance considerations.

And absolutely, I'd welcome any contributions you have towards this effort - there's tons of potential to make Dash apps even more LLM friendly!

Adrian

chgiesse:
Hi @KoolADE85 

Thanks for the response and making it compatible! Looking forward to use it! 

Yes I also think that the functionality without sse is already a good starting point. Maybe providing it together with a native sse callback? :)

dash/mcp/primitives/resources/resource_components.py
```diff
@@ -41,7 +42,7 @@ def read_resource(cls, uri: str = "") -> ReadResourceResult:
         components = sorted(
             [
                 {
-                    "id": str(comp.id),
+                    "id": str(getattr(comp, "id", None)),
```
camdecoster: Doesn't line 49 guarantee that `id` exists?

dash/mcp/primitives/tools/input_schemas/schema_pattern_matching.py
```diff
@@ -66,7 +66,10 @@ def _get_wildcard_type(dep_id: dict) -> str | None:
 
 def _infer_value_schema(param: MCPInput) -> dict[str, Any] | None:
     """Infer the JSON Schema for the ``value`` field from a matching component."""
-    matches = find_matching_components(parse_wildcard_id(param["component_id"]))
+    pattern = parse_wildcard_id(param["component_id"])
+    if pattern is None:
+        return None
+    matches = find_matching_components(pattern)
```
camdecoster:
```suggestion
    matches = find_matching_components(pattern) if pattern else None
```

dash/mcp/primitives/tools/callback_adapter.py
```diff
@@ -361,9 +362,7 @@ def _param_annotations(self) -> list[Any | None]:
         return [hints.get(func_name) for func_name, _ in self._dep_param_map]
 
 
-def _expand_dep(
-    dep: CallbackDependency, value: Any
-) -> CallbackInput | list[CallbackInput]:
+def _expand_dep(dep: CallbackDependency, value: Any) -> CallbackInputs:
     """Attach a concrete value to a callback dependency to produce a valid callback input.
```
camdecoster: Claude 😠!
```suggestion
    """
    Attach a concrete value to a callback dependency to produce a valid callback input.
```

dash/mcp/primitives/tools/callback_adapter.py
```diff
@@ -408,15 +407,19 @@ def _expand_output_spec(
         if pattern is not None:
             concrete_ids = _derive_output_ids(pattern, resolved_inputs)
             if not concrete_ids:
-                concrete_ids = [comp.id for comp in find_matching_components(pattern)]
-            expanded = [{"id": cid, "property": prop} for cid in concrete_ids]
+                concrete_ids = [
+                    getattr(comp, "id") for comp in find_matching_components(pattern)
```
camdecoster, line 412: Could `comp` be `None` here?

dash/mcp/_decorator.py
```diff
@@ -0,0 +1,51 @@
...
+
+        @functools.wraps(fn)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            return fn(*args, **kwargs)
+
+        return wrapper
+
+    if func is not None:
+        return _wrap(func)
+    return _wrap
```
camdecoster, line 51: Do you need to return the wrapping function here?

dash/mcp/_server.py
```diff
@@ -0,0 +1,268 @@
...
+logger = logging.getLogger(__name__)
+
+
+def enable_mcp_server(app: Dash, mcp_path: str) -> None:
+    """Add MCP routes to a Dash app."""
+
+    app.mcp_decorated_functions = dict(MCP_DECORATED_FUNCTIONS)
+    MCP_DECORATED_FUNCTIONS.clear()
+
+    _session_id: str | None = None
```
camdecoster: Could you look into what would happen to this value if multiple users attempt to connect at the same time? Would some users end up with the wrong session id? It's a classic TOCTOU race condition 🙄.

Could multiple gunicorn works get there own session id? You mentioned this during our review.

KoolADE85: This is now addressed in dadc63d

dash/mcp/_decorator.py
KoolADE85: This is because the decorator supports optional arguments:
func is `None` when used with arguments (`@mcp_enabled(name="foo")`)
It's `not None` when used bare (`@mcp_enabled`)

dash/mcp/primitives/tools/callback_adapter.py
KoolADE85: It cannot be None since `find_matching_components` only returns pattern-matching components (i.e. `id` is a dict), so `getattr(comp, "id")` is always a dict.

dash/mcp/primitives/tools/input_schemas/schema_pattern_matching.py
KoolADE85: This one was actually necessary for the linter too.

camdecoster (approved):

PR #3755, opened 2026-04-21
Chore: `percy/dash-table-test` blocks merges when table tests are skipped

KoolADE85:
The percy table tests are required to merge a PR, but the job never runs if the table tests are skipped. This is a small bugfix to unblock PRs that don't trigger the table tests.


T4rk1n (approved):
💃

PR #3756, opened 2026-04-21
Fix #2462: allow MATCH Input when Output has no MATCH

### LINKED ISSUE #2462: [Feature Request] Remove Callback Wildcard Restrictions (MATCH)
In trying to make my applications as modular as possible, I tend to use pattern-matching callbacks just about everywhere. There are many instances, however, where I need to be able to detect which component was triggered (via MATCH) and update a universal component.

I would love for the restriction on the MATCH wildcard that requires the same number of MATCH values in the Input ID's and Output ID's to be lifted, allowing for more wildcards in the input than the output. It's my understanding that today's release of `allow_duplicate` in Dash's Output component should open the door to adding in the aforementioned solution.

The current solution for this use case is to use the ALL wildcard along with the callback context to filter down to the content that I'm ultimately interested in, but that solution does not scale especially well as it requires a lot of information to pass through the network, slowing things down.

bpostlethwaite:
Fixes #2462.

Two years of comments on that issue all land in the same place: people want to use `MATCH` in an `Input`/`State` without repeating the same `MATCH` keys on the `Output`. This PR does exactly that.

## Behavior change

You can now pair a `MATCH` input with a fixed-id Output, no Output, or a wildcard Output that only uses `ALL`:

```python
@callback(
    Output("notifications", "children"),
    Input({"type": "grid", "id": MATCH}, "cellValueChanged"),
)
def on_edit(event):
    return f"edited: {ctx.triggered_id}"
```

Rules kept intact:
- All Outputs of a callback must still share the same `MATCH` keys.
- `ALLSMALLER` still requires a corresponding `MATCH` in an Output — it has no reference otherwise.

## What's in the diff

1. **Validation** (`dependencies.js::findMismatchedWildcards`) — skip the Input/State `MATCH`-subset check when the Output has no `MATCH` keys.
2. **Resolution** (`dependencies.js::addAllResolvedFromOutputs` + `dependencies_ts.ts::getCallbacksByInput`) — when the Output lacks `MATCH`, the triggering Input's `MATCH` values become the callback's `anyVals` so simultaneous MATCH triggers each get a distinct `resolvedId` and don't collapse into one firing during deduplication.

## Tests run locally

- `tests/unit/` — 377 passed, 4 skipped (2 browser-smoke failures are chromedriver/env, unrelated)
- `tests/compliance/` — 13 passed
- `dash-renderer` Karma — 25 passed (includes 9 new cases for validation + `resolvedId` uniquification)
- New Selenium: `test_cbwc009_match_input_fixed_output`, `test_cbwc010_match_input_no_output`, `test_dvcv017_match_input_permitted_no_output_match` — passed
- Regression check: `test_cbwc002_fibonacci_app`, `test_cbwc003_same_keys`, `test_cbwc008_running_match`, `test_dvcv006_inconsistent_wildcards` — all still pass

## Test plan

- [ ] CI green
- [ ] Spot-check a user app that hit the original error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3690](https://github.com/plotly/dash/pull/3690) Fixes Input when min or max is set to None
 - [#3723](https://github.com/plotly/dash/pull/3723) Fix misaligned `dcc.Slider` marks when some labels are empty strings
 - [#3740](https://github.com/plotly/dash/pull/3740) Fix cannot tab into dropdowns in Safari
+- [#2462](https://github.com/plotly/dash/issues/2462) Allow `MATCH` in `Input`/`State` when the callback's `Output` has no wildcards (fixed-id Output, no Output, or `ALL`-only wildcard Output). `ALLSMALLER` still requires a corresponding `MATCH` in an Output.
 
 ## [4.1.0] - 2026-03-23
 
--- a/dash/dash-renderer/src/actions/dependencies.js
+++ b/dash/dash-renderer/src/actions/dependencies.js
@@ -434,24 +434,37 @@ function findMismatchedWildcards(outputs, inputs, state, head, dispatchError) {
             ]);
         }
     });
+    // When the Outputs don't carry any MATCH keys (fixed-id outputs, no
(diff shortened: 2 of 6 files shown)


T4rk1n (approved):
Looks good 💃

PR #3765, opened 2026-05-07
Feature: add partial pattern matching (partial=True) for callbacks

### LINKED ISSUE #3764: [Feature Request] Partial Pattern Matching
## Summary

I use Dash pattern-matching callbacks extensively in my multi-page apps and make heavy use of the `ALL` wildcard. However, the callback architecture of my apps could be simplified further by introducing an option to allow **partial matches**, where a pattern with fewer keys matches components that have a superset of those keys.

This proposes an opt-in `partial=True` flag on `Input`, `Output`, and `State`.

## Motivation

Currently, pattern-matching callbacks require the pattern and component to share **exactly the same set of keys**:

```python
# Pattern: {"type": "btn", "index": ALL}
# Only matches: {"type": "btn", "index": 1}, {"type": "btn", "index": 2}
# Does NOT match: {"type": "btn", "index": 1, "page": "home"}
```

This forces developers to either:
1. Ensure all related components have identical key sets (rigid, often impractical)
2. Create separate callbacks for each key combination (verbose, duplicated logic)
3. Use `callback_context` workarounds to multiplex unrelated callbacks

I find with my apps I frequently have components that share a conceptual "type" but differ in their additional metadata keys — pages, sections, tabs, permissions, etc.

## Proposed API

```python
from dash import Input, Output, State, ALL, MATCH

# Literal partial: matches all components with type="btn", regardless of extra keys
@app.callback(
    Output("output", "children"),
    Input({"type": "btn"}, "n_clicks", partial=True),
)
def handle_all_btns(n_clicks_list):
    # Receives a list (implicitly multi-valued since multiple components match)
    return f"Total: {sum(c or 0 for c in n_clicks_list)}"

# With ALL wildcard: collects all components that have a "type" key
@app.callback(
    Output("output", "children"),
    Input({"type": ALL}, "n_clicks", partial=True),
)
def handle_any_type(n_clicks_list):
    return f"Collected {len(n_clicks_list)} components"

# With MATCH: one-at-a-time on the shared key, ignoring extra keys
@app.callback(
    Output({"type": "display", "index": MATCH}, "children", partial=True),
    Input({"type": "btn", "index": MATCH}, "n_clicks", partial=True),
)
def handle_one(n_clicks):
```

i-murray:
# Feature: Partial Pattern Matching (`partial=True`) for Callbacks

## Description

Dash's pattern-matching callbacks currently require the pattern and component IDs to have **exactly the same set of keys**. This means a pattern like `{"type": "btn"}` will not match a component with ID `{"type": "btn", "index": 1}` — even though the component clearly has the key and value the pattern cares about.

This PR adds an opt-in `partial=True` flag on `Input`, `Output`, and `State` that relaxes the key-matching constraint: a pattern's keys only need to be a **subset** of the component's keys. Extra keys on the component are ignored.

Closes https://github.com/plotly/dash/issues/3764

### Usage

```python
@app.callback(
    Output("output", "children"),
    Input({"type": "btn"}, "n_clicks", partial=True),
    prevent_initial_call=True,
)
def on_btn_click(n_clicks):
    return f"clicked {n_clicks}"
```

This fires for any component whose ID contains `"type": "btn"`, regardless of other keys:
- `{"type": "btn", "index": 1}`
- `{"type": "btn", "page": "home", "section": "main"}`
- `{"type": "btn", "index": 2, "tab": "first"}`

Works with wildcards:
```python
Input({"type": ALL}, "n_clicks", partial=True)    # collects all components with a "type" key
Input({"type": MATCH}, "n_clicks", partial=True)   # matches one-at-a-time on "type"
```

### Design choices

- **Per-dependency opt-in** — not a new wildcard type, not a per-callback flag. Each `Input`/`Output`/`State` independently opts in, so partial and non-partial deps can coexist in the same callback.
- **Extra keys are ignored** — component keys not present in the pattern are treated as if they matched. This is the most permissive and intuitive behavior.
- **Literal-only partial patterns are implicitly multi-valued** — a pattern like `{"type": "btn"}` with `partial=True` can match multiple components, so the callback receives a list (same as `ALL`). Patterns with `MATCH` still resolve one-at-a-time.
- **Zero-cost when unused** — a `hasPartialPatterns` boolean guard on the callback graph ensures apps without `partial=True` pay no additional overhead. All new loops are skipped entirely.

---

## Contributor Checklist

   - [x] Add `partial` attribute to `DashDependency`, `Input`, `Output`, `State` (Python)
   - [x] Update `_id_matches()` for subset key matching when `partial=True`
   - [x] Serialize `partial` flag via `to_dict()` and propagate through `insert_callback()`
   - [x] Add `addPartialPattern()` and `partialIdMatch()` to frontend (`dependencies.js`)
   - [x] Update `computeGraphs()` to detect partial flags and set `hasPartialPatterns`
   - [x] Update `getWatchedKeys()`, `getCallbackByOutput()`, `getUnfilteredLayoutCallbacks()` with partial pattern loops
   - [x] Add `resolvePartialDeps()` and update `getCallbacksByInput()` in `dependencies_ts.ts`
   - [x] Update `isMultiValued()` for implicit multi-valued partial patterns
   - [x] Fix `requestedCallbacks.ts` type compatibility for updated `isMultiValued` signature
   - [x] Add `hasPartialPatterns` guard to all new loops for zero-cost when unused
   - [x] 16 unit tests in `tests/unit/test_partial_matching.py` — subset matching, wildcards, serialization, bidirectional checks, negative cases
   - [x] 6 integration tests in `tests/integration/callbacks/test_partial_wildcards.py` — basic partial match, ALL collection, literal filtering, mixed key sets

### Optionals

- [ ] If this PR needs a follow-up in **dash docs**, **community thread**, I have mentioned the relevant URLs as follows

---

## Performance

Partial matching requires cross-keyStr searches in several hot-path functions that were previously $O(1)$ hash lookups:

| Function | Original | With partial patterns |
|----------|:--------:|:--------------------:|
| `getWatchedKeys()` | $O(1)$ | $O(K \cdot p)$ |
| `getCallbacksByInput()` | $O(1)$ | $O(K \cdot p)$ |
| `getCallbackByOutput()` | $O(1)$ | $O(K \cdot p)$ |
| `getUnfilteredLayoutCallbacks()` | $O(N)$ | $O(N \cdot K \cdot p)$ |

Where $K$ = number of distinct keyStr entries in the pattern index, $p$ = number of keys in the pattern (typically 1–5), and $N$ = number of components in a layout chunk.

### `hasPartialPatterns` guard

To ensure **zero cost for apps that don't use `partial=True`**, `computeGraphs()` sets a `hasPartialPatterns` boolean on the graph in $O(D)$ time (where $D$ = total dependencies). Every new loop is guarded:

```javascript
if (graphs.hasPartialPatterns) {
    for (const patKeyStr in graphs.inputPatterns) { ... }
}
```

| Scenario | Per-event cost |
|----------|:--------------:|
| No partial patterns (common case) | $O(1)$ — single boolean check, same as before |
| With partial patterns | $O(K \cdot p)$ — unavoidable cross-keyStr search |


i-murray:
If, instead of `partial`, you prefer a more descriptive keyword argument for `Input`/`State`/`Output`, I would suggest alternatives such as `partial_match`, `partial_pattern`, `allow_partial` or `exact` (with inverse logic & defaulting to `True`).

Additionally, if this feature is integrated, the relevant Dash documentation [Pattern-Matching Callbacks](https://dash.plotly.com/pattern-matching-callbacks) would need to be updated.

T4rk1n:
> If, instead of `partial`, you prefer a more descriptive keyword argument for `Input`/`State`/`Output`, I would suggest alternatives such as `partial_match`, `partial_pattern`, `allow_partial` or `exact` (with inverse logic & defaulting to `True`).
> 
> Additionally, if this feature is integrated, the relevant Dash documentation [Pattern-Matching Callbacks](https://dash.plotly.com/pattern-matching-callbacks) would need to be updated.

I like this feature, but yea `partial` is too broad in Input kwargs, it should be specific to the pattern matching. As such I think `partial_pattern` would be best.

camdecoster:
@i-murray are you still interested in finishing this PR?

i-murray:
@camdecoster this branch should be ready for a review. I've just renamed the keyword argument from `partial` to `partial_pattern` as suggested by @T4rk1n.

dash/dash-renderer/src/actions/dependencies.js
```diff
@@ -1405,11 +1658,63 @@ export function getUnfilteredLayoutCallbacks(graphs, paths, layoutChunk, opts) {
                 handleOneId(id, graphs.outputMap[id], graphs.inputMap[id]);
             } else {
                 const keyStr = Object.keys(id).sort().join(',');
+                const _keys = Object.keys(id).sort();
```
T4rk1n, line 1694: Second sort of the same keys — `keyStr` above already did `Object.keys(id).sort()`. And `_keys` is only read inside the `hasPartialPatterns` blocks, but it runs for every component regardless. Compute the sorted array once and build `keyStr` from it.

dash/dash-renderer/src/actions/dependencies.js
```diff
@@ -1237,15 +1460,45 @@ export function getWatchedKeys(id, newProps, graphs) {
     const vals = props(keys, id);
     const keyStr = keys.join(',');
     const keyPatterns = graphs.inputPatterns[keyStr];
```
T4rk1n, line 1493: Lost the early return when there are no `keyPatterns`, so we now run the `newProps.filter(...)` even when there's nothing to match. Bail early when there are no `keyPatterns` and no partial patterns; only enter the per-prop loop when there's something to check.

T4rk1n (commented):
The perf section's "zero cost when unused" doesn't hold — some of the new work runs even when no callback sets `partial_pattern`, and a single partial callback flips the global flag so the extra loops then run for all wildcard callbacks. Three spots below to tighten before merge.

dash/dash-renderer/src/actions/dependencies.js
```diff
@@ -945,16 +1006,36 @@ export function computeGraphs(dependencies, dispatchError, config) {
...
+                    addPattern(inputPatternMap, inId, inProp, finalDependency);
+                }
             } else {
                 addMap(inputMap, inId, inProp, finalDependency);
             }
         });
     });
     outputPatterns = offloadPatterns(outputPatternMap, outputPatterns);
     inputPatterns = offloadPatterns(inputPatternMap, inputPatterns);
+    hasPartialPatterns = parsedDependencies.some(
```
T4rk1n: The loops in `getWatchedKeys` / `getCallbacksByInput` / `getCallbackByOutput` / `getUnfilteredLayoutCallbacks` walk every keyStr and only find out it's not partial after the split + subset check. Build `partialInputPatterns` / `partialOutputPatterns` here in the same O(D) pass so those loops scan only the partial keyStrs.

dash/dash-renderer/src/actions/dependencies.js
i-murray: The ID keys are now sorted once and `keyStr` is built from that array, removing the duplicate `Object.keys(id).sort()`.

dash/dash-renderer/src/actions/dependencies.js
i-murray: `getWatchedKeys` now returns early when there are no matching `keyPatterns` or partial patterns. This avoids the per-property filter loop when there is nothing to check.

dash/dash-renderer/src/actions/dependencies.js
i-murray: New `partialInputPatterns` and `partialOutputPatterns` indexes are now built during graph construction. All four cross-key lookup paths use these indexes, so they scan only partial pattern key sets.

T4rk1n (approved):
💃

PR #3767, opened 2026-05-07
MCP Server Part 10: OAuth discovery

KoolADE85:
## Summary

Support OAuth so MCP clients can discover how to authenticate against an authorization server when connecting to the Dash MCP endpoint. Implements the discovery half of the [MCP Authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization).

- New `mcp_authorization_server` constructor arg (and `DASH_MCP_AUTHORIZATION_SERVER` env var). When set, registers an [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728) Protected Resource Metadata endpoint at `/.well-known/oauth-protected-resource/<mcp-path>` advertising the authorization server.
- **Dash serves discovery metadata only — it does not enforce auth.** Token validation, 401 responses, and `WWW-Authenticate` headers are the responsibility of the hosting platform (middleware, reverse proxy, etc.).

### DIFF
--- a/dash/mcp/_server.py
+++ b/dash/mcp/_server.py
@@ -74,6 +74,15 @@ def _setup_mcp_oauth(app: Dash, mcp_path: str, mcp_authorization_server: str) ->
     server. Auth enforcement is the responsibility of the hosting platform
     (e.g. Plotly Cloud gateway, Dash Embedded, or a reverse proxy).
     """
+    if app.config.requests_pathname_prefix != "/":
+        raise ValueError(
+            "`mcp_authorization_server` cannot be used in conjunction with "
+            "`requests_pathname_prefix`. "
+            "Authorization must be implemented at the platform level "
+            "(see https://www.rfc-editor.org/rfc/rfc9728#section-3). "
+            "Remove the mcp_authorization_server parameter."
+        )
+
     well_known_path = urljoin("/.well-known/oauth-protected-resource/", mcp_path)
 
     def _serve_resource_metadata():
--- a/dash/mcp/_server.py
+++ b/dash/mcp/_server.py
@@ -11,7 +11,9 @@ import inspect
 import json
 import logging
 import os
+from functools import reduce
 from typing import TYPE_CHECKING, Any
+from urllib.parse import urljoin
 
 from mcp.types import (
     LATEST_PROTOCOL_VERSION,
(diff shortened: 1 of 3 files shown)


dash/mcp/_server.py
```diff
@@ -47,7 +49,61 @@
...
+
+
+def _setup_mcp_oauth(app: Dash, mcp_path: str, mcp_authorization_server: str) -> None:
+    """Register RFC 9728 Protected Resource Metadata endpoint for MCP.
+
+    Serves discovery metadata so MCP clients can find the authorization
+    server. Auth enforcement is the responsibility of the hosting platform
+    (e.g. Plotly Cloud gateway, Dash Embedded, or a reverse proxy).
+    """
+    well_known_path = urljoin("/.well-known/oauth-protected-resource/", mcp_path)
```
camdecoster, line 86: Could you double check the URL construction? It looks like the final result might not be correct. Shouldn't the pathname prefix be included before the MCP path?

KoolADE85: As discussed offline, I am now raising when `mcp_authorization_server` is used in conjunction with `requests_pathname_prefix` since we cannot accommodate both.

dash/mcp/_server.py
```diff
@@ -47,7 +49,70 @@
...
+    server. Auth enforcement is the responsibility of the hosting platform
+    (e.g. Plotly Cloud gateway, Dash Embedded, or a reverse proxy).
+    """
+    if app.config.requests_pathname_prefix != "/":
+        raise ValueError(
+            "`mcp_authorization_server` cannot be used in conjunction with "
+            "`requests_pathname_prefix`. "
+            "Authorization must be implemented at the platform level "
+            "(see https://www.rfc-editor.org/rfc/rfc9728#section-3). "
+            "Remove the mcp_authorization_server parameter."
```
camdecoster, line 83: It might be worth it to guide the user a bit more on how to fix the issue with a URL template or something. At a minimum, I'd update the last sentence to be a little less of a command.

```suggestion
            "Remove the `mcp_authorization_server` or `requests_pathname_prefix` parameter to avoid this error."
```

dash/mcp/_server.py
```diff
@@ -47,7 +49,70 @@
...
+        )
+
+    well_known_path = urljoin("/.well-known/oauth-protected-resource/", mcp_path)
+
+    def _serve_resource_metadata():
+        return app.backend.make_response(
+            json.dumps(
+                {
+                    "resource": _url_from_path(
+                        app, app.config.requests_pathname_prefix, mcp_path
```
camdecoster, line 93: The prefix will only ever be "/" here, right?

dash/mcp/_server.py
```diff
@@ -47,7 +49,70 @@
...
+
+def _setup_mcp_oauth(app: Dash, mcp_path: str, mcp_authorization_server: str) -> None:
+    """Register RFC 9728 Protected Resource Metadata endpoint for MCP.
+
+    Serves discovery metadata so MCP clients can find the authorization
+    server. Auth enforcement is the responsibility of the hosting platform
+    (e.g. Plotly Cloud gateway, Dash Embedded, or a reverse proxy).
+    """
+    if app.config.requests_pathname_prefix != "/":
+        raise ValueError(
```
camdecoster, line 78: Raising here means that the error will get caught in the `try`/`except` block in dash.py. Would it be better to raise earlier when first checking on config options?

dash/mcp/_server.py
```diff
@@ -207,6 +272,9 @@ def _handle_not_allowed():
     )
     app.routes.append(mcp_url)
 
+    if mcp_authorization_server:
```
camdecoster, line 275: Should this check happen before registering the routes? If the auth server has been configured, then the routes get registered but MCP can't be enabled. It might be better to fail earlier.

camdecoster (approved):
I left some suggestions, but this should work as is.

KoolADE85:
Closing this as discussed offline. We will not be including any auth features for the initial MCP release.

PR #3768, opened 2026-05-07
optimize dropdown search

### LINKED ISSUE #3616: [BUG] dcc.Dropdown in dash 4.0.0 is considerably slower than before

AnnMarieW:
Closes #3616

This PR changes the `dropdownSearch.ts` so it builds the search index once during option sanitization and reuses it instead of rebuilding the index on every search input update.

I tested the example included in issue 3616 with 80,000 options and search performance is now comparable to Dash 3.4.


components/dash-core-components/src/utils/dropdownSearch.ts
```diff
@@ -55,38 +56,48 @@ export function sanitizeDropdownOptions(
...
 
-    return {options: sanitized, indexes, valueSet};
+    // Build the search index ONCE during sanitization
+    const search = new Search('value');
+    search.searchIndex = new UnorderedSearchIndex();
+    search.indexStrategy = new AllSubstringsIndexStrategy();
+    search.tokenizer = TOKENIZER;
+
+    indexes.forEach(index => {
+        search.addIndex(index);
```
camdecoster, line 66: Does it matter if the index include here is the same as the `uidFieldName` passed into the constructor? I ask because `indexes` always contains `'value'` and that's the same as the `uidFieldName` used in the constructor.

components/dash-core-components/src/utils/dropdownSearch.ts
```diff
@@ -55,38 +56,48 @@ export function sanitizeDropdownOptions(
...
+        valueSet,
+        search,
+    };
 }
 
 export function filterOptions(
     options: SanitizedOptions,
     searchValue?: string,
-    search_order?: 'index' | 'original'
+    searchOrder?: string
```
camdecoster: Why remove the union type here? In fact, you could reference the existing type in **components/dash-core-components/src/types.ts**.

camdecoster: I think you'll need to update references to this property throughout the repo.

components/dash-core-components/src/utils/dropdownSearch.ts
```diff
@@ -55,38 +56,48 @@ export function sanitizeDropdownOptions(
...
-    });
+    const results =
+        (options.search.search(searchValue) as DetailedOption[]) || [];
 
-    if (options.options.length > 0) {
-        search.addDocuments(options.options);
-    }
+    // Preserve original option order
+    if (searchOrder === 'original') {
+        const resultSet = new Set(results.map(option => option.value));
```
camdecoster, line 95: Why do you use `map` here? Does it allow for faster lookups when only using `value`?

camdecoster (changes requested):
This seems fine, but you'll need to update references from `search_order` to `searchOrder`.

components/dash-core-components/src/utils/dropdownSearch.ts
AnnMarieW: 'value' in the constructor is only used as the document uid field in js-search. It is not automatically indexed for searching, so we still need search.addIndex('value') here.  
This is unchanged from previous version (old line 75-77)

components/dash-core-components/src/utils/dropdownSearch.ts
AnnMarieW: The `map` just makes it more robust.   It converts the results into a `Set` of option values instead of object references.

components/dash-core-components/src/utils/dropdownSearch.ts
AnnMarieW: I'm not sure what you mean here.  But if the type is defined like this, does it have to change anywhere else?

searchOrder?: DropdownProps['search_order']

AnnMarieW: Good call!  It's better to use the correct prop types.

camdecoster: Ah, never mind. This is an argument not an object key. Your change shouldn't affect existing function calls. It might be worth updating those calls to use the same name casing, but it's not required. Though I think using the existing type is still a good idea.

camdecoster (approved):

PR #3771, opened 2026-05-08
[4.2] Persistent callbacks

### LINKED ISSUE #3411: [Feature Request] allow for callbacks without Inputs
I would like to request a feature enhancement to allow for callbacks to not have an input, but rely on having a State and an initial call / load to perform useful work.

Currently Dash does not allow for callbacks without inputs. I think this makes sense in most cases and the check/limit will aid most novice Dash developers.

My counter example is where we have a callback that does have a State (but no real Inputs), but depending on the value of the State, during the initial load will fetch some values from a external database and update Options and Value of a dcc.Dropdown 

At the moment we solve this be adding a fake-input in the following way:
```python
    layout = html.Div(
        [
            dcc.Dropdown(id=f"my-dropdown"),
            dcc.Interval(id="fake_input", n_intervals=0, max_intervals=0, interval=1),
        ]
    )

    @callback(
        Output("my-dropdown", "options"),
        Output("my-dropdown", "value"),
        Input("fake_input", "n_intervals"),
    )
```
But this makes the callback map much messier and larger then it needs to be, specially if you have multiple drop downs.

I have tried to comment out the 3 JS lines that does the check in `dash_renderer.dev.js` and everything seems to work just fine as intended for my particular use case.

Image

So: keep the check mostly as is, but allow for callbacks that lack Inputs, but have States and have `prevent_initial_call` as False. Optionally, If we think it's to dangerous to allow no-input-callbacks in the general sense, we could add a kwarg to the callback to bypass the check. Something like: `allow_no_inputs = True` for individual edge cases like my counter example.

### LINKED ISSUE #3770: [BUG] no output callbacks doesn't trigger as initial callback.
If you have a callback with no output and an initial input, the initial callback will not fire.

Example:
```python
from dash import Dash, html, Input, callback

app = Dash()

app.layout = html.Div([
    html.Button("Start", id="start-btn", n_clicks=0),
])


@callback(
    Input("start-btn", "n_clicks"),
)
def update(n_clicks):
    print(f"Clicked {n_clicks}")
```

T4rk1n:
This PR adds support for:                                                                                                                                                                                      
  1. Persistent callbacks - Callbacks with persistent=True don't show "Updating..." in the browser title while running                                                                                           
  2. No-output callbacks - Callbacks without outputs that fire on initial load, fixes #3770                                                                                  
  3. No-input callbacks - Callbacks without inputs (only State) that fire on initial load, resolves #3411                                                       
  4. No-input no-output callbacks - Callbacks with neither inputs nor outputs that fire once on initial load

## Examples

### Persistent callback without inputs/outputs

```python
import asyncio                                                                                                                                                                                                 
from dash import Dash, html, callback, set_props                                                                                                                                                               
                                                                                                                                                                                                               
app = Dash(backend="fastapi", websocket_callbacks=True)                                                                                                                                                        
                                                                                                                                                                                                               
app.layout = html.Div([                                                                                                                                                                                        
    html.Div(id="counter", children="0"),                                                                                                                                                                      
])                                                                                                                                                                                                             
                                                                                                                                                                                                               
@callback(persistent=True)                                                                                                                                                                                     
async def continuous_updates():                                                                                                                                                                                
    # Fires once on load, runs forever without showing "Updating..." title                                                                                                                                     
    n = 0                                                                                                                                                                                                      
    while True:                                                                                                                                                                                                
        set_props("counter", {"children": str(n)})                                                                                                                                                             
        n += 1                                                                                                                                                                                                 
        await asyncio.sleep(1)                                                                                                                                                                                 
                                                                                                                                                                                                               
if __name__ == "__main__":                                                                                                                                                                                     
    app.run() 
```

### No-input callback (fires on initial load)

```python
from dash import Dash, html, Output, State, callback                                                                                                                                                           
                                                                                                                                                                                                               
app = Dash()                                                                                                                                                                                                   
                                                                                                                                                                                                               
app.layout = html.Div([                                                                                                                                                                                        
    dcc.Store(id="config", data={"theme": "dark"}),                                                                                                                                                            
    html.Div(id="output"),                                                                                                                                                                                     
])                                                                                                                                                                                                             
                                                                                                                                                                                                               
@callback(                                                                                                                                                                                                     
    Output("output", "children"),                                                                                                                                                                              
    State("config", "data"),                                                                                                                                                                                   
)                                                                                                                                                                                                              
def initialize_from_config(config):                                                                                                                                                                            
    # Fires once on page load with State values                                                                                                                                                                
    return f"Theme: {config['theme']}"                                                                                                                                                                         
                                                                                                                                                                                                               
if __name__ == "__main__":                                                                                                                                                                                     
    app.run() 
```

### No-input no-output callback (side-effect only) 

```python
from dash import Dash, html, callback                                                                                                                                                                          
                                                                                                                                                                                                                 
app = Dash()                                                                                                                                                                                                   
app.layout = html.Div([html.Div(id="output")])                                                                                                                                                                 
                                                                                                                                                                                                               
@callback()                                                                                                                                                                                                    
def log_page_load():                                                                                                                                                                                           
    # Fires once when page loads                                                                                                                                                                               
    print("Page loaded!")                                                                                                                                                                                      
    # Could log to database, initialize connections, etc.                                                                                                                                                      
                                                                                                                                                                                                               
if __name__ == "__main__":                                                                                                                                                                                     
    app.run() 
```

### DIFF
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -170,6 +170,7 @@ def create_callback_id(output, inputs, no_output=False):
             # Get the call site of the @callback decorator
             stack = inspect.stack()
             # Walk up the stack to find the actual callback call site
+            # Fallback to empty hash if no external frame found
             # (skip internal dash package frames)
             dash_package_path = os.path.dirname(__file__)
             for frame_info in stack:
@@ -177,8 +178,7 @@ def create_callback_id(output, inputs, no_output=False):
                 if not frame_info.filename.startswith(dash_package_path):
                     call_site = f"{frame_info.filename}:{frame_info.lineno}"
                     return hashlib.sha256(call_site.encode("utf-8")).hexdigest()
-            # Fallback to empty hash if no external frame found
-            return _hash_inputs()
+
         return _hash_inputs()
 
     if isinstance(output, (list, tuple)):
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -165,6 +165,20 @@ def create_callback_id(output, inputs, no_output=False):
 
     if no_output:
         # No output will hash the inputs.
+        # For no-input callbacks, also include the call site to make each unique
+        if not inputs:
+            # Get the call site of the @callback decorator
+            stack = inspect.stack()
+            # Walk up the stack to find the actual callback call site
+            # (skip internal dash package frames)
+            dash_package_path = os.path.dirname(__file__)
+            for frame_info in stack:
+                # Skip frames from within the dash package itself
+                if not frame_info.filename.startswith(dash_package_path):
+                    call_site = f"{frame_info.filename}:{frame_info.lineno}"
+                    return hashlib.sha256(call_site.encode("utf-8")).hexdigest()
+            # Fallback to empty hash if no external frame found
+            return _hash_inputs()
         return _hash_inputs()
 
     if isinstance(output, (list, tuple)):
--- a/dash/dash-renderer/src/actions/index.js
+++ b/dash/dash-renderer/src/actions/index.js
@@ -109,7 +109,7 @@ function triggerDefaultState(dispatch, getState) {
         }
     );
 
-    // Also include no-output callbacks whose inputs are in the layout
+    // Also include no-output callbacks whose inputs are in the layout (or have no inputs)
     const noOutputCallbacks = (graphs.callbacks || [])
         .filter(cb => cb.noOutput && !cb.prevent_initial_call)
         .map(cb => {
@@ -118,6 +118,10 @@ function triggerDefaultState(dispatch, getState) {
             return resolved;
         })
         .filter(cb => {
+            // If no inputs, always include (fires once on initial load)
+            if (cb.callback.inputs.length === 0) {
+                return true;
+            }
             // Check if any input is in the layout
             const inputs = cb.getInputs(paths);
             return inputs.some(inp =>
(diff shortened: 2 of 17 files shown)


dash/dash-renderer/src/actions/index.js
```diff
@@ -95,12 +100,62 @@ function triggerDefaultState(dispatch, getState) {
...
+            // If no inputs, always include (fires once on initial load)
+            if (cb.callback.inputs.length === 0) {
+                return true;
+            }
+            // Check if any input is in the layout
+            const inputs = cb.getInputs(paths);
+            return inputs.some(inp =>
+                Array.isArray(inp) ? inp.length > 0 : inp
+            );
+        });
```
camdecoster: You iterate through the callbacks three times. What do you think of switching to a `for...of` loop that does everything in one pass? In fact, you could use one loop that handles both `noOutputCallbacks` and `noInputCallbacks`.

dash/_utils.py
```diff
@@ -165,6 +165,20 @@ def _concat(x):
...
+            # Walk up the stack to find the actual callback call site
+            # (skip internal dash package frames)
+            dash_package_path = os.path.dirname(__file__)
+            for frame_info in stack:
+                # Skip frames from within the dash package itself
+                if not frame_info.filename.startswith(dash_package_path):
+                    call_site = f"{frame_info.filename}:{frame_info.lineno}"
+                    return hashlib.sha256(call_site.encode("utf-8")).hexdigest()
+            # Fallback to empty hash if no external frame found
+            return _hash_inputs()
```
camdecoster:
```suggestion
```

dash/_utils.py
```diff
@@ -165,6 +165,20 @@ def _concat(x):
 
     if no_output:
         # No output will hash the inputs.
+        # For no-input callbacks, also include the call site to make each unique
+        if not inputs:
+            # Get the call site of the @callback decorator
+            stack = inspect.stack()
+            # Walk up the stack to find the actual callback call site
+            # (skip internal dash package frames)
```
camdecoster, line 174:
```suggestion
            # Fallback to empty hash if no external frame found
            # (skip internal dash package frames)
```

camdecoster (commented):
I left a few comments, but this is looking good.

PR #3779, opened 2026-05-11
fix(testing): return [] instead of None from get_logs for non-Chrome browsers

troclaux:
## Summary

`Browser.get_logs()` returns `None` when the driver is not Chrome. This breaks test assertions like `assert dash_duo.get_logs() == []` on Firefox/Safari.

The Chrome path already returns `[]` (an empty list comprehension) when there are no log entries, so returning `[]` on other drivers is the consistent behaviour. The warning message is updated to reflect the change.

## Changes

`dash/testing/browser.py`: changed `return None` → `return []` and updated the warning text accordingly.

## Test plan

- [x] Existing tests that assert `dash_duo.get_logs() == []` will no longer raise a `TypeError` on non-Chrome drivers


T4rk1n (approved):
💃

PR #3783, opened 2026-05-13
batch/debounce websocket set_props

T4rk1n:
When you use set_props in a loop inside a websocket callback, it would create single updates in the renderer and can result in lag.
This PR add a batching and debounce mechanism to set_props, default to 5ms/200hz configurable with
```python
Dash(..., websocket_batch_delay=0.005)  # 5ms (default) - set to 0 to disable batching
````

### DIFF
--- a/@plotly/dash-websocket-worker/src/MessageRouter.ts
+++ b/@plotly/dash-websocket-worker/src/MessageRouter.ts
@@ -64,9 +64,16 @@ export class MessageRouter {
 
     /**
      * Handle a message from the WebSocket server.
-     * @param message The message from the server
+     * Messages may be batched as arrays for efficiency.
+     * @param message The message from the server (single message or array)
      */
     public handleServerMessage(message: unknown): void {
+        // Handle batched messages (array of messages)
+        if (Array.isArray(message)) {
+            this.handleBatchedMessages(message);
+            return;
+        }
+
         const msg = message as WorkerMessage;
         const rendererId = msg.rendererId;
 
@@ -89,6 +96,53 @@ export class MessageRouter {
 
             default:
                 console.warn(`Unknown message type from server: ${msg.type}`);
+        }
+    }
+
+    /**
+     * Handle a batch of messages from the server.
+     * Groups set_props by renderer and forwards as a single batch message.
+     * @param messages Array of messages
+     */
+    private handleBatchedMessages(messages: unknown[]): void {
+        // Group set_props by renderer, keep others separate
+        const setPropsPayloadsByRenderer: Map<string, SetPropsMessage['payload'][]> = new Map();
+        const otherMessages: WorkerMessage[] = [];
+
+        for (const message of messages) {
+            const msg = message as WorkerMessage;
+            if (msg.type === WorkerMessageType.SET_PROPS) {
+                const setPropsMsg = msg as SetPropsMessage;
+                const rendererId = setPropsMsg.rendererId;
+                if (!setPropsPayloadsByRenderer.has(rendererId)) {
+                    setPropsPayloadsByRenderer.set(rendererId, []);
+                }
+                setPropsPayloadsByRenderer.get(rendererId)!.push(setPropsMsg.payload);
+            } else {
+                otherMessages.push(msg);
+            }
+        }
+
+        // Forward batched set_props to each renderer
+        for (const [rendererId, payloads] of setPropsPayloadsByRenderer) {
+            const port = this.renderers.get(rendererId);
+            if (port) {
+                try {
+                    port.postMessage({
+                        type: WorkerMessageType.SET_PROPS_BATCH,
+                        rendererId,
+                        payload: payloads
+                    });
+                } catch (error) {
+                    console.warn(`Failed to forward batch to renderer ${rendererId}, removing`);
+                    this.renderers.delete(rendererId);
+                }
+            }
+        }
+
+        // Forward other messages individually
+        for (const msg of otherMessages) {
+            this.handleServerMessage(msg);
         }
     }
 
--- a/@plotly/dash-websocket-worker/src/MessageRouter.ts
+++ b/@plotly/dash-websocket-worker/src/MessageRouter.ts
@@ -111,6 +111,10 @@ export class MessageRouter {
 
         for (const message of messages) {
             const msg = message as WorkerMessage;
+            // Skip heartbeat_ack - already handled by WebSocketManager
+            if ((msg as any).type === 'heartbeat_ack') {
+                continue;
+            }
             if (msg.type === WorkerMessageType.SET_PROPS) {
                 const setPropsMsg = msg as SetPropsMessage;
                 const rendererId = setPropsMsg.rendererId;
--- a/@plotly/dash-websocket-worker/src/MessageRouter.ts
+++ b/@plotly/dash-websocket-worker/src/MessageRouter.ts
@@ -70,7 +70,7 @@ export class MessageRouter {
     public handleServerMessage(message: unknown): void {
         // Handle batched messages (array of messages)
         if (Array.isArray(message)) {
-            this.handleBatchedMessages(message);
+            this.handleBatchedMessages(message as WorkerMessage[]);
             return;
         }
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: @plotly/dash-websocket-worker/src/MessageRouter.ts [hunk 1]
```diff
@@ -92,6 +99,57 @@ export class MessageRouter {
         }
     }
 
+    /**
+     * Handle a batch of messages from the server.
+     * Groups set_props by renderer and forwards as a single batch message.
+     * @param messages Array of messages
+     */
+    private handleBatchedMessages(messages: unknown[]): void {
```
camdecoster: Why is `messages` of type `unknown[]`? Why not `WorkerMessages[]`?

@plotly/dash-websocket-worker/src/MessageRouter.ts
```diff
@@ -92,6 +99,57 @@ export class MessageRouter {
...
+        const setPropsPayloadsByRenderer: Map<string, SetPropsMessage['payload'][]> = new Map();
+        const otherMessages: WorkerMessage[] = [];
+
+        for (const message of messages) {
+            const msg = message as WorkerMessage;
+            // Skip heartbeat_ack - already handled by WebSocketManager
+            if ((msg as any).type === 'heartbeat_ack') {
+                continue;
+            }
+            if (msg.type === WorkerMessageType.SET_PROPS) {
```
camdecoster, line 122: If you use `WorkerMessage[]` instead of `unknown[]`, you can get rid of the casts here.

@plotly/dash-websocket-worker/src/MessageRouter.ts
```diff
@@ -92,6 +99,57 @@ export class MessageRouter {
...
+                const setPropsMsg = msg as SetPropsMessage;
+                const rendererId = setPropsMsg.rendererId;
+                if (!setPropsPayloadsByRenderer.has(rendererId)) {
+                    setPropsPayloadsByRenderer.set(rendererId, []);
+                }
+                setPropsPayloadsByRenderer.get(rendererId)!.push(setPropsMsg.payload);
```
camdecoster: Non null assertions are a code smell. Could this be rewritten to avoid it?

@plotly/dash-websocket-worker/src/MessageRouter.ts
```diff
@@ -92,6 +99,57 @@ export class MessageRouter {
...
+            } else {
+                otherMessages.push(msg);
+            }
+        }
+
+        // Forward batched set_props to each renderer
+        for (const [rendererId, payloads] of setPropsPayloadsByRenderer) {
+            const port = this.renderers.get(rendererId);
+            if (port) {
```
camdecoster: Should a message be logged if `port` is falsy?

@plotly/dash-websocket-worker/src/MessageRouter.ts
camdecoster, line 1: Does the message order matter? If you batch everything, the order will change. `[set_prop_1, otherMessage, set_prop_2]` gets sent out in the order of `[set_prop_1, set_prop_2]`, `[otherMessage]`.

@plotly/dash-websocket-worker/src/WebSocketManager.ts
```diff
@@ -195,7 +195,23 @@ export class WebSocketManager {
         try {
             const data = JSON.parse(event.data);
```
camdecoster: Could you cast this as a `WorkerMessage` (or more appropriate type)?

@plotly/dash-websocket-worker/src/WebSocketManager.ts
```diff
@@ -195,7 +195,23 @@ export class WebSocketManager {
...
+            // Handle batched messages - check for heartbeat_ack in the batch
+            if (Array.isArray(data)) {
+                for (const msg of data) {
+                    if (msg && msg.type === 'heartbeat_ack') {
+                        this.clearHeartbeatTimeout();
+                        break;
+                    }
+                }
+                // Track activity and forward batch for processing
+                this.lastActivityTime = Date.now();
```
camdecoster: Could you have a case where the message array is just one `'heartbeat_ack'` message? If so, then you're updating `lastActivityTime`, which differs from how you handle a single heartbeat message.

dash/backends/ws.py
```diff
@@ -184,29 +185,67 @@ def create_ws_context(
...
-    Messages are pre-serialized strings (using to_json).
+    Messages are pre-serialized strings (using to_json). For efficiency,
+    this function batches messages that arrive within batch_delay of each
+    other, sending them as a JSON array. When no message arrives within
+    the window, all collected messages are sent immediately.
 
     Args:
         send_text: Async function to send text data over WebSocket
         outbound_queue: janus.Queue instance for receiving messages (strings)
+        batch_delay: Time in seconds to wait for additional messages (default: 5ms)
```
camdecoster: The OP says use `0` if you want to disable batching. Maybe that should be included here? Would `None` be a better option? Would it be better to call say this disables debouncing rather than disabling batching?

dash/backends/ws.py
```diff
@@ -184,29 +185,67 @@ def create_ws_context(
...
     try:
         while True:
-            msg = await outbound_queue.async_q.get()
-            if msg == SHUTDOWN_SIGNAL:
-                break
-            await send_text(msg)
+            # Wait indefinitely for first message, then use timeout for batching
+            timeout = batch_delay if messages else None
+            try:
+                msg = await asyncio.wait_for(q.get(), timeout=timeout)
```
camdecoster, line 215: Would it be more appropriate to get a message and send it immediately with `send_text` if `batch_delay` is 0?

dash/dash-renderer/src/observers/websocketObserver.ts
```diff
@@ -129,6 +130,20 @@ export async function initializeWebSocket(
         }
     };
 
+    // Handle single SET_PROPS message
+    workerClient.onSetProps = (payload: SetPropsPayload) => {
+        processSetProps(payload);
+    };
```
camdecoster: Would this also work?

```suggestion
    workerClient.onSetProps = processSetProps;
```

dash/backends/ws.py
```diff
@@ -184,29 +185,67 @@ def create_ws_context(
...
+    other, sending them as a JSON array. When no message arrives within
+    the window, all collected messages are sent immediately.
 
     Args:
         send_text: Async function to send text data over WebSocket
         outbound_queue: janus.Queue instance for receiving messages (strings)
+        batch_delay: Time in seconds to wait for additional messages (default: 5ms)
     """
+    q = outbound_queue.async_q
+    messages: list = []
```
camdecoster: You could tighten this type if you want:

```suggestion
    messages: list[str] = []
```

dash/dash.py
```diff
@@ -486,6 +486,7 @@ def __init__(  # pylint: disable=too-many-statements, too-many-branches
         websocket_callbacks: Optional[bool] = False,
         websocket_allowed_origins: Optional[List[str]] = None,
         websocket_inactivity_timeout: Optional[int] = 300000,
+        websocket_batch_delay: Optional[float] = 0.005,
```
camdecoster, line 489: What happens if a user explicitly passes in `None` for this value?

camdecoster (commented):
Could you add a changelog entry? I'd also like to see a test related to the new batching mechanism.

dash/backends/ws.py
T4rk1n: 0 matches the same pattern as the inactivity config. Debouncing is the just the mechanism by which the batching happens, so batching is more relevant than debounce imo.

dash/backends/ws.py
T4rk1n: Maybe it saves one iteration :+1:

dash/dash.py
T4rk1n: It doesn't pass typing? Optional[float] requires a float and has float as default.

camdecoster (approved):

PR #3785, opened 2026-05-15
Fix graph patch & duplicate clicks

T4rk1n:
- Fix patch not working with dcc.Graph figure
- Fix dcc.Graph duplicate clicks not triggering props updates.

### DIFF
--- a/components/dash-core-components/src/fragments/Graph.react.js
+++ b/components/dash-core-components/src/fragments/Graph.react.js
@@ -317,6 +317,8 @@ class PlotlyGraph extends Component {
         if (!layout) {
             return layout;
         }
+        // Clone layout to avoid mutating the original (important for Patch)
+        layout = {...layout};
         const override = this.getLayoutOverride(responsive);
         const {override: prev_override, originals: prev_originals} = this.state;
         // Store the original data that we're about to override
@@ -414,6 +416,8 @@ class PlotlyGraph extends Component {
         gd.on('plotly_click', eventData => {
             const clickData = filterEventData(gd, eventData, 'click');
             if (!isNil(clickData)) {
+                // Add timestamp to ensure each click is unique (for DashWrapper deduplication)
+                clickData.timestamp = Date.now();
                 setProps({clickData});
             }
         });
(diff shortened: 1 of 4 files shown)


components/dash-core-components/src/fragments/Graph.react.js
```diff
@@ -317,6 +317,8 @@ class PlotlyGraph extends Component {
         if (!layout) {
             return layout;
         }
+        // Clone layout to avoid mutating the original (important for Patch)
+        layout = {...layout};
```
camdecoster: You might want to use a deep clone here to avoid mutating child objects.

components/dash-core-components/src/fragments/Graph.react.js
```diff
@@ -444,6 +450,29 @@ class PlotlyGraph extends Component {
...
                 setProps({relayoutData: relayout});
             }
+            // Sync shapes from gd.layout to figure when shapes are modified by user
+            // This is needed because getLayout() clones layout to prevent mutation issues
+            if (eventData && gd.layout) {
+                const hasShapeChanges = Object.keys(eventData).some(
+                    key => key === 'shapes' || key.startsWith('shapes[')
+                );
+                if (hasShapeChanges) {
+                    const {figure} = this.props;
```
camdecoster: Could `figure` be undefined here? You could guard against that with a default value:


```suggestion
                    const {figure = {}} = this.props;
```

components/dash-core-components/src/fragments/Graph.react.js
```diff
@@ -444,6 +450,29 @@ class PlotlyGraph extends Component {
             if (!isNil(relayout) && !equals(relayout, relayoutData)) {
                 setProps({relayoutData: relayout});
             }
+            // Sync shapes from gd.layout to figure when shapes are modified by user
+            // This is needed because getLayout() clones layout to prevent mutation issues
```
camdecoster, line 455: Do you need to do the same thing for annotations, selections, etc.?

camdecoster (approved):
This seems fine. I left a few comments. Could you add a changelog entry?

components/dash-core-components/src/fragments/Graph.react.js
T4rk1n: I don't think so, it's really meant for toggle clicks in something like a heatmap, you could toggle one cell and toggle it back again later.

PR #3796, opened 2026-05-26
Implement configuration function to customize MCP server

KoolADE85:
This PR adds a function to configure the MCP server behaviour:

Rather than add a growing list of constructor args around the MCP server, this PR implements a single `configure_mcp_server()` function that allows users to toggle various parts of the server with arguments:

```
def configure_mcp_server(
    *,
    include_layout: bool = True,
    include_callbacks: bool = True,
    include_clientside_callbacks: bool = True,
    include_pages: bool = True,
    expose_callback_docstrings: bool = False,
) -> None:
```

The existing `mcp_expose_docstrings` constructor arg has been removed and replaced by this function.

### DIFF
--- a/dash/dash.py
+++ b/dash/dash.py
@@ -492,7 +492,6 @@ class Dash(ObsoleteChecker):
         websocket_batch_delay: Optional[float] = 0.005,
         enable_mcp: Optional[bool] = None,
         mcp_path: Optional[str] = None,
-        mcp_expose_docstrings: Optional[bool] = None,
         **obsolete,
     ):
 
@@ -574,9 +573,6 @@ class Dash(ObsoleteChecker):
             hide_all_callbacks=False,
             csrf_token_name=csrf_token_name,
             csrf_header_name=csrf_header_name,
-            mcp_expose_docstrings=get_combined_config(
-                "mcp_expose_docstrings", mcp_expose_docstrings, False
-            ),
         )
         self.config.set_read_only(
             [
@@ -620,6 +616,7 @@ class Dash(ObsoleteChecker):
         self._callback_list: list = []
         self.callback_api_paths: dict = {}
         self.mcp_decorated_functions: dict = {}
+        self.mcp_callback_map: Any = None
 
         # list of inline scripts
         self._inline_scripts: list = []
--- a/dash/mcp/_configure.py
+++ b/dash/mcp/_configure.py
@@ -6,6 +6,8 @@
 
 from __future__ import annotations
 
+from typing import Optional
+
 from dash import get_app
 from dash.exceptions import AppNotFoundError
 from dash.mcp.primitives.resources import _RESOURCE_PROVIDERS as MCP_RESOURCE_PROVIDERS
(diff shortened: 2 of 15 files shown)


dash/mcp/_configure.py
```diff
@@ -0,0 +1,90 @@
...
+
+    updated_resources = list(_ALL_MCP_RESOURCE_PROVIDERS)
+    if not include_layout:
+        updated_resources.remove(LayoutResource)
+        updated_resources.remove(ComponentsResource)
+    if not include_clientside_callbacks:
+        updated_resources.remove(ClientsideCallbacksResource)
+    if not include_pages:
+        updated_resources.remove(PagesResource)
+        updated_resources.remove(PageLayoutResource)
```
camdecoster: What do you think of creating lists to keep things tidy? Something like this:
```py
LAYOUT_RESOURCES = [ComponentsResource, LayoutResource]
if not include_layout:
  updated_resources.remove(item) for item in LAYOUT_RESOURCES
```

dash/mcp/_configure.py
```diff
@@ -0,0 +1,90 @@
...
+    MCP_RESOURCE_PROVIDERS[:] = updated_resources
+
+    updated_tools = list(_ALL_MCP_TOOL_PROVIDERS)
+    if not include_layout:
+        updated_tools.remove(GetDashComponentTool)
+    MCP_TOOL_PROVIDERS[:] = updated_tools
+
+    get_app().mcp_callback_map = None
```
camdecoster: Do you need to wrap this in a try/except block too?

tests/integration/mcp/conftest.py
camdecoster, line 1: Do you need to reset `MCP_DECORATED_FUNCTIONS` like you do in **tests/unit/mcp/conftest.py**?
```py
app_context.set(app)
if MCP_DECORATED_FUNCTIONS:
    app.mcp_decorated_functions = dict(MCP_DECORATED_FUNCTIONS)
    MCP_DECORATED_FUNCTIONS.clear()
```

dash/mcp/_configure.py
```diff
@@ -0,0 +1,90 @@
...
+    include_callbacks: bool = True,
+    include_clientside_callbacks: bool = True,
+    include_pages: bool = True,
+    expose_callback_docstrings: bool = False,
+) -> None:
+    """
+    Configure which content the Dash MCP server exposes.
+
+    Any parameter that is omitted will be reset to its default value. Calling
+    with no args will reset all configuration to its default state.
```
camdecoster: IMO, it seems safer to only mutate the config values when they're specifically passed in. On first call they get set with defaults or the provided values, then subsequent calls only update the specific values passed in.

dash/dash.py
```diff
@@ -616,6 +612,7 @@ def __init__(  # pylint: disable=too-many-statements, too-many-branches
         self._callback_list: list = []
         self.callback_api_paths: dict = {}
         self.mcp_decorated_functions: dict = {}
+        self.mcp_callback_map: Any = None
```
camdecoster, line 619:
```suggestion
        self.mcp_callback_map: Optional["CallbackAdapterCollection"] = None
```

dash/mcp/_configure.py
```diff
@@ -0,0 +1,90 @@
...
+            include_callbacks=False,
+            include_clientside_callbacks=False,
+            include_pages=False,
+        )
+    """
+    try:
+        if get_app().backend.has_request_context():
+            raise RuntimeError("MCP server can't be configured within a callback")
+    except AppNotFoundError:
+        ...
```
camdecoster:
```suggestion
        pass
```

camdecoster (approved):
I left a couple of suggestions, but nothing huge. Could you add a changelog entry for this update? Specifically describing the way the user facing API changed.

dash/mcp/_configure.py
KoolADE85: Done! I used sets instead of lists to be clear that the _order_ of resources isn't re-defined here: only their inclusion.

dash/mcp/_configure.py
KoolADE85: Good catch!

tests/integration/mcp/conftest.py
KoolADE85: Done - added this as a teardown step after the test runs.

dash/mcp/_configure.py
KoolADE85: Agreed. I switched this to a "patch" style configuration as you described. 👍

dash/dash.py
KoolADE85: This is actually needs to be `Any` for now because of Python < 3.10.
`CallbackAdapterCollection` imports from the `mcp` module which cannot be installed on 3.8. Therefore, if we import it here for typing, it will break on older python versions.

camdecoster: Stupid backward compatibility!

PR #3797, opened 2026-05-26
improved websocket connection management

### LINKED ISSUE #3788: [BUG][4.2] Persitent websocket callback lost updates.
When a websocket disconnect on the frontend by cause of inactivity for example, the websocket is reconnect when any other callback happen. But if you have a persistent websocket callback running in the background, it will not be able to send set_props or get_props request and the is_shutdown is not changed since it was reconnected.

This suggest the queue handling is lost when the websocket reconnect.

### LINKED ISSUE #3790: [BUG][4.2.0rc1] backends._fastapi: unhandled WebSocketDisconnect cascades into 3-frame ASGI traceback on idle-WS reap
When a Dash 4.2 WebSocket-served callback handler tries to send a response on a socket the client has already closed (very common for long-idle browser tabs,
  mobile background tabs, or proxy/edge reaps), dash.backends._fastapi doesn't catch starlette.websockets.WebSocketDisconnect around websocket.send_json(...).
  The exception escapes the task, the next iteration of the receive loop then calls websocket.receive_json() on the already-closed socket, and a second handler
  re-tries send_json after Starlette has emitted its close frame — producing three back-to-back tracebacks per disconnected client and an ERROR: Exception in 
  ASGI application from uvicorn. The client experience is unaffected (it's already gone), but each stale-WS reap costs ~50 log lines and surfaces as an ASGI
  error in observability tooling.

  Environment

  - dash==4.2.0rc1 (backend="fastapi", websocket_callbacks=True)
  - Python 3.12
  - starlette (current FastAPI dep), uvicorn (websockets impl), websockets package
  - Render.com web service behind their edge proxy (i.e. proxy may reap long-idle sockets)
  - Optional but present in the stack: dash-clerk-auth ASGI middleware (not the source of the bug — exception originates inside dash.backends._fastapi; included
  only because it appears in the traceback)

  Observed behavior

  After a browser tab sat idle on a Dash page for ~7 hours, the WebSocket dropped (proxy reap) and the server logged:

  future: <Task finished name='Task-2146'
    coro=<FastAPIDashServer.serve_websocket_callback.<locals>.websocket_handler.<locals>.execute_callback_task() done,
    defined at /usr/local/lib/python3.12/site-packages/dash/backends/_fastapi.py:727>
    exception=WebSocketDisconnect()>
    File ".../uvicorn/protocols/websockets/websockets_sansio_impl.py", line 433, in send
      self.conn.send_text(text_data.encode())
    File ".../websockets/protocol.py", line 337, in send_text
      raise InvalidState(f"connection is {self.state.name.lower()}")
  websockets.exceptions.InvalidState: connection is closing

  During handling of the above exception, another exception occurred:
    File ".../starlette/websockets.py", line 86, in send
      await self._send(message)
    ...
    File ".../uvicorn/protocols/websockets/websockets_sansio_impl.py", line 452, in send
      raise ClientDisconnected()
  uvicorn.protocols.utils.ClientDisconnected

  During handling of the above exception, another exception occurred:
    File ".../dash/backends/_fastapi.py", line 735, in execute_callback_task
      await websocket.send_json(
    File ".../starlette/websockets.py", line 176, in send_json
      await self.send({"type": "websocket.send", "text": text})
    File ".../starlette/websockets.py", line 89, in send
      raise WebSocketDisconnect(code=1006)
  starlette.websockets.WebSocketDisconnect

  Immediately after, a sibling task hits the same dead socket on the retry path:

  2026-05-20 03:17:05,883 ERROR Task exception was never retrieved
  future: <Task finished name='Task-2147' ... exception=RuntimeError('Cannot call "send" once a close message has been sent.')>
    File ".../dash/backends/_fastapi.py", line 735, in execute_callback_task
      await websocket.send_json(
    File ".../starlette/websockets.py", line 98, in send
      raise RuntimeError('Cannot call "send" once a close message has been sent.')
  RuntimeError: Cannot call "send" once a close message has been sent.

  And finally the receive loop in the handler itself tries to read from the now-fully-closed socket:

  ERROR:    Exception in ASGI application
    ...
    File ".../dash/backends/_fastapi.py", line 748, in websocket_handler
      message = await websocket.receive_json()
    File ".../starlette/websockets.py", line 134, in receive_json
      raise RuntimeError('WebSocket is not connected. Need to call "accept" first.')
  RuntimeError: WebSocket is not connected. Need to call "accept" first.

  The full ASGI middleware chain (FastAPI → custom middleware → auth._asgi → SessionMiddleware → dash.backends._fastapi.__call__ → starlette router →
  dash.backends._fastapi.websocket_handler) propagates the final RuntimeError all the way to uvicorn, which logs it as ERROR: Exception in ASGI application.

  Triggering scenario (deterministic)

  1. Client opens a Dash 4.2 page with backend="fastapi" and websocket_callbacks=True.
  2. The WebSocket completes the initial dependency exchange and goes idle.
  3. The client disappears in a way that the server learns about between the dispatch of a callback and its response — most reliably reproduced by a proxy/edge
  reap on a long-idle socket, but also by simply closing the tab while a callback is in flight, or by mobile background-tab kill.
  4. The server tries to deliver the callback response → cascade begins.

  Root cause analysis

  In dash/backends/_fastapi.py (rc1 line numbers — please confirm against current HEAD):

  - Line 735 (execute_callback_task): await websocket.send_json(...) is not wrapped in a handler for starlette.websockets.WebSocketDisconnect or
  uvicorn.protocols.utils.ClientDisconnected. When the underlying socket is in closing/closed, starlette.websockets.WebSocket.send raises
  WebSocketDisconnect(code=1006); the task exits with exception=WebSocketDisconnect() and is never retrieved, so Python emits the "Task exception was never
  retrieved" warning and the framework keeps trying.
  - Line 735 (a second task): a second concurrent execute_callback_task invocation hits send_json after starlette has already emitted its close frame for the
  same socket, so it raises RuntimeError('Cannot call "send" once a close message has been sent.').
  - Line 748 (websocket_handler): the outer receive loop calls await websocket.receive_json() without first checking websocket.client_state /
  websocket.application_state; on the now-closed socket starlette raises RuntimeError('WebSocket is not connected. Need to call "accept" first.'), which escapes
  to uvicorn and surfaces as ERROR: Exception in ASGI application.

  Net: a single client disconnect produces (1) an unhandled task exception, (2) a "Cannot send after close" follow-up, (3) an unhandled receive on a closed
  socket — each with a full multi-frame traceback.

  Suggested fix

  Conceptually, every send_json/receive_json on the WS in _fastapi.py should treat WebSocketDisconnect and RuntimeError-on-closed-socket as a clean terminal 
  condition:

  # dash/backends/_fastapi.py — around the existing send/receive sites
  from starlette.websockets import WebSocketDisconnect, WebSocketState

  _DISCONNECT_EXC = (WebSocketDisconnect,)

  # line ~735 in execute_callback_task
  try:
      if websocket.application_state == WebSocketState.CONNECTED:
          await websocket.send_json(payload)
      else:
          logger.debug("WS callback response dropped: socket already closed")
  except _DISCONNECT_EXC:
      logger.debug("WS callback response: client disconnected mid-send")
  except RuntimeError as exc:
      # 'Cannot call "send" once a close message has been sent.' /
      # 'WebSocket is not connected. ...' — both reduce to "client gone".
      if "close" in str(exc).lower() or "not connected" in str(exc).lower():
          logger.debug("WS callback response: %s", exc)
      else:
          raise
  
  # line ~748 in websocket_handler
  while True:
      if websocket.application_state != WebSocketState.CONNECTED:
          break
      try:
          message = await websocket.receive_json()
      except _DISCONNECT_EXC:
          logger.debug("WS callback receive: client disconnected")
          break
      except RuntimeError as exc:
          if "not connected" in str(exc).lower():
              break
          raise
      ...
  
  The key invariants:

  1. Never call send_json without checking application_state == CONNECTED first, or alternatively, always wrap it in try/except WebSocketDisconnect. Currently
  neither is true at line 735.
  2. Catch WebSocketDisconnect in the receive loop, and exit the loop cleanly. The current receive_json at line 748 is bare.
  3. Treat RuntimeError messages mentioning "close" or "not connected" as the same terminal condition — starlette uses RuntimeError (not WebSocketDisconnect) for
   the "send after close-frame-sent" and "receive before accept" cases, which is a starlette API quirk worth absorbing here rather than asking every framework
  user to.

  Impact
  
  - No user-visible regression (the client is already gone in every case observed).
  - Significant log/observability noise: each stale-WS reap produces ~50 lines of traceback across three tasks, surfaces as ERROR: Exception in ASGI application
  from uvicorn, and triggers Python's "Task exception was never retrieved" warning. At scale this will dominate error dashboards for any deployment behind a
  reaping proxy (Render, Heroku, Cloud Run, Cloudflare, etc.) or with mobile clients.
  - Mitigation in user code is awkward: a wrapping ASGI middleware can swallow the exceptions, but only by also masking real RuntimeErrors in handlers — I thought the
  right fix belongs inside dash.backends._fastapi.

T4rk1n:
Fix #3788 and fix #3790

- Disable/reset inactivity timeout when the app tab is active.
- Tabbing back restarts the ws connection instead of waiting for a callback.
- Persistent Callback with no ouputs/inputs that were running while the app disconnected will be restarted automatically.
- Improved error handling when the websocket disconnect/reconnect.


dash/backends/_fastapi.py
```diff
@@ -753,6 +762,22 @@ async def websocket_handler(websocket: WebSocket):
                         renderer_id = message.get("rendererId", "")
                         payload = message.get("payload", {})
 
+                        # Update current renderer ID for cleanup
+                        current_renderer_id = renderer_id
+
+                        # Adopt connection for this renderer (allows reconnection)
+                        # Called for every callback to ensure registry entry exists
+                        # (entry may have been cleaned up after previous callback)
+                        registry.adopt_connection(
```
camdecoster: Is this idempotent?

dash/backends/_fastapi.py
```diff
@@ -753,6 +762,22 @@ async def websocket_handler(websocket: WebSocket):
...
+                            renderer_id,
+                            outbound_queue,
+                            pending_get_props,
+                            shutdown_event,
+                        )
+
+                        # Register this callback with the registry
+                        registry.register_callback(renderer_id)
```
camdecoster: Why is this required?

T4rk1n:
@camdecoster This is ready for review, I removed the callback adoption that was not working for multi worker production environment.

dash/dash-renderer/src/observers/websocketObserver.ts
```diff
@@ -175,13 +177,53 @@ export async function initializeWebSocket(
...
+            );
+            const state = store.getState();
+            const {graphs} = state;
+
+            if (graphs?.callbacks) {
+                const persistentCallbacks = graphs.callbacks.reduce(
+                    (acc: any[], cb: any) => {
+                        // Only re-trigger no-output callbacks with no inputs
+                        // These are the "persistent" callbacks that should restart
+                        if (cb.noOutput && cb.inputs.length === 0) {
```
camdecoster, line 200: This checks for persistent callbacks, but why doesn't it look at the `persistent` flag (from `isLoading.ts`)?

dash/dash-renderer/src/observers/websocketObserver.ts
```diff
@@ -201,6 +243,24 @@ export async function initializeWebSocket(
     } catch (error) {
         console.error('[Dash] Failed to connect to WebSocket worker:', error);
     }
+
+    // Handle tab visibility changes
+    document.addEventListener('visibilitychange', () => {
```
camdecoster, line 248: Is there a situation where you'd need to remove the event listener?

tests/websocket/test_ws_reconnect.py
```diff
@@ -0,0 +1,253 @@
...
+        ws = ctx.websocket
+        set_props("status", {"children": "running"})
+
+        # Properly check is_shutdown in the loop
+        for i in range(1, 11):
+            if ws and ws.is_shutdown:
+                # Exit gracefully on disconnect
+                raise PreventUpdate
+            set_props("progress", {"children": str(i * 10)})
+            await asyncio.sleep(0.2)
```
camdecoster, line 241: Will this hit the 2s timeout when it makes a new call every 0.2s?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3669](https://github.com/plotly/dash/pull/3669) Selection for DataTable cleared with custom action settings
 - [#3680](https://github.com/plotly/dash/pull/3680) Added `search_order` prop to `Dropdown` to allow users to preserve original option order during search 
 - Added `csrf_token_name` and `csrf_header_name` config options to allow configuring the CSRF cookie and header names. Fixes [#729](https://github.com/plotly/dash/issues/729)
+- [#3797](https://github.com/plotly/dash/pull/3797) Improved websocket callback management.
```
camdecoster, line 12: Will the changes in this fix be documented somewhere besides this changelog entry?

dash/dash-renderer/src/observers/websocketObserver.ts
T4rk1n: No, it's for the complete page lifecycle.

dash/dash-renderer/src/observers/websocketObserver.ts
T4rk1n: No it's actually just the no Output and no Inputs since those cannot be restarted after startup. Other cases can use reload button or other mechanism.

CHANGELOG.md
T4rk1n: A more comprehensive guide for the websocket shutdown management can be found on the WebsocketCallback docstring and in the `.ai/ARCHITECTURE.md`

camdecoster (approved):

PR #3799, opened 2026-05-28
fix: dropdown search crash

### LINKED ISSUE #3707: [BUG] 4.1.0 dropdown search crash
**Describe your context**
```
dash                 4.1.0
```

**Describe the bug**

When typing a subset search pattern that matches a **large** subset of the items, then it causes the dropdown to crash

**Expected behavior**

The filtered items should be visible instead of a crash

**Minimal Working Code Example**
```

app = Dash(__name__)

options = [
    {"label": html.Div(["Item", html.Span(f"#{i}")]),
     "value": f"item_{i}",
     "search": f"item_#{i}"}
    for i in range(1, 5001)
]

app.layout = html.Div(
    [
        html.H3("Dropdown that crashes after typing a number"),
        dcc.Dropdown(
            id="big-dropdown",
            options=options,
            value="item_1",
            placeholder="Select an item",
            searchable=True,
            clearable=True,
            style={"width": "500px"},

        ),
    ],
)

if __name__ == "__main__":
    app.run(debug=True)
```

If you type a number in the dropdown search bar then the application crashes and the search input is frozen

AnnMarieW:
closes #3707

This fixes a virtualization issue introduced in Dash 4.1 for large searchable option lists with component labels.

react-window uses the row index as the default key. After filtering, the same index may point to a different option, causing stale component rows to be reused. With component labels rendered through ExternalWrapper, this could result in errors like:

> Cannot read properties of undefined (reading 'props')

Adding a stable `itemKey` based on the option value makes sure the rows are remounted correctly when the filtered options change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3768](https://github.com/plotly/dash/pull/3768) Improved `Dropdown` search performance for large options lists
 - [#3759](https://github.com/plotly/dash/pull/3759) Fix the issue where `Patch` objects cannot be updated via `set_props()` in `websocket` callback. Fix [#3742](https://github.com/plotly/dash/issues/3742)
 - [#3789](https://github.com/plotly/dash/pull/3789) Fixed extra wrapper in `DatePickerRange` and `DatePickerSingle` causing styling and layout issues.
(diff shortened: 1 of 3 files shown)


camdecoster:
Good find! Could you double check linting? Once that's fixed, I'll approve and this can be merged.

camdecoster (approved):

PR #3802, opened 2026-05-30
Feature #2717: dcc.DateRangeSlider and dcc.DateSlider

### LINKED ISSUE #2717: [Feature Request] Add date support for dcc.Slider/dcc.RangeSlider

Ideally, something like:
```
import datetime
from dateutil.relativedelta import relativedelta
# ...
dcc.RangeSlider(
        type='date',
        min=datetime.datetime(2020, 5, 20), 
        max=datetime.datetime(2023, 12, 20), 
        step=relativedelta(months=+6),
        value=[datetime.datetime(2022,10,20), datetime.datetime(2023,5,20)]
)
```
Less ideally, but understandable since dates are not json serializable (I think?):
```
dcc.RangeSlider(
        type='date',
        dateformat='%Y-%m-%d',
        min='2020-05-20', # str
        max='2023-12-20',
        step='0-6-0', # 6 months in string format
        # this string information would be transformed to do the calculations in the background with something like
        value=['2022-10-20', '2023-5-20']
)
```

```
from dash import Dash, html, dcc, dash_table, Input, State, Output, callback
import plotly.express as px
from dateutil.relativedelta import relativedelta
import pandas as pd
from datetime import datetime

app = Dash(__name__)

df = px.data.stocks(datetimes=True)
all_dates = df["date"]
# transform dates to string in the month/year format
all_dates_my = all_dates.apply(lambda x: x.strftime("%m/%Y")).unique()

app.layout = html.Div([
    # slider for months
    dcc.RangeSlider(
        min=0, 
        max=len(all_dates_my)-1, 
        value=[2,4],
        step=1,
        marks={
            i:{'label': all_dates_my[i]} for i in range(len(all_dates_my))
        },
        id='date-slider-01'),
    # slider for weeks
    html.Div(id='out')
])

@callback(
    Output('out', 'children'),
    Input('date-slider-01', 'value'),
    State('date-slider-01', 'marks')
)
def slider_value(val, marks):
    selected_dates_str = [marks[str(v)]['label'] for v in val]
    # selected_dates = [datetime.strptime(marks[str(v)]['label'], format="%m/%Y")  for v in val]
    filtered_df = df.query(f"date >= '{selected_dates_str[0]}' and date < '{selected_dates_str[1]}'")
    # convert the selected dates into datetime format
    selected_dates = [datetime.strptime(v, "%m/%Y")  for v in selected_dates_str]
    print(selected_dates)

    return [
        html.Div(f"the selected range is from {selected_dates_str[0]} to {selected_dates_str[1]}"),
        dash_table.DataTable(filtered_df.to_dict('records'), [{"name": i, "id": i} for i in filtered_df.columns])
    ]

if __name__ == '__main__':
    app.run(debug=True)

```

**Additional context**
Example case where this would be useful instead of using DatePicker:

> The datepicker range can only move on a monthly basis and the calendar is redundant. My purpose is to specify the learning period for a machine learning model, and there is no need for calendar (day of the week) information; the key is to choose a range of 24 months.

ines-om:
This PR fixes #2717.

We implement a `dcc.DateRangeSlider` and a `dcc.DateSlider` that wraps the existing sliders, making them date compatible. 

The features developed include:
- Disabling specific dates, date ranges or events (like weekends and specific weekdays), through `disable_flags` and `disabled_dates` props;
- An optional disabled dates indicator, which visually signals to the user where disabled dates are in the slider, through the `disabled_dates_indicator` prop;
- A new property that prevents selected ranges from spanning over disabled dates, through the `no_disabled_in_between` prop, which wasn't in the original plan but made sense to include.

We initially proposed the ability to disable holidays, but decided against it because it would have meant maintaining holiday data for different countries, which felt like too much added complexity for limited benefit (users can just pass in the dates they want to disable).

These components make use of the previous sliders’ props, like steps (now year, month and day based), marks, min, max, vertical, etc.

## Contributor Checklist

   -  [x] Created  `fragments/DateRangeSlider.tsx` wrapping  `dcc.RangeSlider'
   -  [x] Created  `components/DateRangeSlider.tsx`
   -  [x] Created  `DateSlider.tsx` wrapping  `dcc.DateRangeSlider`
   -  [x] Changed `sliders.css` to contain date slider logic and addition of disabled dates indicator
   -  [x] Expanded `types.ts` to include date sliders' props
   -  [x] Added `helpers.ts` functions for logic of disabling dates, stepping dates, dealing with disabled date ranges, enforcing the optional no disabled dates in between rule, etc.
    - [x] Created ` utils/computeDateSliderMarkers.ts` 
    -  [x] Added tests to ` helpers.test.ts` and `computeDateSliderMarkers.test.ts`

### optionals


AnnMarieW:
Hi @ines-om 

Nice work on this PR. Date-based controls are  tricky to implement, and date sliders would be a nice addition to Dash. I spent some time testing it and wanted to share some initial feedback. I'll add more comments as I continue trying different features.

- `allow_direct_input` prop should be removed since there are no input fields
- Types for dates (value, min, max etc)  need to accept datetime objects.  See datepickers for an example.
- Need to add integration tests using [dash testing](https://dash.plotly.com/testing).  See the other [slider tests for examples](https://github.com/plotly/dash/tree/dev/components/dash-core-components/tests/integration/sliders)

- if marks are defined, it only allows dates at the predefined marks.  To be consistant with other slider components, that should only be the case when `step=None`
- Rather than having `delta_days`, `delta_months` `delta_years`, it would be better to have `step` (number | null) and `step_unit` ("day" | "month" | "year")
- Note - It's common to select end-of-month dates for financial reports.  That doesn't work currently. Try setting `dcc.DateSlider(min="2025-01-31, max="2025-12-31, delta_months=1)`


- The labels overflow the containers at the min and max points:

image


--------------------------------


- at the max point there are two labels that overlap:

-------------------------

image



-------------------

- default format for tooltip should be yyyy-mm-dd (same as the default for the marks)
- date shown in tooltip is off by one compared to value as shown in the callback

---------------------------

image

-----------------

That's all I've found so far. Overall, this is a solid start!

Francisc0C:
Hi @AnnMarieW,

Thank you very much for the feedback and for taking the time to test the PR so thoroughly. We really appreciate it.

@ines-om and I will work on addressing these issues as soon as possible. We would also greatly appreciate any additional feedback you may have.

Regarding the issue where the date shown in the tooltip is off by one compared to the value shown in the callback, would it be possible for you to share which specific test or scenario caused the error? That would help us verify the fix properly.

Thank you again!

AnnMarieW:
Hi @ines-om and @Francisc0C 

Here is a simple example app for testing:

```python
from dash import Dash, Input, Output,  html, dcc
import datetime

app = Dash()

app.layout = html.Div([
    dcc.DateSlider(
        value="2025-02-01",       
        min=datetime.date(2025, 1, 1),
        max=datetime.date(2025, 12, 31),
        id="date-slider",
        tooltip={"placement": "top", "always_visible": True},
      #  marks={"2025-01-01": "2025-01-01", "2025-02-01": "2025-02-01", "2025-03-01": "2025-03-01"},
      #  delta_months=1,
        delta_days=1,
    ),
    html.Div(id="date-slider-output", style={"marginTop": 20}),
], style={"marginTop": 60})

@app.callback(
    Output("date-slider-output", "children"),
    Input("date-slider", "value")
)
def update(value):
    return f" You selected: {value}"


if __name__ == "__main__":
    app.run(debug=True)


```

Francisc0C:
Hi,

We made a new commit that changes delta to step and step unit, fixes the end-of-month jumps, removes the direct input prop, and sets the default format to yyyy-mm-dd.

We’re going to continue improving the sliders, but for now this is what we were able to do. Feel free to let us know if you find anything else.

Regarding the end-of-month jump behavior, this currently only works properly if the slider minimum is set to a day 31. If it is set to day 30, the slider will keep jumping from the 30th to the 30th (or 28/29 in the case of February). Do you think it would make sense to add an anchor prop to control which day is used for these jumps? This way we could set the anchor to the 31st and consistently get end-of-month behavior. Right now, it behaves as if the minimum value is being used as the anchor.

Thanks!

AnnMarieW:
Thanks for the update @ines-om and @Francisc0C 


The month-end steps work great now (as long as the min day is 31)   How about checking if the min date is a month-end date and apply the same logic?

I suggested to remove the direct input prop because I couldn't get it  working in `DateSlider` and I thought that feature wasn't available.  However, I just tried the `DateRangeSlider` and it worked great in that component!  It's cool to have a `DatePicker` as the direct input - that's a nice UI.  Sorry to advise you to remove the prop.  Is it possible to add that feature  to the `DateSlider`?

The issue with the overlapping labels at the max point seems to be fixed in this version.

The default date format in the tooltip looks good now.

The new props for `step` and `step_unit` works well.  It also fixes the issue where it was only allowing to select dates with custom marks.  

More info for the off-by-one error between the tooltip and the value: it only seems to happen when the `step=1` and `step_unit="days"`

ines-om:
Hello! We have implemented the rest of the changes:
- Date Slider now accepts direct input prop;
- No overlay between dates and input or between marks themselves (We decided that if the max mark overlaps with the previous mark, the max mark should not show);
- Types for dates now accept datetime objects;

We are missing the integration tests but we will be sending them later this week. If you need us to change anything let us know! Thank you!

ines-om:
Hi @AnnMarieW, we noticed merging dev changed some of the the UI for the sliders and have fixed the visual issues. We'll send the fixes with the tests!

AnnMarieW:
@ines-om 
Thanks for the note about the visual issues.  The width of the sliders and date picker input fields looked odd in this version.  

I'd also recommend avoiding changes to the existing Slider CSS and, if possible, adding any new styles under a dedicated `date-slider` class. That will reduce the risk of introducing regressions in the other `Slider` components.

For consistency with the existing `Slider` components:

* When using a single-value slider (not a range slider), the input field should be positioned to the right of the slider.
* In narrow containers, the date input fields should be hidden. The CSS likely needs to be updated [here](https://github.com/ines-om/dash/blob/d573c446cac1a8b4881ea62f5c1d079d6126c43c/components/dash-core-components/src/components/css/sliders.css#L248).

### Disabled dates

I started testing the disabled dates feature and found several issues. Rather than addressing them all in this PR, it may be worth discussing whether disabled dates should be removed from the current scope and implemented as a separate feature later.

Some of the issues I found:

* Disabled dates can still be selected from the date picker calendar.
* When navigating with the keyboard arrow keys, reaching a disabled date prevents further movement of the slider handle.
* The visual indicators for disabled dates on the track are difficult to see unless the selected date range is very short (for example, one month instead of one year).
* The tooltip provides no indication that a date is disabled. The user only discovers this because selecting the date is not allowed.

Francisc0C:
Hi,

In terms of the disabled dates, most of the issues you mentioned should be relatively straightforward to address. The main exception is the keyboard navigation behaviour with the arrow keys, as that is tied to some of the decisions around how disabled dates are handled.

That said, @ines-om and I agree that removing the disabled dates logic would significantly simplify the slider and make the implementation easier to maintain.

If the team feels that disabled dates would be better handled as a separate feature, we would be happy to move them out of the scope of this PR and revisit them later.

Please let us know what you think.

AnnMarieW:
OK, we can give time for others to chime in on the disabled dates feature. 

In the meantime, there is another use case described in issue 2717: selecting a fixed range (for example 24 months) within the full min and max date range.

Right now, if you use `step=24` and `step_unit="months"`, the slider only allows values in  24-month intervals starting from `min_date`. That means you cannot pick arbitrary start or end dates inside the range.

This use case fits better with `RangeSlider`   `pushable` prop, possibly extended with a date unit like `pushable_unit`.

>`pushable` (number | boolean; optional): pushable could be set as True to allow pushing of surrounding handles when moving an handle. When set to a number, the number will be the minimum ensured distance between handles.

This would let you enforce a minimum range like 24 months, while still allowing any valid start date within the full range.

components/dash-core-components/src/fragments/DateRangeSlider.tsx
```diff
@@ -0,0 +1,551 @@
...
+                    <DatePickerSingle
+                        key={`min-input-${resetKey}`}
+                        className="dash-range-slider-min-input"
+                        date={displayValues[0]}
+                        setProps={({date}) => handleDateInputChange(0, date)}
+                        placeholder="Start date"
+                        min_date_allowed={minStr}
+                        max_date_allowed={maxStr}
+                        display_format={display_format}
+                    />
```
KoolADE85: The datepicker input is such a nice touch here :)
I think it needs a `flex: 0` to prevent it from expanding. I noticed it crowds out the slider and causes the labels to overlap
Image

ines-om:
@AnnMarieW Thank you for letting us know!

We just pushed the tests and UI updates, namely separating the CSS, making the input fields responsive, and resolving some sizing issues.

The keyboard test file currently has one remaining failure that we believe to be a bug. Fixing that, along with looking into this use case, will take us a bit more time.

Thank you for all the help!

ines-om:
Hello @AnnMarieW,

This commit adds additional integration tests and makes the existing ones less fragile. It also fixes the keyboard test bug we mentioned previously.

@Francisc0C and I also have a few questions regarding `pushable`.
To support `pushable`, we need to convert the `pushable` value together with its unit into a timestamp value and pass that to the underlying range slider. Unfortunately, this cannot be handled directly in the `DateRangeSlider` fragment. Without going through the range slider logic, we can only adjust the selected value itself, similarly to how disabled dates are currently handled.
This approach works correctly for days, but it becomes problematic for months and years because their duration in milliseconds is not constant. As a result, any timestamp-based conversion is only an approximation and can lead to incorrect behavior.

At the moment, we see the following options:
1. Restrict `pushable` to days only.
2. Allow approximate behavior for months and years. However, this may cause the slider to skip dates that should actually be valid, so we would prefer not to go with this option.
3. Enforce `pushable` only on commit for months and years by validating the selected range in the `DateRangeSlider`. This would partially defeat the purpose of pushable, since the constraint would not be enforced while dragging.
4. Not adding `pushable` support to DateRangeSlider.

We would appreciate any thoughts on which approach would be preferred.
That said, there may be an alternative approach that we have overlooked in our analysis. If the team is aware of a better way to implement this, or sees any flaws in our reasoning, we would greatly appreciate your feedback. 

Thank you!

AnnMarieW:
Hi @ines-om and @Francisc0C 

I think it would be reasonable to limit the `pushable` to be a minimum number of days only.   Adding support for months/years is more complex, and could be added in a future feature request.

components/dash-core-components/src/utils/calendar/helpers.ts
```diff
@@ -162,23 +190,96 @@ export function isDateInRange(
...
+/**
+ * Converts milliseconds since epoch to a YYYY-MM-DD date string.
+ * Always treats dates as UTC to avoid timezone shifts.
+ */
+export function timestampToDateString(
+    timestamp: number | undefined
+): string | undefined {
+    if (!timestamp) {
+        return undefined;
+    }
```
KoolADE85: Small bug: this will early-return if the timestamp is `0` (i.e. a date of 1970-01-01) and causes the slider handle to disappear.

KoolADE85 (commented):
This is a fantastic addition to DCC, I'm really liking the implementation overall.

Regarding the disabled_dates: it seems like there's a lot of room for edge cases to complicate the idea. I would be okay to exclude it for this PR and add it later.

But if we want to include it, I think we need to ensure:
* focus can never return to the `body` element when using the keyboard
* there should be integration tests that cover keyboard interaction (including edge cases) to avoid future regressions. This is an area that's easy to forget about when we update things.
* more subjective: but the visual appearance of the disabled dates should use existing CSS variables for colours for consistent theming

components/dash-core-components/src/fragments/DateRangeSlider.tsx
AnnMarieW: Does this apply to the latest version?  I think this has been fixed already.

KoolADE85: ah, you are right! I was looking at outdated code when I saw this 🤦‍♂️

components/dash-core-components/src/components/css/datesliders.css
```diff
@@ -0,0 +1,132 @@
+.date-slider-container,
+.date-range-slider-container {
+    display: flex;
+    align-items: center;
+    gap: 10px;
+    width: 100%;
+    container-type: inline-size;
```
KoolADE85: We shouldn't use the `container-type` rule here since it causes the popup calendar to be rendered in the wrong place, especially when the page is scrolled down.

ines-om:
Hi @AnnMarieW @KoolADE85,
Thank you so much for the guidance!
We are going to remove the `disabled_dates` functionality from this PR to make it cleaner and should be able to send the version with the mentioned fixes implemented and without the `disabled_dates` in the following days.

ines-om:
Hi @AnnMarieW @KoolADE85,

Just pushed the latest updates! We used a `ResizeObserver` to make the inputs disappear. We tried different ways of grouping the slider and inputs into a single container, but it kept causing rendering issues with the calendar box.

We also removed the `disabled_dates` logic entirely. It was quite a big removal, but the tests all pass and our date sliders' Dash scripts are working well.

If there's anything you need us to change, let us know!

ines-om:
@AnnMarieW Noticed some issues in helpers, but should be an easy fix! Will send the correct version soon.

ines-om:
@AnnMarieW Should be good now! Sorry about that, we hadn't removed the `disabled_dates` tests from the `helpers.test.ts` .

components/dash-core-components/src/components/DateRangeSlider.tsx
```diff
@@ -0,0 +1,40 @@
...
+import dateRangeSlider from '../utils/LazyLoader/dateRangeSlider';
+
+import './css/datesliders.css';
+
+const RealDateRangeSlider = lazy(dateRangeSlider);
+
+/**
+ * A date range slider component.
+ * Used for specifying a range of dates with optional disabled date indicators
+ * and calendar-aware stepping.
```
AnnMarieW, line 12:
```suggestion
 * Used for specifying a range of dates with calendar-aware stepping.
```

components/dash-core-components/src/components/DateSlider.tsx
```diff
@@ -0,0 +1,174 @@
...
+import {strAsDate, snapToStep} from '../utils/calendar/helpers';
+import dateRangeSlider from '../utils/LazyLoader/dateRangeSlider';
+import './css/datesliders.css';
+
+const RealSlider = lazy(dateRangeSlider);
+const narrowWindow = 500;
+
+/**
+ * A slider component for selecting a single date.
+ * This is a wrapper around DateRangeSlider that handles date values.
```
AnnMarieW, line 27:
```suggestion

```

Best to remove the comment about the wrapper since it doesn't add value when users see the docstring

components/dash-core-components/src/fragments/DateRangeSlider.tsx
```diff
@@ -0,0 +1,457 @@
...
+    snapToStep,
+} from '../utils/calendar/helpers';
+import {autoGenerateDateMarks} from '../utils/computeDateSliderMarkers';
+import rangeSlider from '../utils/LazyLoader/rangeSlider';
+
+const RealSlider = lazy(rangeSlider);
+
+/**
+ * Slider component for selecting a date.
+ * This is a wrapper around RangeSlider that handles date-to-timestamp conversions
```
AnnMarieW, line 36:
```suggestion
 *  handles date-to-timestamp conversions
```

components/dash-core-components/src/fragments/DateRangeSlider.tsx
```diff
@@ -0,0 +1,457 @@
...
+    useEffect(() => {
+        mappedValueRef.current = mappedValue;
+    }, [mappedValue]);
+
+    // Converts what comes back from the RangeSlider (timestamps) into date strings to show the user
+    const mappedSetProps: RangeSliderProps['setProps'] = useCallback(
+        newProps => {
+            const {value, drag_value} = newProps;
+            const mappedProps: Partial<DateRangeSliderProps> = omit(
+                ['min', 'max', 'step', 'value', 'drag_value', 'setProps'],
```
AnnMarieW, line 194: does `step_unit` need to be included here?

components/dash-core-components/src/types.ts
```diff
@@ -618,6 +618,248 @@ export interface RangeSliderProps extends BaseDccProps<RangeSliderProps> {
...
+export type DisableDatesFlag =
+    | 'weekends'
+    | 'weekdays'
+    | 'sundays'
+    | 'mondays'
+    | 'tuesdays'
+    | 'wednesdays'
+    | 'thursdays'
+    | 'fridays'
+    | 'saturdays';
```
AnnMarieW, line 634: this can be removed

components/dash-core-components/src/types.ts
```diff
@@ -618,6 +618,248 @@ export interface RangeSliderProps extends BaseDccProps<RangeSliderProps> {
...
+     * Unit by which steps are made
+     */
+    step_unit?: DateStepUnit | null;
+
+    /**
+     * Marks on the slider.
+     * The key determines the position (a string for date),
+     * and the value determines what will show.
+     * If you want to set the style of a specific mark point,
+     * the value should be an object which
```
AnnMarieW, line 664:
```suggestion
     * the value should be a dictionary which
```

components/dash-core-components/src/types.ts
```diff
@@ -618,6 +618,248 @@ export interface RangeSliderProps extends BaseDccProps<RangeSliderProps> {
...
+     * Unit by which steps are made
+     */
+    step_unit?: DateStepUnit | null;
+
+    /**
+     * Marks on the slider.
+     * The key determines the position (a number),
+     * and the value determines what will show.
+     * If you want to set the style of a specific mark point,
+     * the value should be an object which
```
AnnMarieW, line 769:
```suggestion
     * the value should be a dictionary which
```

components/dash-core-components/src/types.ts
```diff
@@ -618,6 +618,248 @@ export interface RangeSliderProps extends BaseDccProps<RangeSliderProps> {
...
+     */
+    display_format?: string;
+
+    /**
+     * If true, the selected range will automatically clamp to exclude any
+     * disabled dates when the range is expanded. The boundary closest to
+     * the disabled date will be adjusted to stop just before it.
+     * Requires `value` to have exactly two dates (a start and end).
+     */
+    no_disabled_in_between?: boolean;
```
AnnMarieW, line 860: remove prop

components/dash-core-components/tests/unit/calendar/helpers.test.ts
```diff
@@ -276,3 +279,147 @@ describe('parseYear', () => {
...
+    });
+
+    it('returns undefined for missing date or step', () => {
+        expect(stepDate(undefined, 1, 'days')).toBeUndefined();
+        expect(stepDate(baseDate, undefined, 'days')).toBeUndefined();
+        expect(stepDate(baseDate, undefined, undefined)).toBeUndefined();
+    });
+});
+
+describe('isDateDisabled', () => {
```
AnnMarieW, line 314: A few more disabled dates tests in this file

AnnMarieW:
Nice update!
It can be sad to remove big features like the disabled dates, but it will all be there in the git history  and will provide a nice head start for a future PR  :slightly_smiling_face: 

When using the `DateRangeSlider` in an app, I found the following issues  (need to check the `DateSlider` as well):


- if `step` is > 1 and the `step_unit` is not defined, it is not possible to move the handles with the keyboard. (`step_unit` should default to "days")
- if `step` > 1 and the last step is greater than `max` date, the handles get stuck when moving the handle with the keyboard.
Related to the above point, same result for any way the last step is greater than max date when using the keyboard, like `step_unit="months"`

- When `marks` are defined, but the marks are not between the min and the max, all the marks are displayed as "2T"
- With `marks` and `step=None` to restrict to marks, the keyboard navigation overrides the restriction
- With `marks` and `step=None` to restrict to marks, you can still select out of range values on the date pickers

- Need to include the `allowCross` prop in range slider

ines-om:
Hello @AnnMarieW,

Sorry for the silence on this PR. @Francisc0C is currently working and I have to finish pre-study modules before my master's begins next month. 
We expect to have more free time to pick this back up between August and September. Thanks so much for your patience, and sorry again for the delay!

PR #3805, opened 2026-06-02
fix fastapi middleware swallowing user routes bodies

### LINKED ISSUE #3801: [BUG] dash 4.2.0-Dash 4.3.0rc0 POST deadlock
'''
Minimal reproduction: Dash 4.2.0-4.3.0rc  + FastAPI POST deadlock

Start server:
    python dash_bug_server.py

Test from another terminal:
    # GET — works fine
    curl -s -m 3 http://127.0.0.1:9999/api/echo

    # POST — times out (BUG)
    curl -s -m 5 -X POST http://127.0.0.1:9999/api/echo \
        -H "Content-Type: application/json" \
        -d '{"hello":"world"}'

Expected: GET returns JSON immediately; POST hangs until timeout.
'''
```
import os
from dash import Dash, html
from fastapi import FastAPI, Request
from fastapi.responses import JSONResponse

fastapi_app = FastAPI()
dash_app = Dash(__name__, server=fastapi_app)
dash_app.layout = html.Div("Dash is running")

@fastapi_app.get("/api/echo")
async def echo_get():
    return JSONResponse({"method": "GET", "ok": True})

@fastapi_app.post("/api/echo")
async def echo_post(request: Request):
    body = await request.json()     # ← HANGS here
    return JSONResponse({"echo": body})

if __name__ == "__main__":
    dash_app.run(host="127.0.0.1", port=9999, debug=False)
```

T4rk1n:
Fixes #3801

### DIFF
--- a/dash/backends/_fastapi.py
+++ b/dash/backends/_fastapi.py
@@ -223,7 +223,14 @@ class DashMiddleware:  # pylint: disable=too-few-public-methods
             await self.app(scope, receive, send)
             return
 
-        # HTTP/WebSocket request handling
+        # Non-Dash routes pass through to avoid consuming body stream
+        path = scope["path"]
+        prefix = self.dash_app.config.routes_pathname_prefix
+        if "_dash-" not in path and path != prefix and path != prefix.rstrip("/"):
+            await self.app(scope, receive, send)
+            return
+
+        # HTTP request handling
         request = Request(scope, receive=receive)
         token = set_current_request(request)
 
--- a/dash/backends/_fastapi.py
+++ b/dash/backends/_fastapi.py
@@ -226,7 +226,12 @@ class DashMiddleware:  # pylint: disable=too-few-public-methods
         # Non-Dash routes pass through to avoid consuming body stream
         path = scope["path"]
         prefix = self.dash_app.config.routes_pathname_prefix
-        if "_dash-" not in path and path != prefix and path != prefix.rstrip("/"):
+        dash_prefix = prefix.rstrip("/") + "/_dash-"
+        if (
+            not path.startswith(dash_prefix)
+            and path != prefix
+            and path != prefix.rstrip("/")
+        ):
             await self.app(scope, receive, send)
             return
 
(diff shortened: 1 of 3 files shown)


dash/backends/_fastapi.py
```diff
@@ -223,7 +223,14 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             await self.app(scope, receive, send)
             return
 
-        # HTTP/WebSocket request handling
+        # Non-Dash routes pass through to avoid consuming body stream
+        path = scope["path"]
+        prefix = self.dash_app.config.routes_pathname_prefix
+        if "_dash-" not in path and path != prefix and path != prefix.rstrip("/"):
+            await self.app(scope, receive, send)
+            return
```
camdecoster, line 236: Returning here skips `_run_before_hooks` and `_run_after_hooks`. Is that intentional? You could move the check inside the try/except block and only call `_setup_timing` if it's a Dash route.

dash/backends/_fastapi.py
```diff
@@ -223,7 +223,14 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             await self.app(scope, receive, send)
             return
 
-        # HTTP/WebSocket request handling
+        # Non-Dash routes pass through to avoid consuming body stream
+        path = scope["path"]
+        prefix = self.dash_app.config.routes_pathname_prefix
+        if "_dash-" not in path and path != prefix and path != prefix.rstrip("/"):
```
camdecoster: This would match a path like `"_dash-my-awesome-path"`, which would be a false positive. It might be worth making this check a bit more specific. Though it seems like no one would actually do this.

dash/backends/_fastapi.py
T4rk1n: Yes intentional, those are only for emulating the flask api for the dash setup, not intended for the user to use.

dash/backends/_fastapi.py
T4rk1n: Yea better put the prefix and startwith check. Other it could catch some random route like `/my_dash-favorite`

camdecoster (approved):

Issue #3812, opened 2026-06-10
[BUG] `RuntimeError: No active request in context` on catchall route (FastAPI backend)


WheatMaThink:
### Summary

When using Dash with the FastAPI backend, accessing any non-Dash HTTP path (e.g. `/login`, `/about`, or any custom route) that falls through to the `catchall` route raises a `RuntimeError: No active request in context`, causing a 500 Internal Server Error.

### Affected Version

- **Dash version:** 4.2.0+ (including latest `main` branch as of 2026-06-11)
- **Python version:** 3.12
- **Backend:** FastAPI
- **OS:** Linux

### Description

The `DashMiddleware` in `dash/backends/_fastapi.py` only sets the request context (`set_current_request`) for Dash-specific routes (`_dash-*` paths and the root prefix). However, the `_setup_catchall()` method registers a `{path:path}` catch-all route that matches **all** GET paths and calls `dash_app.index()`, which internally requires a request context via `get_current_request()`.

When a non-Dash path is requested, the middleware skips setting the request context and passes through. The catchall route then tries to render `dash_app.index()` without a context, triggering the `RuntimeError`.

### Steps to Reproduce

1. Create a Dash app with FastAPI backend:

```python
from fastapi import FastAPI
from dash import Dash, html
from starlette.testclient import TestClient

fastapi_app = FastAPI()
dash_app = Dash(__name__, server=fastapi_app)
dash_app.layout = html.Div("Hello World")

client = TestClient(fastapi_app)
```

2. Access the root path — works correctly:

```python
resp = client.get("/")
assert resp.status_code == 200  # ✅ OK
```

3. Access any non-Dash path — crashes:

```python
resp = client.get("/login")
# ❌ RuntimeError: No active request in context
```

### Expected Behavior

The catchall route should successfully render the Dash index HTML for all paths, allowing Dash's client-side router to handle navigation (e.g., redirect to login page, show 404, etc.).

### Actual Behavior

```
RuntimeError: No active request in context

Traceback (most recent call last):
  File ".../fastapi/routing.py", line 134, in app
    await wrap_app_handling_exceptions(app, request)(scope, receive, send)
  ...
  File ".../dash/backends/_fastapi.py", line 333, in catchall
    return Response(content=dash_app.index(), media_type="text/html")
                            ^^^^^^^^^^^^^^^^
  File ".../dash/dash.py", line 1269, in index
    request = self.backend.request_adapter()
[...]

PR #3815, opened 2026-06-12
Fix missing request context (cookies/headers) in WebSocket callbacks

### LINKED ISSUE #3814: [BUG] WebSocket callbacks do not expose request cookies/headers on callback_context
**Describe your context**

```
dash                      4.2.0
dash-enterprise-auth      (current)
```

- Backend: FastAPI (Dash(__name__, backend="fastapi")); the same issue exists for the Quart backend.

**Describe the bug**

When a callback runs over the WebSocket transport (declared with `websocket=True`, or via `websocket_callbacks=True`), the request `cookies` and `headers` are **not** available on `dash.callback_context`. They are empty.

The HTTP transport populates these correctly (via `Dash._initialize_context`), but the WebSocket context builder (`create_ws_context` in `dash/backends/ws.py`) never sets `cookies`, `headers`, `args`, `path`, `remote`, or `origin`.

This breaks any authentication helper that reads the request context. In our case, running our apps on **Dash Enterprise**, `dash_enterprise_auth.get_user_data()` reads the `kcIdToken` cookie (or the `Plotly-User-Data` header on older Dash Enterprise) from `callback_context.cookies`. Over WebSocket it receives `{}`.

**Steps to reproduce**

```python
from dash import Dash, html, dcc, Input, Output, ctx

app.layout = html.Div([
    dcc.Input(id="q", type="text"),
    html.Div(id="out"),
])

@app.callback(Output("out", "children"), Input("q", "value"), websocket=True)
def show_cookies(value):
    # Over HTTP this is populated; over WebSocket it is empty.
    return f"cookies={dict(ctx.cookies)} headers={bool(ctx.headers)}"

if __name__ == "__main__":
    app.run(debug=True)
```

1. Run the app and open it in a browser (so the request carries cookies).
2. Type into the input to trigger the `websocket=True` callback.
3. Observe `ctx.cookies` is `{}` and `ctx.headers` is empty.
4. Change the callback to a normal HTTP callback (remove `websocket=True`) and observe the cookies/headers are present.

**Expected behavior**

`callback_context.cookies` and `callback_context.headers` (and `args`, `path`, `remote`, `origin`) should be populated for WebSocket callbacks just as they are for HTTP callbacks, so authentication/session helpers behave consistently across transports.

**Actual behavior**

`callback_context.cookies`/`headers` are empty for WebSocket callbacks, causing auth helpers such as `dash_enterprise_auth.get_user_data()` to fail.

**Root cause**

`create_ws_context` in `dash/backends/ws.py` builds the callback context but omits the request metadata that `Dash._initialize_context` sets for the HTTP path. The WebSocket handshake request (which carries the cookies/headers) is available in the backend WebSocket handlers (`_fastapi.py` / `_quart.py`) but is never forwarded into the context.

i-murray:
Propagate cookies, headers, args, path, remote, and origin from the WebSocket handshake onto the callback context in create_ws_context, mirroring the HTTP path. This fixes auth helpers (e.g. `dash_enterprise_auth.get_user_data`) failing over the WebSocket transport because callback_context.cookies/headers were empty.

Closes #3814 

## Contributor Checklist

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 - [#3805](https://github.com/plotly/dash/pull/3805) Fix FastAPI POST routes deadlock caused by middleware consuming request body. Fixes [#3801](https://github.com/plotly/dash/issues/3801).
+- [#3815](https://github.com/plotly/dash/pull/3815) Fix missing request context (cookies/headers) in websocket callbacks.
 
 ## [4.2.0] - 2026-06-01 - *The Freedom Update*
 
--- a/dash/_utils.py
+++ b/dash/_utils.py
@@ -218,6 +218,23 @@ def inputs_to_dict(inputs_list):
     return inputs
 
 
+def populate_request_metadata(g, adapter):
+    """Copy request metadata from a request adapter onto a context object.
+
+    Shared by the HTTP path (``Dash._initialize_context``) and the WebSocket
+    path (``dash.backends.ws.create_ws_context``) so that both transports expose
+    identical request context (cookies, headers, args, path, remote, origin) on
+    ``callback_context``.
+    """
+    g.cookies = dict(adapter.cookies)
+    g.headers = dict(adapter.headers)
+    g.args = adapter.args
+    g.path = adapter.full_path
+    g.remote = adapter.remote_addr
+    g.origin = adapter.origin
+    return g
+
+
 def convert_to_AttributeDict(nested_list):
     new_dict = []
     for i in nested_list:
--- a/dash/backends/_fastapi.py
+++ b/dash/backends/_fastapi.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from contextlib import contextmanager
 from contextvars import copy_context, ContextVar
 import asyncio
 import concurrent.futures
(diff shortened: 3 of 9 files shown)


T4rk1n:
We have another PR #3819 that uses set_context_request on the fastapi part for the same effect, I think we need both solution consolidated.

i-murray:
> We have another PR #3819 that uses set_context_request on the fastapi part for the same effect, I think we need both solution consolidated.

**Re: #3819**: It addresses a different bug. #3819 fixes when a non-Dash path falls through to the FastAPI catch-all route (fixes #3812) by wrapping the `index()` render in `set_current_request(request)`. It doesn't touch WebSocket callbacks, so it doesn't populate `callback_context.cookies` / `headers` / `args` / `path` / `remote` / `origin` for them, which is the gap this PR targets. The two changes are independent, so this one is still needed.

**Re: consolidating the mechanism**: I've refactored so WebSocket callbacks now reuse the same request-context path as HTTP callbacks instead of building a parallel dict:

- Extracted the metadata population into a shared `populate_request_metadata(g, adapter)` helper in _utils.py, used by both the HTTP path (`_initialize_context`) and the WebSocket path (`create_ws_context`).
- For FastAPI, the WebSocket handler now activates the request via `set_current_request(websocket)` / `reset_current_request(token)` inside the worker thread, and reads metadata through `FastAPIRequestAdapter`, exactly like the HTTP flow. (`ContextVars` don't propagate into the `ThreadPoolExecutor`, so activation happens via an `activate_request()` context-manager seam passed into `run_callback_in_executor`, which enters it inside the worker thread.)
- For Quart, it captures a request snapshot from the handshake and yields it through the same `activate_request()`.
- `callback_context.cookies` / `headers` / `args` / `path` / `remote` / `origin` now work identically in WebSocket and HTTP callbacks (so downstream helpers like `dash_enterprise_auth.get_user_data` work), with no duplicated population logic.

**Tests**: added unit coverage for the adapter/threading seam and end-to-end tests for both FastAPI and Quart asserting cookies/headers are visible inside a WebSocket callback.

BSd3v:
See my comment here:
https://github.com/plotly/dash/issues/3814#issuecomment-4813305506

robertclaus:
Given the feedback from the maintainers, we don't recommend this change. Please re-open if we misinterpreted anything!

PR #3817, opened 2026-06-14
Fix background callback context serialisation for non-dict request args

### LINKED ISSUE #3816: [BUG] Background callbacks fail to serialise request `args` on FastAPI/Quart backends (`QueryParams is not JSON serializable`)
# Background callbacks fail to serialise request `args` on FastAPI/Quart backends (`QueryParams is not JSON serializable`)

## Summary

When a `background=True` callback is dispatched on a non-Flask backend (e.g. the FastAPI or Quart adapters), Dash attempts to serialise the callback context to the background worker (Celery via kombu, or Diskcache). Serialisation fails because the request `args` stored on the context is a backend-specific multi-dict `starlette.datastructures.QueryParams` on FastAPI, which is **not JSON serialisable**.

The result is a runtime error such as:

```
TypeError: Object of type QueryParams is not JSON serializable
```

(or the equivalent kombu/`EncodeError` wrapping it) whenever a background callback is triggered. The same problem applies to the Quart adapter, whose `args` is a `werkzeug`/`quart` `MultiDict`.

## Environment

- **Dash version:** 4.2.0
- **Backend:** FastAPI (also affects Quart)
- **Background callback manager:** Celery (`CeleryManager`); Diskcache is likely affected for the same reason
- **Python:** 3.12

## Steps to reproduce

1. Configure a Dash app on the FastAPI backend with a Celery background callback manager.
2. Define any callback with `background=True`.
3. Trigger the callback from the browser.

```python
from dash import Dash, Input, Output, html
from dash.background_callback import CeleryManager
from celery import Celery

celery_app = Celery(__name__, broker=..., backend=...)
manager = CeleryManager(celery_app)

app = Dash(__name__, background_callback_manager=manager)  # FastAPI backend
app.layout = html.Div([
    html.Button("Run", id="run"),
    html.Div(id="out"),
])

@app.callback(
    Output("out", "children"),
    Input("run", "n_clicks"),
    background=True,
    prevent_initial_call=True,
)
def slow(n):
    return f"done {n}"
```

Clicking **Run** dispatches the job and raises the serialisation error.

## Expected behaviour

A background callback should be dispatched to the worker successfully, regardless of which backend (Flask, FastAPI, Quart) is in use. The request context handed to the worker must be JSON serialisable.

## Actual behaviour

Dispatch fails because the callback context contains a non-serialisable `args` object.

## Root cause analysis

During request setup, `Dash._initialize_context` stores the request data on the context. Unlike `cookies` and `headers`, which are explicitly coerced to plain dicts, `args` is stored **as-is**:

`dash/dash.py` (`_initialize_context`):

```python
g.cookies = dict(adapter.cookies)
g.headers = dict(adapter.headers)
g.args = adapter.args            # <-- not coerced to a plain dict
```

This is intentional for the synchronous HTTP path: several code paths rely on the multi-dict interface, e.g. `callback_context.args.getlist("cancelJob")` and `adapter.args.getlist("oldJob")`, which a plain `dict` does not provide.

On the FastAPI backend, `adapter.args` is the raw Starlette query params:

`dash/backends/_fastapi.py`:

```python
@property
def args(self):
    return self._request.query_params   # starlette.datastructures.QueryParams
```

For `background=True` callbacks, `_setup_background_callback` (`dash/_callback.py`) copies the context value and dispatches it to the worker via `callback_manager.call_job_fn(...)`. The background callback manager serialises this context (Celery → kombu → JSON), and serialisation fails on the `QueryParams` (FastAPI) / `MultiDict` (Quart) object.

On the Flask backend the issue does not surface in the same way because the worker receives a JSON-serialisable representation.

i-murray:
## Description

Fixes a serialisation failure that breaks **all** `background=True` callbacks on the FastAPI backend (and, for the same reason, Quart).

When a background callback is dispatched, Dash copies the callback context and hands it to the background worker (Celery via kombu, or Diskcache), where it must be serialised. The request `args` stored on the context is a backend-specific multi-dict `starlette.datastructures.QueryParams` on FastAPI, a `werkzeug`/`quart` `MultiDict` on Quart — which is **not JSON serialisable**, so dispatch fails with:

```
TypeError: Object of type QueryParams is not JSON serializable
```

### Root cause

During request setup, `Dash._initialize_context` (`dash/dash.py`) stores the request data on the context. Unlike `cookies` and `headers`, which are coerced to plain dicts, `args` is stored as-is:

```python
g.cookies = dict(adapter.cookies)
g.headers = dict(adapter.headers)
g.args = adapter.args            # backend-specific multi-dict, not coerced
```

This is intentional for the synchronous HTTP path, which relies on the multi-dict interface (e.g. `adapter.args.getlist("oldJob")` and `callback_context.args.getlist("cancelJob")`). The problem only appears when that object is carried into the background worker and serialised.

### Fix

`_setup_background_callback` (`dash/_callback.py`) builds a **copy** of the context for the worker. We coerce `args` to a plain `dict` **only on that copy**, leaving the live request context (and its `getlist` interface) untouched:

```python
ctx_value = AttributeDict(**context_value.get())   # copy handed to the worker
...
args_value = ctx_value.get("args")
if args_value is not None and not isinstance(args_value, dict):
    ctx_value["args"] = dict(args_value)
```

Because this mutates only the worker-bound copy:

- `adapter.args.getlist("oldJob")` (in `_get_callback_manager`) is unaffected, it reads the request adapter directly.
- `callback_context.args.getlist("cancelJob")` (the synchronous HTTP cancel path) is unaffected, it reads the live `context_value`, not the copy.
- The only consumer that sees the coerced `dict` is the background worker, which does not use `getlist` and otherwise could not receive the object at all.

Closes #3816

## Contributor Checklist

   -  [x] Coerce non-dict request `args` to a plain `dict` on the worker-bound context copy in `_setup_background_callback`
   -  [x] Add regression tests covering the serialisation behaviour

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [UNRELEASED]
 
 ### Fixed
+- [#3817](https://github.com/plotly/dash/pull/3817) Fix background callback context serialisation for non-dict request args on the FastAPI and Quart backends. Fixes [#3816](https://github.com/plotly/dash/issues/3816).
 - [#3805](https://github.com/plotly/dash/pull/3805) Fix FastAPI POST routes deadlock caused by middleware consuming request body. Fixes [#3801](https://github.com/plotly/dash/issues/3801).
 
 ## [4.2.0] - 2026-06-01 - *The Freedom Update*
--- a/dash/_callback.py
+++ b/dash/_callback.py
@@ -429,6 +429,10 @@ def _setup_background_callback(
     ctx_value.pop("background_callback_manager")
     ctx_value.pop("dash_response")
 
+    args_value = ctx_value.get("args")
+    if args_value is not None and not isinstance(args_value, dict):
+        ctx_value["args"] = dict(args_value)
+
     job = callback_manager.call_job_fn(
         cache_key,
         job_fn,
--- a/tests/unit/library/test_background_callback_serialisation.py
+++ b/tests/unit/library/test_background_callback_serialisation.py
@@ -1,0 +1,162 @@
+import json
+from unittest import mock
+
+import pytest
+
+from dash import _callback
+from dash._callback import _setup_background_callback, context_value
+from dash._utils import AttributeDict
+
+
+class FakeMultiDict:
+    """Minimal multi-dict standing in for a backend query-params object.
+
+    It mimics ``starlette.datastructures.QueryParams`` for these tests: it is
+    convertible to a ``dict`` and exposes ``getlist``, but is **not** a ``dict``
+    subclass and is **not** JSON serialisable.
+
+    Parameters
+    ----------
+    items : list of tuple
+        The ``(key, value)`` pairs held by the multi-dict.
+    """
+
+    def __init__(self, items):
+        self._items = list(items)
+
+    def getlist(self, key):
+        """Return all values stored under ``key``."""
+        return [value for stored_key, value in self._items if stored_key == key]
+
+    def keys(self):
+        """Return the keys, enabling ``dict(self)`` coercion."""
+        return [key for key, _ in self._items]
+
+    def __getitem__(self, key):
+        for stored_key, value in self._items:
+            if stored_key == key:
+                return value
+        raise KeyError(key)
+
+
+def _make_context(args_value):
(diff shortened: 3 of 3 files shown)


T4rk1n (approved):
💃

PR #3819, opened 2026-06-15
Bugfix: Fix no context in catchall route

### LINKED ISSUE #3812: [BUG] `RuntimeError: No active request in context` on catchall route (FastAPI backend)

KoolADE85:
Fixes a 500 `RuntimeError: No active request in context`  that breaks custom paths which fall through to the catch-all route.

The catch-all view already receives the request, but we ignored it. This PR registers it into the request context for the duration of the render, mirroring how we handle Dash routes themselves.

Fixes #3812 

## Contributor Checklist

### optionals


T4rk1n (approved):
💃

T4rk1n:
:confused:

PR #3824, opened 2026-06-17
Bugfix: Fix quart threadedrunner stop graceful shutdown

### LINKED ISSUE #3823: [BUG] dash.testing ThreadedRunner.stop() hangs at teardown for Quart backend apps
**Describe your context**

```
dash                      4.2.0
dash-testing-stub         0.0.2
```
- Python 3.12 (Windows 10)
- Python 3.14 (Windows 10, Ubuntu).

**Describe the bug**

When a Dash app uses the Quart backend (`Dash(__name__, backend="quart")`) and is served through `dash.testing`'s `ThreadedRunner` (the default integration-test runner, server in a worker thread), `ThreadedRunner.stop()` can hang at teardown — the test passes but the runner never returns.

`stop()` only has a graceful-shutdown branch for **FastAPI**, keyed on `hasattr(self._app, "_uvicorn_server")` (`dash/testing/application_runners.py`). A Quart app has no such attribute, so it falls into the `else` branch, which does:

The hypercorn server thread is parked in a blocking syscall (IOCP on Windows, epoll on POSIX), so the injected `SystemExit` isn't delivered until the loop next runs Python bytecode. The unbounded `join()` then blocks. Meanwhile the Quart backend's *own* cooperative shutdown switch — `backend._ws_shutdown_event`, awaited by `serve(shutdown_trigger=...)` — is never signalled, because only the main-thread signal handler sets it and the server is running in a worker thread.

**Minimal reproduction**:

```python
import threading
from dash import Dash, Input, Output, dcc, html
from dash.testing.application_runners import ThreadedRunner

app = Dash(__name__, backend="quart")
app.layout = html.Div([dcc.Input(id="in", value="x"), html.Div(id="out")])
app.callback(Output("out", "children"), Input("in", "value"))(lambda v: v)

runner = ThreadedRunner(stop_timeout=3)
runner.start(app, host="127.0.0.1")

done = threading.Event() # flag; set only if stop() actually returns

# run stop() on a daemon thread so a hang wedges that thread, not the script
threading.Thread(target=lambda: (runner.stop(), done.set()), daemon=True).start()

# wait up to 10s (> stop_timeout); False means stop() never returned = the hang
print("stop() returned" if done.wait(10) else "HANG: stop() did not return in 10s")
```

Output: `HANG: stop() did not return in 10s` (FastAPI or Flask print `stop() returned`).

**Expected behavior**

`ThreadedRunner.stop()` should shut a Quart app down cleanly and return within `stop_timeout`, the same as it does for FastAPI/Flask.

**Screenshots**

If applicable, add screenshots or screen recording to help explain your problem.

joschrag:
`dash.testing`'s `ThreadedRunner.stop()` only has a graceful-shutdown branch for FastAPI — it keys off `hasattr(self._app, "_uvicorn_server")`. A Quart app has no such attribute, so it falls through to the `else` branch: `thread.kill()` injects an asynchronous `SystemExit`, then `thread.join()` is called with no timeout.

the Quart backend already serves with `serve(shutdown_trigger=...)` awaiting `backend._ws_shutdown_event`; only the main-thread signal handler ever sets it, and in tests the server runs in a worker thread. `stop()` now sets that event itself — thread-safely, via `loop.call_soon_threadsafe(event.set)` on the server's own loop — then joins bounded by `stop_timeout`. 

Closes #3823 

## Contributor Checklist

   -  [x] Add a graceful Quart shutdown branch to ThreadedRunner.stop()
   -  [x] Add a regression test covering the teardown hang

### optionals


camdecoster:
Thanks for the PR! Our team will take a look and follow up.

dash/testing/application_runners.py
```diff
@@ -218,11 +218,24 @@ def run():
...
         if self._app and hasattr(self._app, "_uvicorn_server"):
-            server = self._app._uvicorn_server  # pylint: disable=protected-access
+            server = self._app._uvicorn_server
             server.should_exit = True
             self.thread.join(timeout=self.stop_timeout)  # type: ignore[reportOptionalMemberAccess]
+        # For Quart apps, signal hypercorn's cooperative shutdown event. Only the
+        # main-thread signal handler sets it, but in tests the server runs in a
+        # worker thread, so we set it ourselves -- thread-safely, on the server's
+        # own loop (the event binds its loop on first await) -- then join bounded.
+        elif quart_shutdown_event is not None:
```
T4rk1n: Could change these checks to be more clear and not dependent on private variables checks, we have the server_type attribute
```
server_type = getattr(self._app.backend, "server_type", "flask")
if server_type == "fastapi":
    ...
elif server_type == "quart":
    ...
else:
    ...
```

tests/backend_tests/test_threaded_runner_stop.py
```diff
@@ -0,0 +1,73 @@
+import threading
+import time
+
+import pytest
+
+from dash import Dash, Input, Output, dcc, html
+from dash.testing.application_runners import ThreadedRunner
+
+
+def test_quart_threaded_runner_stop_is_graceful_and_bounded():
```
T4rk1n: Think we could make this test generic and parametrize with the three backends?

joschrag: The test is now adjusted and covers all three backends.

T4rk1n (approved):
💃 Looks good thank you.

PR #3826, opened 2026-06-17
Add websocket_max_workers and run async callbacks directly on the loop

### LINKED ISSUE #3825: WebSocket callback concurrency is bounded by an unconfigurable thread pool
WS callbacks are dispatched to a `ThreadPoolExecutor` sized `min(32, cpu_count+4)` ([`base_server.py:194-209`](https://github.com/plotly/dash/blob/ffd8e96a88f07c84fc3bafa7f46cd452b45dd8a6/dash/backends/base_server.py#L194-L209)). Digging into performance implications, this can pose issues because it means WS callback scaling is:

1. Not configurable. Both call sites ([`_fastapi.py:735`](https://github.com/plotly/dash/blob/ffd8e96a88f07c84fc3bafa7f46cd452b45dd8a6/dash/backends/_fastapi.py#L735), [`_quart.py:563`](https://github.com/plotly/dash/blob/ffd8e96a88f07c84fc3bafa7f46cd452b45dd8a6/dash/backends/_quart.py#L563)) invoke `get_callback_executor()` with no `max_workers`, and no constructor param exposes it. Hitting the ~20-connection cap leaves "deploy more workers" as the only escape.
2. Async callbacks run one-per-thread.[`ws.py:435`](https://github.com/plotly/dash/blob/ffd8e96a88f07c84fc3bafa7f46cd452b45dd8a6/dash/backends/ws.py#L432-L437) calls `asyncio.run()` inside the worker thread, spinning a fresh event loop per callback. A `persistent=True` callback holds its thread (~8 MB stack) for the whole session while mostly `await`-ing

Users need to brute force scaling by adding more, and more, and more CPU cores, when one CPU core could handle many more threads than are currently allocated.

Proposed changes (independent):

- Quick win: expose `websocket_max_workers` on the `Dash` constructor, threaded to the two call sites via `getattr(dash_app, "_websocket_max_workers", None)`. ~5 lines, backward-compatible (`None` = current default). Lets users raise the cap past 32. This would scale memory linearly — 1000 threads ≈ 8 GB stacks.
- Durable fix: route `inspect.iscoroutine` results onto the backend's existing event loop as tasks; keep the thread pool only for sync callbacks. Takes async-callback concurrency from "threads × 8 MB" to "coroutines × KB."

Docs (https://dash.plotly.com/websocket-callbacks#connection-limits-and-scaling) accurately describe current behavior and would need updating.

T4rk1n:
- Add `websocket_max_workers` argument, default to 4.
- async callbacks runs directly on loop, sync uses the threadpool, warn if persistent callback is sync. fixes #3825.


dash/backends/_fastapi.py
```diff
@@ -725,16 +724,28 @@ async def websocket_handler(websocket: WebSocket):
...
+            # tasks and to resolve get_props futures.
+            loop = asyncio.get_running_loop()
+
             # Create janus queue for outbound messages (main loop context)
             outbound_queue: janus.Queue[str] = janus.Queue()
-            # Track pending get_props requests with standard queue.Queue for responses
-            pending_get_props: Dict[str, queue.Queue] = {}
+            # Track pending get_props requests. Values are queue.Queue (threadpool /
+            # sync path) or asyncio.Future (event-loop / async path).
+            pending_get_props: Dict[str, Any] = {}
```
camdecoster: Could this type be tightened up a bit?

dash/backends/_fastapi.py
```diff
@@ -725,16 +724,28 @@ async def websocket_handler(websocket: WebSocket):
 
             await websocket.accept()
 
+            # The connection's event loop, used to dispatch async callbacks as
+            # tasks and to resolve get_props futures.
+            loop = asyncio.get_running_loop()
```
camdecoster, line 734: What do you think of enabling debug mode for the loop here (and in a similar place for Quart)? You could use the existing debug config variable to turn this on. This way, users would get warning when a callback is executing slowly. You could also add a way for users to configure `loop.slow_callback_duration`.

camdecoster, line 734: Another option (maybe in tandem) would be to add a watchdog thread outside of the pool that could keep an eye out for bottlenecked callbacks.

dash/backends/_fastapi.py
```diff
@@ -725,16 +724,28 @@ async def websocket_handler(websocket: WebSocket):
...
-            # Get thread pool executor
-            executor = self.get_callback_executor()
-            # Track pending callback futures
-            pending_callbacks: Dict[str, concurrent.futures.Future] = {}
+            # Create a per-connection thread pool executor so that long-lived
+            # callbacks on one connection cannot starve worker threads for others.
+            # pylint: disable=protected-access
+            executor = self.create_callback_executor(
+                getattr(dash_app, "_websocket_max_workers", 4)
+            )
```
camdecoster: What do you think of moving this inside the `try` below (along with the assignment to `sender_task`? If an error occurs before the `try`, the executor is never shut down.

dash/backends/_fastapi.py
```diff
@@ -725,16 +724,28 @@ async def websocket_handler(websocket: WebSocket):
...
+            pending_get_props: Dict[str, Any] = {}
             # Shutdown event to signal connection closure to worker threads
             shutdown_event = threading.Event()
-            # Get thread pool executor
-            executor = self.get_callback_executor()
-            # Track pending callback futures
-            pending_callbacks: Dict[str, concurrent.futures.Future] = {}
+            # Create a per-connection thread pool executor so that long-lived
+            # callbacks on one connection cannot starve worker threads for others.
+            # pylint: disable=protected-access
```
camdecoster: Delete these stale pylint comments.

camdecoster (commented):
Are there any protections in place for the case of a server getting hammered and increasing the thread count continuously (when an app fires four sync callbacks right away)?

dash/backends/_fastapi.py
T4rk1n: Good idea, maybe as a follow up.

dash/backends/base_server.py
```diff
@@ -209,7 +215,7 @@ def get_callback_executor(
         return self._callback_executor
 
     def shutdown_executor(self, wait: bool = True) -> None:
```
camdecoster, line 217: Is this function needed anymore?

dash/backends/_fastapi.py
```diff
@@ -777,66 +789,78 @@ async def websocket_handler(websocket: WebSocket):
                             dash_app._websocket_callbacks,
                         )
 
-                        # Create WebSocket callback instance
+                        # Async callbacks (incl. persistent ones) run as tasks on
+                        # the event loop; sync callbacks go to the threadpool.
+                        # pylint: disable=protected-access
```
camdecoster, line 794: I think you can still remove these stale pylint comments.

camdecoster (approved):
This seems good overall. I think the debug addition would be good to add later, but this addresses the underlying issue.

PR #3830, opened 2026-06-18
Fix claude code connectivity timeouts

KoolADE85:
Recent versions of Claude Code have intermittent issues connecting to a Dash MCP server when run on localhost.
This PR adds a workaround that allows Claude to connect reliably.

Issue:
The MCP server previously responded to GET requests with a `405 Not Allowed`. While the spec does allow this, the dev server automatically closes connections with this response. Depending on the timing, Claude would attempt to reuse the already-closed connection which caused connectivity failures.

Solution:
The 405 response code is replaced with a 200 and an empty, completed stream which Claude accepts and correctly moves on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ## Fixed
 - [#3805](https://github.com/plotly/dash/pull/3805) Fix FastAPI POST routes deadlock caused by middleware consuming request body. Fixes [#3801](https://github.com/plotly/dash/issues/3801).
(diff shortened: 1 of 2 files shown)


ndrezn:
Is an e2e test connecting a Dash MCP to Claude Code in scope for our test suite?

T4rk1n (approved):
💃

KoolADE85:
> Is an e2e test connecting a Dash MCP to Claude Code in scope for our test suite?

No, we don't have any plumbing in place to connect Claude to the MCP server in an integration test. I think that will be a heavier lift.

Issue #3835, opened 2026-06-20
[Feature Request] Drop `mcp` dependency in Dash 4.3.0 to avoid excess dependencies


ghaarsma:
When upgrading from a clean 4.2 Dash (24 dependencies) to 4.3, it requires an additional 26 newly installed dependencies to perform the upgrade. See original install log and 4.2 -> 4.3 upgrade log below.

It seems that a lot of the new installs are related to the new MCP feature. Would it be possible to make this an optional install for the next release?

```
Resolving dependencies... (14.1s)

Package operations: 24 installs, 0 updates, 0 removals

  - Installing colorama (0.4.6)
  - Installing markupsafe (3.0.3)
  - Installing blinker (1.9.0)
  - Installing charset-normalizer (3.4.7)
  - Installing certifi (2026.6.17)
  - Installing click (8.4.1)
  - Installing idna (3.18)
  - Installing itsdangerous (2.2.0)
  - Installing jinja2 (3.1.6)
  - Installing narwhals (2.22.1)
  - Installing packaging (26.2)
  - Installing urllib3 (2.7.0)
  - Installing werkzeug (3.1.8)
  - Installing zipp (4.1.0)
  - Installing flask (3.1.3)
  - Installing plotly (6.8.0)
  - Installing typing-extensions (4.15.0)
  - Installing requests (2.34.2)
  - Installing setuptools (82.0.1)
  - Installing janus (2.0.0)
  - Installing retrying (1.4.2)
  - Installing importlib-metadata (9.0.0)
  - Installing nest-asyncio (1.6.0)
  - Installing dash (4.2.0)


(venv) PS C:\Users\gxvh\Python\dash_test5> poetry update
Updating dependencies
Resolving dependencies... (24.3s)

Package operations: 26 installs, 1 update, 0 removals

  - Installing attrs (26.1.0)
  - Installing pycparser (3.0)
  - Installing rpds-py (2026.5.1)
  - Installing annotated-types (0.7.0)
  - Installing h11 (0.16.0)
  - Installing cffi (2.0.0)
  - Installing anyio (4.14.0)
  - Installing pydantic-core (2.46.4)
  - Installing referencing (0.37.0)
  - Installing typing-inspection (0.4.2)
  - Installing cryptography (49.0.0)
  - Installing jsonschema-specifications (2025.9.1)
  - Installing pydantic (2.13.4)
  - Installing python-dotenv (1.2.2)
  - Installing starlette (1.3.1)
[...]

PR #3838, opened 2026-06-22
Drop mcp package dependency, inline protocol types

ndrezn:
## Summary

- The `mcp` package was imported only for its types — no server, transport, or client code was used
- Removes `mcp>=1.23.0` from `requirements/install.txt` (a core install dependency that every Dash user had to pull in)
- Adds `dash/mcp/types/protocol.py` with ~24 equivalent Pydantic v2 models matching the MCP wire format exactly
- Updates all import sites from `from mcp.types import ...` to `from dash.mcp.types import ...`
- Resource URI fields simplified from `AnyUrl` to `str` (dash:// URIs are already validated at the app layer)

## Test plan

- [x] Existing `tests/unit/mcp/` pass
- [x] Existing `tests/integration/mcp/` pass
- [x] Verify `pip install dash` no longer pulls in the `mcp` package


ndrezn:
This would address #3835

KoolADE85 (approved):

PR #3850, opened 2026-06-28
Feature: Integrate dash uploader

rakrsh:

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals


robertclaus:
@rakrsh thank you for looking into this! Adding an uploader component directly to Dash is an interesting idea. We took a quick look at the implementation in this PR and have some suggestions.

1. The dash-uploader code this is porting over hasn't been maintained in a while; so we would want to re-implement the functionality if we're introducing it as a new core component.
2. We're on the fence about whether a component this complex should be part of the Dash core repo vs a standalone package. I think we could be convinced either way, but a standalone package benefits from being easier to maintain.
3. There are some considerations for how this would integrate with Plotly's app hosting services on Plotly Cloud and Dash Enterprise; so if we do want to add it to the core libraries it might be easier for Plotly to develop given that additional context.

Since the PR has been sitting for a few weeks, we're also going to close it for now. Your branch should still be available, so feel free to re-open if you continue to work on this!

Issue #525, opened 2019-01-05
Creating a spectrum plot with callback


yaronogen:
Hello all, 
I have a csv file with 15 soil samples. each sample has some parameters like: OM and Clay and in addition it has a spectrum (starting from 400 until 750).
Im trying to create a dashboard with 3 sections: 
1. Plot of Clay vs. Sand
2. stats data
3. spectrum graph from 400 until 750.

and i would like the stats data and the spectrum graph to be updated automatically when i hover over a certain point in the first plot.
I was able to do so with the stats data, however, I was not able to do it with the spectrum.
I dont know if i was reading the data correctly or I had a problem with the callback.

Im new to python and dash so I would need your assistance
Thank you

import dash
import dash_core_components as dcc
import dash_html_components as html
import plotly
import plotly.plotly as py
import plotly.graph_objs as go
import plotly.offline as pyo
import pandas as pd
import cufflinks as cf
from dash.dependencies import Input, Output

app = dash.Dash()

df = pd.read_csv('test_GitHub.csv', index_col=None, na_values=['NA'],sep=',',low_memory=False)

app.layout = html.Div([ #global title
                html.Div([ #header
                        html.H1('Title'),
                        html.H3('Additional title')
                        ]),
                html.Div([
                    dcc.Graph(
                        id = 'test_scatter',
                        figure = {
                            'data': [go.Scatter(
                                x = df['Clay'],
                                y = df['Sand'],
                                text = df['Soil'],
                                hoverinfo = ['text','x','y'],
                                mode = 'markers'
                                )],
                            'layout': go.Layout(
                                title = 'Plot',
                                xaxis = {'title': 'Clay'},
                                yaxis = {'title': 'Sand'},
                                hovermode = 'closest'
                                )}),
                    dcc.Markdown(
                        id = 'stats',
                        ),
                    dcc.Graph(
                        id = 'spec',
                        figure = {
                            'data': [go.Scatter(
                                x = df['400':'750'],
                                y = df.iloc[soil_index]['400':'750'],
                                text = df['Soil'],
                                hoverinfo = ['text','x','y'],
                                mode = 'markers'
                                )],
                            'layout': go.Layout(
                                title = 'Spectrum',
                                xaxis = {'title': 'Wavelength (nm)'},
                                yaxis = {'title': 'Value'},
                                hovermode = 'closest'
                                )})
                     ],style = {'width': '50%', 'height': '50%', 'display': 'inline-block'})])

@app.callback(
[...]

PR #3880, opened 2026-07-06
Support React 19

T4rk1n:
Port PR #3646 (react19 branch) onto current dev, with fixes for the third-party component library issues reported there:

- Add React 19.2.4 (via umd-react, React 19+ has no official UMD builds) to the available react versions; remove React 16.
- New react-shim.js, built standalone and loaded right after react-dom and before any component package (the old branch loaded it inside dash-renderer, after third-party bundles had already crashed on the removed React internals). It stubs ReactCurrentOwner, redirects the legacy 'react.element' $$typeof symbol so libraries that pre-bundled a React <=18 jsx-runtime don't hit React error #525, and provides window.ReactJSXRuntime.
- Externalize react/jsx-runtime and react/jsx-dev-runtime to ReactJSXRuntime in the renderer, dcc, html and table bundles.
- Use umd-react@19.2.4 rather than 19.2.0, which silently bundles React 19.1.0.
- CI: run main, html and dcc test jobs against React 18.3.1 and 19.2.4; keep Percy pinned to 18.3.1.


AnnMarieW:
Hi @T4rk1n 
Nice to see this PR! Here's some early feedback from testing.

I ran a few minimal apps using this branch with React 19 enabled.

* The latest versions of DMC and DBC now render without errors, so the issues reported in  PR 3646 appear to be resolved.
* I tried a few components from DBC, and everything I tested worked correctly. (Comprehensive testing is still needed.)
* With the current DMC build, I found a number of issues, even though there are no console errors. Components that use popovers (Select, DatePicker, Menu, etc.) render their dropdowns in the top-left corner of the page. Slider responds to keyboard input but cannot be dragged with the mouse. Carousel does not change slides, and RichTextEditor crashes the browser.
* I rebuilt DMC with webpack changes similar to those in this PR (adding the react/jsx-runtime and react/jsx-dev-runtime externals), and those components worked correctly.

T4rk1n:
> * I rebuilt DMC with webpack changes similar to those in this PR (adding the react/jsx-runtime and react/jsx-dev-runtime externals), and those components worked correctly.

Yea this is a gotcha of the current shim, if you use features from jsx-runtime you need to put it in externals. Think might need to do the same in DBC and other libraries.

AnnMarieW:
> Yea this is a gotcha of the current shim, if you use features from jsx-runtime you need to put it in externals. Think might need to do the same in DBC and other libraries.

Another issue is that the new DMC build is not compatible with previous dash versions.     The app (running default React 18 and dash 4.3) doesn't render - and has this error message:
`Cannot read properties of undefined (reading 'jsx')`

T4rk1n:
> > Yea this is a gotcha of the current shim, if you use features from jsx-runtime you need to put it in externals. Think might need to do the same in DBC and other libraries.
> 
> Another issue is that the new DMC build is not compatible with previous dash versions. The app (running default React 18 and dash 4.3) doesn't render - and has this error message: `Cannot read properties of undefined (reading 'jsx')`

I made it backward compatible but need webpack config in libraries for the full support.

Can refer to dcc webpack config for how to adapt the JSXRuntime:

https://github.com/plotly/dash/blob/0548848db7cda6ca3e44045183c374423a52864d/components/dash-core-components/webpack.config.js#L10-L48

AnnMarieW:
@T4rk1n 
> I made it backward compatible but need webpack config in libraries for the full support.

I tried this in DMC and it worked great - Thanks!

components/dash-core-components/tests/integration/input/test_number_input.py
camdecoster, line 1: Could you let me know why these changes were made?

components/dash-core-components/webpack.config.js
```diff
@@ -7,6 +7,46 @@ const packagejson = require('./package.json');
 
 const dashLibraryName = packagejson.name.replace(/-/g, '_');
 
+// Externalized react/jsx-runtime. Newer Dash provides window.ReactJSXRuntime
+// (see dash-renderer/src/react-shim.js) backed by the React version loaded on
+// the page; the inline fallback keeps this bundle working on older Dash,
+// which does not define the global. Keep in sync with react-shim.js.
+const jsxRuntimeExternal = `var (window.ReactJSXRuntime || (window.ReactJSXRuntime = (function (React) {
```
camdecoster: What do you think of extracting this into a separate file and importing in all the places that use it?

camdecoster (approved):

PR #3883, opened 2026-07-07
Fix callbacks double-registered when app file is loaded by import string

### LINKED ISSUE #3818: [BUG] Callbacks double-registered when using uvicorn `--reload` under backend="fastapi"
**Context**

- Python 3.12.13
- uvicorn 0.49.0
- Linux (Docker: `python:3.12-slim`)
-  Package version:
```
dash                    4.2.0
```

**Description**

When using `Dash(backend="fastapi")` with uvicorn's `--reload` flag for development, **all callbacks are registered twice** after the first file-change reload. This produces `Duplicate callback outputs` and `Overlapping wildcard callback outputs` errors.

**Root cause:** `GLOBAL_CALLBACK_LIST` and `GLOBAL_CALLBACK_MAP` in `dash._callback` are module-level globals that persist across uvicorn's reload cycle. When uvicorn detects a file change, it calls `importlib.reload()` on the application module **within the same process**. This re-executes all `@callback` decorators, appending duplicates to the never-cleared global lists.

This did not happen with the Flask backend because Werkzeug's reloader spawns a **fresh subprocess** (clean `sys.modules`, clean globals). Uvicorn's reloader operates in-process via `importlib.reload()`.

Example:

This code below:

```python
import os
from dash import Dash, html, dcc, callback, Output, Input

app.layout = html.Div([
    html.H1("Reload Bug Demo"),
    dcc.Input(id="input", value="Hello"),
    html.Div(id="output"),
])

@callback(Output("output", "children"), Input("input", "value"))
def update(value):
    return f"You typed: {value}"
 
app.enable_dev_tools(debug=True, dev_tools_hot_reload=True)

server = app.server

if __name__ == "__main__":
    import uvicorn
    uvicorn.run("app:server", host="0.0.0.0", port=8050, reload=True)
```

Run with:

```ps
pip install dash[fastapi]==4.2.0
python app.py
```

Will result the following error:

```
In the callback for output(s):
  output.children
Output 0 (output.children) is already in use.
To resolve this, set `allow_duplicate=True` on duplicate outputs, or combine the outputs into one callback function, distinguishing the trigger by using `dash.callback_context` if necessary.
```

Currently I have to manually clearing the private globals before registering callbacks:

```python
from dash._callback import GLOBAL_CALLBACK_LIST, GLOBAL_CALLBACK_MAP
GLOBAL_CALLBACK_LIST.clear()
GLOBAL_CALLBACK_MAP.clear()
```

This works but relies on private internals (`dash._callback`) that could break in any future release.

T4rk1n:
When the app module runs as a script, a server that loads it by import string (e.g. uvicorn.run("app:server", reload=True)) executes the same file twice in the worker process: multiprocessing spawn re-runs it as __mp_main__, then the import string imports it again under its real name. Both passes run the `@callback` decorators, duplicating every spec in _dash-dependencies and triggering "Duplicate callback outputs" errors in the renderer.

Dash() now pre-registers the running main module in sys.modules under its canonical import name (only when that name resolves to the same file and isn't already imported), so the second import reuses the already-executed module instead of re-executing the file.

Fixes #3818

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+## Fixed
+- [](https://github.com/plotly/dash/pull/) Fix callbacks being registered twice when running the app file as a script with a server that loads it by import string, e.g. `uvicorn.run("app:server", reload=True)` with `backend="fastapi"`. The spawned worker re-executes the main module as `__mp_main__` and the import string then executed the same file a second time, duplicating every callback in `_dash-dependencies` and triggering `Duplicate callback outputs` errors in the renderer. `Dash()` now pre-registers the running main module in `sys.modules` under its canonical import name so the second import reuses it instead of re-executing the file. Fixes [#3818](https://github.com/plotly/dash/issues/3818).
+
 ## [4.4.0] - 2026-07-03
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 
 ## Fixed
-- [](https://github.com/plotly/dash/pull/) Fix callbacks being registered twice when running the app file as a script with a server that loads it by import string, e.g. `uvicorn.run("app:server", reload=True)` with `backend="fastapi"`. The spawned worker re-executes the main module as `__mp_main__` and the import string then executed the same file a second time, duplicating every callback in `_dash-dependencies` and triggering `Duplicate callback outputs` errors in the renderer. `Dash()` now pre-registers the running main module in `sys.modules` under its canonical import name so the second import reuses it instead of re-executing the file. Fixes [#3818](https://github.com/plotly/dash/issues/3818).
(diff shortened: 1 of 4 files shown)


dash/dash.py
```diff
@@ -154,6 +154,37 @@
...
+    try:
+        spec = find_spec(import_name)
+        if (
+            spec is not None
+            and spec.origin is not None
+            and os.path.samefile(spec.origin, module_file)
+        ):
+            sys.modules[import_name] = module
+    except (ImportError, ValueError, OSError):
+        pass
```
camdecoster: Should a warning be logged here?

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+## Fixed
+- [](https://github.com/plotly/dash/pull/) Fix callbacks being registered twice when running the app file as a script with a server that loads it by import string, e.g. `uvicorn.run("app:server", reload=True)` with `backend="fastapi"`. The spawned worker re-executes the main module as `__mp_main__` and the import string then executed the same file a second time, duplicating every callback in `_dash-dependencies` and triggering `Duplicate callback outputs` errors in the renderer. `Dash()` now pre-registers the running main module in `sys.modules` under its canonical import name so the second import reuses it instead of re-executing the file. Fixes [#3818](https://github.com/plotly/dash/issues/3818).
```
camdecoster: This entry is pretty verbose. I'd suggest trimming it a bit.

```suggestion
- [3883](https://github.com/plotly/dash/pull/3883) Fix callbacks being registered twice when running the app file as a script with a server that loads it by import string, e.g. `uvicorn.run("app:server", reload=True)` with `backend="fastapi"`. The spawned worker re-executes the main module as `__mp_main__` and the import string then executed the same file a second time, duplicating every callback in `_dash-dependencies` and triggering `Duplicate callback outputs` errors in the renderer. `Dash()` now pre-registers the running main module in `sys.modules` under its canonical import name so the second import reuses it instead of re-executing the file. Fixes [#3818](https://github.com/plotly/dash/issues/3818).
```

camdecoster (approved):
I left a couple of suggestions. The CHANGELOG entry is missing the PR number.

dash/dash.py
T4rk1n: I added a comment explaining why it's needed.

PR #3888, opened 2026-07-09
Streaming callbacks + shared storage (multiplexed streaming transport)

T4rk1n:
## Summary

This PR adds **streaming callbacks** and, to support them well across worker processes, a new backend-agnostic **shared storage** primitive.

- **Streaming callbacks** — decorate an `async def` generator; each `yield` is pushed to the browser as it is produced (same shape as a normal return; `dash.Patch` yields apply incrementally, e.g. LLM token streaming). No opt-in keyword — a generator streams by definition. Sync generators are rejected at registration.
- **Shared storage** (`dash.ctx.shared_storage` / `app.shared_storage`) — a cross-process key/value store + ordered publish/subscribe, usable by apps for cross-callback/session state, and used internally as the broker for streaming.
- **Multiplexed HTTP streaming** — all of a page's streams share **one** downlink connection instead of one per callback, so they no longer hit the browser's ~6-connections-per-host ceiling. Works on Flask, Quart, and FastAPI, no WebSocket required.

## Streaming callbacks

```python
@callback(Output("out", "children"), Input("go", "n_clicks"))
async def stream(n):
    async for token in llm.stream(prompt):
        patch = Patch()
        patch += token
        yield patch
```

## Shared storage

```python
app = Dash(__name__)                     # shared_storage=LocalSharedStorage by default
# inside a callback:
dash.ctx.shared_storage.set("k", value)  # cross-process KV (JSON-compatible values)
dash.ctx.shared_storage.publish("topic", msg)
```

The default `LocalSharedStorage` elects a single owner process per machine (AF_UNIX socket on POSIX, TCP loopback on Windows — the bind is the lease, re-elected on owner death) and serves the others. A single-process deployment is its own owner and pays no socket overhead. Subscriptions are ordered and replayable: a reconnecting consumer resumes from its last-seen sequence out of a bounded buffer, and a buffer overrun surfaces as an explicit gap rather than a silent loss. Wire codec is **msgspec (never pickle)**; pass `shared_storage=None` to disable, or a `BaseSharedStorage` subclass/instance to swap the backend.

## Architecture — multiplexed streaming

A streaming callback no longer holds its own HTTP connection. Its POST returns a fast ack; its frames are pumped onto a shared-storage topic and relayed over the page's single downlink, routed back to the right callback by `requestId`. Because the frames travel through shared storage, **the worker that runs a callback and the worker that holds the downlink need not be the same process** — shared storage is the broker.

```mermaid
flowchart TB
    subgraph browser["Browser — one page"]
        cbs["streaming callbacks<br/>(async def generators)"]
        sc["StreamClient<br/>single downlink per page"]
        cbs --> sc
    end

    subgraph server["Dash server — any number of worker processes"]
        wa["worker A<br/>runs callback, pumps frames"]
        wb["worker B<br/>serves the downlink"]
    end

    store[("Shared storage owner process<br/>KV + ordered pub/sub<br/>topic per connection")]

    sc -->|"1 · uplink POST, fast ack<br/>streamConnection = conn + requestId"| wa
    wa -->|"2 · publish frames, tagged requestId + seq"| store
    sc -->|"3 · single downlink<br/>streamDownlink = conn, from = seq"| wb
    store -->|"4 · subscribe, replay from seq"| wb
    wb -->|"5 · NDJSON frames"| sc
    sc -->|"6 · route by requestId, apply"| cbs
```

**Transport selection** (renderer): a streaming callback rides the **WebSocket** transport when websocket callbacks are enabled; otherwise the **multiplexed HTTP** transport when shared storage is available; otherwise falls back to today's **one NDJSON connection per callback**. So nothing changes for apps that don't opt into shared storage.

**Scheduler:** long-lived streams no longer consume the renderer's concurrent-request budget, and clientside callbacks are exempt from it — a page full of streams no longer starves other callbacks.

## Testing

- Shared storage: engine semantics + **cross-process** KV, pub/sub, reconnect-replay, owner re-election.
- Streaming transport: uplink/downlink end-to-end over real HTTP on **Flask, Quart, and FastAPI**.
- Renderer: `StreamClient` routing, multiplexing, reconnect-from-cursor, keepalive (karma).
- **Browser validation** (Selenium): the streaming integration suite runs over the multiplexed transport, plus an 8-stream demo where all streams and a clientside clock run simultaneously with a single downlink.

## Notes

- Adds `msgspec` to install requirements (wire codec; JSON fallback).
- Shared-storage values must be JSON-compatible (like `dcc.Store`).
- The in-memory backend is not durable: if the owner process dies, a survivor re-elects with an empty store (durable backends can follow).

## Follow-up (not in this PR)

- Host the downlink in a **SharedWorker** so it's one connection per *browser* (shared across tabs) rather than per page. `StreamClient` is written host-agnostic for this; the per-page transport already solves the connection-limit problem.


dash/_callback.py
```diff
@@ -878,7 +914,219 @@ async def async_add_context(*args, **kwargs):
...
+        stream = _kwargs.get("stream", False)
+        is_gen_func = inspect.isgeneratorfunction(func)
+        is_async_gen_func = inspect.isasyncgenfunction(func)
+        if stream:
+            if not (is_gen_func or is_async_gen_func):
+                raise StreamCallbackError(
+                    f"stream=True callback '{callback_id}' must be a generator "
+                    "function (or async generator function) that yields output "
+                    "updates."
+                )
```
KoolADE85: This looks like two "sources of truth" that must agree with each other (`stream=True` and `async`/`generator`).
What do you think about removing the `stream` flag and just detect streaming automatically based on `isasyncgenfunction`/ `isgeneratorfunction`? That would allow devs to just start yielding values without needing to manage flags or understand our API (because there would be no API at all).

KoolADE85 (changes requested):
The implementation here will easily starve the server of resources.
Run the sample app with `gunicorn -w 1` and observe:
1. It will terminate streams that take longer than 30s (by default)
2. The number of simultaneous requests is limited to the number of workers you spawn. After that limit, the server becomes unresponsive and/or terminates worker threads.

dash/_callback.py
T4rk1n: Good catch, only need a yield, stream keyword is redundant.

T4rk1n:
> The implementation here will easily starve the server of resources. Run the sample app with `gunicorn -w 1` and observe:
> 
>     1. It will terminate streams that take longer than 30s (by default)
> 
>     2. The number of simultaneous requests is limited to the number of workers you spawn. After that limit, the server becomes unresponsive and/or terminates worker threads.

With a sync flask backend, it's not really possible to fix at the app level. Need to run with `--worker-class gthread --threads 8 --timeout 0` for a true fix. Will add support in cloud and DE.

There is also a warning recommending a different backend when streaming, but maybe we should not allow it?

chgiesse:
@T4rk1n I solved that issue with my event streaming package by limiting the event stream to the normal 30 sec server timeout. With that, the stream stays in the scope of a normal http request while still providing its advantages.

T4rk1n:
> @T4rk1n I solved that issue with my event streaming package by limiting the event stream to the normal 30 sec server timeout. With that, the stream stays in the scope of a normal http request while still providing its advantages.

This is too limiting for LLM purposes, conversations can last a long time.

It's entirely out of the application scope since we now support multiple backends and there is a big difference in how uvicorn and gunicorn handle this by default:

- gunicorn: default is a flat 30s timeout, no reset on messages. 
- uvicorn: 60s seconds resets on every message, a keep alive pattern is all you need.

Then there is also reverse proxy (nginx & such) settings that might get in the way.

We'll configure plotly-cloud/enterprise to accept the streaming properly without user configuration and have some documentation for flask apps and gunicorn.

dash/_callback.py
```diff
@@ -935,7 +979,205 @@ async def async_add_context(*args, **kwargs):
...
+            callback_map[callback_id]["stream"] = True
+            if is_gen_func:
+                # A sync generator stream occupies a server worker (or WS
+                # executor thread) for the whole stream duration; recommend
+                # async so it runs on the event loop.
+                warnings.warn(
+                    f"Streaming callback '{callback_id}' is a synchronous "
+                    "generator; it will occupy a server worker for the whole "
+                    "stream. Define it with 'async def' so it runs on the "
+                    "event loop instead.",
```
KoolADE85: I think this warning understates the potential problem.
```suggestion
                warnings.warn(
                    f"Callback '{callback_id}' is a synchronous generator; "
                    "it can slow down or stall the entire app under heavy load. "
                    "Consider 'async def' instead for better performance.",
```

dash/backends/ws.py
```diff
@@ -506,6 +512,105 @@ def on_done(f: "concurrent.futures.Future | asyncio.Future") -> None:
...
+    if marker.is_async:
+        # Defensive: an async generator that ended up on the thread path is
+        # driven on a private event-loop thread.
+        frames = sync_iter_asyncgen(marker.frames)
+    else:
+        frames = iter_stream_frames(marker)
+    try:
+        for frame in frames:
+            if ws_callback.is_shutdown:
+                return {"status": "prevent_update"}
```
KoolADE85, line 583: 2 bugs around a similar root cause here (also applies to the async version below):

1. hot-reloading is broken while any stream is alive
2. streams are left running orphaned after a server shutdown

Here we handle client disconnect - maybe it's also a good place to handle server shutdowns?

chgiesse:
> > @T4rk1n I solved that issue with my event streaming package by limiting the event stream to the normal 30 sec server timeout. With that, the stream stays in the scope of a normal http request while still providing its advantages.
> 
> 
> 
> This is too limiting for LLM purposes, conversations can last a long time.
> 
> 
> 
> It's entirely out of the application scope since we now support multiple backends and there is a big difference in how uvicorn and gunicorn handle this by default:
> 
> 
> 
> - gunicorn: default is a flat 30s timeout, no reset on messages. 
> 
> - uvicorn: 60s seconds resets on every message, a keep alive pattern is all you need.
> 
> 
> 
> Then there is also reverse proxy (nginx & such) settings that might get in the way.
> 
> 
> 
> We'll configure plotly-cloud/enterprise to accept the streaming properly without user configuration and have some documentation for flask apps and gunicorn.

Yes that's true. Long-running streams wouldn't be possible. But having the option to progressively push ui updates from a regular callback is pretty handy I think. And you wouldnt have to rely on the server timeout. You can wrap the generator in its own timer.

PR #3897, opened 2026-07-13
Close feature gaps between clientside_callback and regular callbacks

T4rk1n:
Add support to `clientside_callback` for features that previously only worked with server-side `@callback`:

- Flexible/grouped signatures: `output=/inputs=/state=` may be dicts or nested lists. Dict groupings pass a single destructurable object to the JS function; the return value is validated against the output grouping. Grouping indices are serialized into the callback spec only when non-trivial, so flat signatures keep identical specs.
- `running=`: on/off side updates applied around execution, reverting on success, `PreventUpdate`, and error.
- `on_error=`: JS error handler (source string or `ClientsideFunction`) whose return value becomes the outputs (undefined -> no_update).
- `callback_context` parity: `triggered_prop_ids`, `args_grouping`, `outputs_grouping`, `using_args_grouping`, `using_outputs_grouping`.
- `optional=` passthrough.

`clientside_callback` now raises `CallbackException` for unsupported kwargs (background=, progress=, cancel=, manager=, ...) instead of silently ignoring them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](https://semver.org/).
 
+## Unreleased
+
+### Added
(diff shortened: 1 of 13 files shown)


KoolADE85 (approved):
Tested each of the features and all are working well here. 💃

PR #3902, opened 2026-07-17
callback args hardening.

T4rk1n:
Fix background callbacks trusting the client-supplied `job`/`oldJob`/`cancelJob` and `cacheKey` query parameters verbatim, which let an unauthenticated client terminate an arbitrary process (with `DiskcacheManager`, by sending its PID) or read and delete arbitrary result-cache entries. The `cacheKey`/`job` handles are now HMAC-signed by the server and bound to a per-page-load token, so forged or replayed values are rejected. Handles stay opaque to the renderer, so no app or callback code changes are required.


camdecoster (approved):
This seems fine, but I think @KoolADE85 should give it look before you merge.

KoolADE85 (approved):
I tested this locally, and it's working well for me.

PR #3915, opened 2026-07-22
Bugfix: MCP server returns error when chrome not available for kaleido

### LINKED ISSUE #3914: MCP: server responds with error data when using kaleido
If Kaleido is installed but chrome is unavailable, Dash's MCP server will respond with a Python error when attempting to render plotly graphs.
The actual data is returned as expected, but is accompanied with the error message that agents typically surface to the user:

Image

`MCP error -32603: RuntimeError: Chromium.get_cli() received invalid args: dict_keys(['headers'])`

KoolADE85:
Fixes https://github.com/plotly/dash/issues/3914


T4rk1n (approved):
💃

camdecoster (approved):

Issue #3464, opened 2025-09-30
[Feature Request] Allow dcc.Upload to Ingest a Folder


pw-jjeffers:
**Is your feature request related to a problem? Please describe.**
Currently my users have to individually upload files from multiple subfolders, increasing the work and potential for errors.

**Describe the solution you'd like**
I would like my users to be able to drag a folder into the drop zone and upload all of the files, using the same output API as it does with multiple files. This is supported by react-dropzone, the react component that powers dcc.Upload, as you can see by dragging a folder to the example here in there site: https://react-dropzone.js.org/#section-basic-example.

**Describe alternatives you've considered**
Rewrapping react-dropzone as new dash component but that creates a maintenance overhead that should not be necessary since it is already the component that powers dcc.Upload. I will also in the meantime have multiple dropzones to upload files, but this is less than ideal for normal usage.

**Additional context**
This was a request on the previous dcc repo: https://github.com/plotly/dash-core-components/issues/1003

PR #3916, opened 2026-07-24
Fix multiple file drag-and-drop in dcc.Upload

### LINKED ISSUE #3671: [BUG] Dash upload components behaves differently on multiple file upload via drag & drop and file selector
**Ccontext**
A basic minimalistic dash app with an upload component that support multiple files upload of type ".pdf" on windows 11, but developing with wsl ubuntu 22

```
dash                 4.0.0
dash-core-components 2.0.4
dash-html-components 2.0.0
dash-table           5.0.0
```

    - OS: Windows
    - Browser: Chrome and Edge
    - Version 144 and 146

**Describe the bug**

When dragging multiple files and dropping them on the upload component, only the first one triggers the callback.

**Expected behavior**

Both files to be sent in the callback in the backend

AnnMarieW:
closes #3671 

Fixes a regression introduced in #3464

Fix drag-and-drop of multiple files in `dcc.Upload`. Collect all `DataTransferItem` entries before the first async operation to avoid losing items during drag event processing.

I'm not sure how to do a test for this, but when I run it locally, this fixes the multiple file drag/drop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 - [#3646](https://github.com/plotly/dash/pull/3646) Remove React 16 support (`16.14.0` is no longer an accepted value for `REACT_VERSION` / `_set_react_version`).
 
+### Fixed
+- [#3916](https://github.com/plotly/dash/pull/3916) Fixed a regression where dragging multiple files into `dcc.Upload` would upload only the first file when `multiple=True`
+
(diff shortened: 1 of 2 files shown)


KoolADE85 (commented):
Looking good! I tested on Firefox and Chrome and it works well. Just needs a CHANGELOG entry.

AnnMarieW:
@KoolADE85   Thanks for the review!  I added the changlog entry.

KoolADE85 (approved):

PR #3922, opened 2026-07-28
Fix number input spinner when only min is set

### LINKED ISSUE #3921: dcc.Input(type="number", min=...) erases value on spinner step (regression of #3689)
Setting `min` on a number `dcc.Input` makes the native spinner (or arrow keys) erase the value — it gets cast to `None`. This is the same symptom as #3689, which was fixed and closed, but it reproduces again on dash 4.4.1.

### Reproduction

```python

app = Dash()
app.layout = html.Div([
    dcc.Input(type="number", value=40, min=0),    # spinner/arrows blank the value
    dcc.Input(type="number", value=40, step=1),    # works fine
])

if __name__ == "__main__":
    app.run(debug=True)
```

Click the up/down spinner (or focus + arrow keys) on the first input → the value clears. The second input is unaffected.

### Notes

- Only `min` triggers it; `step` and `max` alone are fine.
- Browser/spinner-dependent — did not reproduce in headless Chromium.

### Environment

- dash 4.4.1, Python 3.12

AnnMarieW:
Closes #3921 

Fix `dcc.Input(type="number")` stepper  behavior when only `min` is set.

Fixed typo in number input stepper constraint check from  PR #3690 

Also I noticed that `onKeyPress` is deprecated and replaced it with  `onKeyDown`

### DIFF
--- a/components/dash-core-components/tests/integration/input/test_number_input.py
+++ b/components/dash-core-components/tests/integration/input/test_number_input.py
@@ -283,4 +283,3 @@ def test_inni011_min_max_bug(dash_dcc, min):
     dash_dcc.wait_for_text_to_equal("#output", "16")
 
     assert dash_dcc.get_logs() == []
-
--- a/components/dash-core-components/tests/integration/input/test_number_input.py
+++ b/components/dash-core-components/tests/integration/input/test_number_input.py
@@ -258,7 +258,7 @@ def test_inni010_valid_numbers(dash_dcc, ninput_app):
     assert dash_dcc.get_logs() == []
 
 
-@pytest.mark.parametrize("min", [None, 1])
+@pytest.mark.parametrize("min", [None, 0, 1])
 def test_inni011_min_max_bug(dash_dcc, min):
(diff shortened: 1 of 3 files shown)


components/dash-core-components/tests/integration/input/test_number_input.py
```diff
@@ -258,13 +258,14 @@ def test_inni010_valid_numbers(dash_dcc, ninput_app):
     assert dash_dcc.get_logs() == []
 
 
-def test_inni011_min_max_bug(dash_dcc):
-    """Test that decrement increment button works correctly with min/max set to None."""
+@pytest.mark.parametrize("min", [None, 1])
```
KoolADE85: I think 0 would be a useful test case here too!
```suggestion
@pytest.mark.parametrize("min", [None, 0, 1])
```

KoolADE85 (approved):
Thanks AnnMarie, looks good here 💃

AnnMarieW:
@KoolADE85 I made one more change after you reviewed - to replace the deprecated `onKeyPress`.  Is that ok?

T4rk1n (approved):

PR #3923, opened 2026-07-29
fix(callbacks): use . for dots in dict ID callback IDs to fix JSON.parse SyntaxError

### LINKED ISSUE #3480: [BUG] Error when using . in dict ids
**Describe your context**

```
dash                             3.2.0
dash_ag_grid                     31.3.1
dash_auth                        2.3.0
dash-bootstrap-components        2.0.4
dash-bootstrap-templates         2.1.0
dash-extensions                  2.0.4
dash-iconify                     0.1.2
```

    - OS: iOS
    - Browser chrome
    - Version 141.0.7390.66 (Official Build) (arm64)
   
**Describe the bug**

Trying to use dict id in a callback with a "." character in it leads to an error.
**Expected behavior**

It does not error.
**MWE**
Opening this page leads to a blank page, with error in the console.
```
from dash import Dash, Input, Output, html, callback

layout = [
    html.Button("Button", id={"component": "button.dot"}),
]


@callback(
    Output({"component": "button.dot"}, "children"),
    Input({"component": "button.dot"}, "n_clicks"),
    prevent_initial_call=True,
)
def test(n):
    return "test"


app = Dash(__name__)
app.layout = layout
app.run()
```
Error:
```
hook.js:608 SyntaxError: Bad escaped character in JSON at position 21 (line 1 column 22)
    at JSON.parse (<anonymous>)
    at dash_renderer.v3_2_0m1756799785.min.js:2:75910
    at $n (dash_renderer.v3_2_0m1756799785.min.js:2:75920)
    at Hn (dash_renderer.v3_2_0m1756799785.min.js:2:75726)
    at dash_renderer.v3_2_0m1756799785.min.js:2:77864
    at Pe (dash_renderer.v3_2_0m1756799785.min.js:2:54296)
    at dash_renderer.v3_2_0m1756799785.min.js:2:55241
    at dash_renderer.v3_2_0m1756799785.min.js:2:52978
    at e (dash_renderer.v3_2_0m1756799785.min.js:2:44876)
    at dash_renderer.v3_2_0m1756799785.min.js:2:77829
overrideMethod @ hook.js:608
react-dom@18.v3_2_0m1756799785.3.1.min.js:79 Uncaught SyntaxError: Bad escaped character in JSON at position 21 (line 1 column 22)
    at JSON.parse (<anonymous>)
    at dash_renderer.v3_2_0m1756799785.min.js:2:75910
    at $n (dash_renderer.v3_2_0m1756799785.min.js:2:75920)
    at Hn (dash_renderer.v3_2_0m1756799785.min.js:2:75726)
    at dash_renderer.v3_2_0m1756799785.min.js:2:77864
    at Pe (dash_renderer.v3_2_0m1756799785.min.js:2:54296)
    at dash_renderer.v3_2_0m1756799785.min.js:2:55241
    at dash_renderer.v3_2_0m1756799785.min.js:2:52978
    at e (dash_renderer.v3_2_0m1756799785.min.js:2:44876)
    at dash_renderer.v3_2_0m1756799785.min.js:2:77829
```

Mukller:
## Problem

Fixes #3480.

When a dict component ID contains a dot in one of its string values (e.g.
id={"component": "button.dot"}), opening the app results in a blank page and
SyntaxError: Bad escaped character in JSON in the browser console.

### Root cause

In _utils.py, create_callback_id calls x.component_id_str().replace(".", "\.")
on every component ID. For string IDs this is fine, but dict IDs are serialized
as JSON. After the replacement, a value like "button.dot" becomes "button\.dot"
inside the JSON fragment. The frontend then calls JSON.parse on this fragment
(in parsePMCId and callbacks.ts) and encounters \. which is NOT a valid JSON
escape sequence -> SyntaxError.

### Fix

For dict IDs, replace . with . (the standard JSON Unicode escape for
U+002E FULL STOP) instead of \.:

- . is valid inside a JSON string.
- JSON.parse transparently decodes it back to ., so the frontend receives
  the original dict value unchanged.
- No changes are needed on the JavaScript side.

### Before / After

component_id = {"component": "button.dot"}, property = "children"

Before (broken): {"component":"button\.dot"}.children
  -> JSON.parse throws SyntaxError

After (fixed): {"component":"button.dot"}.children
  -> JSON.parse returns {"component": "button.dot"} correctly

Closes #3480.


Mukller (commented):
## Self-Review

### Root Cause

 escapes dots in component IDs with  so they don't collide
with the  separator between component-id and property in the callback ID string.
This is correct for **string IDs** (the JS side un-escapes  when splitting).

For **dict IDs**,  returns valid JSON (e.g.
). After , the dot inside
the JSON string value becomes  — an **invalid JSON escape sequence** — and
 throws .

### Why  is the right fix

 is a standard JSON Unicode escape (U+002E FULL STOP). It is
valid inside JSON strings and  decodes it back to  automatically.
The frontend receives the original dict value unchanged. No JS changes are needed.

### Correctness matrix

| ID type | Value | Callback ID (before) | Callback ID (after) |
|---|---|---|---|
| string |  |  | unchanged |
| dict |  |  — BROKEN |  ✓ |
| dict multi-dot |  |  — BROKEN |  ✓ |
| dict, no dot |  | unchanged | unchanged |

### What does NOT change

-  — unchanged; still used for runtime input matching
- Pattern-matching wildcards (//) — no dots → unaffected  
- Multi-output format () — structural separators unaffected
-  — still ; callers that parse the JSON portion
  use  which correctly decodes  → 
- No JavaScript changes required —  handles  natively

camdecoster:
Thanks for the PR! Our team will review and follow up.

T4rk1n (changes requested):
Can you add a test for this?

Mukller:
Added tests in tests/unit/test_callback_unit.py:

- test_create_callback_id_escapes_dots_in_string_id — confirms string ids keep the existing `\.` escaping
- test_create_callback_id_escapes_dots_in_dict_id_as_json_unicode — confirms dict ids with dots use `\u002e` and that the id portion round-trips through json.loads() back to the original dict
- test_create_callback_id_dict_id_without_dots_unaffected — sanity check for dict ids with no dots

Also pushed a branch update to pick up latest main.

tests/unit/test_callback_unit.py
```diff
@@ -1,128 +1,165 @@
...
-        interval=1000,
-    )
-    assert callable(decorator)
-
-
-def test_callback_module_export():
-    """Test that callback is properly exported from dash module."""
-    assert hasattr(dash, "callback")
-    assert dash.callback is callback
+"""Unit tests for callback decorator behavior - no browser required."""
```
T4rk1n, line 1: This file got converted to CRLF, which is why it shows as a full rewrite (132 deletions) instead of just the three added tests. Can you convert it back to LF so the diff is only the new tests?

T4rk1n (commented):
Fix is right. Escaping has to stay (the `...` multi-output separator in `split_callback_id` would otherwise trip on dots inside a dict value), and `\u002e` keeps the escape while staying valid JSON — `parsePMCId` keys off the last `}` and `rsplit(".", 1)` on the backend both round-trip cleanly. `create_callback_id` is the only place that does this escaping, so the fix is complete.

Can you add a `CHANGELOG.md` entry under Fixed, linking back to #3480? One inline note below.

Issue #3924, opened 2026-07-29
[Feature Request] Option to compress request payload in a callback


datenzauberai:
**Problem**
While the payload of the HTTP response bodies can be compressed in case of transmitting large amounts of data from the server side to the client, this is not possible for the actual HTTP request body of a callback that sends this data to the server side.

Grids are notoriously using huge JSON structures of several megabytes. Transmitting this data via the network can make the application feel sluggish.

**Solution**
We could have an option to compress the body of a server side callback on the client side. We could introduce two new parameters for the callback:

`compress_payload: bool = False`: If True, the callback request payload will be compressed using gzip compression before being sent to the server. Defaults to False.
`compress_threshold: int = 5_000`: The size threshold in bytes above which the payload will be compressed when `compress_payload` is True. Set to 0 to always compress regardless of size. Defaults to 5,000 bytes (5 kB).

**Alternatives**
This can partially be simulated by using client-side callbacks that compress and copy the data to a compressed `dcc.Store` and then rewire the callbacks to use that store and decompress on the server side. However, this bloats the code and application state.

See referenced PR for more details: https://github.com/plotly/dash/pull/3925

AnnMarieW:
Just adding a link to a related discussion that might be helpful:  https://github.com/plotly/dash/pull/3456

PR #3925, opened 2026-07-29
Add options to compress callback request payload

datenzauberai:
This implements an option to compress the payload of a callback's request (see https://github.com/plotly/dash/issues/3924).

This is useful when transmitting large amounts of data from the client-side to the server-side (server to client is already covered by standard HTTP compression via Content-Encoding).

The basic idea is that the server side callbacks can be marked for payload compression with two parameters `compress_payload: bool` and `compress_threshold: int` that indicate if the request body should be compressed and the minimum size of body before compressing.

```python
@app.callback(
    Output("my-plot", "figure"),
    Input("refresh-button", "n_clicks"),
    State("my-grid", "rowData"),
    State("my-grid", "columnDefs"),
    State("my-store", "data"),
    compress_payload=True,
    compress_threshold=10_000, # optional, default would be 5_000
)
def update_plot(n, gridRowData, gridColumnDefs, storeData):
```

On the client side the JSON of the body is constructed as before. When the callback is marked for request compression and the request body exceeds the compression threshold (the default is 5,000 bytes), the body is compressed and send as binary with the header `Content-Encoding=gzip`. The compression is done using the Compression Streams API which is [widely available across major browsers](https://caniuse.com/wf-compression-streams).

On the server side the request is checked for the content encoding header and the request body is decompressed when necessary.

This new feature is able to speed-up application performance for use-cases where loads of data are transmitted. Depending on your connection and payload you can expect roundtrip response time improvements up to 10x. Some numbers showing the average speed-up ratio of the roundtrip response time depending on different connections and the payload sizes (realistic grid data as a JSON).

newplot

This seems to work fine across all backends (flask, fastapi, and quart).

Advantage of this solution is that it does not depend on compressing individual properties and can be used without compromising client-side callbacks to be able to access properties.

I first tried sending the compressed payload as Base64-encoded data inside JSON, but sending it as raw binary is significantly more efficient. It is also less proprietary, avoids adding a base64 package to the dash renderer, and allows a reverse proxy to decompress the payload transparently (which would reset/remove the content encoding header).

The decompression on the server is protected against ZIP bomb attacks with a default limit of 64 MB uncompressed payload. This limit can be changed by setting the environment variable `DASH_MAX_PAYLOAD_SIZE_MB` (limit in MB, e.g. `128` to set a limit of 128 MB).

## Contributor Checklist

   -  [x] Added parameters `compress_payload` and `compress_threshold` to callbacks
   -  [x] Implemented client-side compression and server-side decompression of the callback payload

### optionals


Aaron-Wrote-This:
Doesn't dash already support this via the `compress` option?  https://dash.plotly.com/reference

Maybe adding support for fast api's compression directly in dash could be helpful too: https://fastapi.tiangolo.com/advanced/middleware/#gzipmiddleware 
No new library's needed.

And ideally you don't really want dash running compression on the main thread in most cases. Having the server do it instead is generally better/faster/more flexible, such as nginx https://docs.nginx.com/nginx/admin-guide/web-server/compression/

datenzauberai:
> Doesn't dash already support this via the `compress` option? https://dash.plotly.com/reference
> 
[…]
> And ideally you don't really want dash running compression on the main thread in most cases. Having the server do it instead is generally better/faster/more flexible, such as nginx https://docs.nginx.com/nginx/admin-guide/web-server/compression/

This was also my first thought when I looked into this! Unfortunately, all these standard compression mechanism will only work for the HTTP response from the server to the client downloading the data. This PR would address the other direction when uploading the data which is not covered by the standard means. I enhanced the description to make this a little more clear.

Aaron-Wrote-This:
Oh, that's cool! I learned something new, thanks!
Nice feature, I'll definitely check it out.

T4rk1n:
> Downside is that a new dependency has to be introduced to actually compress the payload in the browser (I opted for https://github.com/101arrowz/fflate).

I think we should be using https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API instead of the new dependency. The compressions are async based and won't block the render thread.

```js
const stream = new Blob([newBody]).stream()
    .pipeThrough(new CompressionStream('gzip'));
fetchBody = await new Response(stream).blob();
```

There is also a server side zip bomb concern, need to put a decompression cap. (e.g., `decompressobj` with bounded `max_length` loops, tied to a max-payload config)

datenzauberai:
> > Downside is that a new dependency has to be introduced to actually compress the payload in the browser (I opted for https://github.com/101arrowz/fflate).
> 
> I think we should be using https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API instead of the new dependency. The compressions are async based and won't block the render thread.
> 
Replaced the lib with the standard API.

> 
> There is also a server side zip bomb concern, need to put a decompression cap. (e.g., `decompressobj` with bounded `max_length` loops, tied to a max-payload config)

Good find. Added a max_size that would be 64 MB by default that should be enough for most use-cases and easily handled by most servers.

datenzauberai:
@T4rk1n, is there anything missing from my side or could I somehow support moving this forward (clarifications, changes, updates, additions, ...)?

T4rk1n:
> @T4rk1n, is there anything missing from my side or could I somehow support moving this forward (clarifications, changes, updates, additions, ...)?

Just need to run `npm run format` and add a changelog entry then I think it's good to go.

datenzauberai:
@T4rk1n Formatting is done and I have added the change log entry!

T4rk1n (approved):
💃

PR #3930, opened 2026-07-31
Shared storage: backend-agnostic state manager + pub/sub

T4rk1n:
Split out from #3888 (Streaming callbacks + shared storage) so it can be reviewed on its own. This PR is **phase 1**: the backend-agnostic shared storage layer. The streaming work builds on top of it and stays on #3888 for now.

## What this adds

Shared storage: a backend-agnostic state manager available on every app via `dash.ctx.shared_storage` (and `app.shared_storage`).

- A cross-process key/value store (`get`/`set`/`delete`) and ordered publish/subscribe (`publish`/`subscribe`) for sharing state between callbacks or across worker processes without an external service.
- The default `LocalSharedStorage` elects a single owner process per machine (binding an `AF_UNIX` socket on POSIX, TCP loopback on Windows — the bind is the lease, re-elected on owner death) and serves the rest; a single-process deployment is its own owner and pays no socket overhead.
- Started lazily on first use, so it costs nothing until touched and never binds in a gunicorn preload master or the Flask reloader parent. Pass `shared_storage=None` to `Dash(...)` to disable, or a `BaseSharedStorage` subclass/instance to swap the backend.
- Subscriptions are ordered and replayable: a consumer that reconnects resumes from its last-seen message out of a bounded buffer, and a buffer overrun surfaces as an explicit gap rather than a silent loss.
- The wire codec is msgspec (msgpack), never pickle, so bytes off the socket cannot execute code; values must be JSON-compatible (like `dcc.Store`).

## Contributor Checklist

- [x] I have run the tests locally and they passed (`pytest tests/shared_storage` — 17 passed).
- [x] I have added tests, or extended existing tests, to cover any new features in this PR (unit + cross-process integration under `tests/shared_storage/`, plus a dedicated CI job).
- [x] I have added an entry in the `CHANGELOG.md`.


dash/_shared_storage/_codec.py
```diff
@@ -0,0 +1,37 @@
+"""Codec for the shared-storage socket transport.
+
+Prefers ``msgspec`` (msgpack: fast, compact) and falls back to stdlib ``json``.
```
KoolADE85: This looks like a Claude-ism: is there some condition that demands a real reason to "fall back"? I'd argue we should pick 1 implementation and keep it DRY.

dash/_shared_storage/_engine.py
```diff
@@ -0,0 +1,115 @@
...
+
+import threading
+import time
+from collections import deque
+from typing import Any, Deque, Dict, List, NamedTuple, Tuple
+
+# Per-topic replay buffer size. Sized to cover a reconnect window comfortably;
+# a producer that outruns a disconnected consumer past this raises a gap rather
+# than dropping messages silently.
+DEFAULT_BUFFER = 2048
```
KoolADE85: I'm a bit uneasy about this: it looks like this means we will keep old messages around in a buffer? We are going to have to maintain code paths for this, and it has a limited use-case.

In a traditional callback, if there's a network blip, you must re-trigger the callback anyway. It's a design choice that I feel we should not change just because we're streaming now.

As implemented, the feature is "general purpose" (very cool btw!). It will also be useful for publishing larger datasets from a shared location. A callback that publishes a 2MB dataset will silently retain 2048 of them (4GB per topic!). Could OOM or fill up disk space, and it's not obvious that this will happen.

I'm inclined to suggest that old data just be discarded - no buffer. If the client falls behind for some reason, then they need to trigger the callback again.
Alternatively, I suggest making the buffer `0` by default: safe out-of-the-box but can be enabled if you know what you're doing.

T4rk1n: The buffer is to ensure the frames are not dropped, there can be a slight delay before a subscribe actually start or burst of events (diskcache has polling, so it need buffer). We can't set the buffer to 0 since that would drop all messages. I think 32 would be a good middle ground.

KoolADE85: OK, I agree. I just want to call out that the buffer is bounded by a per-topic "count" rather than "total bytes" so it's certainly possible to OOM. This may be something we just document rather than work around it here.

KoolADE85 (approved):
I've tested quite a few scenarios and this works well on my end. Great feature! 💃

PR #3931, opened 2026-07-31
Streaming callbacks with multiplexed transport

T4rk1n:
> Split out of #3888 and **stacked on #3930 (shared storage)** — this is the **streaming** half. It targets `feat/shared_storage`; review it as the diff on top of that PR, and it will be retargeted to `dev` once #3930 merges.

## Summary

This PR adds **streaming callbacks** and the **multiplexed HTTP transport** that carries them across worker processes, built on the backend-agnostic **shared storage** primitive from #3930.

- **Streaming callbacks** — decorate an `async def` generator; each `yield` is pushed to the browser as it is produced (same shape as a normal return; `dash.Patch` yields apply incrementally, e.g. LLM token streaming). No opt-in keyword — a generator streams by definition. Sync generators are rejected at registration.
- **Multiplexed HTTP streaming** — all of a page's streams share **one** downlink connection instead of one per callback, so they no longer hit the browser's ~6-connections-per-host ceiling. Works on Flask, Quart, and FastAPI, no WebSocket required.
- **Shared storage** (`dash.ctx.shared_storage` / `app.shared_storage`, from #3930) — a cross-process key/value store + ordered publish/subscribe, used internally here as the broker for streaming.

## Streaming callbacks

```python
@callback(Output("out", "children"), Input("go", "n_clicks"))
async def stream(n):
    async for token in llm.stream(prompt):
        patch = Patch()
        patch += token
        yield patch
```

## Shared storage (base PR #3930)

```python
app = Dash(__name__)                     # shared_storage=LocalSharedStorage by default
# inside a callback:
dash.ctx.shared_storage.set("k", value)  # cross-process KV (JSON-compatible values)
dash.ctx.shared_storage.publish("topic", msg)
```

The default `LocalSharedStorage` elects a single owner process per machine (AF_UNIX socket on POSIX, TCP loopback on Windows — the bind is the lease, re-elected on owner death) and serves the others. A single-process deployment is its own owner and pays no socket overhead. Subscriptions are ordered and replayable: a reconnecting consumer resumes from its last-seen sequence out of a bounded buffer, and a buffer overrun surfaces as an explicit gap rather than a silent loss. Pass `shared_storage=None` to disable, or a `BaseSharedStorage` subclass/instance to swap the backend (`DiskcacheSharedStorage`, `RedisSharedStorage`, ... — see #3930).

## Architecture — multiplexed streaming

A streaming callback no longer holds its own HTTP connection. Its POST returns a fast ack; its frames are pumped onto a shared-storage topic and relayed over the page's single downlink, routed back to the right callback by `requestId`. Because the frames travel through shared storage, **the worker that runs a callback and the worker that holds the downlink need not be the same process** — shared storage is the broker.

```mermaid
flowchart TB
    subgraph browser["Browser — one page"]
        cbs["streaming callbacks<br/>(async def generators)"]
        sc["StreamClient<br/>single downlink per page"]
        cbs --> sc
    end

    subgraph server["Dash server — any number of worker processes"]
        wa["worker A<br/>runs callback, pumps frames"]
        wb["worker B<br/>serves the downlink"]
    end

    store[("Shared storage owner process<br/>KV + ordered pub/sub<br/>topic per connection")]

    sc -->|"1 · uplink POST, fast ack<br/>streamConnection = conn + requestId"| wa
    wa -->|"2 · publish frames, tagged requestId + seq"| store
    sc -->|"3 · single downlink<br/>streamDownlink = conn, from = seq"| wb
    store -->|"4 · subscribe, replay from seq"| wb
    wb -->|"5 · NDJSON frames"| sc
    sc -->|"6 · route by requestId, apply"| cbs
```

**Transport selection** (renderer): a streaming callback rides the **WebSocket** transport when websocket callbacks are enabled; otherwise the **multiplexed HTTP** transport when shared storage is available; otherwise falls back to today's **one NDJSON connection per callback**. So nothing changes for apps that don't opt into shared storage.

**Scheduler:** long-lived streams no longer consume the renderer's concurrent-request budget, and clientside callbacks are exempt from it — a page full of streams no longer starves other callbacks.

## Testing

- Streaming transport: uplink/downlink end-to-end over real HTTP on **Flask, Quart, and FastAPI**.
- Renderer: `StreamClient` routing, multiplexing, reconnect-from-cursor, keepalive (karma).
- Shared-storage broker: the `iter_with_seq` / `aiter_with_seq` sequence-aware subscriptions the downlink resumes from, across all backends (`test_stream_hub.py`).
- **Browser validation** (Selenium): the streaming integration suite runs over the multiplexed transport, plus an 8-stream demo where all streams and a clientside clock run simultaneously with a single downlink.

## Notes

- Shared-storage values must be JSON-compatible (like `dcc.Store`).
- The default in-memory backend is not durable: if the owner process dies, a survivor re-elects with an empty store. For multi-pod deployments behind a load balancer use `RedisSharedStorage` (see #3930).

## Follow-up (not in this PR)

- Host the downlink in a **SharedWorker** so it's one connection per *browser* (shared across tabs) rather than per page. `StreamClient` is written host-agnostic for this; the per-page transport already solves the connection-limit problem.


dash/_stream_hub.py
```diff
@@ -0,0 +1,151 @@
...
+    These frames feed a ``StreamedCallbackResponse`` so the existing NDJSON
+    response path serializes and keep-alives them -- no bespoke endpoint. It is
+    long-lived: it carries frames for every callback on the connection, not one
+    stream, and ends when the client hangs up. ``replay_from`` resumes a
+    reconnecting downlink from its last cursor. Each envelope carries its ``seq``
+    so the client can resume from it after a reconnect without losing frames.
+    """
+    with storage.subscribe(stream_topic(connection_id), replay_from) as sub:
+        for seq, message in sub.iter_with_seq():
+            yield {**message, "seq": seq}
```
KoolADE85, line 102: The loop here ends on client disconnect, but not always on server shutdown. As a result, streams continue even after a Ctrl+C (see the repro app below).
We should respond to a shutdown event the way `ws.py` does.

```python
import asyncio
from dash import Dash, Input, Output, callback, html

server = app.server
app.layout = html.Div([
    html.H3("If you click the button, Ctrl+C won't stop this server"),
    html.Button("stream", id="btn", n_clicks=0),
    html.Div(id="out"),
])

@callback(
    Output("out", "children"),
    Input("btn", "n_clicks"),
    prevent_initial_call=True
)
async def long_stream(n):
    for i in range(600):
        await asyncio.sleep(0.5)
        print(f"tick {i}")
        yield f"tick {i}"

if __name__ == "__main__":
    app.run(debug=True)

```

dash/backends/_flask.py
```diff
@@ -251,8 +264,74 @@ def add_redirect_rule(self, app, fullname, path):
...
+                headers=dict(STREAM_HEADERS),
+            )
+
+        def _serve_downlink(downlink, with_request_ctx):
+            # The client's single multiplexed streaming connection: relay this
+            # connection's frames (published by streaming callbacks, possibly on
+            # other workers, via shared storage) as an ordinary NDJSON stream.
+            storage = dash_app.shared_storage
+            frames = subscribe_envelopes(
+                storage, downlink["connectionId"], downlink.get("from")
```
KoolADE85: Here we are implicitly trusting the `connectionId` sent by a client.
If the ID is somehow leaked or derived, then anyone can use the ID to join someone else's stream and receive its data. Suggest generating the ID server-side and tying it to a cookie or session.

See also `_serve_uplink` where a connectionId could also be used to inject frames into someone else's stream.

This also applies to the fastapi & quart backends.

dash/dash-renderer/src/utils/streamClient.ts
```diff
@@ -0,0 +1,219 @@
...
+    new Promise(resolve => setTimeout(resolve, ms));
+
+export class StreamClient {
+    private connectionId = genId();
+    private pending = new Map<string, PendingStream>();
+    private counter = 0;
+    // Last sequence applied; the downlink resumes from here on reconnect. Starts
+    // at 0 so the first connect replays anything published before it subscribed
+    // (the uplink POST and the downlink open race).
+    private cursor = 0;
```
KoolADE85, line 70: This cursor is per-page, monotonic, and is never reset.
On reconnect, the client sends `from: this.cursor` (line 160 below).
This seems correct as long as the server is running, but a server restart introduces a flaw: the server's cursor must catch up to the client before the stream can resume.

See this example app (works with gunicorn as well as the dev server)

```python
import time

from dash import Dash, Input, Output, callback, dcc, html

app = Dash(__name__, backend="flask")
server = app.server

app.layout = html.Div(
    children=[
        html.H3("Streamed clock stalls silently after a server restart"),
        html.P("Watch it tick a few seconds, then restart the server."),
        html.P("Observe that it stays frozen for as many seconds as the previous server was alive."),
        html.Pre(id="clock", children="waiting for first tick…"),
        dcc.Interval(id="kick", interval=1000),
    ],
)


@callback(
    Output("clock", "children"),
    Input("kick", "n_intervals"),
    prevent_initial_call=True,
)
async def stream_clock(_):
    yield f"serverside clock: {time.strftime('%H:%M:%S')}"


if __name__ == "__main__":
    app.run(debug=True, port=8050)

```

KoolADE85 (changes requested):

T4rk1n:
@KoolADE85 This is ready for another look.

T4rk1n:
### Streaming transport load test (42d618f2d)

One streaming callback per simulated browser, yielding every 0.5 s; "frame delay" is server yield → client receipt. 8-core dev box, with the load clients running on the same machine (so the gunicorn numbers are pessimistic).

| server | browsers | frame delay p50 / p95 |
|---|---|---|
| uvicorn, 1 worker (FastAPI) | 1000 | 3 ms / 22 ms |
| uvicorn, 4 workers | 1000 | 1 ms / 3 ms |
| gunicorn `-w 4 --threads 8` (Flask, poll mode) | 300 | 105 ms / 200 ms |
| gunicorn `-w 8 --threads 8` | 1000 | 180 ms / 3.6 s (CPU-bound) |
| gunicorn `-w 1` (sync worker) | 50 | 50 ms / 100 ms |

Takeaways:
- ASGI (Quart/FastAPI) keeps one open downlink per browser with no thread behind it; a single uvicorn worker holds 1000 streaming browsers with constant latency.
- Flask/WSGI now polls (`Dash(stream_poll_interval=100)`, default): each downlink request returns queued frames and ends, so no thread is pinned per browser and `gunicorn --threads 2` no longer hangs on the second browser. The cost is one cheap request per browser per interval, so plan roughly one gunicorn worker per ~150 concurrently streaming browsers. For thousands of concurrent streams, use an ASGI backend.
- All tabs of a browser share the downlink through a SharedWorker (`dash-stream-worker.js`), so streams no longer count against the ~6 connections-per-host browser limit; closing a tab cancels its streams server-side.

Details are in `.ai/ARCHITECTURE.md` → "Multiplexed downlink and the stream SharedWorker".

MattSkewbie:
Docs PR is located here: https://github.com/plotly/ddk-dash-docs/pull/4103.

KoolADE85 (approved):
I've tested this thoroughly with all backend types and various streaming scenarios. It's working well here. Congrats @T4rk1n ! 💃

Issue #3929, opened 2026-07-31
[BUG] dbc Tabs not showing default first tab anymore with Dash 4.2.0 and higher


ghaarsma:
For any Dash version > 4.1.0, the dash-bootstrap-components (dbc) Tabs do not show the default 1st tab anymore.

With dash==4.1.0 and dash-bootstrap-components==2.0.4

<img width="1225" height="558" alt="Image" src="https://github.com/user-attachments/assets/b554bfb7-271c-4862-898b-748787656614" />

With dash==4.4.1 and dash-bootstrap-components==2.0.4

<img width="1104" height="505" alt="Image" src="https://github.com/user-attachments/assets/22bedc4c-ec29-4d6a-af07-dcf34ee94a59" />

With the following trivial example:
```python
from dash import Dash, html
import dash_bootstrap_components as dbc

app = Dash(__name__, title="Dash Tab Test", external_stylesheets=[dbc.themes.BOOTSTRAP])

children = [dbc.Tab(html.H2(f"Test Tab {i}"), label=f"Tab {i}") for i in range(3)]

app.layout = (
    html.Div(
        [
            dbc.Tabs(children=children)
        ]
    ),
)


if __name__ == "__main__":
    app.run(debug=True)
```

Note: if you switch on the tabs back and forth, the tab contents appear again. It's just the first tab does not show up anymore by default. See also dbc [documentation](https://www.dash-bootstrap-components.com/docs/components/tabs/).

AnnMarieW:
@ghaarsma Thanks for reporting and the good write-up!

There's a new PR to fix the issue (thanks @BSd3v). There are a few related issues that are being looked at as well, so if there's a delay before the fix is merged,  a workaround is to  give each `Tab` a `tab_id` and setting the initial tab to render in the  `active_tab` prop on `Tabs`:

```python
     dbc.Tabs(
            [
                dbc.Tab(label="Tab 1", tab_id="tab-1"),
                dbc.Tab(label="Tab 2", tab_id="tab-2"),
            ],
            id="tabs",
            active_tab="tab-1",
        )

```

robertclaus:
Thanks for reporting this. I was able to reproduce in another setting as well.

PR #3934, opened 2026-07-31
fixes aggressive remounting

BSd3v:
fixes:
- https://github.com/plotly/dash/issues/3846
- https://github.com/plotly/dash/issues/3929

Issue was that on the first render of a component, it would trigger it to remount and clear the `layoutHashes` since it wasnt protected to see if it was the very first time that the component was rendering.

### DIFF
--- a/dash/dash-renderer/src/wrapper/DashWrapper.tsx
+++ b/dash/dash-renderer/src/wrapper/DashWrapper.tsx
@@ -446,7 +446,7 @@ function DashWrapper({
     };
 
     useEffect(() => {
-        if (_newRender) {
+        if (_newRender && freshRenders.current > 1) {
             dispatch(
                 resetComponentState({
                     itempath: componentPath


KoolADE85 (approved):
Good catch, such a sneaky issue!

Mukller (commented):
Verified the mechanism against `DashWrapper.tsx` on master: `freshRenders.current += 1` lives in the `useMemo(..., [_newRender])` block, so on first mount with `_newRender` set the counter reaches 1 *before* the `useEffect` runs — hence `> 1` correctly suppresses the spurious `resetComponentState` dispatch that caused #3846/#3929. The core fix looks right.

Two edge cases worth considering, plus one request:

**1. StrictMode / double-invoked renders.** In development React StrictMode can invoke the render pass twice; if both invocations increment the ref before effects flush, `freshRenders.current` hits 2 on the very first mount and the remount bug resurfaces in dev builds. Does dash-renderer run wrappers under StrictMode anywhere? If yes, an escape hatch like comparing against a "seen first commit" boolean ref instead of counting renders might be more robust:

```tsx
const didInitialRender = useRef(false);
useEffect(() => {
    if (_newRender && didInitialRender.current) {
        dispatch(resetComponentState({ itempath: componentPath }));
    }
    didInitialRender.current = true;
}, [...]);
```

**2. Instance reuse across paths.** `freshRenders` never resets when `componentPath` changes (`renderedPath.current = componentPath` updates, counter doesn't). If Dash ever recycles a wrapper instance for a different component, the new component's legitimate first `_newRender` will have counter > 1 and get an immediate state reset. Probably rare, but cheap to guard by resetting the counter (or using the boolean above) when `renderedPath.current !== componentPath`.

**3. Tests.** The diff is 1 line — could you add a renderer test reproducing the #3846 sequence (initial render must not dispatch `resetComponentState`) so it stays fixed?

Nice find either way — the root cause (reset firing on first render) matches both linked issues precisely.

camdecoster:
Closing this in favor of the stack that @T4rk1n created starting with #3881.

PR #3936, opened 2026-08-02
Use the proxied url as the Jupyter server url

### LINKED ISSUE #3401: [BUG] DASH_PROXY not used by JupyterDash
**Describe your context**

-  replace the result of `pip list | grep dash` below (using `pixi` so this is the conda package version)
```
dash                           3.2.0           pyhd8ed1ab_0          5.3 MiB    conda  https://conda.anaconda.org/conda-forge/
```

**Describe the bug**
When I set `DASH_PROXY` so that Dash reports a different URL than the listen address it is not respected by Dash's Jupyter support.

**Expected behavior**

The external URL and iframe that Dash renders should use the `DASH_PROXY`

https://github.com/benmoss/jupyter-dash-proxy demonstrates the issue, just needs the https://pixi.sh/ package manager to install and run.

Image

Sanjays2402:
Fixes #3401.

When `DASH_PROXY` (or `run(proxy=...)`) is set, the app is served on `host:port` but reached through the proxy. `run()` already used the proxied url for the "Dash is running on" log line, but passed nothing to `jupyter_dash.run_app()`, so JupyterDash fell back to `http://host:port` for both the external url and the inline iframe — unreachable behind the proxy. `server_url` is now derived from the proxied half of the proxy string when the caller did not pass `jupyter_server_url` explicitly.

## Contributor Checklist

   -  [x] pass the proxied url through to `jupyter_dash.run_app()`
   -  [x] add a regression test

`tests/unit/test_configs.py` passes (the 2 unrelated failures in that file are missing built JS assets in my source checkout, identical before and after this change). The new `test_proxy_jupyter_server_url` fails without the fix (`server_url` is `None`) and passes with it.

### optionals

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Fixed
 - [#3916](https://github.com/plotly/dash/pull/3916) Fixed a regression where dragging multiple files into `dcc.Upload` would upload only the first file when `multiple=True`
 - [#3922](https://github.com/plotly/dash/pull/3922) Fix `dcc.Input(type="number")` stepper behavior when only `min` is set.
+- [#3925](https://github.com/plotly/dash/pull/3925) Use the proxied url as the Jupyter server url so `DASH_PROXY` is honored by the external url and inline iframe in notebooks.
 
 ## [4.4.1] - 2026-07-21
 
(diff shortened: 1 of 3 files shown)


T4rk1n (approved):
Looks good, thank you. 💃

Issue #3937, opened 2026-08-03
[BUG] dcc.Patch() wipes persistence of sibling pattern-matching elements


Aaron-Wrote-This:
Sibling of: https://github.com/plotly/dash/issues/3681
Environment

```
dash                 4.4.0
```

When patching a new element to list of elements with pattern matched ids, persistence is wiped from all sibling IDs.

Minimal example:

```python
from dash import ALL, Dash, Input, Output, Patch, dcc, html, clientside_callback, callback

app = Dash(__name__)

def make_input(index):
    return dcc.Input(
        id={"type": "inp", "index": index},
        value="initial",
        persistence=True,
        persistence_type="local",
    )


app.layout = html.Div(
    [
        html.Button("add", id="add"),
        html.Button("clear", id="clear"),
        html.Div([make_input(0), make_input(1)], id="container"),
        html.Div(id="display"),
        html.Pre(id="storage"),
        dcc.Interval(id="local_storage_poll", interval=250),
        dcc.Input("Hello, I don't wipe", id='non-pattern-matching-id', persistence=True, persistence_type='local')
    ]
)


@callback(
    Output("container", "children"),
    Input("add", "n_clicks"),
    prevent_initial_call=True,
)
def add_input(n):
    patch = Patch()
    patch.append(make_input(n + 1))
    return patch


@callback(Output("display", "children"), Input({"type": "inp", "index": ALL}, "value"))
def show(values):
    return "|".join(values)


# Debug info, show local storage to make it easier to noticed when it's wiped
clientside_callback(
    """
    function display_persisted_vals() {
        return Object.keys(window.localStorage)
            .filter(k => k.startsWith('_dash_persistence.'))
            .map(k => k + ' = ' + window.localStorage.getItem(k))
            .join('\\n') || '(nothing persisted)';
    }
    """,
    Output("storage", "children"),
    Input("local_storage_poll", "n_intervals"),
)

# Debug helper, clear local storage to do a clean test
# (helpful after the fix when the storage doesn't wipe constantly, lol)
clientside_callback(
    """
    function clear_persistence() {
        Object.keys(window.localStorage)
            .filter(k => k.startsWith('_dash_persistence.'))
            .forEach(k => window.localStorage.removeItem(k));
        window.location.reload();
        return window.dash_clientside.no_update;
    }
    """,
    Output("clear", "n_clicks"),
    Input("clear", "n_clicks"),
    prevent_initial_call=True,
)


if __name__ == "__main__":
[...]

PR #3938, opened 2026-08-03
Fix dcc.Patch() disturbing pre-existing components (re-run callbacks + wiped persistence)

### LINKED ISSUE #3681: [BUG] dash.Patch() reruns inital callback for all matching elements
Environment
```
dash                 4.0.0
```

Hello, I've been getting into the wonderful world of extensibility, dash.Patch() and pattern matching callbacks. 
Click a button, a new input appears, ect. But I have run into an issue when using pattern matching callbacks. 

When a new element is appended via dash. Patch(), the initial callbacks run against every element of that type.

I have created a minimum test app to show the issue. 

When clicking the "Add Slider" button, I would expect to see the `print_callback_value` function run once **per click**. Instead I see it run multiple times, once **per element.**
This can be seen in the console as well as the debug dash callbacks view.
10 slider creations cause 56 init callbacks.

Image

This causes performance issues with multiple sliders. It creates flickering as the elements are setup, and may have errors related to unexpected reinit callbacks, they have no ctx.triggered.

I have also created this PR as a suggested fix. https://github.com/plotly/dash/pull/3682

```python
from dash import Dash, dcc, html, Input, Output, callback, State, ctx, Patch, MATCH, no_update, clientside_callback, ALL

app = Dash(__name__)

app.layout = html.Div(
    children=[
        dcc.Button(id='add_slider_btn', children='Add Slider'),
        html.Div(id='slider_container', children=[])
])


@callback(
    Output({'type': 'slider','aio_id': ALL}, 'value'),
    Input({'type': 'slider','aio_id': ALL}, 'value'),
    prevent_initial_call=False
)
def print_callback_value(val):
    print(f'{ctx.triggered_id=} = {val}')
    raise PreventUpdate


@callback(Output("slider_container", "children"),
          Input("add_slider_btn", "n_clicks"),
          prevent_initial_call=True,
          )
def add_new_slider(n_clicks):
    sliders_container_patch = Patch()
    sliders_container_patch.append(
        dcc.Slider(id={
            'type': 'slider',
            'aio_id': str(n_clicks)
        },
            step=0.1,
            persistence=True, persistence_type='local'
        )
    )
    return sliders_container_patch


if __name__ == '__main__':
    app.run(debug=True)
```

Aaron-Wrote-This:
Fixes [#3681](https://github.com/plotly/dash/issues/3681) and [#3937](https://github.com/plotly/dash/issues/3937)

## Problem

When a `dcc.Patch()` added or rebuilt components inside a container, two separate bugs trigger, 
effecting components which already existed before the Patch:

1. Initial callbacks refire for every matching `MATCH`/`ALL` output, so appending one new
   pattern matching id'd `dcc.Slider` reruns the initial callback for every preexisting slider
2. Persisted user edits are wiped, so edits to persistence enabled components are cleared
   because the patch added a new sibling

I tracked both items back to the way Patch resolution distinguishes components carried over from the
pre-Patch layout from freshly built ones

## Fix

1. Snapshot the paths table before a Patch is resolved

2. Callbacks can suppress the initial call for carried-over components, by detecting props
reference identity (`child.props === oldPropsRef`). ramda's `assocPath` is structurally immutable, 
so untouched nodes keep their exact `props` reference while rebuilt/replaced node gets a new one

3. `applyPersistence` now skips components that already existed pre-Patch, instead of re-running 
`modProp` and misreading the carried over value as a "server override" that clears the stored edit

## Contributor Checklist

I've been doing a lot of work with dynamic UIs and creating extensible elements. I am loving the patch system, but I found some significant drawbacks to the both in performance and usability.

At first, I thought the fix for the initial callbacks was simple and just checking the ID more carefully would be enough, but it ended up being requiring adding an 'audit trail' for patches, which also ended up fixing the persistence issues.

Javascript isn't my 'first language', so let me know if there are any things I'm doing awkwardly, or best practices I'm missing. Happy to fix those up.
Or I understand if this is too complicated and not exactly needed. I looked for simpler options, but I couldn't find a way to differentiate patches and understand their impact in any other way, but it would be great if one existed and I just didn't see it.

   -  [x] Record patch operations
   -  [x] Fix patch running against existing components with matching pattern matching IDs
   -  [x] Fix persisted values being wiped from existing components during a patch
- [x] I have run the tests locally and they passed

### optionals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3916](https://github.com/plotly/dash/pull/3916) Fixed a regression where dragging multiple files into `dcc.Upload` would upload only the first file when `multiple=True`
 - [#3922](https://github.com/plotly/dash/pull/3922) Fix `dcc.Input(type="number")` stepper behavior when only `min` is set.
 - [#3925](https://github.com/plotly/dash/pull/3925) Use the proxied url as the Jupyter server url so `DASH_PROXY` is honored by the external url and inline iframe in notebooks.
+- [#3938](https://github.com/plotly/dash/pull/3938) Fix `dcc.Patch()` re-running the initial callbacks of components that were already on the page, including every matching (`MATCH`/`ALL`) element, and wiping their user-edited persisted values. Fixes [#3681](https://github.com/plotly/dash/issues/3681) and [#3937](https://github.com/plotly/dash/issues/3937)
 
 ## [4.4.1] - 2026-07-21
 
--- a/dash/dash-renderer/src/actions/callbacks.ts
+++ b/dash/dash-renderer/src/actions/callbacks.ts
@@ -32,6 +32,7 @@ import {
     BackgroundCallbackInfo,
     CallbackResponse,
     CallbackResponseData,
+    PatchedOutputs,
     SideUpdateOutput
 } from '../types/callbacks';
 import {isMultiValued, stringifyId, isMultiOutputProp} from './dependencies';
@@ -41,7 +42,8 @@ import {createAction, Action} from 'redux-actions';
 import {addHttpHeaders} from '../actions';
 import {notifyObservers, updateProps} from './index';
 import {CallbackJobPayload} from '../reducers/callbackJobs';
-import {parsePatchProps} from './patch';
+import {isPatch, parsePatchProps} from './patch';
+import {createPatchAnalysis} from './patchAnalysis';
 import {computePaths, getPath} from './paths';
 
 import {requestDependencies} from './requestDependencies';
@@ -244,6 +246,10 @@ const zipIfArray = (a: any, b: any) => {
 
 function cleanOutputProp(property: string) {
     return property.split('@')[0];
+}
+
+function patchedResultFields(patchedOutputs: PatchedOutputs) {
+    return keys(patchedOutputs).length ? {patchedOutputs} : {};
 }
 
 async function handleClientside(
@@ -970,13 +976,22 @@ export function executeCallback(
                         );
                         // Patch methodology: always run through parsePatchProps for each output
                         const currentLayout = getState().layout;
+                        const patchedOutputs: PatchedOutputs = {};
                         flatten(outputs).forEach((out: any) => {
                             const propName = cleanOutputProp(out.property);
                             const outputPath = getPath(paths, out.id);
-                            const dataPath = [stringifyId(out.id), propName];
+                            const idStr = stringifyId(out.id);
+                            const dataPath = [idStr, propName];
                             const outputValue = path(dataPath, data);
                             if (outputValue === undefined) {
                                 return;
+                            }
+                            if (isPatch(outputValue)) {
+                                // One analysis per output, shared by all of its
+                                // patched props
+                                patchedOutputs[idStr] =
+                                    patchedOutputs[idStr] ||
+                                    createPatchAnalysis();
                             }
                             const oldProps =
                                 path(
@@ -985,7 +1000,8 @@ export function executeCallback(
                                 ) || {};
                             const newProps = parsePatchProps(
                                 {[propName]: outputValue},
-                                oldProps
+                                oldProps,
+                                patchedOutputs[idStr]
                             );
                             data = assocPath(
                                 dataPath,
@@ -993,7 +1009,11 @@ export function executeCallback(
                                 data
                             );
                         });
-                        return {data, payload};
+                        return {
+                            data,
+                            payload,
+                            ...patchedResultFields(patchedOutputs)
+                        };
                     } catch (error: any) {
                         return {error, payload};
                     }
@@ -1079,13 +1099,22 @@ export function executeCallback(
                         // Layout may have changed.
                         // DRY: Always run through parsePatchProps for each output
                         const currentLayout = getState().layout;
+                        const patchedOutputs: PatchedOutputs = {};
                         flatten(outputs).forEach((out: any) => {
                             const propName = cleanOutputProp(out.property);
                             const outputPath = getPath(paths, out.id);
-                            const dataPath = [stringifyId(out.id), propName];
+                            const idStr = stringifyId(out.id);
+                            const dataPath = [idStr, propName];
                             const outputValue = path(dataPath, data);
                             if (outputValue === undefined) {
                                 return;
+                            }
+                            if (isPatch(outputValue)) {
+                                // One analysis per output, shared by all of its
+                                // patched props
+                                patchedOutputs[idStr] =
+                                    patchedOutputs[idStr] ||
+                                    createPatchAnalysis();
                             }
                             const oldProps =
                                 path(
@@ -1094,7 +1123,8 @@ export function executeCallback(
                                 ) || {};
                             const newProps = parsePatchProps(
                                 {[propName]: outputValue},
-                                oldProps
+                                oldProps,
+                                patchedOutputs[idStr]
                             );
 
                             data = assocPath(
@@ -1111,7 +1141,11 @@ export function executeCallback(
                             );
                         }
 
-                        return {data, payload};
+                        return {
+                            data,
+                            payload,
+                            ...patchedResultFields(patchedOutputs)
+                        };
                     } catch (res: any) {
                         lastError = res;
                         if (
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Fixed
 - [#3916](https://github.com/plotly/dash/pull/3916) Fixed a regression where dragging multiple files into `dcc.Upload` would upload only the first file when `multiple=True`
 - [#3922](https://github.com/plotly/dash/pull/3922) Fix `dcc.Input(type="number")` stepper behavior when only `min` is set.
+- [#???](https://github.com/plotly/dash/pull/???) Fix `dcc.Patch()` re-running the initial callbacks of components that were already on the page, including every matching (`MATCH`/`ALL`) element, and wiping their user-edited persisted values. Fixes [#3681](https://github.com/plotly/dash/issues/3681) and [#3937](https://github.com/plotly/dash/issues/3937)
```
T4rk1n: Replace with PR number

dash/dash-renderer/src/actions/patchAnalysis.ts
```diff
@@ -0,0 +1,95 @@
...
+/*
+ * The analysis, but only if `property` is one of the props it describes.
+ * A callback can return a Patch for one prop and a full value for another, and
+ * a full value is a fresh sub-tree in which nothing was carried over
+ */
+export function analysisForProp(
+    analysis: PatchAnalysis | undefined,
+    property: string
+): PatchAnalysis | undefined {
+    return analysis && analysis.patchedProps[property] ? analysis : undefined;
```
T4rk1n: Prefer using optional chaining:

```js
return analysis?.patchedProps[property] ? analysis : undefined;
```

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -1079,22 +1099,32 @@ export function executeCallback(
...
                                 path(
                                     outputPath.concat(['props']),
                                     currentLayout
                                 ) || {};
                             const newProps = parsePatchProps(
                                 {[propName]: outputValue},
-                                oldProps
+                                oldProps,
+                                patchedOutputs[idStr]
                             );
```
T4rk1n: This code is duplicated from L981-1016, might be worth it now to refactor into a function.

dash/dash-renderer/src/actions/patchAnalysis.ts
```diff
@@ -0,0 +1,95 @@
...
+ *   restored
+ * * `writtenProps`: props the patch wrote directly on a component that already
+ *   existed (`p[0]['props']['value'] = 5`). The component was not recreated,
+ *   but the server did provide a new value for that prop, which persistence
+ *   needs to know about to detect a server override
+ *
+ * Everything else in the resulting tree was carried over from the previous
+ * layout, whatever its `props` reference says
+ */
+export type PatchAnalysis = {
```
T4rk1n, line 23: One question about how the analysis gets consumed, though:

The two consumers use different predicates:

- `handleOneId` (callback suppression) uses `isCarriedOverByPatch`, which only checks `freshIds`
- `persistenceMods` uses `isUntouchedByPatch`, which checks `freshIds` **and** `writtenProps`

So a component whose prop the patch *wrote* (but didn't recreate) counts as "carried over" on the callback side, and its initial calls get suppressed. I think that can strand a dependent component:

```python
# both slider and badge already exist inside the patched container
p = Patch()
p[0]["props"]["children"][0]["props"]["value"] = 5  # writes slider.value
return p
```

with a callback `Input(slider, "value") -> Output(badge, "children")` where the badge is also inside the patched chunk:

- output side: badge isn't in `freshIds`, so `isCarriedOverByPatch` suppresses its initial call
- input side: all of the callback's outputs are inside the chunk, so the existing `chunkPath` filter drops it too

On `dev` today the over-firing bug incidentally keeps this callback alive, so the badge updates. With this PR I believe it never fires and the badge goes stale. `test_cbwc014` writes `deep-input.value` but nothing listens to that value, so this path isn't exercised.

The write here lands on the callback's input (slider.value), while the stranded callback outputs to badge. So the fix can't live on the output side — isUntouchedByPatch is keyed on the output id (badge), which the patch neither created nor wrote, so it would still read as carried-over. The gap is really in the input-side chunkPath filter: it drops a callback whose outputs are all inside the chunk because it assumes the chunk's initial-call path will cover it, but that path is (correctly) suppressed for the carried-over output, so nothing fires. Could the filter make an exception when the triggering input prop is in writtenProps? Either way, a test with a listener on the patch-written prop (not just a write nothing consumes, like test_cbwc014) would pin the semantics down.

This is an app to repro:
```python
"""
Repro for the freshIds/writtenProps asymmetry question on plotly/dash#3938.

Scenario:
  - `num` (an input) and `badge` both live inside `container` and exist
    before any patch
  - a callback listens to Input(num.value) and writes Output(badge.children);
    both ends are INSIDE the patched chunk
  - clicking "Patch" returns a Patch() on container.children that
      1. writes num.value          -> writtenProps case (component NOT recreated)
      2. appends a brand-new div   -> freshIds case (control: proves the
                                      suppression isn't just total)

What to look for after clicking "Patch" once:

  On dev / released dash (over-firing bug present):
      num shows the patched value AND badge updates to match
      (kept alive incidentally by the spurious initial-call refire)

  On the PR branch, IF the asymmetry is real:
      num shows the patched value but badge keeps its old text -> STALE
      - output side: badge is not in freshIds -> initial call suppressed
      - input side:  all the callback's outputs are inside the chunk
                     -> the existing chunkPath filter drops it too

  On the PR branch, IF something else covers it:
      badge updates anyway -> soften the review comment into
      "can we add a test pinning this down?"

The on-page counters make it visible without opening devtools:
  - badge callback count stuck at 1 after patching  = stale (only page load)
  - new-div MATCH count should track patches applied (control staying healthy)
"""

import threading

from dash import ALL, Dash, Input, Output, Patch, dcc, html

lock = threading.Lock()
calls = {"badge": 0, "new_div": 0}


def bump(name):
    with lock:
        calls[name] += 1


app = Dash(__name__, suppress_callback_exceptions=True)

app.layout = html.Div(
    [
        html.Button("Patch", id="patch-btn", n_clicks=0),
        html.Div(
            [
                dcc.Input(id="num", type="number", value=1),
                html.Div("badge: (initial)", id="badge"),
            ],
            id="container",
        ),
        html.Hr(),
        html.Div(id="badge-calls"),
        html.Div(id="new-div-calls"),
        html.Div("patch not applied yet", id="done"),
        dcc.Interval(id="tick", interval=500),
    ]
)


@app.callback(
    Output("done", "children"),
    Input("patch-btn", "n_clicks"),
    prevent_initial_call=True,
)
def do_patch(n):
    p = Patch()
    # writtenProps: write a prop on the EXISTING input (index 0 of children)
    p[0]["props"]["value"] = 100 + n
    # freshIds: append a genuinely new component as a control
    p.append(html.Div("appended (waiting)", id={"type": "new-div", "index": n}))
    return p, f"patch #{n} applied (num should now read {100 + n})"


# The callback under test: input AND output both inside the patched chunk.
@app.callback(Output("badge", "children"), Input("num", "value"))
def update_badge(value):
    bump("badge")
    return f"badge: num value is {value}"


# Control: initial call for the appended (fresh) components should still fire.
@app.callback(
    Output({"type": "new-div", "index": ALL}, "children"),
    Input({"type": "new-div", "index": ALL}, "id"),
    prevent_initial_call=False,
)
def new_div_initial(ids):
    bump("new_div")
    return [f"appended #{i['index']} (initial call ran)" for i in ids]


@app.callback(
    Output("badge-calls", "children"),
    Output("new-div-calls", "children"),
    Input("tick", "n_intervals"),
)
def show_counts(_):
    with lock:
        return (
            f"badge callback fired: {calls['badge']} time(s) "
            "(stuck at 1 after patching = STALE; increments = updated)",
            f"new-div ALL callback fired: {calls['new_div']} time(s) "
            "(should increment with each patch)",
        )


if __name__ == "__main__":
    app.run(debug=True)
```

Aaron-Wrote-This: Yuuup, good point. Thanks!

I added a check for if it was written by a patch based off of the analysis

I also wrote up some tests to prove to myself it was working, we can drop those if they're unnecessary to keep around in the repo

T4rk1n (commented):
Looks good, just need to run `npm run format` then we can merge.

Aaron-Wrote-This:
Formatting fixed, hopefully good to go now! Thanks!

T4rk1n (approved):
Looks good, thank you 💃 

This is part of fixes for #3846 performance regression.

PR #3941, opened 2026-08-04
Open FastAPI/Quart websocket only when a websocket callback needs it

### LINKED ISSUE #3939: [BUG] Websocket connection initiated by front end even if not enabled in back end
**Describe your context**
Websockt connections are created even if unused.

```
dash                 4.4.0
```
Tested in Edge and Firefox.

**Describe the bug**

I am not sure if this is intended behaviour or a bug. The Dash front end attempts to established a WS connection even if `websocket_callbacks=False` is set in `dash.Dash` and no callbacks are set to use websockets.

**Expected behavior**
It would be preferable that no connection is attempted unless the app is configured to actually use websockets.

T4rk1n:
Closes #3939

## Problem

The FastAPI and Quart backends always advertise WebSocket infrastructure (`url`/`worker_url`) in the page config, because they have `websocket_capability = True` regardless of whether any WebSocket callback exists. The renderer keyed its "open the socket" decision on the mere *presence* of that infra, so **every** app on these backends opened a WebSocket on page load — even apps with no WebSocket callbacks at all. `config.websocket.enabled` (`bool(self._websocket_callbacks)`) was already computed but ignored for the connect decision.

Repro: a plain `Dash(backend="fastapi")` app with only HTTP callbacks opens a socket on load (`WebSocket /_dash-ws-callback [accepted]`).

## Fix (renderer only)

Split "register handlers" from "open the socket":

- **`observers/websocketObserver.ts`** — `initializeWebSocket` still registers all message handlers (`onSetProps`, `onGetPropsRequest`, `onConnected`, …) whenever the infra is present, but only opens the socket eagerly when `config.websocket.enabled` is true. Keeping handlers registered is what lets the lazy path work: a per-callback `websocket=True` opens the socket on first dispatch via the existing `handleWebsocketCallback → workerClient.ensureConnected` path, and the handlers are already there to route server-pushed `set_props`/`get_props`.
- Guarded the `visibilitychange` reconnect with `wasDisconnected`, so tab focus never opens a *first* connection for an app that never had one.
- **`AppProvider.react.tsx`** — the effect now registers the observer whenever the infra is available; the enabled/lazy decision lives entirely inside `initializeWebSocket`.

## Resulting behavior

| Setup | Connects |
|---|---|
| `websocket_callbacks=True` (global) | eagerly on load (unchanged) |
| only per-callback `websocket=True` | lazily, on first such dispatch |
| no WebSocket callbacks | never |

## Tests

`tests/websocket/test_ws_lazy_connect.py` adds three regression tests that use a `websocket_connect` hook as a precise server-side probe for "did a socket actually open": never-connect (HTTP only), lazy-connect (per-callback), eager-connect (global). Verified they **fail** on the pre-fix bundle (the never/lazy cases saw a socket open on load) and pass with the fix. Existing `test_ws_basic.py` + `test_ws_hooks.py` (15 tests) still pass.


KoolADE85 (approved):
Works well! 💃

PR #3944, opened 2026-08-06
Fix backend detection in testing runners

### LINKED ISSUE #3942: [BUG] dash.testing.ThreadedRunner misdetects a subclassed/wrapped FastAPI app as Flask (e.g. under OpenTelemetry auto-instrumentation) -> TypeError: Config.__init__() got an unexpected keyword argument 'threaded'

**Describe your context**

```
dash                                      4.4.0
fastapi                                   0.136.3
uvicorn                                   0.49.0
```

Also reproduced against dash 4.4.1 (latest release as of this report) — the relevant code in `dash/testing/application_runners.py` is unchanged between 4.4.0 and 4.4.1.

- OS: Windows 11
- Python: 3.14.5
- Not frontend-related (pure `dash.testing` / backend issue)

**Describe the bug**

`dash.testing.application_runners.ThreadedRunner.start()` picks the FastAPI/Quart vs. Flask `app.run()` branch by string-sniffing the server's class module:

```python
module = app.server.__class__.__module__
# FastAPI support
if module.startswith("fastapi"):
    app.run(**options)
# Quart support (ASGI - runs its own async event loop)
elif module.startswith("quart"):
    app.run(**options)
# Flask fallback (WSGI - needs threaded mode)
else:
    app.run(threaded=True, **options)
```

This breaks for any `fastapi.FastAPI` **subclass** defined outside a module literally named `fastapi.*` — which is exactly what several common FastAPI instrumentation/wrapping libraries do, e.g. `opentelemetry-instrumentation-fastapi`'s `FastAPIInstrumentor`, which replaces `fastapi.FastAPI` process-wide with `opentelemetry.instrumentation.fastapi._InstrumentedFastAPI` (a `FastAPI` subclass whose `__module__` is `"opentelemetry.instrumentation.fastapi"`, not `"fastapi.*"`).

When that happens, `ThreadedRunner` falls through to the **Flask** branch and calls `app.run(threaded=True, ...)`. For the FastAPI backend, that `threaded` kwarg flows straight into `uvicorn.Config(...)` (`dash/backends/_fastapi.py`), which doesn't accept it:

```
TypeError: Config.__init__() got an unexpected keyword argument 'threaded'
```

...which `ThreadedRunner` then reports as `DashAppLoadingError: threaded server failed to start`. This makes `dash[testing]` (and anything built on `dash_duo`/`dash_thread_server`, e.g. Playwright-based test suites using `dash.testing.plugin`) unusable for any FastAPI-backend app that has OpenTelemetry auto-instrumentation (or any other subclassing wrapper) applied to it.

**Minimal reproduction** (no OpenTelemetry install required — just simulates the same "FastAPI subclass defined in another module" shape):

```python
import fastapi
from dash import Dash, html
from dash.testing.application_runners import ThreadedRunner


class WrappedFastAPI(fastapi.FastAPI):
    """Stands in for any third-party FastAPI subclass defined in another
    module -- e.g. opentelemetry.instrumentation.fastapi._InstrumentedFastAPI."""


WrappedFastAPI.__module__ = "some_other_package.wrapper"

server = WrappedFastAPI()
app = Dash(__name__, server=server)
app.layout = html.Div("hello")

runner = ThreadedRunner()
runner.start(app, start_timeout=5)  # raises DashAppLoadingError
print("started:", runner.started, runner.url)
runner.stop()
```

Real-world trigger, for reference:

```python
from fastapi import FastAPI
from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor
from dash import Dash

FastAPIInstrumentor().instrument()  # swaps fastapi.FastAPI -> _InstrumentedFastAPI

server = FastAPI()  # now actually an _InstrumentedFastAPI instance
app = Dash(__name__, server=server)
# any dash.testing fixture that goes through ThreadedRunner now fails
```

**Expected behavior**

`ThreadedRunner` should recognize a FastAPI (or Quart) app regardless of what module the concrete class is defined in — e.g. via `isinstance(app.server, fastapi.FastAPI)` / `isinstance(app.server, quart.Quart)` instead of a `__module__` string prefix check. `isinstance` correctly handles subclasses defined anywhere, which a module-name string check fundamentally cannot.

Suggested fix in `dash/testing/application_runners.py::ThreadedRunner.start()`:

```python
try:
    import fastapi as _fastapi
except ImportError:
    _fastapi = None
try:
    import quart as _quart
except ImportError:
    _quart = None

if _fastapi is not None and isinstance(app.server, _fastapi.FastAPI):
    app.run(**options)
elif _quart is not None and isinstance(app.server, _quart.Quart):
    app.run(**options)
else:
    app.run(threaded=True, **options)
```

(`MultiProcessRunner.start()` in the same file has the identical pattern and would benefit from the same fix.)

**Workaround** we're using in the meantime (test-fixture-only, not a real fix): retag the instrumented class's `__module__` before starting the server, e.g.

```python
from opentelemetry.instrumentation.fastapi import _InstrumentedFastAPI
_InstrumentedFastAPI.__module__ = "fastapi.instrumented"
```

...but this is fragile (breaks if OTel's internal class name changes, and mutates third-party class state) and doesn't help anyone not already deep in `dash.testing` internals — hence this report.

cyphercodes:
Fixes #3942.

`ThreadedRunner` and `MultiProcessRunner` determined whether to pass Flask's `threaded=True` option by inspecting the concrete server class module. That misclassified wrapped or subclassed FastAPI/Quart servers whose classes are defined elsewhere.

Use Dash's already-resolved `app.backend.server_type` instead, while preserving the Flask fallback. The regression coverage includes a wrapped FastAPI-shaped server and the FastAPI, Quart, and Flask option paths.

## Contributor Checklist

   - [x] Detect runner behavior from the resolved Dash backend
   - [x] Add regression coverage for wrapped ASGI servers and Flask fallback

### optionals

### Local verification

- `python -m pytest tests/unit/test_app_runners.py -k 'wrapped_fastapi or uses_backend_type' -q` — 4 passed
- `python -m pytest tests/unit/test_testing.py -q` — 1 passed
- Black, Flake8, Pylint, Pyright (production file), and `git diff --check`


dash/testing/application_runners.py
```diff
@@ -28,6 +28,14 @@
 logger = logging.getLogger(__name__)
 
 
+def _run_app(app, options):
+    server_type = getattr(getattr(app, "backend", None), "server_type", "flask")
+    if server_type in ("fastapi", "quart"):
+        app.run(**options)
+    else:
+        app.run(threaded=True, **options)
```
T4rk1n: We lost the comments indicating that we need the threaded=True for flask testing.

dash/testing/application_runners.py
```diff
@@ -28,6 +28,14 @@
 logger = logging.getLogger(__name__)
 
 
+def _run_app(app, options):
+    server_type = getattr(getattr(app, "backend", None), "server_type", "flask")
```
T4rk1n: This is duplicated code at line 221-222, could potentially extract that into a `_server_type` function to prevent drift.

T4rk1n (commented):
Looks good, just some minor comments and need to add a changelog entry.

cyphercodes:
Addressed the review feedback:
- restored the Flask `threaded=True` context comment
- extracted shared backend server type detection to `_server_type()`
- added a CHANGELOG entry for #3944

Verification:
- `python -m pytest tests/unit/test_app_runners.py -q -k 'wrapped_fastapi or run_app_uses_backend_type'`
- `git diff --check`

Note: the full `tests/unit/test_app_runners.py` file still has pre-existing local environment/setup failures in this fresh venv (`dash.dcc._js_dist` missing for smoke setup and process server timeout), so I used the focused backend detection tests above.

Mukller (approved):
Reviewed the refactor — this is the right direction. Module-path sniffing (`__class__.__module__.startswith("fastapi")`) breaks for any wrapped/instrumented server (APM agents, test wrappers), while reading `app.backend.server_type` is semantic and survives decoration. The deduplication into `_run_app`/`_server_type` also removes the two copies of the same ladder that could drift apart. Unit test with the fake `instrumentation.wrapper` module covers exactly the regression scenario. 👍

Two observations, neither blocking:

1. **Unregistered ASGI apps are still misdetected** — a raw FastAPI/Quart app without Dash's `.backend` attribute falls through to the Flask branch and gets `threaded=True`, which `uvicorn.run(**options)` will choke on (`TypeError: unexpected keyword`). Out of scope for this PR since Dash-managed backends always set the attribute, but worth either a docs note in `dash.testing` or a friendlier error when `threaded` reaches an unknown `app.run`.

2. **Follow-up sweep**: are there other `startswith("fastapi")` / `startswith("quart")` ladders left in the codebase (e.g., in `dash/testing/plugin.py` or composite runners)? If yes, a follow-up PR reusing `_server_type` there would finish the job.

Changelog entry format matches the neighboring entries. Nice, tight PR.

T4rk1n (approved):
💃

Issue #3946, opened 2026-08-11
[BUG] Callback chain slows down progressively when a Patch-appended container grows (regression in 4.2.0 vs 4.1.0)


mlipatro:
## Summary

In an app that combines:

1. a **self-retriggering callback chain** (a callback that writes its own `dcc.Store` `Input` via `set_props`),
2. a **`dcc.Interval` polling in parallel**, and
3. a **container appended to via `Patch()` that is never cleared**,

the per-round-trip callback overhead grows steadily with each subsequent run of the chain.

The growing container in our case is a **log view**: a background `dcc.Interval` fetches newly produced log records once per second and appends them to a scrollable `html.Div` with `Patch().append(...)`. The log is intentionally never truncated, so its child count only grows during a page session.

The effect is that a user-visible operation performing a *fixed* number of callback round-trips takes longer every time it is repeated in the same page session — about **5× - 100× slower after 11 repetitions** in the minimal example below, and considerably worse in our real application, where it became unusable after a handful of operations.

This does **not** happen on **Dash 4.1.0**, where per-step overhead stays flat. We had to pin to 4.1.0 because of it.

Disabling the `dcc.Interval` (so nothing is appended to the log view) removes the slowdown entirely — that is how we originally isolated and found the cause, but the problem itself seems to be in the `Patch` itself.

## Expected behavior

Per-round-trip overhead should stay almost roughly constant across repeated runs. Appending to a growing container should cost time proportional to the *appended* nodes, not to the total number of nodes already present, and it should not make unrelated callbacks progressively slower.

## Actual behavior

Per-round-trip overhead drastically grows, monotonically with the number of nodes accumulated in the `Patch`-updated container.

## Minimal reproducible example

```python
import time

from dash import Dash, Input, Output, Patch, callback, ctx, dcc, html, no_update, set_props

STEPS = 50          # callback round-trips per run
LOGS_PER_STEP = 5   # log records appended to the growing container per step

LOG: list[str] = []     # log records produced by the chain, drained by the interval
CURSOR = 0              # how much of LOG the client already has
REPORT: list[str] = []  # measurement history

app = Dash(__name__)
app.layout = html.Div([
    html.Button("Run", id="run", n_clicks=0),
    html.Pre(id="report"),
    dcc.Store(id="store", data={"step": None, "t0": 0, "run": 0}),
    dcc.Interval(id="tick", interval=1000),
    html.Div(id="log", style={"height": "200px", "overflowY": "auto"}),
])


[...]

PR #3948, opened 2026-08-12
Fix O(total) Patch().append() slowdown into a growing container

T4rk1n:
### Summary

`dcc.Patch().append()` (and `.extend()`) into a container that keeps growing gets **progressively slower as the container fills** — a fixed-size append starts costing seconds. This is a regression from 4.2.0, distinct from #3846/#3881 (remount) and #3938 (Patch re-firing callbacks / wiping persistence); none of those fix it.

Root cause: on every patch that updates `children`, two places did O(total-children) work regardless of how much the patch actually changed:
- **`DashWrapper` re-hydrated the *entire* children array** — re-running `Registry.resolve`, prop hydration and recursive child-wrapping for every pre-existing child — even though a ramda `[...old, new]` append leaves the first N children **reference-identical** and unchanged. This dominated (>99% of the wall-clock ramp).
- **`computePaths` rebuilt the whole id→path table** by re-crawling every existing child (asymptotically O(total), though small in absolute terms).

### The fix

- **`DashWrapper.tsx`** — a child that is the *same object reference* as the previous render skips re-hydration and reuses its cached render. Gated to a **settled component reconciling in place** (`Boolean(renderH) && !newRender.current`) — never a first render or a forced remount (`dash.remount()` / identity change), where re-rendering everything below is the intended behavior. Because only reference-identical children are skipped, anything a real change touches (insert, prepend, reorder, replace, prop/persistence write) produces a *new* object at that index and still renders normally.
- **`patchAnalysis.ts` / `patch.ts`** — track per-property "pure tail-append" eligibility (`tailAppends`); any `Insert`/`Prepend`/`Delete`/`Reverse`/nested write disqualifies it.
- **`paths.js`** — `appendPaths()` updates the path table for only the appended tail slice, used when the patch is a provable pure tail-append.

### Measured (append cost for one 250-node append vs. container size)

| existing nodes | before (dev) | after | 4.1.0 ref |
|---:|---:|---:|---:|
| 0 | 95 ms | 82 ms | 66 ms |
| 1000 | 767 ms | 94 ms | 67 ms |
| 2000 | 2419 ms | 85 ms | 82 ms |
| 2750 | 4491 ms | 85 ms | 103 ms |

Flat, back in the 4.1.0 band — **~53× faster at 2750 nodes**. Benchmarked headless (Playwright) on a source build; numbers reproduced on the exact branch build.

### Tests

14 new renderer tests (`patch.test.js`, `paths.test.js`): tail-append gating (Append/Extend counted; Insert/Prepend/Delete/nested-write disqualify) and `appendPaths` ↔ `computePaths` equivalence, including pattern-matching (dict) ids and nested children. Full renderer suite: **50/50 pass**, including #3938's patch/persistence tests.

### Stacking

Stacked on **#3881** (`fix/remounts`, which already contains merged #3938). Base is set to `fix/remounts`; **retarget to `dev` once #3881 merges.**

> Repros the scenario from the (closed-as-dup) #3946 report. Worth confirming against that reporter's app before release.


KoolADE85 (changes requested):
This PR does fix top-level appends, but it does not fix nested patch locations.
This example app is speedy on `4.1.0` but on this branch each click gets slower as children accumulate:

```python
from dash import Dash, Input, Output, Patch, callback, html

app = Dash(__name__)

app.layout = html.Div(
    [
        html.Button("Append 250 children", id="btn"),
        html.Span([html.Span([])], id="container"),
    ]
)


@callback(
    Input("btn", "n_clicks"),
    prevent_initial_call=True,
)
def append_nested(n):
    p = Patch()
    p[0]["props"]["children"].extend(
        [html.Span(f"[batch {n} child {i}] ") for i in range(250)]
    )
    return p


if __name__ == "__main__":
    app.run(debug=True)
```

dash/dash-renderer/src/wrapper/DashWrapper.tsx
```diff
@@ -87,9 +87,22 @@ function DashWrapper({
...
+    const remountedThisRender = useRef(false);
     const renderedIdentity: MutableRefObject<string | null> = useRef(null);
     const hasFreshRendered = useRef(false);
     const renderedPath = useRef<DashLayoutPath>(componentPath);
+    // Previous array value of each children-like prop this component
+    // hydrated, keyed by childrenPath. Lets wrapChildrenProp recognize
+    // items that are the exact same object as last time (eg. every
+    // pre-existing item in a list a Patch only appended to), so they can
+    // keep reconciling in place instead of being forced to re-hydrate their
+    // whole subtree just because a sibling was added.
```
KoolADE85: nit: I don't find these comments very helpful. Maybe can just remove them?

dash/dash-renderer/src/wrapper/DashWrapper.tsx
```diff
@@ -218,18 +234,38 @@ function DashWrapper({
...
+                        // An item that's the exact same object as before
+                        // didn't change - not even props a shallow patch
+                        // analysis might miss - so it doesn't need to be
+                        // forced into a fresh hydrate just because this array
+                        // as a whole did (eg. a sibling got appended, or a
+                        // deeper list this item sits above grew: ramda's
+                        // assocPath rebuilds the array and this item's parent
+                        // on the way to the change, but `concat` keeps the
+                        // pre-existing items themselves the same object).
+                        const unchanged = allowRefSkip && prevArray?.[i] === n;
```
KoolADE85: 🪓

dash/dash-renderer/src/wrapper/DashWrapper.tsx
```diff
@@ -516,7 +552,17 @@ function DashWrapper({
...
+                    : 0,
+                // Check items one by one against the previous render whenever
+                // we're not remounting. Even a component getting a fresh
+                // hydrate (eg. a Patch appended a deep sibling, so its parent
+                // was rebuilt on the path down) can keep the pre-existing
+                // items - the exact same objects - reconciling in place
+                // instead of re-hydrating the whole list every time. A first
+                // render simply has no previous array to match, and a remount
+                // must rebuild everything, so both fall through to a fresh
+                // hydrate.
```
KoolADE85: 🪓

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
...
+- Fix `dcc.Patch().append()` (and `.extend()`) into a growing container getting progressively slower as the container fills — each append re-hydrated the entire children array (re-running `Registry.resolve` and prop hydration for every pre-existing child) and rebuilt the whole id→path table, making a single append cost O(total children) instead of O(appended). Appending 250 nodes to a 3000-node container dropped from ~8.6s back to a flat ~0.2s, matching pre-4.2.0 behavior. This now also covers appending to a *nested* list (eg. `p[0]['props']['children'].extend(...)`), which previously fell back to the full re-hydrate. Children that are the same object reference as the previous render skip re-hydration (unless the component is being remounted), and pure tail-appends — at any depth — update the path table incrementally. Any change that reorders, inserts, replaces or writes a child still re-renders it normally.
```
KoolADE85: I think we should avoid AI generated changelogs - they are incomprehensible to the point of being useless.

```suggestion
- [#3948](https://github.com/plotly/dash/issues/3948) Fix page getting progressively slower as callbacks append children
```

KoolADE85 (approved):
This is working well (even faster than `4.1.0`!) on my end.
Just a few nits around useless AI text.

CHANGELOG.md
T4rk1n: Yes, think I might have used opus 5 on Hermes for this run, it would explain the gibberish comments.

PR #3950, opened 2026-08-14
Add dedicated Graph config schema test

### LINKED ISSUE #3735: Extract graph schema validation into properly named test
In test: test_tagr002_tabs_render_without_selected we do

```
    # do some extra tests while we're here
    # and have access to Graph and plotly.js
    check_graph_config_shape(dash_dcc)
```

This check that we have all the props defined, but the test is random and parametrized so the same check is run twice. It should be in it's own test.

Nice6042:
Closes #3735

Adds a dedicated Graph integration test that recursively compares the live Plotly config schema with the generated `dcc.Graph` config metadata. Known Plotly-internal paths remain excluded, and assertion messages identify the exact config path when the schemas drift.

Tests:
- `python -m pytest components/dash-core-components/tests/integration/graph/test_graph_basics.py --headless -q` — 14 passed, 1 skipped
- `python -m black --check components/dash-core-components/tests/integration/graph/test_graph_basics.py` — passed

## Contributor Checklist

   - [x] add a dedicated Graph config-schema test
   - [x] verify schema drift produces a path-specific failure

### optionals

### DIFF
--- a/components/dash-core-components/tests/integration/graph/test_graph_basics.py
+++ b/components/dash-core-components/tests/integration/graph/test_graph_basics.py
@@ -11,6 +11,52 @@ from dash import Dash, Input, Output, dcc, html
 import dash.testing.wait as wait
 
 
+def test_grbs013_graph_config_matches_plotly_schema(dash_dcc):
+    app = Dash(__name__)
+    app.layout = dcc.Graph(id="graph")
+    dash_dcc.start_server(app)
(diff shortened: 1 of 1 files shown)


T4rk1n (commented):
Test looks good, but I am not sure this should run along with the rest of tests, but rather a weekly job on ci that alerts us if it fails by creating an issue.
Let me know if you want to do that or prefer if you want I can set that up.

PR #3954, opened 2026-08-18
Performance benchmark suite + O(1) wildcard resolution + faster layout crawl

T4rk1n:
Stacked on #3948 (base branch `perf/patch-append-rehydration`), so the diff here is only the benchmark work and the two optimizations it surfaced. Review/merge #3948 first.

### 1. A platform performance benchmark harness (`benchmarks/`)
Standalone, **not** part of the pytest suite — timing is noisy, so it reports rather than flaking the test matrix. 10 scenarios across the platform (initial/deep hydration, Patch append top-level + nested, scalar Patch update, full-list replacement, callback fan-out, `ALL`-wildcard resolution, deep callback chain). Each runs in its own subprocess against the **production** bundle, driven by headless Chrome, timed with in-page `performance.now()`, aggregated as median/p90/max plus a `growth` ratio that flags O(total) creep. `--profile <scenario>` captures a Chrome CPU profile + a hottest-functions table.

### 2. CI job (`.github/workflows/benchmarks.yml`)
Runs on PRs touching `dash/`, `benchmarks/`, or components. Builds the production renderer, runs vs the committed `baseline.json`, **hard-fails only on an order-of-magnitude regression**, **warns without failing** on smaller drift, and **upserts a sticky PR comment** with the results table.

### 3. Wildcard resolution O(n²) → O(1) — found by the harness
`getPath` for a pattern-matching (dict) id did a linear deep-equality scan over every component sharing the id's key set, called once per resolved component — so an `ALL` dispatch over N components was O(N²). Added an `objIndex` hash map (`valuesKey → path`) maintained inline in `computePaths` (copy-on-write) and `appendPaths`; the ordered `objs` array stays for pattern iteration. **~580ms → ~140ms (≈4x)** on `ALL` over 400 components.

### 4. Ramda currying overhead in the per-node layout crawl — also found by the harness
`crawlLayout` (run on every component on every path recompute + callback gather) used curried `path`/`pathOr` per node. Replaced with direct property access on the hot common path. Same-machine A/B: **patch_append ~16%**, initial render + wildcard a few %. No behavior change.

`.ai/PERFORMANCE.md` documents the methodology, how to profile, and the findings (including two that are *not* fixed here: `callback_chain` is network-bound, and layouts deeper than ~250 fail to serialize).

## Testing
- Renderer unit: 55/55 (adds nested/paths coverage).
- Integration green across `test_wildcards`, `test_basic_callback`, `test_multiple_callbacks`, `test_layout_paths_with_callbacks`, `test_patch`, `test_children_reorder`.
- Both optimizations verified with same-machine A/B benchmark runs.

## Note
`baseline.json` was generated locally; absolute thresholds work anywhere, but the baseline-ratio comparison is best regenerated on `ubuntu-latest` (adopt the first CI run's `results.json`).


KoolADE85 (approved):
Nice addition! 💃

PR #3955, opened 2026-08-18
Unpin selenium to fix CI flakiness against current Chrome

T4rk1n:
## Problem

Recent pushes and PRs have been going red intermittently across the browser-based integration tests. The failures are **scattered across unrelated Selenium tests** (`test_persistence`, `test_csp`, `test_multi_output`, `test_derived_props`, async callbacks, table server tests…), a *different* subset each run, all surfacing as `StaleElementReferenceException` / `TimeoutException`. That pattern is environmental flakiness, not a bad merge (which would fail the same test deterministically).

## Root cause

- `requirements/testing.txt` pinned `selenium>=3.141.0,<=4.2.0` (selenium 4.2.0 is from 2022).
- Every CI browser job uses `browser-actions/setup-chrome@v1` with `chrome-version: stable` — **unpinned** — so CI now installs **Chrome 151**.
- Selenium 4.2 predates **Selenium Manager** (added in 4.6), so it cannot reliably provision/drive current Chrome, producing timing/staleness flakiness.

Two amplifiers: the dependabot pip bump that would have raised selenium never landed on `dev`, and the recent React 18/19 test matrix roughly doubled the browser shards, so a single flake reddens the whole run more often.

## Fix

Bump the pin to `selenium>=4.11.0,<=4.46.0`. The `>=4.11.0` floor guarantees a mature Selenium Manager that auto-provisions a chromedriver matching whatever stable Chrome CI installs (this is why `install-chromedriver: false` in the setup step remains correct).

## Compatibility checks

- No removed `find_element_by_*` APIs anywhere in `dash/` (those were dropped in selenium 4.3).
- Driver construction already uses the modern API: `webdriver.Chrome(options=...)` / `webdriver.Remote(command_executor=..., options=...)`.
- Verified the full API surface `dash/testing/browser.py` uses, plus Selenium Manager availability, against selenium 4.46.0.

## Follow-ups (not in this PR)

- A few tests exhaust all 3 reruns (e.g. `test_async_cbsc001_simple_callback`) and may be genuinely broken rather than flaky — worth a targeted look once this settles the noise.
- Optionally pin Chrome for fully reproducible runs; modern selenium tracks `stable` fine either way.


T4rk1n:
Follow-up commit `1845c32`: the first CI run surfaced two **deterministic** breaks that the old `selenium<=4.2.0` cap had been masking (both now fixed):

1. `dash/testing/browser.py` set the Firefox `marionette` capability, which modern selenium/geckodriver reject (`InvalidArgumentException: marionette is not the name of a known capability`). It's the implicit protocol now — removed. This is what failed the **Lint & Unit** jobs (`test_browser_smoke[Firefox]`).
2. Three test modules used the `find_element(s)_by_*` helper methods that selenium removed in 4.3 (`AttributeError: 'WebElement' object has no attribute 'find_element_by_tag_name'`). Migrated to `find_element(s)(By.*, ...)`. This failed a **Main Dash** Chrome group.

Remaining `StaleElementReferenceException` flakes in that first run should be resolved by the selenium bump itself (matching driver for Chrome 151). Re-running CI to confirm.

T4rk1n:
## CI verified — systemic flakiness fixed

Full run + targeted re-runs of the flaky jobs. The two structural causes of the cross-PR flakiness are resolved:

- **selenium/Chrome mismatch** — the pinned selenium 4.2.0 couldn't drive the current stable Chrome (151); bumped to `>=4.11.0` (Selenium Manager auto-provisions a matching driver). Exposed and fixed four deterministic selenium-4 breaks along the way: `marionette` capability, removed `find_element_by_*`, `move_to_element_with_offset` center-origin, and `TimeoutException.args[0]`→`.msg`.
- **"Setup virtual display" hang** — the step ran `apt-get install -y xvfb`, which intermittently blocked on the runner's dpkg lock. xvfb is preinstalled on the runners, so the apt-get calls were removed; the step now just starts Xvfb. 26 display-step instances ran with zero hangs.

### Remaining reds are pre-existing flakes, not regressions
On re-run, the DCC and Main Dash failures (`test_rdcap003`, `test_msmh003` markdown re-render race, `test_inni004`) all passed — confirmed flaky. The one stubborn failure, `test_async_cbsc001_simple_callback` (`AssertionError: initial count + each key stroke`), **is already failing on `dev` HEAD** (run 32175191617), so it predates this branch.

Recommend merging this PR (it strictly improves CI) and tracking the async-callback test failure + the racy per-test flakes in a separate issue.

T4rk1n:
## ✅ Full green run

Run [32283425251](https://github.com/plotly/dash/actions/runs/32283425251) passed with zero failed jobs after the anti-hang + gating fixes.

**The two systemic problems are gone:**
- **No more hangs.** The 'Run Async Callback Tests' step was wedging until the job timeout. Root cause: `ThreadedRunner.stop()`'s Flask path did an unbounded `self.thread.join()`; if the SystemExit thread-kill fails to unwind a worker, teardown blocks forever. Bounded it (FastAPI/Quart already were). Added `pytest-timeout` (180s/test, via the `[ci]` extra) so any future hang fails with a stack dump naming the test instead of stalling the step.
- **Callback-count flakes fixed.** cbsc001/cbsc008 (async + sync) asserted an exact one-callback-per-keystroke count the renderer never guaranteed — it coalesces same-identity callbacks still queued in `requested` state. Rewrote them to gate each keystroke on the previous callback having executed (`until(...)`), so coalescing can't happen; dropped the Lock/sleep/@flaky. Async step now `26 passed`, background `36 passed`.

Remaining occasional reds (`test_lcbc017`, `test_persistence`, `test_rdcap003`, `test_dveh002`) are the pre-existing low-frequency flake tail — they cleared on re-run and predate this branch.

Optional follow-up (not in this PR): add `if: ${{ !cancelled() }}` to the Async step so a Background flake stops skipping it, and the graceful-Flask-shutdown change to cut the teardown-kill noise at the source.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3646](https://github.com/plotly/dash/pull/3646) Remove React 16 support (`16.14.0` is no longer an accepted value for `REACT_VERSION` / `_set_react_version`).
 
 ### Fixed
+- Unpin `selenium` in the testing requirements (was capped at `<=4.2.0`, from 2022) and require `>=4.11.0`. The old cap predated Selenium Manager, so the pinned selenium could not drive the current stable Chrome (151+) that CI installs, producing widespread `StaleElementReferenceException`/`TimeoutException` flakiness across the browser-based integration tests. Modern selenium auto-provisions a matching chromedriver, restoring stable CI runs.
```
camdecoster: Are these entries supposed to be sorted by the PR number?

```suggestion
- [#3955](https://github.com/plotly/dash/pull/3955) Unpin `selenium` in the testing requirements (was capped at `<=4.2.0`, from 2022) and require `>=4.11.0`. The old cap predated Selenium Manager, so the pinned selenium could not drive the current stable Chrome (151+) that CI installs, producing widespread `StaleElementReferenceException`/`TimeoutException` flakiness across the browser-based integration tests. Modern selenium auto-provisions a matching chromedriver, restoring stable CI runs.
```

camdecoster (approved):
Looks fine. Would it be worth pinning Chrome to keep a stable test environment?

T4rk1n:
> Looks fine. Would it be worth pinning Chrome to keep a stable test environment?

I'd rather not pin chrome, when it's pinned it accumulates regressions we don't see on tests and it takes longer to fix afterwards (This took 3 days because selenium was pinned for so long).

PR #3957, opened 2026-08-23
Fix running= with an ALL wildcard crashing when no components match

### LINKED ISSUE #3297: [BUG] Using pattern matching ALL for running argument fails when the list is empty
When using pattern matching for the `running` argument (ex: `running=[(Output({'type': 'loading', 'id': ALL}, 'display'), 'show', 'hide')]`), if there are no components on the page dash throws the error: `state.paths.objs[idKey] is undefined`

**Describe your context**

```
dash               3.0.4
```

**Describe the bug**

When using pattern matching for the `running` argument (ex: `running=[(Output({'type': 'loading', 'id': ALL}, 'display'), 'show', 'hide')]`), if there are no components on the page dash throws the error: `state.paths.objs[idKey] is undefined`

**Expected behavior**

No error, no side updates (since there are no components)

**Minimum Working Example**

```
import dash
from dash import Dash, html, dcc, callback, Output, Input, Patch, ALL, callback_context
import time
import random


app = Dash(use_pages=True, pages_folder="")

home_page_layout = html.Div([
    dcc.Loading(
        html.Div(id={'type':'scorecard', 'id': 'home1'}),
        id={'type': 'loading', 'id': 'home1'}
    ),
    dcc.Loading(
        html.Div(id={'type':'scorecard', 'id': 'home2'}),
        id={'type': 'loading', 'id': 'home2'}
    )
])
other_page_layout = html.Div('Other')

dash.register_page('home', path='/', layout=home_page_layout)
dash.register_page('blank', path='/blank', layout=other_page_layout)

app.layout = html.Div([
    dcc.Store(id='data'),
    html.Div([
        html.Span([
            dcc.Link(f"{page['name']} - {page['path']}", href=page["relative_path"]), ' | '
        ]) for page in dash.page_registry.values()
    ]),
    html.Button('Reload Data', id='reload-data'),
    dash.page_container
])


@callback(
    Output('data', 'data'),
    Input('reload-data', 'n_clicks'),
    prevent_initial_call=True,
    running=[(Output({'type': 'loading', 'id': ALL}, 'display'), 'show', 'hide')]
)
def get_data(n_clicks):
    time.sleep(3)
    return random.randint(0, 10)

@callback(
    Output({'type': 'scorecard', 'id': 'home1'}, 'children'),
    Input('data', 'data'),
    prevent_initial_call=True
)
def scorecard1(data):
    return f'Data: {data}'

@callback(
    Output({'type': 'scorecard', 'id': 'home2'}, 'children'),
    Input('data', 'data'),
    prevent_initial_call=True
)
def scorecard2(data):
    return f'Data squared: {data * data}'

if __name__ == '__main__':
    app.run(debug=True)
```

In this example, pressing the "Reload Data" button on the second page (/blank) throws the error as the loading elements are all on the first page

Mohit-Ak:
Callbacks that use a pattern-matching id in `running=` crash the renderer as soon as you're on a page where none of those components exist. From #3297:

```python
@callback(
    Output('data', 'data'),
    Input('reload-data', 'n_clicks'),
    running=[(Output({'type': 'loading', 'id': ALL}, 'display'), 'show', 'hide')],
)
```

Click the button while you're on the home page and it works. Navigate to `/blank`, which renders none of the `{'type': 'loading', ...}` components, click again, and the browser console throws `state.paths.objs[idKey] is undefined`. The callback never gets to run.

## What was going wrong

`getAllPMCIds` indexes `state.paths.objs[idKey]` and calls `.map` on the result. `paths.objs` only has a key for an id shape that is actually rendered somewhere, so on a page with none of those components the lookup returns `undefined` and `.map` throws. The two other places that do this same lookup — `resolveDeps` in `dependencies_ts.ts` and `getPath` in `paths.js` — both guard it and return an empty result, so this was the odd one out. Guarding it the same way is the direct fix.

That turned out not to be the whole story though. `replacePMC` decides whether a wildcard was expanded by checking `extras.length`:

```js
if (extras.length) {
    return extras;
}
return [replaced];
```

An `ALL` that legitimately matches zero components expands to an empty list, which is falsy, so control falls through to `return [replaced]` — and `replaced` only ever received the *non-wildcard* keys, because the wildcard branch writes to `extras` instead. So once the crash is gone you get `[{type: 'loading'}]` back: an id with the `id` key missing entirely. `sideUpdate` then treats that as a real component id and tries to update it. So I tracked expansion with an explicit flag rather than inferring it from the length, and made `sideUpdate` drop pattern-matching outputs that resolve to nothing, since in that case `componentId` still holds the unresolved pattern and isn't safe to use.

The "matches nothing" case is the correct outcome here, not an error — there is simply nothing to show or hide, and the callback should proceed normally. That's also why this doesn't route through the existing "ID running component not found in layout" error: that one is for a concrete id that isn't in the layout, whereas a wildcard matching zero components is expected behaviour when you're on a different page.

## Testing

Added `dash/dash-renderer/tests/patternMatching.test.js`, which had no unit coverage before. It covers `MATCH`, `ALL`, plain concrete ids, and the two empty-page cases.

Before the source change, with the tests in place:

```
getAllPMCIds returns an empty list when no component uses that id shape FAILED
	TypeError: Cannot read properties of undefined (reading 'map')
	    at getAllPMCIds (build/commons.js:3121:34)
replacePMC yields no ids when ALL matches nothing on the current page FAILED
	TypeError: Cannot read properties of undefined (reading 'map')
replacePMC yields no ids when ALLSMALLER matches nothing on the current page FAILED
	TypeError: Cannot read properties of undefined (reading 'map')
TOTAL: 3 FAILED, 40 SUCCESS
```

That's the same error as the reported one, and the `MATCH`/`ALL`/concrete-id cases passed throughout, which is what confines the bug to the empty-match path.

After:

```
$ npm test
TOTAL: 43 SUCCESS
```

Lint is clean on the touched files (`eslint` 0 errors, `prettier --list-different` empty). `tsc --noEmit` reports the same set of pre-existing `node_modules` type errors before and after — none from these files.

Fixes #3297


camdecoster:
Thanks for the PR! Our team will take a look and provide some feedback.

dash/dash-renderer/src/actions/callbacks.ts
```diff
@@ -443,16 +443,25 @@ function sideUpdate(outputs: SideUpdateOutput, cb: ICallbackPayload) {
...
                     [componentId, propName] = parsePMCId(id);
                     replacedIds = replacePMC(componentId, cb, i, getState);
+                    isPatternMatching = true;
                 } else if (id.includes('.')) {
                     [componentId, propName] = id.split('.');
                 }
 
                 const props = propName ? {[propName]: value} : value;
 
+                if (isPatternMatching && replacedIds.length === 0) {
```
T4rk1n, line 458: This early return is the half of the fix that keeps the unresolved pattern (`{type: 'loading'}` with no `id`) out of `updateComponent`, and it has no test. The unit tests stop at `replacePMC` returning `[]` and never reach this branch. Add an integration test modeled on `test_cbwc008_running_match` that reproduces #3297 directly: multi-page app, navigate to a page with none of the `ALL` components, click, then assert the console is clean and the callback output updated. That exercises this line and proves the actual crash is gone.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3646](https://github.com/plotly/dash/pull/3646) Remove React 16 support (`16.14.0` is no longer an accepted value for `REACT_VERSION` / `_set_react_version`).
 
 ### Fixed
+- [#3957](https://github.com/plotly/dash/pull/3957) Fix a `running` argument using a pattern-matching `ALL`/`ALLSMALLER` id crashing the renderer with `state.paths.objs[idKey] is undefined` when none of the matching components are on the current page (for example after navigating to another page in a multi-page app). The wildcard now resolves to an empty set of components and the callback proceeds without any side updates. Fixes [#3297](https://github.com/plotly/dash/issues/3297).
```
T4rk1n, line 20: Trim to two lines. Something like: "Fix `running=` with a pattern-matching `ALL`/`ALLSMALLER` id crashing the renderer when none of the matching components are on the current page. Fixes #3297." The navigation example and the resolve-to-empty detail belong in the commit body or `.ai/`.

T4rk1n (commented):
Fix is right, but the fix isn't fully tested yet. The diagnosis is accurate: `getAllPMCIds` was the one caller of `state.paths.objs[idKey]` that didn't guard the lookup (`resolveDeps` and `getPath` both do), and the `extras.length` overload in `replacePMC` did fall through to a partial id on a zero match. The explicit `expanded` flag is the correct split between "a wildcard expanded" and "the expansion was empty," and all the side-update kinds (`running`, `runningOff`, `progress`, `set_props`) go through the same guard, so the empty match is handled symmetrically.

The gap is coverage. This is two changes and only one is tested. The new unit tests exercise `getAllPMCIds`/`replacePMC`, but the `callbacks.ts` guard, the half that actually stops the unresolved pattern from reaching `updateComponent`, has no test at all, and nothing here reproduces #3297 end to end. The reported failure is a navigate-then-click crash in a running app; the PR proves the helpers in isolation but never proves the crash is gone. Please add a renderer integration test next to `test_cbwc008_running_match`: a two-page app, navigate to a page that renders none of the wildcard components, click the button, then assert the browser console is clean and the callback's own output still updates. That fails before this change and passes after, and it covers the `callbacks.ts` early return, not just the helpers.

Also tighten the CHANGELOG entry to two lines (house style, detail goes in `.ai/`): the crash plus the fix, drop the parenthetical.

PR #3960, opened 2026-08-25
Bugfix: Fix dcc.Markdown issues with <dccLink />

### LINKED ISSUE #3951: [BUG] dcc.Markdown with a `<dccLink children="..." />` does not render children
As of v4.4.0 `dcc.Markdown` does not render dccLinks properly when they use a `children` prop. The link is rendered using the `href` instead.

With this code:
```
dcc.Markdown(
    '<dccLink href="/page-2" children="Go to page 2" />',
    dangerously_allow_html=True,
),
```

Prior to v4.4.0: 
Image

On v4.4.0:

Image

This is a result of upgrading `react-jsx-parser` (1.21.0 → 2.4.1) which changed its behaviour under the hood.

KoolADE85:
`dcc.Markdown` used dependencies that were fundamentally incompatible with our `<dccLink />` implementation + React 19 deps.
This PR rewrites the Markdown component to use markdown internals that are compatible.
Along the way, it's converted to Typescript.

`dcc.Link` is also converted to Typescript (no functional change) so as not to introduce TS compile errors.

Testing notes:
* while this is a rewrite, I intend for this to be included in a minor release: You can compare against the old `Markdown.react.js` implementation for parity
* I added plenty of missing integration tests

Fixes #3951 

## Contributor Checklist

### optionals


components/dash-core-components/src/fragments/Markdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const renderMarkdown = (text: string) => (
+        <MarkdownContent
+            dedent={dedent}
+            dangerously_allow_html={dangerously_allow_html}
+            mathjax={mathjax}
+            link_target={link_target}
+        >
+            {text}
+        </MarkdownContent>
+    );
```
T4rk1n: Think we can put this in a useCallback

components/dash-core-components/src/utils/LazyLoader/third-party/hljs.ts
```diff
@@ -0,0 +1,17 @@
...
+
+declare global {
+    interface Window {
+        hljs?: HLJSApi;
+    }
+}
+
+// Reuse a highlight.js instance already on `window` (e.g. supplied by the host
+// page), otherwise lazily load the bundled third-party build.
+export default (): Promise<HLJSApi> =>
```
T4rk1n: Put a name for the export.

components/dash-core-components/src/utils/LazyLoader/third-party/mathjax.ts
```diff
@@ -0,0 +1,20 @@
...
+    config: {startup: {typeset: boolean}};
+}
+
+declare global {
+    interface Window {
+        MathJax?: MathJaxObject;
+    }
+}
+
+export default (mathjax?: boolean): Promise<MathJaxObject | undefined> =>
```
T4rk1n: Name for function.

components/dash-core-components/src/fragments/Markdown.tsx
KoolADE85: Done!

T4rk1n (approved):
💃

PR #3852, opened 2026-06-29
Enable  callbacks for Plotly hoveranywhere and clickanywhere events

AnnMarieW:
This PR adds support for Plotly hoveranywhere and clickanywhere events in dcc.Graph by including `xvals` and `yvals` in the `clickData` and `hoverData` props.  Also adds  xaxis_id  and yaxis_id so this can be used with subplots.

closes #3851

To do (changes needed in plotly.js):
 1. Add pixel coordinates to make it possible to use dcc.Tooltip to create custom tooltips.    In the [PR 7707](https://github.com/plotly/plotly.js/pull/7707/changes#diff-754e3dde48df955de63f6c76eb1c43fed92a3931b9d7f51035a2d400029c8063)  to add this feature in Plotly.js, , `xPixel` and `yPixel` are added to the event data, but it's only available on points.   You can see this when running the sample app below.
 2. It's not currently possible to clear the hover data on unhover - it always shows the last hovered point.   I'd like to suggest adding an unhover event in plotly.js when the cursor leaves the plot area.

-------------------------

Here's a minimal app for testing:

```python

import json
from dash import Dash, dcc, html, Input, Output, callback
import plotly.express as px
import plotly
print("plotly version installed:", plotly.__version__,  "hoveranywhere requires >=6.7" )

df = px.data.iris()
fig = px.scatter(df, x="sepal_width", y="sepal_length", color="species")
fig.update_layout(hoveranywhere=True, clickanywhere=True)

app=Dash()

styles = {
    'pre': {
        'border': 'thin lightgrey solid',
        'overflowX': 'scroll'
    }
}

app.layout = html.Div([
    dcc.Graph(
        id='basic-interactions',
        figure=fig,
    ),

    html.Div([
        html.Div([
            dcc.Markdown("""
                **Hover Data**

                Mouse over values in the graph.
            """),
            html.Pre(id='hover-data', style=styles['pre'])
        ], className='three columns'),

        html.Div([
            dcc.Markdown("""
                **Click Data**

                Click on points in the graph.
            """),
            html.Pre(id='click-data', style=styles['pre']),
        ], className='three columns'),
    ])
])


@callback(
    Output('hover-data', 'children'),
    Input('basic-interactions', 'hoverData'))
def display_hover_data(hoverData):
    return json.dumps(hoverData, indent=2)


@callback(
    Output('click-data', 'children'),
    Input('basic-interactions', 'clickData'))
def display_click_data(clickData):
    return json.dumps(clickData, indent=2)


[...]

PR #3961, opened 2026-08-27
Include xPixel and yPixel in hoverdata and clickdata in `dcc.Graph`

AnnMarieW:
This is a continuation of PR #3852 

> To do (changes needed in plotly.js):

> - Add pixel coordinates to make it possible to use dcc.Tooltip to create custom tooltips. In the PR 7707 to add this feature in Plotly.js, , xPixel and yPixel are added to the event data, but it's only available on points. You can see this when running the sample app below.
> - It's not currently possible to clear the hover data on unhover - it always shows the last hovered point. I'd like to suggest adding an unhover event in plotly.js when the cursor leaves the plot area.

Both of these issues were fixed in https://github.com/plotly/plotly.js/pull/7966  and are now available in Plotly 7. Thanks @emilykl :slightly_smiling_face: 

The Plotly.js PR #7966 adds `xPixel` and `yPixel` (the cursor coordinates) to all hover and click events, regardless of whether `hoveranywhere` or `clickanywhere` are enabled.

This PR includes `xPixel` and `yPixel` data in the `dcc.Graph` `hoverData` and `clickData` props only when  `hoveranywhere` or `clickanywhere` option is enabled in the figure. This avoids triggering Dash callbacks when the cursor coordinates aren't needed.

I'll provide some sample apps as a separate comment.


AnnMarieW:
@T4rk1n - The test fails because it requires Plotly 7 -  what's the best way to handle that?

T4rk1n:
> @T4rk1n - The test fails because it requires Plotly 7 - what's the best way to handle that?

You can either install plotly.py 7.0.0 to install it globally (Probably best), or if it's just a few tests can put the plotly.js version in assets and set it for the test.

AnnMarieW:
Here's a sample app - the `xPixel` and `yPixel` are used to create a floating custom tooltip with `dcc.Tooltip`:


image

image


```python
import numpy as np
import pandas as pd

app = Dash(__name__)

np.random.seed(42)
dates = pd.date_range("2024-01-01", periods=500, freq="B")

price_changes = np.random.normal(0, 1.5, len(dates))
prices = 100 + np.cumsum(price_changes)

df = pd.DataFrame(
    {
        "date": dates,
        "price": prices,
    }
)

fig = go.Figure(
    go.Scatter(
        x=df["date"],
        y=df["price"],
        mode="lines",
        name="Stock",
    )
)

fig.update_layout(
    hoveranywhere=True,
    xaxis=dict(
        title="Date",
        showspikes=False,
    ),
    yaxis=dict(
        title="Price",
        showspikes=True,
        spikemode="toaxis+across",
        spikesnap="cursor",
        spikecolor="#FF4136",
        spikethickness=1.5,
    ),
)

app.layout = html.Div(
    [
        dcc.Graph(
            id="stock-chart",
            figure=fig,
            clear_on_unhover=True
        ),
        dcc.Tooltip(
            id="tooltip",
            direction="top",
        ),
    ]
)


@app.callback(
    Output("tooltip", "show"),
    Output("tooltip", "bbox"),
    Output("tooltip", "children"),
    Input("stock-chart", "hoverData"),
)
def update_tooltip(hover_data):
    if not hover_data:
        return False, {}, None

    price = hover_data["yvals"][0]

    count = (df["price"] > price).sum()

    children = html.Div(
        [
            html.Div(
                f"Price: ${price:.2f}",
                style={"fontWeight": "bold"},
            ),
            html.Div(
                f"Times above this price: {count}",
            ),
        ]
    )

    return (
        True,
        {
            "x0": hover_data["xPixel"],
            "x1": hover_data["xPixel"] + 10,
            "y0": hover_data["yPixel"],
            "y1": hover_data["yPixel"] + 10,
        },
        children,
    )


if __name__ == "__main__":
    app.run(debug=True)


```

AnnMarieW:
@T4rk1n - Looks like this is ready for review now :slightly_smiling_face:

T4rk1n (approved):
💃

PR #3973, opened 2026-09-04
Run every CI job when shared dependency or config files change

T4rk1n:
Follow-up to #3972. The dependabot bump in #3966 went green on its PR because the Table, DCC, HTML and Backend Callback jobs were skipped by the path filters: `requirements/**` was only listed for the background, websocket and streaming filters, while every job installs from those files. The flake8 F824 and mimesis Python 3.9 failures only showed up after the merge to dev.

This adds a `shared_paths` list that every filter includes (dorny/paths-filter supports YAML anchors for this):

- `requirements/**`, `setup.py`
- `pytest.ini`, `.flake8`, `.pylintrc`
- `package.json`, `package-lock.json`
- `dash/testing/**` (the harness all component tests use)
- `.github/workflows/testing.yml`

Since this PR touches the workflow file itself, its own run exercises every job.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -28,28 +28,47 @@ jobs:
         id: filter # Give an ID to this step to reference its outputs
         uses: dorny/paths-filter@v3
         with:
+          # Every job installs from requirements/ and setup.py and runs under
+          # the shared lint and pytest config, so a change there must run every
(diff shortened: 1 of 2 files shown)


KoolADE85 (approved):
Looks fine here.
I'm expecting that the CI issues will resolve themselves once we merge and sync #3972

PR #3976, opened 2026-09-05
Add configurable scroll-to-top behavior to dcc.Link

### LINKED ISSUE #3974: [Feature Request] Condition dcc.Link's window.scrollTo(0,0) on a component prop
Occasionally a dcc.Link will update the URL, though only some portion of the layout will be updated. However, the page will still scroll to the top.

[dcc.Link currently calls](https://github.com/plotly/dash/blob/731afc926ad236e55f38548cbbd07b8e7dfbc62d/components/dash-core-components/src/components/Link.tsx#L16-L44) window.scrollTo(0,0) unconditionally. It could be nice for that to be conditional, perhaps controlled by a prop.

None.

**Additional context**
None.

CNFeffery:
This PR makes the automatic scroll-to-top behavior of `dcc.Link` configurable.

- Add a new boolean `scrollToTop` prop that defaults to `true`, preserving the existing behavior.
- Only call `window.scrollTo(0, 0)` when `scrollToTop` is enabled.

Closes #3974


AnnMarieW:
Nice PR!
Just chiming in here...

What do you think about naming the prop `scrollTo` (or `scroll_to`) and allowing the arguments supported by [Element.scrollTo](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo) ?

It could default to the current behavior, `scrollTo(0, 0)`, and also allow the options dict with `top`, `left`, and `behavior`.

CNFeffery:
> Nice PR! Just chiming in here...
> 
> What do you think about naming the prop `scrollTo` (or `scroll_to`) and allowing the arguments supported by [Element.scrollTo](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo) ?
> 
> It could default to the current behavior, `scrollTo(0, 0)`, and also allow the options dict with `top`, `left`, and `behavior`.

Since for the `dcc.Link` component, its built‑in behavior policy only has two states: whether to automatically scroll back to the top. Introducing more complex scroll‑position parameter controls does not seem necessary at this stage.

T4rk1n (approved):
Looks good.

PR #3977, opened 2026-09-07
Add Patch-style partial reads to WebSocket get_prop

### LINKED ISSUE #3975: [Feature Request] Support Patch-style partial reads in WebSocket get_prop()

WebSocket `get_prop()` already provides a convenient way to read a component property during a callback. It currently returns the complete property value, which works well as a general default.

For large object- or list-shaped props, it could be useful to also support a more flexible partial-read option when only one nested value is needed. For example, a custom button might expose a detailed `onClickEvent` object, while the callback only needs:

~~~python
onClickEvent["node1"]["node1-1"]["target"]["value"]
~~~

With the current API, the callback retrieves the complete value and then selects this node in Python. An optional client-side path could make this use case more efficient by returning only the requested part.

If feasible, it would be helpful for `get_prop()` to accept an optional partial-read path using the same location types already supported by Dash's [`Patch`](https://dash.plotly.com/partial-properties): string keys and integer list indices.

For example:

~~~python
# Nested object value
value = await ws.get_prop(
    "button",
    "onClickEvent",
    path=["node1", "node1-1", "target", "value"],
)

# A list item nested inside an object
record = await ws.get_prop(
    "button",
    "onClickEvent",
    path=["node1", "records", 5],
)
~~~

The browser could resolve the path before serialization and return only the selected value or subtree over the WebSocket.

Potential behavior:

- Support nested combinations of string keys and integer indices, matching `Patch` location semantics.
- Keep the current full-property behavior when `path` is omitted.
- Preserve the existing asynchronous and `timeout` behavior.
- Document the behavior for missing keys or invalid indices.

The existing approach of retrieving the complete prop and selecting the value in Python remains straightforward. A clientside callback and a small `dcc.Store` can also provide similar behavior. A built-in optional path would offer a more concise approach for this use case.

**Additional context**

- [WebSocket callbacks — Reading Props with `get_prop`](https://dash.plotly.com/websocket-callbacks#reading-props-with-get_prop)
- [Partial Property Updates and the `Patch` class](https://dash.plotly.com/partial-properties)

This could complement the existing `get_prop()` API with a focused read-side counterpart to `Patch`: optionally retrieve a nested location without returning the complete property.

CNFeffery:
This PR adds optional Patch-style partial reads to WebSocket `get_prop()`.

- Add a keyword-only `path` argument that accepts string keys and integer list indices, including negative indices.
- Resolve the selected value client-side before WebSocket serialization, while preserving existing full-property reads when `path` is omitted, `None`, or `[]`.
- Return `None` for missing or structurally incompatible locations; reject malformed path arguments before sending a request.
- Add unit and WebSocket integration coverage, and document the API in the architecture guide.

### Usage examples

```python
# Existing behavior: read the complete property.
data = await ws.get_prop("store", "data")

# Read a nested object value.
value = await ws.get_prop(
    "store", "data", path=["event", "target", "value"]
)

# Read a list item by its zero-based index.
first_value = await ws.get_prop(
    "store", "data", path=["records", 0, "value"]
)

# Read from a list; negative indices count from the end.
last_value = await ws.get_prop(
    "store", "data", path=["records", -1, "value"]
)

# A string key is not coerced to a list index, so this returns None.
not_found = await ws.get_prop(
    "store", "data", path=["records", "0"]
)
assert not_found is None

# An explicit empty path is also a full-property read.
data = await ws.get_prop("store", "data", path=[])
```

`path` accepts string dictionary keys and integer list indices. A returned
`None` may represent a real `null` value, a missing location, or a
structurally incompatible path. Object keys must be strings and list indices
must be integers (so `"0"` is not coerced to `0`); malformed path arguments
raise before a request is sent.

Closes #3975


T4rk1n (commented):
Nice, looks good, CI is currently broken, I'll finish the review once it's green.

T4rk1n (approved):
Looks good!

PR #3980, opened 2026-09-12
Serialize `Dash._setup_server` under a lock and publish its flag last

### LINKED ISSUE #3971: [BUG] Race condition in Dash._setup_server(): guard flag is set before the work it protects
**Describe your context**

```
dash          4.4.1
dash_ag_grid  35.3.0     (only used by the first script, to have a component bundle to fetch;
                          the second script needs dash alone)
```

Flask 3.1.3 · Werkzeug 3.1.8 · gunicorn 26.2.0 (`gthread`, 4 workers × 4 threads) · Python 3.13

This is a server-side race, not a frontend bug, although the visible symptom is in the browser.

- OS: Linux aarch64; also reproduced independently on Linux x86-64 in Docker

**Describe the bug**

`Dash._setup_server()` runs as a `before_request` hook and is meant to execute once per
process. It sets its guard flag **before** performing the work that flag protects:

```python
def _setup_server(self):                                    # dash.py:1702
    if self._got_first_request["setup_server"]:
        return                                              # <-- thread B leaves here
    self._got_first_request["setup_server"] = True          # <-- flag set BEFORE the work
    ...
    _validate.validate_layout(self.layout, self._layout_value())   # 1724
    self._generate_scripts_html()                                  # 1726, fills registered_paths
    ...
    self.callback_map[k] = _callback.GLOBAL_CALLBACK_MAP.pop(k)    # 1737, fills callback_map
```

This is a TOCTOU. On any WSGI server that serves more than one request at a time in a single
process — `gunicorn -k gthread --threads N`, waitress, werkzeug with `threaded=True`, uWSGI
with threads — a second thread can enter while the first is still inside `_setup_server`, find
the flag already set, return immediately, and then read state the first thread has not finished
writing. It only affects the **first page load after a process starts**, and a refresh makes it
go away, which is why it is easy to dismiss as a caching or network glitch.

Both symptoms below are reproduced deterministically by the scripts linked further down.

**1. `registered_paths` still empty → every component-suites request 500s.** A browser asks for
the index and its JS bundles on parallel connections. The index thread is inside
`_setup_server`; the bundle threads skip it and reach `validate_js_path` against an empty
allowlist:

```
dash.exceptions.DependencyException: Error loading dependency. "dash_ag_grid" is not a
registered library.
Registered libraries are:
[]
```

The browser gets HTML where it expected JavaScript (`Refused to execute script ... MIME type
('text/html')`), so grids and clientside code are silently broken on that load. This is what we
hit in production.

**2. `callback_map` still incomplete → callbacks 500 with `KeyError`.** Callbacks registered
with `dash.callback` (module level — how every pages-based app registers them) are moved from
`GLOBAL_CALLBACK_MAP` into `self.callback_map` only inside `_setup_server`. A thread that skips
the setup carries on to `_prepare_callback`:

```
  File "dash/dash.py", line 1628, in _prepare_callback
    cb = self.callback_map[output]
KeyError: 'out.children'
...
KeyError: "Callback function not found for output 'out.children'."
```

The second script includes the control that makes this meaningful: the *same* request, replayed
once initialisation has finished, returns 200.

**Possibly related, not reproduced.** #2925 reports `RuntimeError: dictionary changed size
during iteration` with an identical signature ("occurs on startup and initial access … a full
refresh (Ctrl+F5) resolves the issue … affecting the first user to visit the app after startup
… becomes more frequent as callback numbers increase"). In 4.4.1 the only iteration of
`callback_map` on that path is `validate_background_callbacks` (`dash.py:1754`, called from
inside `_setup_server`; it was `validate_long_callbacks` in the 2.x line #2925 reports), so that
error would require *two* threads inside the body at once — possible in principle, since the
check and the set are separate statements and both threads can pass the check before either one
sets the flag. I could not trigger that: 0 occurrences in 400 trials × 8 threads released
through a barrier with `sys.setswitchinterval(1e-6)`. I mention it because the two symptoms I
did reproduce show that this state really is observable half-written, but I am not claiming
#2925 is proven to be the same bug.

*Steps to reproduce.* Deterministic, with no monkeypatching of Dash. The only "slow" ingredient
is a layout function that takes a moment — an ordinary application property, and the pattern
Dash documents as `app.layout = serve_layout`. It matters because `_setup_server` evaluates the
layout *before* filling `registered_paths` and `callback_map`, so it widens the pre-existing
window. Ordering of the requests is arranged on the client side only.

Reproduction scripts: https://gist.github.com/a-tram/1a522f6f2beb388543de7d5f18644962

```
pip install dash dash-ag-grid
python repro_setup_server_race.py       # symptom 1
python repro_callback_map_race.py       # symptom 2, with a control
```

Output on dash 4.4.1, identical on 5 consecutive runs each:

```
index: 200
bundle x40: {500: 40}
REPRODUCED
```

```
index: 200
update-component x20: {500: 20}
control (same request, setup finished): 200
REPRODUCED
```

Every one of those bundle 500s is the `DependencyException` quoted above, with `Registered
libraries are: []`.

Note that gunicorn's default worker class is `sync` (one request at a time per worker), so the
canonical `gunicorn app:server -w 4` deployment is immune. Only threaded deployments are
exposed, which probably explains why this is not reported more often. `--preload` does not help
either: `_setup_server` is a `before_request` hook, so it still runs after the fork, in every
worker.

**Expected behavior**

The first request to a freshly started process should not be able to observe half-initialised
state. Concurrent first requests should either wait for initialisation to finish or all see it
complete — never a raised flag next to an empty `registered_paths` or a partially filled
`callback_map`.

*Suggested fix.* Moving the flag after the work is not sufficient on its own: two threads would
then both run the full setup, and the `GLOBAL_CALLBACK_MAP.pop(k)` loop is not safe to run
twice. Double-checked locking keeps the warm path lock-free, and also closes the second window
noted above — today the check and the set are separate statements, so the flag does not prevent
two threads from running the whole body at once:

```python
# in __init__
self._setup_lock = threading.Lock()

def _setup_server(self):
    if self._got_first_request["setup_server"]:
        return
    with self._setup_lock:
        if self._got_first_request["setup_server"]:
            return
        ...  # existing body, unchanged
        self._got_first_request["setup_server"] = True   # set LAST
```

The same check-then-set-then-work shape appears for the `"pages"` key in `dash.py`
(`router_async` and `router_sync`), so it may be worth auditing that one as well.

*Workaround for anyone hitting this now.* Force the initialisation at worker start, where no
other thread can race it — for gunicorn, in `post_worker_init`:

```python
with app.server.test_request_context("/"):
    app._setup_server()
```

**Screenshots**

Not applicable — both reproductions are deterministic and print their result.

mokashang:
## Contributor Checklist

### optionals

#### Reference issue

Closes #3971.

#### What does this implement/fix?

`Dash._setup_server` is a `before_request` hook that is meant to execute once per process. It sets its `_got_first_request["setup_server"]` guard flag **before** performing the work that flag protects: populating `registered_paths` via `_generate_scripts_html`, copying the `GLOBAL_CALLBACK_MAP` entries into `callback_map`, running background-callback validation, and so on. On a multi-threaded WSGI worker such as `gunicorn -k gthread`, waitress, or `flask run --threaded`, a second request can enter `_setup_server` between the "set flag" and "do the work" statements, see the flag already set, skip setup, and then read `registered_paths` and validate against `callback_map` while both are still empty. The visible symptom is the first burst of `_dash-component-suites/<lib>/...` bundle requests after a restart returning 500 with `Error loading dependency. "<lib>" is not a registered library. Registered libraries are: []`.

The fix:

- Add a per-instance `threading.Lock` next to `_got_first_request` and wrap the setup body with double-checked locking: the fast-path early-return still fires without any lock cost once setup is done; a raced-in second thread waits on the lock, then sees the flag set by whichever thread won and returns without redoing the work.
- Move the `_got_first_request["setup_server"] = True` publication to the very end of the guarded block so no other thread can observe it while any side effect is still in flight.

#### Additional information

Scope is deliberately narrow. The same TOCTOU pattern exists in the `pages` guards inside `router_async` / `router_sync` (`if self._got_first_request["pages"]: return` followed by `self._got_first_request["pages"] = True`); those are left for a follow-up because the async version has an `await get_layouts()` inside the section that needs guarding, which needs `asyncio.Lock` rather than the plain `threading.Lock` used here.

Regression test in `tests/unit/test_setup_server_race.py` slows `_generate_scripts_html` by 100ms and runs four concurrent `_setup_server` calls behind a `threading.Barrier`. Before the fix, three of the four threads returned from `_setup_server` with `app.registered_paths` still empty; with the fix all four see it populated. `pylint tests/unit -d all -e C0410,C0413,W0109 --rcfile=.pylintrc` and `pylint dash --rcfile=.pylintrc` both stay clean, and `black dash tests --exclude 'metadata_test.py|node_modules' --check` passes.


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - [#3646](https://github.com/plotly/dash/pull/3646) Remove React 16 support (`16.14.0` is no longer an accepted value for `REACT_VERSION` / `_set_react_version`).
 
 ### Fixed
+- Fix `Dash._setup_server` publishing its "already done" guard flag before the setup work behind it had run. Under a multi-threaded WSGI worker such as `gunicorn -k gthread` a second request arriving mid-setup could observe the flag already set, skip setup, then read `registered_paths` / `callback_map` while they were still empty, causing the first burst of component bundle requests after a restart to 500 with `Error loading dependency. "<lib>" is not a registered library`. The setup body now runs under a lock and only publishes the flag after all work completes. Fixes [#3971](https://github.com/plotly/dash/issues/3971).
```
T4rk1n, line 20: Neighbor entries lead with the PR link (`[#3944](...)`, `[#3955](...)`). Prefix this with `[#3980](https://github.com/plotly/dash/pull/3980)` to match. The `Fixes #3971` at the end is the issue, not the PR.

T4rk1n (commented):
The `_setup_server` fix is correct: flag published last under the lock, double-checked so the post-setup path stays lock-free, with a real regression test. But the same fix needs to land on the pages router in this PR.

`router_sync` / `router_async` (`dash/dash.py:2737` and `dash/dash.py:2658`) are `before_request` hooks with the identical set-flag-before-work TOCTOU, so a raced request skips router registration and the app serves with the `_ID_CONTENT` callback unregistered and `validation_layout` unbuilt: the same class of first-burst-after-restart breakage, just in pages.

- `router_sync` (`dash/dash.py:2737`): a straight mirror of this change. Wrap the body in `threading.Lock` double-checked locking and publish the flag last, same as `_setup_server`. No reason to defer it.
- `router_async` (`dash/dash.py:2658`): needs an `asyncio.Lock` because of the `await get_layouts()` inside the guarded region (a `threading.Lock` held across the await would deadlock the loop), and on 3.9 that lock has to be lazily created inside the coroutine (before the first await) rather than at `__init__`.

One CHANGELOG nit inline. Fix the pages guards and this is good.

CHANGELOG.md
mokashang: Done in 6ad48ede7 - CHANGELOG entry now leads with `[#3980](...)`. Same commit also folds in the router_sync / router_async lock changes.

mokashang:
Thanks for the review. Pushed [6ad48ede7](https://github.com/plotly/dash/pull/3980/commits/6ad48ede7) which folds the same fix into both pages hooks:

- **`router_sync`** (`dash/dash.py:2736`): mirrors `_setup_server` verbatim - new per-instance `self._pages_lock = threading.Lock()`, double-checked read of `_got_first_request["pages"]`, flag published only after the callback and `validation_layout` are in place.
- **`router_async`** (`dash/dash.py:2657`): uses an `asyncio.Lock` per your guidance. The lock is stored as `self._pages_async_lock: Optional[asyncio.Lock] = None` and lazily constructed inside the coroutine (`if self._pages_async_lock is None: self._pages_async_lock = asyncio.Lock()`) before the first await, so on 3.9 it binds to the running event loop rather than whatever loop was current at `Dash()` construction. The check-and-assign is safe because asyncio only yields at await points.
- **CHANGELOG**: renamed to lead with `[#3980](https://github.com/plotly/dash/pull/3980)` and rewrote the entry to cover all three hooks.

Added `tests/unit/test_pages_router_race.py` with a regression test per hook. Both are structured to reproduce the buggy state under normal Python scheduling:

- **`test_router_sync_is_atomic_under_concurrent_requests`**: monkey-patches `dependencies.Input.__init__` on the first call to sleep 100ms, opening the window between the flag being set (on the pre-fix path) and the router callback being registered. 4 threads race in behind a `Barrier` and each snapshots `app.callback_map` at the moment its `router_sync()` call returns; the test asserts every snapshot contains the `_pages_content` router entry. Before the fix, the raced-in threads' snapshots were empty of that entry.
- **`test_router_async_is_atomic_under_concurrent_tasks`**: registers a page whose `layout=` is an `async` function that awaits 50ms, forcing an await inside `get_layouts()` in the guarded region. Two tasks are gathered concurrently and each returns the state visible at the moment its `router_async()` returned. Before the fix the second task returned with `validation_layout is None`; after, both tasks see it populated and the router callback registered exactly once.

T4rk1n:
Seems like there is an issue with serving the favicon in this PR:
```
{'level': 'SEVERE', 'message': 'http://localhost:58136/favicon.ico - Failed to load resource: the server responded with a status of 500 (INTERNAL SERVER ERROR)', 'source': 'network', 'timestamp': 1790011073837}
```
Also need to run `npm run format`

PR #3984, opened 2026-09-16
Post benchmark results comment on fork PRs

T4rk1n:
## Problem

On community (fork) PRs the Performance Benchmarks workflow fails to post its results comment:

```
RequestError [HttpError]: Resource not accessible by integration
POST /repos/plotly/dash/issues/{n}/comments  ->  403
```

Fork PRs run with a read-only `GITHUB_TOKEN`, so the `pull-requests: write` declared in `benchmarks.yml` is silently capped and the inline `actions/github-script` comment step 403s. This can't be fixed inside that workflow: GitHub caps the token because the run executed fork-authored code.

## Fix

Same pattern the repo already uses in `post-test-status.yml`: do the privileged work in a `workflow_run` follow-up, which runs in the base-repo context with a real read/write token even for forks.

- `benchmarks.yml`: drop the inline comment step (and the now-unused top-level `pull-requests: write`). Record the PR number into the `benchmark-results` artifact, since `workflow_run`'s payload omits it for fork PRs.
- New `post-benchmark-comment.yml`: on Performance Benchmarks completion, download the artifact and post/update the sticky comment.

The benchmark job still runs and still fails the check on a hard regression; only the comment posting moves to where it has permission.

Guards: the follow-up skips cleanly (no red check) when a cancelled/superseded run left no artifact, and the PR-number step is `always()` so a pre-benchmark CI flake doesn't lose it.

## Notes

`workflow_run` reads the workflow file from the default branch, so the follow-up won't fire until this merges to `dev`. It can't be exercised end-to-end from this branch alone.


.github/workflows/post-benchmark-comment.yml
```diff
@@ -0,0 +1,69 @@
+name: Post Benchmark Comment
+
+# Posts the benchmark results comment in the base-repo context (with a
+# read/write token) after the Performance Benchmarks workflow finishes. Fork
+# PRs run that workflow with a read-only GITHUB_TOKEN, so its own comment step
+# would 403 ("Resource not accessible by integration"). Same pattern as
+# post-test-status.yml. The results and the PR number arrive via the
+# benchmark-results artifact.
```
camdecoster, line 8: The comment step has been removed, so this comment should be updated.

camdecoster (commented):
Looks good overall. Could you pin the actions to commit hashes?

PR #3985, opened 2026-09-16
Refresh JS dependencies within range, hold breaking majors

T4rk1n:
## What

A single consolidated dependency-update PR to replace the batch of stale, failing dependabot PRs. All six npm lockfiles are regenerated to the latest versions **within the existing semver ranges** (node 24 / npm 11, preserving `lockfileVersion 2`): root, `dash/dash-renderer`, `components/dash-core-components`, `components/dash-table`, `components/dash-html-components`, and `@plotly/dash-component-plugins`.

**No direct dependency crosses a major boundary.** The open dependabot PRs fail because they bundle breaking majors (eslint 9, typescript 6/7, webpack-cli 7, react-window 2, Babel 8). Per a conservative policy, this PR takes every safe minor/patch/security bump and holds the breakers.

## Notable changes

- **renderer:** the pinned `@types/ramda@0.29.1` transitively pulls `types-ramda`, which within its range jumped 0.29.2 -> 0.29.10 and tightened `keys()`'s generic to `extends object`, breaking `keys<string>(...)` in `src/actions/callbacks.ts` with 37 `tsc` errors. Pinned `types-ramda` to 0.29.2 via a **scoped** `overrides` entry (dev-only typings, no source change). Also bumped `webpack-cli` `^5` -> `^7` to match the component packages (verified building).
- **dash-table:** `webpack-dev-server` 5.2.4 -> 5.2.6 (security, closes #3868).
- **dependabot.yml:** added `ignore` entries so the majors that break the build stop reopening red PRs: `typescript >=6`, `@babel/* >=8`, `babel-loader >=10` (all npm dirs), `react-window >=2` (dcc), `@types/ramda >=0.30` (renderer, to stay in sync with the override). eslint/prettier/@typescript-eslint were already locked by the existing allow-lists.

## Verification (all on node 24, CI's version)

- **renderer:** webpack build (0 errors), eslint (0 errors), prettier, karma 85/85
- **dash-core-components:** build, es-check ES9, eslint, jest 85/85
- **dash-table:** build, es-check ES6, eslint, prettier
- **dash-html-components:** metadata extract (react-docgen 5) + build + eslint
- **@plotly/dash-component-plugins:** build
- No source-file drift.

## Known held vulnerabilities

Two dev/build-only advisories remain, fixable only by breaking majors, so held under the conservative policy:
- `elliptic` (via `node-polyfill-webpack-plugin`, browser crypto polyfill; needs the plugin's v4 major)
- `serialize-javascript` (via `mocha` test runner; needs mocha 12)

## Supersedes

Stale dependabot PRs: #3868, #3869, #3890, #3904, #3905, #3906, #3910, #3917, #3918, #3919, #3920. These should be closed once this merges.

## Follow-ups (not in this PR)

- `@plotly/dash-component-plugins` has no `dependabot.yml` entry (its 4 stale PRs #3917-#3920 predate the current allow-lists). Its lock is refreshed here, but it has no ongoing dependabot coverage; adding a restricted entry is a separate decision.
- Several other workspace packages (`@plotly/eslint-config-dash`, test-component packages, etc.) are likewise not covered by dependabot. No open PRs today; noting the gap.


dash/dash-renderer/package.json
```diff
@@ -80,9 +80,14 @@
     "ts-loader": "^9.5.2",
     "typescript": "^5.8.3",
     "webpack": "^5.104.1",
-    "webpack-cli": "^5.1.4",
+    "webpack-cli": "^7.0.3",
```
camdecoster, line 83: This is a major update, contradicting the PR description. Was this intentional?

camdecoster (commented):
This all seems fine. I just had one question before approving.

dash/dash-renderer/package.json
T4rk1n: > . Also bumped webpack-cli ^5 -> ^7 to match the component packages (verified building).

I think it's fine, the other packages were already on 7, this one had drifted.

camdecoster (approved):

PR #3986, opened 2026-09-16
Modify before-request hooks to handle results for fastapi

BSd3v:
For fastapi, the before hooks are a middleware insert.

Flask and Quart both allow for an early return when using before_request hooks, this just allows it to follow the same pattern.


T4rk1n (approved):
💃

PR #3988, opened 2026-09-20
Test graph PDF output with more than 1000 points

### LINKED ISSUE #3889: [BUG] Graphs excluded from print pages if they have over 1000 datapoints
**Context**
When saving a Dash app as a pdf  `dcc.Graph` behaves wonky if the graph contains more than 1000 datapoints (starting at 1001 points). The actual graph doesn't appear properly in the printed page. Instead, an empty graph is shown (or it could be that the background is printed over the graph).

I tested the behavior in Firefox, Chrome and Edge, and the behavior is different per browser. I tested both printing the page via the browser print page functionality and via the html2pdf library. 

In Chrome and Edge, the browser print page functionality works fine with more than 1000 datapoints. In Firefox, the datapoints are gone, and the print preview shows an empty plot.

Using the html2pdf library, all browser have issues. In Chrome and Edge, the plot line seems to have a different scale and is hidden behind the plot background. In Firefox the plot line is completely gone, just showing an empty plot. It could be that it is hidden behind the plot background as well but that the scale is smaller so we don't see it.

It's possibly a bit of an edge case that might not be due to Dash. If it is not, please consider documenting this limitation somewhere. 

**Expected behavior**
The plot should be visible on the printed page regardless how many points are in the plot.

**Example app**
To reproduce the behavior, take this example app for generating a simple plot and play around with `num_points`. 

```python
import dash
from dash import dcc, html, clientside_callback, Input
import plotly.express as px
import numpy as np

app = dash.Dash(__name__)

num_points = 2000
x = np.linspace(0, 10, num_points)
y = np.sin(x) * 100 + np.random.normal(0, 5, num_points)

fig = px.line(
    x=x,
    y=y,
    title="Line Chart",
)

app.layout = html.Div(
    id="main_container",
    children=[
        html.H1("Plotly Dash App"),
        dcc.Graph(figure=fig),
        html.Div(
            children=[
                dcc.Button(
                    id="save_as_pdf",
                    children="Save as PDF",
                )
            ],
        ),
    ]
)

clientside_callback(
    f"""
    function (button_clicked) {{
        if (button_clicked && button_clicked > 0) {{
            var mainContainerElement = document.getElementById("main_container");

            var opt = {{
                // margin units are those that are defined in jsPDF key below.
                margin: 10,
                filename: 'test_file.pdf',
                image: {{ type: 'jpeg', quality: 0.98 }},
                html2canvas: {{
                    scale: 3,
                    width: 1920,
                    dpi: 300,
                }},
                jsPDF: {{ unit: 'mm', format: 'A4', orientation: 'p' }},
                // pagebreak: {{ mode: ['avoid-all'] }}
            }};
            html2pdf().from(mainContainerElement).set(opt).save();
        }}
    }}
    """,
    Input(component_id="save_as_pdf", component_property="n_clicks"),
)

if __name__ == "__main__":
    app.run(debug=True)
```

**Setup**
```
dash 4.0.0
dash-core-components 2.0.0
dash-html-components 2.0.0
dash-table 5.0.0
plotly 6.5.2
numpy 2.4.4
```
Take `html2pdf` from here: https://github.com/eKoopmans/html2pdf.js/blob/main/dist/html2pdf.bundle.min.js and put it in the `assets` folder next to the example app.

-  Browser, Version and OS

    - OS: Windows
    - Browser: 
      - Chrome: 150.0.7871.114
      - Firefox: 152.0.4
      - Edge: 150.0.4078.48

**Screenshots**
Edge/Chrome with 2000 points

Print page (this is also how it shows in all cases when number of points <= 1000)
Image

html2pdf
Image

Firefox with 2000 points

Print page
Image

html2pdf
Image

lorenzozanee:
Adds a deterministic browser integration fixture for 1000 and 1001 point graphs. The test emulates print media and exercises Chrome Page.printToPDF while preserving the existing graph and browser-log assertions.

The assertion verifies a valid non-empty PDF and a visible SVG trace; it does not inspect rendered PDF pixels, so cross-browser validation remains for CI and maintainers.

## Contributor Checklist

   -  [ ] task 1
   -  [ ] task 2

### optionals

    -  [ ] here is the show and tell thread in Dash community

Fixes #3889


components/dash-core-components/tests/integration/graph/test_graph_print.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@pytest.mark.parametrize("point_count", [1000, 1001])
+def test_grpr001_graph_survives_print(point_count, dash_dcc):
+    app = Dash(__name__)
+    points = list(range(point_count))
+    app.layout = dcc.Graph(
+        id="graph",
+        figure={
+            "data": [{"type": "scatter", "mode": "lines", "x": points, "y": points}],
```
T4rk1n: `type: "scatter"` never goes down the path that breaks. 1000 and 1001 points both render as SVG, so the parametrize doesn't test the threshold at all. The switch at 1001 happens because `px` picks `scattergl`. Can you use `scattergl` (or build the figure with `px.line`) so the test hits the WebGL path?

components/dash-core-components/tests/integration/graph/test_graph_print.py
```diff
@@ -0,0 +1,36 @@
...
+    dash_dcc.start_server(app)
+    dash_dcc.wait_for_element("#graph .main-svg")
+
+    driver = dash_dcc.driver
+    if not hasattr(driver, "execute_cdp_cmd"):
+        pytest.skip("print-to-PDF regression requires Chrome DevTools Protocol")
+
+    driver.execute_cdp_cmd("Emulation.setEmulatedMedia", {"media": "print"})
+    pdf = driver.execute_cdp_cmd(
+        "Page.printToPDF", {"printBackground": True, "preferCSSPageSize": True}
```
T4rk1n: The issue says Chrome's native print works fine at 2000 points. The failing cases are Firefox print and html2pdf/html2canvas. A Chrome `Page.printToPDF` check passes whether the bug is there or not. What case is this meant to fail on?

components/dash-core-components/tests/integration/graph/test_graph_print.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+
+    pdf_bytes = base64.b64decode(pdf["data"])
+    assert pdf_bytes.startswith(b"%PDF")
```
T4rk1n: These assertions don't look at what got printed. Any page makes a PDF with a `%PDF` header that's over 1000 bytes, and the `is_displayed()` check reads the live DOM, not the PDF. An empty plot in the PDF (the reported symptom) passes all three. If you want to test print output, can you render the PDF, or a canvas snapshot of the graph, and check that the trace pixels are there?

T4rk1n (commented):
This doesn't fix #3889 and the test would pass on `dev` as-is, so it can't guard against the bug. The 1000-point cutoff in the issue comes from `px.line`: with `render_mode="auto"`, plotly express switches to `scattergl` once you pass 1000 rows (`px.line(x=range(1001), y=range(1001)).data[0].type == "scattergl"`). So the actual bug is about WebGL canvas content not making it into print or html2canvas output. This test hardcodes `type: "scatter"`, which renders SVG at any point count, and it uses Chrome native print, which the issue already says works. I'd block on this: please drop the `Fixes #3889` line. Either turn this into a real repro (scattergl, checking the printed output) or close it and leave the issue open.

